PR #1, opened 2016-08-12
Registry stuff

shadowfacts:
Remapped most of the registry stuff.

Used Mojang names where possible (only `DefaultMappedRegistry`)

### DIFF
--- a/mappings/net/minecraft/util/registry/IDRegistry.json
+++ b/mappings/net/minecraft/util/registry/IDRegistry.json
@@ -1,0 +1,106 @@
+{
+  "obf": "db",
+  "name": "net/minecraft/util/registry/IDRegistry",
+  "field": [
+    {
+      "obf": "a",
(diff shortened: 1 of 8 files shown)


mappings/net/minecraft/util/registry/IDRegistry.json
```diff
@@ -0,0 +1,106 @@
+{
+  "obf": "db",
+  "name": "net/minecraft/util/registry/IDRegistry",
```
asiekierka: IdRegistry. We don't capitalize both letters of ID.

asiekierka:
Could you squash the two commits? I normally wouldn't ask that, but the change here is so small a squashing would be good for future viewers.

shadowfacts:
GitHub can [squash and merge](https://github.com/blog/2141-squash-your-commits).

asiekierka:
Woah.

PR #4, opened 2016-08-13
Misc

shadowfacts:
Block stuff, state stuff, tile entity stuff, creative tabs.


mappings/net/minecraft/block/Block.mapping
```diff
@@ -1,19 +1,30 @@
 CLASS none/akw net/minecraft/block/Block
 	FIELD A stateFactory Lnone/ask;
 	FIELD a id Lnone/kp;
+	FIELD b creativeTab Lnone/adq;
 	FIELD c defaultState Lnone/asj;
+	FIELD d unlocalizedName Ljava/lang/String;
```
asiekierka: Any better ideas?

mappings/net/minecraft/block/Block.mapping
```diff
@@ -1,19 +1,30 @@
...
+	METHOD a getUnlocalizedName ()Ljava/lang/String;
 	METHOD a deserializeState (I)Lnone/asj;
 		ARG 0 meta
 	METHOD a register (ILjava/lang/String;Lnone/akw;)V
 	METHOD a register (ILnone/kp;Lnone/akw;)V
+	METHOD a setCreativeTab (Lnone/adq;)Lnone/akw;
+		ARG 0 creativeTab
 	METHOD a getBlockFromItem (Lnone/aeo;)Lnone/akw;
+		ARG 0 item
+	METHOD a addSubItems (Lnone/aeo;Lnone/adq;Lnone/dc;)V
```
asiekierka: Wouldn't "addItemsToTab" be closer to its purpose?

mappings/net/minecraft/block/Block.mapping
```diff
@@ -1,19 +1,30 @@
...
+	METHOD a activate (Lnone/aiu;Lnone/cn;Lnone/asj;Lnone/aak;Lnone/qz;Lnone/cu;FFF)Z
```
asiekierka: So we're not doing onX?

mappings/net/minecraft/block/Block.mapping
shadowfacts: It's not only for creative tabs though. JEI uses this method with a `null` creative tab to get all the sub items.

mappings/net/minecraft/block/Block.mapping
shadowfacts: *shrugs* I went with `unlocalizedName` because that's what it is, the block's name before it's been translated for the current locale.

mappings/net/minecraft/block/Block.mapping
shadowfacts: Either is fine however I prefer more concise names and `onBlockActivated` or `onActivated` doesn't convey any more information than `activate` does.

mappings/net/minecraft/block/Block.mapping
asiekierka: No, think of the vanilla context. It is intended only for creative tabs in vanilla.

I'd debate this on IRC if I were you.

mappings/net/minecraft/block/Block.mapping
```diff
@@ -30,14 +41,21 @@ CLASS none/akw net/minecraft/block/Block
 	METHOD b (Lnone/asj;Lnone/aiu;Lnone/cn;)Lnone/bcp;
 		ARG 0 state
 		ARG 1 world
+	METHOD b getBoundingBox (Lnone/asj;Lnone/aiy;Lnone/cn;)Lnone/bcp;
+	METHOD c getLocalizedName ()Ljava/lang/String;
```
shadowfacts: Either works, I just went with unlocalized and localized because that's what they are and those are the de facto standard.

mappings/net/minecraft/block/Block.mapping
asiekierka: I suppose. I thought of "getTranslationKey/getTranslatedName" just now...

jamierocks: `translationKey` / `translatedValue`?

asiekierka:
Awaiting further input on this one - see line notes.

mappings/net/minecraft/block/Block.mapping
asiekierka: I'm not sure. Technically, just doing block.activate() does not activate the block _per se_ - it skips things like future event hooks we might want to add, or item overrides.

mappings/net/minecraft/block/Block.mapping
asiekierka: getStacksOnDisplay? says the modmuss50

PR #10, opened 2016-08-15
Map some blocks, entities, and related methods and classes.

darkevilmac:
Let me know if anything does not match guidelines and needs to be changed.


mappings/net/minecraft/entity/impl/EntityAreaEffectCloud.mapping
```diff
@@ -1 +1,4 @@
 CLASS none/sc net/minecraft/entity/impl/EntityAreaEffectCloud
+	METHOD a readEntityTagCompound (Lnone/dt;)V
```
asiekierka: We want to be doing serializeEntityTag/deserializeEntityTag I think

mappings/net/minecraft/entity/impl/EntityAreaEffectCloud.mapping
```diff
@@ -1 +1,4 @@
...
+	METHOD b writeEntityTagCompound (Lnone/dt;)V
+	METHOD m updateEntity ()V
```
asiekierka: Why not just update? I know entities have two update functions, so... not sure.

darkevilmac:
@asiekierka Fixed.

PR #11, opened 2016-08-15
Map remaining entities as well EntityRegistry.

darkevilmac:
Parent classes of entities still not mapped, let me know if things should be moved. The names should be correct though.


darkevilmac:
#12

PR #13, opened 2016-08-16
Improve text mappings

jamierocks:
- Mojang give away many mappings for free via their #toString() we should use those

- impl packages are horrid

- net/minecraft/text/event vs net/minecraft/event/text

- AbstractTextComponent vs TextComponentBase

- TextStyle vs Style

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ The current version targets Minecraft version **16w32b**.
 
 ## Contributing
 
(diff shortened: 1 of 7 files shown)


shadowfacts:
1. Mojang does given away some names through `toString`, but @asiekierka and myself agreed that we should ignore them and follow the `ThingX` instead of `XThing` convention.
2. `impl` packages are the standard used everywhere else, text components should adhere to the standard as well.
3. Ok
4. You didn't actually rename `TextComponentBase`.
5. Ok

asiekierka:
I disagree with the changes above, bar net/minecraft/text/event being a thing.

PR #14, opened 2016-08-16
Map sound things, tile renderers, and other misc things.

darkevilmac:
Feel free to leave feedback, notable changes are mapping of Sounds for reference, and Sound which includes sound registration.

TileEntityRenderer has also been mapped which includes it's implementations.


mappings/net/minecraft/block/Block.mapping
```diff
@@ -135,6 +135,7 @@ CLASS none/akw net/minecraft/block/Block
 	METHOD a getRawIdFromBlock (Lnone/akw;)I
 		ARG 0 block
 	METHOD a (Lnone/akw;Lnone/akw;)Z
+	METHOD a setSounds (Lnone/apu;)Lnone/akw;
```
asiekierka: Are you sure "setSounds" is good? I don't like it. Perhaps "setSoundType"? "setSoundSet"? @shadowfacts

mappings/net/minecraft/block/impl/BlockFire.mapping
```diff
@@ -7,10 +7,12 @@ CLASS none/amp net/minecraft/block/impl/BlockFire
 	FIELD f UP Lnone/asv;
 	METHOD a deserializeState (I)Lnone/asj;
 	METHOD a getDropAmount (Ljava/util/Random;)I
+	METHOD a setBurnLevels (Lnone/akw;II)V
```
asiekierka: setBurnLevel*s*? Please define the parameters so we can decide on the name further.

mappings/net/minecraft/block/impl/BlockFire.mapping
```diff
@@ -7,10 +7,12 @@ CLASS none/amp net/minecraft/block/impl/BlockFire
...
 	METHOD a getCollisionBox (Lnone/asj;Lnone/aiy;Lnone/cn;)Lnone/bcp;
 	METHOD b createStateFactory ()Lnone/ask;
 	METHOD b randomTick (Lnone/aiu;Lnone/cn;Lnone/asj;Ljava/util/Random;)V
 	METHOD c getRenderingState (Lnone/asj;Lnone/aiy;Lnone/cn;)Lnone/asj;
+	METHOD e init ()V
```
asiekierka: This doesn't init BlockFire, it inits burn values - initX, where X is the renamed setBurnLevels' latter half, would fit better.

mappings/net/minecraft/sound/BlockSound.mapping
```diff
@@ -0,0 +1,29 @@
+CLASS none/apu net/minecraft/sound/BlockSound
```
asiekierka: BlockSound? see setSounds comment above

mappings/net/minecraft/block/Block.mapping
shadowfacts: Depends what the class itself is called. If the class is called Sounds setSounds is fine, if the class is call SoundType then it should be setSoundType

darkevilmac:
@asiekierka Fixed mentioned issues.

PR #16, opened 2016-08-18
Update some Block mappings

copygirl:
> **Fields:** friction
> **Methods:** dropItems, dropStack, dropExperience
> **Parameters:** fortune
> 
> Renamed EntityXPOrb to EntityExperienceOrb

Just a small PR to see if I'm doing things right.
I suppose there's some open questions if you're willing to discuss them:
- `dropItems` could perhaps get a better name. (Just `drop`?)
- `dropStack` and `dropExperience` are static "helper" methods, they could be called `dropXYZFromBlock` though they're already called as `Block.dropXYZ` so I felt like `FromBlock` is unnecessary.
- `dropExperience` could be called `dropExperienceOrbs`, though I personally don't see any added value from being more specific.
- `fortune` could be called `fortuneLevel`
- I didn't know what the float parameter on `dropItems` is used for.

### DIFF
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -13,6 +13,7 @@ CLASS none/akw net/minecraft/block/Block
 	FIELD v soundGroup Lnone/apu;
 	FIELD x material Lnone/ayo;
 	FIELD y mapColor Lnone/ayp;
+	FIELD z friction F
 	METHOD <init> (Lnone/ayo;)V
 		ARG 0 material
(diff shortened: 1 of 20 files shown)


asiekierka:
- Just "drop" is a bit unintuitive - drop what? A falling block of sand?
- dropExperienceOrbs avoids confusion between experience level and experience entity.

shadowfacts:
The changes look good to me.

In response to your questions:
- `dropItems` is fine
- I think `dropStack` and `dropExperience` are fine
- Either is fine
- `fortune` is better IMO, as `fortuneLevel` doesn't provide any additional information

PR #19, opened 2016-08-18
Update Material mappings

copygirl:
### Changes
- Named almost all of the materials
- Named most fields and methods
- Named all Material classes (`Air`, `Liquid`, `NonSolid` and `Portal`)
- Named `PistonBehavior`, moved into `net.minecraft.block.piston`
- Moved Material classes into `net.minecraft.block.material`
### Notes
- `v` isn't used anywhere so I left it obfuscated.
- `DECORATION` is the best name I could come up with for redstone, rails, torches, buttons, flower pots. Another name might be more fitting.
- `OTHER` is similar: It includes clay, silverfish and slime blocks.
- `PLANT`, `PLANT_FLAMMABLE` and `PLANT_BLOCK` seemed like a good way to differenciate between the different types of plants blocks. Though, I suppose the latter could be named more appropriately (it's for melon and pumpkin blocks).
- Some materials are named directly after the only block they're used for. Could be renamed if used for other blocks in the future.
- I left `p()` and `Q` obfuscated since they don't seem to be used, though it might be related to blocks that let light through?
- Unsure about the use of `b()`.
- I hope the methods `setPistonBlocking` and `setPistonBreakable` are named ok.
- I went for `harvestable` regarding the "harvestability" (proper tool required to get drops).
- Unsure about the transparency / opacity / solidity fields and methods. We should try to [choose appropriate names for these](https://twitter.com/ShadowfactsDev/status/761263864159887360). It is worth noting that the community has [already accepted a definition for some of these](http://minecraft.gamepedia.com/Opacity).

Additionally, I apologize for the PR possibly being borked and including more than the latest commit. I don't know what I did wrong.


mappings/net/minecraft/block/material/Material.mapping
```diff
@@ -0,0 +1,61 @@
+CLASS none/ayo net/minecraft/block/material/Material
+	CLASS none/ayo$1
+		METHOD c isSolid ()Z
+	FIELD A CACTUS Lnone/ayo;
+	FIELD B OTHER Lnone/ayo;
+	FIELD C PLANT_BLOCK Lnone/ayo;
```
shadowfacts: What's the difference between the `PLANT` and `PLANT_BLOCK` materials? I'm wondering if there's a better names as currently, it's a bit confusing.

copygirl: Essentially, `PLANT_BLOCK` is meant for full blocks: Melon and pumpkin. I addressed this in the notes, and I agree.

darkevilmac: why not map the plant mapping as crop and change plant block to plant?

copygirl: It does in¢lude flowers. The main difference between `PLANT` and `PLANT_FLAMMABLE` is the flammability. The main difference between those (which are `MaterialNonSolid`) and `PLANT_BLOCK` (regular `Material`) is that the latter uses full cubes.

asiekierka: PLANT_BLOCK looks ugly to me still. PLANT_CUBOID? PLANT_CUBE?

PR #23, opened 2016-08-20
Mapped out Enchantment and related classes

Darkhax:
This PR maps the Enchantment class further, and also maps related classes. Things like implementations, or classes that are used/referenced by the Enchantment class.

### DIFF
--- a/mappings/net/minecraft/enchantment/Enchantment.mapping
+++ b/mappings/net/minecraft/enchantment/Enchantment.mapping
@@ -1,17 +1,53 @@
 CLASS none/ahm net/minecraft/enchantment/Enchantment
+	CLASS none/ahm$a Weight
+		FIELD a COMMON Lnone/ahm$a;
+		FIELD b UNCOMMON Lnone/ahm$a;
+		FIELD c RARE Lnone/ahm$a;
+		FIELD d LEGENDARY Lnone/ahm$a;
+		FIELD e WEIGHT I
+		METHOD a getWeight ()I
+	FIELD a slotTypes [Lnone/sj;
 	FIELD b REGISTRY Lnone/db;
+	FIELD c type Lnone/ahn;
 	FIELD d translationName Ljava/lang/String;
+	FIELD e weight Lnone/ahm$a;
+	METHOD <init> (Lnone/ahm$a;Lnone/ahn;[Lnone/sj;)V
+		ARG 0 weight
+		ARG 1 type
+		ARG 2 slotTypes
 	METHOD a getTranslationKey ()Ljava/lang/String;
+	METHOD a getProtectionAmount (ILnone/rp;)I
+		ARG 0 level
+		ARG 1 source
+	METHOD a getAdditionalDamge (ILnone/ss;)F
+		ARG 0 level
+		ARG 1 entityType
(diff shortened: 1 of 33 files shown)


mappings/net/minecraft/enchantment/Enchantment.mapping
```diff
@@ -1,17 +1,53 @@
 CLASS none/ahm net/minecraft/enchantment/Enchantment
+	CLASS none/ahm$a Weight
+		FIELD a COMMON Lnone/ahm$a;
+		FIELD b UNCOMMON Lnone/ahm$a;
+		FIELD c RARE Lnone/ahm$a;
+		FIELD d LEGENDARY Lnone/ahm$a;
```
darkevilmac, line 6: Maybe the weights should be named in a more standard way like: EXTREME, HIGH, MEDIUM, LOW.

Darkhax, line 6: In this case the weighting is used to describe the rarity of the enchantment.

asiekierka, line 6: I can't believe you used LEGENDARY. Pfft. It's fine, though.

mappings/net/minecraft/enchantment/Enchantment.mapping
```diff
@@ -1,17 +1,53 @@
...
+	METHOD a onTargetDamaged (Lnone/sn;Lnone/se;I)V
+		ARG 0 user
+		ARG 1 target
+		ARG 2 level
+	METHOD b getHighestLevel ()I
 	METHOD b getEnchantmentById (Ljava/lang/String;)Lnone/ahm;
 		ARG 0 id
 	METHOD b getRawIdFromEnchantment (Lnone/ahm;)I
 		ARG 0 enchantment
+	METHOD b onUserDamaged (Lnone/sn;Lnone/se;I)V
```
asiekierka, line 40: I assume those are reactive and not active? (That is, they don't actually damage things, just react to a damage happening)

asiekierka:
Other than that question, LGTM

mappings/net/minecraft/enchantment/Enchantment.mapping
copygirl, line 40: `User` is a bit generic. Is this just for armor? I'd suggest `Wearer`, even if it sounds a bit odd.

shadowfacts, line 40: This method is for armor only right? If so, I concur with @copygirl, Wearer would make more sense.

shadowfacts:
Other than my one comment, this looks good to me.

mappings/net/minecraft/enchantment/Enchantment.mapping
Darkhax, line 40: This is actually not limited to armor. Enchantments like Thorns use this method for their effect. While thorns is an armor enchantment, the effect also works if the enchantment is on the held item, or the offhand item.

asiekierka:
LGTM

mappings/net/minecraft/enchantment/Enchantment.mapping
copygirl, line 40: How about `Owner`?

PR #25, opened 2016-08-20
Mapped out tools and related things

Darkhax:
This maps out some of the methods and fields that are related to tool classes.

### DIFF
--- a/mappings/net/minecraft/item/tool/ItemHoe.mapping
+++ b/mappings/net/minecraft/item/tool/ItemHoe.mapping
@@ -1,2 +1,19 @@
 CLASS none/aen net/minecraft/item/tool/ItemHoe
+	FIELD a material Lnone/aeo$a;
+	FIELD b swingSpeed F
+	METHOD <init> (Lnone/aeo$a;)V
+		ARG 0 material
 	METHOD a activate (Lnone/aak;Lnone/aiu;Lnone/cn;Lnone/qz;Lnone/cu;FFF)Lnone/rb;
+	METHOD a tillSoil (Lnone/aeq;Lnone/aak;Lnone/aiu;Lnone/cn;Lnone/asj;)V
+		ARG 0 stack
+		ARG 1 user
+		ARG 2 world
+		ARG 3 pos
+		ARG 4 newState
+	METHOD a onEntityDamaged (Lnone/aeq;Lnone/sn;Lnone/sn;)Z
+		ARG 0 stack
+		ARG 1 target
+		ARG 2 user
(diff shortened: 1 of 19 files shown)


mappings/net/minecraft/item/tool/ItemTool.mapping
```diff
@@ -1,5 +1,35 @@
...
+		ARG 1 state
+	METHOD a onEntityDamaged (Lnone/aeq;Lnone/sn;Lnone/sn;)Z
+		ARG 0 stack
+		ARG 1 target
+		ARG 2 user
+	METHOD a getAttributeModifiers (Lnone/sj;)Lcom/google/common/collect/Multimap;
+		ARG 0 slotType
+	METHOD c getEnchantability ()I
 	METHOD g getMaterial ()Lnone/aeo$a;
+	METHOD h getMaterial ()Ljava/lang/String;
```
asiekierka: Two methods with the same name which output different things? Looks like a bug.

mappings/net/minecraft/item/tool/ItemHoe.mapping
```diff
@@ -1,2 +1,19 @@
 CLASS none/aen net/minecraft/item/tool/ItemHoe
+	FIELD a material Lnone/aeo$a;
+	FIELD b swingSpeed F
+	METHOD <init> (Lnone/aeo$a;)V
+		ARG 0 material
 	METHOD a activate (Lnone/aak;Lnone/aiu;Lnone/cn;Lnone/qz;Lnone/cu;FFF)Lnone/rb;
+	METHOD a tillSoil (Lnone/aeq;Lnone/aak;Lnone/aiu;Lnone/cn;Lnone/asj;)V
```
asiekierka, line 7: Is there a simpler way to word this? Is it specific to the Hoe?

Darkhax, line 7: Yes, this method is only used by the hoe, and is only in the hoe class. It's basically a fancy way for the how to set a block and play it's sound.

mappings/net/minecraft/item/tool/ItemTool.mapping
Darkhax: This is a typo on my end. The one which returns String should be getMaterialName. Will fix.

mappings/net/minecraft/item/tool/ItemHoe.mapping
shadowfacts, line 7: I think `tillSoil` is fine, that's what the method does.

shadowfacts:
LGTM

PR #27, opened 2016-08-21
Mapped a bunch of misc stuff

Darkhax:
I had hoped this would have been more organized, however these mappings are a bit all over the place. Decided to look at what we had, and what was missing names. For example we had two map classes with a common base, but that base had not been mapped. 

The gist of what was done
- Mapped PlayerModelParts 
- Mapped out maps, Item#isMap and ItemMap
- partially mapped ItemArmor
- Named ItemChorusFruit
- Mapped ParticleType enum
- Named recipes RecipeCopyMap, RecipeRepair, RecipeScaleMap and shield recipes


Darkhax:
Since everyone else is sleeping, I will try to hit a bunch of random misc targets.

mappings/net/minecraft/dispenser/impl/BehaviorDispenseItem.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS none/cr net/minecraft/dispenser/impl/BehaviorDispenseItem
```
asiekierka: Why not DispenserBehaviorItem? The names, IMHO, should aid in seeing the hierarchy of inheritance.

mappings/net/minecraft/entity/Living.mapping
```diff
@@ -0,0 +1 @@
+CLASS none/sd net/minecraft/entity/Living
```
asiekierka: Are you sure on that name?

mappings/net/minecraft/entity/unsorted/Mob.mapping
```diff
@@ -0,0 +1 @@
+CLASS none/zi net/minecraft/entity/unsorted/Mob
```
asiekierka: Are you sure on that name?

mappings/net/minecraft/item/ItemArmor.mapping
```diff
@@ -6,13 +6,29 @@ CLASS none/acw net/minecraft/item/ItemArmor
...
 		FIELD f name Ljava/lang/String;
-		METHOD c getResourceItem ()Lnone/aeo;
+		FIELD h protectionAmounts [I
+		FIELD i enchantability I
+		FIELD j equipSound Lnone/nk;
+		FIELD k toughness F
+		METHOD <init> (Ljava/lang/String;ILjava/lang/String;I[IILnone/nk;F)V
+			ARG 3 name
+		METHOD a getEnchantability ()I
+		METHOD b getEquipSound ()Lnone/nk;
```
asiekierka, line 16: Comment request: getEquipSound vs getSoundEquip?

mappings/net/minecraft/particle/ParticleType.mapping
```diff
@@ -0,0 +1,70 @@
...
+		ARG 5 ignoreRange
+	METHOD a getNames ()Ljava/util/Set;
+	METHOD a getParticleById (I)Lnone/dd;
+		ARG 0 id
+	METHOD a getParticleByName (Ljava/lang/String;)Lnone/dd;
+		ARG 0 name
+	METHOD b getName ()Ljava/lang/String;
+	METHOD c getId ()I
+	METHOD d getArgumentCount ()I
+	METHOD e ignoreRange ()Z
```
asiekierka: shouldIgnoreRange?

mappings/net/minecraft/recipe/impl/ShieldRecipes.mapping
```diff
@@ -1,8 +1,8 @@
-CLASS none/agz
-	CLASS none/agz$a
+CLASS none/agz net/minecraft/recipe/impl/ShieldRecipes
```
asiekierka, line 1: I would have probably called it RecipeShield and RecipeShield.Design, respectively.

mappings/net/minecraft/reference/Achievements.mapping
```diff
@@ -12,6 +12,10 @@ CLASS none/np net/minecraft/reference/Achievements
 	FIELD K FULL_BEACON Lnone/no;
 	FIELD L EXPLORE_ALL_BIOMES Lnone/no;
 	FIELD M OVERPOWERED Lnone/no;
+	FIELD a lowestY I
```
asiekierka, line 15: lowest/highest vs. min/max?

mappings/net/minecraft/util/math/BlockPositionImpl.mapping
```diff
@@ -0,0 +1,13 @@
+CLASS none/cp net/minecraft/util/math/BlockPositionImpl
```
asiekierka, line 1: Might get confusing with BlockPos

mappings/net/minecraft/world/biome/impl/BiomeVoide.mapping
```diff
@@ -0,0 +1 @@
+CLASS none/akj net/minecraft/world/biome/impl/BiomeVoide
```
asiekierka: Voide?

mappings/net/minecraft/entity/Living.mapping
copygirl: Yeah, all entity classes should be prefixed with `Entity`, just like items, blocks and such.

mappings/net/minecraft/item/ItemArmor.mapping
copygirl, line 16: I'm 60/40 on this. `getXSound` sounds much nicer but `getSoundX` feels more proper, just like we wouldn't call blocks `DirtBlock`, but `BlockDirt`. I'd gravitate towards the latter.
( React on this comment with :+1: for `getSoundEquip`, :-1: for `getEquipSound`. )

mappings/net/minecraft/entity/player/PlayerModelParts.mapping
```diff
@@ -0,0 +1,16 @@
+CLASS none/aal net/minecraft/entity/player/PlayerModelParts
```
asiekierka: Client-side-only classes belong in client/. In this case, probably client/model/, which I had just created.

mappings/net/minecraft/entity/Living.mapping
asiekierka: That is not a class. That is an interface. Not to mention, all interfs should start with "I" in our system.

mappings/net/minecraft/dispenser/impl/BehaviorDispenseItem.mapping
copygirl: Just like to the `PistonBehavior` enum.

mappings/net/minecraft/entity/Living.mapping
copygirl: Oh, so seeing `CLASS` in the mapping files doesn't mean it's actually a class? Could be an interface?

Darkhax: This class is the interface that _all_ living entities implement. It doesn't really do anything so it's hard to give it an accurate name. Didn't know we were keeping the I prefix though, will change that.

mappings/net/minecraft/entity/unsorted/Mob.mapping
Darkhax: This one is similar to the one above. It's the interface used by _all_ living entities which are not peaceful. Will add the I prefix.

mappings/net/minecraft/recipe/impl/ShieldRecipes.mapping
Darkhax, line 1: My concern with calling the main class RecipeShield is that it's not actually a Recipe. It has an init method which adds the basic shield recipe, along with the inner recipe to the registry. The inner class is actually a recipe though.

mappings/net/minecraft/reference/Achievements.mapping
Darkhax, line 15: These values adjust every time a new achievement is added. So if the past lowestY was -25 and you put one at -300, the lowestY would become -300. min/max implies (to me) that it is a limitation, however this is just keeping track of the highest and lowest coordinates.

asiekierka:
LGTM, but we need @shadowfacts to fix the merge conflicts

PR #28, opened 2016-08-21
Basic command manager mappings

modmuss50:
I think I have done this correct. This will allow commands to be added.

### DIFF
--- a/mappings/net/minecraft/command/ICommandManager.mapping
+++ b/mappings/net/minecraft/command/ICommandManager.mapping
@@ -2,4 +2,4 @@ CLASS none/m net/minecraft/command/ICommandManager
 	METHOD a runCommand (Lnone/n;Ljava/lang/String;)I
 		ARG 0 commandSender
 		ARG 1 command
-	METHOD b getCommands ()Ljava/util/Map;
+	METHOD b getRegisteredCommands ()Ljava/util/Map;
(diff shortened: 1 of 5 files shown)


mappings/net/minecraft/command/ICommandManager.mapping
```diff
@@ -0,0 +1,2 @@
+CLASS none/m net/minecraft/command/ICommandManager
+	METHOD b getCommands ()Ljava/util/Map;
```
darkevilmac: I prefer getRegisteredCommands or getManagedCommands. That's more opinion than anything though.

modmuss50: Humm, yeah that would make more sense

mappings/net/minecraft/command/ICommandManager.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS none/m net/minecraft/command/ICommandManager
+	METHOD a runCommand (Lnone/n;Ljava/lang/String;)I
```
modmuss50: processCommand might be better thinking about it

asiekierka: executeCommand?

shadowfacts: `execute`, IMO, we already know it's a command

modmuss50:
Ok, this should be ok now

mappings/net/minecraft/reference/Commands.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS none/bc
+CLASS none/bc net/minecraft/reference/Commands
```
modmuss50: Fuck, not this

PR #30, opened 2016-08-22
GlHandler classes and whatnot

iamgreaser:
Should be ready for pulling once reviewed and hit with a truck.

This implements mappings for pretty much everything. What's missing:
- `GlHandler$1`, which appears to create an array to map `TexCoord.[STRQ]` to `[1234]` as a static field. I don't know how to integrate this, and this class only appears to be used by one function which _does_ have a mapping.
- `GlHandler.I`, which sets all fields in the colour buffer to `-1.0f`. I have no idea why you would do such a thing, and I cannot get a name for this.
- A couple of fields from the completely unused stencil buffer state tracker. As the tracker appears to be incomplete, the meanings of these fields cannot be confirmed.
- Anything else I missed.

Comments will be appreciated.

### DIFF
--- a/mappings/net/minecraft/client/GlHandler.mapping
+++ b/mappings/net/minecraft/client/GlHandler.mapping
@@ -14,7 +14,7 @@ CLASS none/bos net/minecraft/client/GlHandler
 		METHOD b disable ()V
 		METHOD a setState (Z)V
 			ARG 0 newState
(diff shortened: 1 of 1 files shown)


asiekierka:
> GreaseMonkey
> GlHandler

_scrolls directly to comment section_ LGTM

PR #33, opened 2016-08-22
More mappings

Darkhax:
This PR adds in a bunch of stuff that was missing from my last PR. Only the deletions were commited for some reason. I also map out a few more misc things like horse armor enum, and some item method.

### DIFF
--- a/mappings/net/minecraft/client/player/EntityPlayerClient.mapping
+++ b/mappings/net/minecraft/client/player/EntityPlayerClient.mapping
@@ -5,6 +5,7 @@ CLASS none/boe net/minecraft/client/player/EntityPlayerClient
 	METHOD a swingHand (Lnone/qz;)V
 		ARG 0 hand
 	METHOD c getBlockPos ()Lnone/cn;
+	METHOD c setCurrentHand (Lnone/qz;)V
 	METHOD g sendChatMessage (Ljava/lang/String;)V
 		ARG 0 message
 	METHOD m update ()V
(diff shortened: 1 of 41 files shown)


mappings/net/minecraft/client/player/EntityPlayerClient.mapping
```diff
@@ -5,6 +5,7 @@ CLASS none/boe net/minecraft/client/player/EntityPlayerClient
 	METHOD a swingHand (Lnone/qz;)V
 		ARG 0 hand
 	METHOD c getBlockPos ()Lnone/cn;
+	METHOD c setCurrentHand (Lnone/qz;)V
```
asiekierka, line 8: Are you sure it's not just a client-side hand? I remember having problems with all the "Hand" function/variable confusions with 1.9 and Charset...

mappings/net/minecraft/client/util/ClientUtils.mapping
```diff
@@ -0,0 +1,13 @@
+CLASS none/h net/minecraft/client/util/ClientUtils
+	CLASS none/h$a OperatingSystem
```
asiekierka, line 2: OperatingSystem vs. Platform?

asiekierka:
other than those two questions LGTM

mappings/net/minecraft/client/util/ClientUtils.mapping
shadowfacts: I think `OperatingSystem` is fine.

mappings/net/minecraft/client/player/EntityPlayerClient.mapping
shadowfacts: Agreed, if it's the client hand, the name should reflect that as to prevent confusion and incorrect side errors.

Darkhax: This method is inherited from [EntitySomethingBase](https://github.com/Darkhax-Forked/pomf/blob/74504345865945d8a0f2c4af4162e65e5eb6ac20/mappings/net/minecraft/sortme/EntitySomethingBase.mapping#L10).

PR #34, opened 2016-08-23
Map ChunkGenerators, some decorators, BiomeGeneratorSettings, and some missing block properties

darkevilmac:
Made some comments on the things I want a second opinion on, let me know if things are missing or should be named differently that I haven't commented on.


mappings/net/minecraft/block/Block.mapping
```diff
@@ -1,11 +1,11 @@
...
 	FIELD A stateFactory Lnone/ask;
-	FIELD a id Lnone/kp;
+	FIELD a ID Lnone/kp;
 	FIELD b creativeTab Lnone/adq;
 	FIELD c defaultState Lnone/asj;
 	FIELD d unlocalizedName Ljava/lang/String;
-	FIELD h registry Lnone/cs;
-	FIELD i blockstateList Lnone/cx;
+	FIELD h REGISTRY Lnone/cs;
+	FIELD i BLOCKSTATELIST Lnone/cx;
```
darkevilmac: I had seen these were not named the same as other static final things like the item registry.

mappings/net/minecraft/world/decorator/impl/DecoratorDeposit.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS none/awc net/minecraft/world/decorator/impl/DecoratorDeposit
```
darkevilmac, line 1: This is named DecoratorDeposit and not DecoratorOre, reason being it's used for things like gravel and dirt as well as ores.

mappings/net/minecraft/block/impl/BlockStone.mapping
```diff
@@ -1,6 +1,24 @@
...
+		FIELD c SMOOTH_GRANITE Lnone/aqc$a;
+		FIELD d DIORITE Lnone/aqc$a;
+		FIELD e SMOOTH_DIORITE Lnone/aqc$a;
+		FIELD f ANDESITE Lnone/aqc$a;
+		FIELD g SMOOTH_ANDESITE Lnone/aqc$a;
+		FIELD h TYPES [Lnone/aqc$a;
+		FIELD i id I
+		FIELD j color Ljava/lang/String;
+		FIELD k name Ljava/lang/String;
+		FIELD l translationKey Lnone/ayp;
```
darkevilmac: Haven't looked at a lot of the localization mappings I believe this is correct though.

mappings/net/minecraft/world/biome/BiomeGenerator.mapping
```diff
@@ -1 +1,13 @@
 CLASS none/ajk net/minecraft/world/biome/BiomeGenerator
+	FIELD a isDecorating Z
```
darkevilmac: decorating or isDecorating?

mappings/net/minecraft/block/Block.mapping
copygirl: Should be `BLOCKSTATE_LIST` then. Not sure if all static final things should be written like that or only fields that are immutable.

mappings/net/minecraft/world/biome/BiomeGenerator.mapping
copygirl: As per my suggestion / addition to [the guidelines](https://docs.google.com/document/d/15fHL-WgK0uMPAy-WJbQtxrfOVLJNlfasjsnt_wruOXA/edit), it should be `decorating`.

mappings/net/minecraft/block/Block.mapping
```diff
@@ -1,11 +1,11 @@
 CLASS none/akw net/minecraft/block/Block
 	FIELD A stateFactory Lnone/ask;
-	FIELD a id Lnone/kp;
+	FIELD a ID Lnone/kp;
```
copygirl: Only used for `Block.AIR`, so it should actually be `AIR_ID`?

mappings/net/minecraft/world/biome/BiomeGeneratorSettings.mapping
```diff
@@ -0,0 +1,158 @@
...
+		FIELD C waterLakeChance I
+		FIELD D useLavaLakes Z
+		FIELD E lavaLakeChance I
+		FIELD F useLavaOceans Z
+		FIELD G fixedBiome I
+		FIELD H biomeSize I
+		FIELD I riverSize I
+		FIELD J dirtSize I
+		FIELD K dirtCount I
+		FIELD L dirtMinHeight I
```
copygirl, line 14: Height? Is that the vertical size or the Y position it spawns in the world? If the latter, maybe it should be `Y` or `YLevel` instead?

mappings/net/minecraft/world/biome/BiomeGenerator.mapping
iamgreaser: There's a guideline doc?

mappings/net/minecraft/world/biome/BiomeGeneratorSettings.mapping
darkevilmac: This stuff here is actually specified with raw strings when it's serialized so unless there's a typo all these fields are correct.

mappings/net/minecraft/world/biome/BiomeGenerator.mapping
asiekierka: There's anything actually on the guideline doc?

mappings/net/minecraft/block/Block.mapping
asiekierka: DEFAULT_ID, really, as that's the intention. or EMPTY_ID?

mappings/net/minecraft/world/biome/BiomeGeneratorSettings.mapping
asiekierka: We don't really... particularly concern ourselves with how Mojang names things. Mojang's names are usually decent, but if you can name something better, by absolutely all means do so.

darkevilmac: As it currently is, I still think minHeight and maxHeight makes more sense and is at least as easy to understand as minYLevel and maxYLevel.

mappings/net/minecraft/block/Block.mapping
```diff
@@ -1,11 +1,11 @@
 CLASS none/akw net/minecraft/block/Block
 	FIELD A stateFactory Lnone/ask;
-	FIELD a ID Lnone/kp;
+	FIELD a DEFAULT_ID Lnone/kp;
 	FIELD b creativeTab Lnone/adq;
 	FIELD c defaultState Lnone/asj;
 	FIELD d unlocalizedName Ljava/lang/String;
 	FIELD h REGISTRY Lnone/cs;
-	FIELD i BLOCKSTATELIST Lnone/cx;
+	FIELD i BLOCKSTATE_IDLIST Lnone/cx;
```
darkevilmac: The field is an IDLIST, so I named it accordingly. Instead of just BLOCKSTATE_LIST.

mappings/net/minecraft/world/biome/BiomeGenerator.mapping
```diff
@@ -1,5 +1,5 @@
 CLASS none/ajk net/minecraft/world/biome/BiomeGenerator
-	FIELD a isDecorating Z
+	FIELD a decorating Z
```
darkevilmac, line 2: I don't really like this all that much myself, I think it should be prefixed with 'is' or 'can' when there's no getters or setters and it's a public field. If there are getters and setters then I think the guidelines make sense.

Thoughts?

mappings/net/minecraft/block/Block.mapping
copygirl: Then it should be `BLOCKSTATE_ID_LIST`, since if it were a non-static field you'd name it `blockstateIdList`.

mappings/net/minecraft/world/biome/BiomeGenerator.mapping
copygirl: Consistensy is more important to me personally. Don't want to end up with some fields named like this and others like that, same with methods.

mappings/net/minecraft/world/biome/BiomeGeneratorSettings.mapping
copygirl: Height is often used to describe the size of something. What about altitude?

mappings/net/minecraft/block/Block.mapping
```diff
@@ -1,11 +1,11 @@
...
 	FIELD A stateFactory Lnone/ask;
-	FIELD a id Lnone/kp;
+	FIELD a DEFAULT_ID Lnone/kp;
 	FIELD b creativeTab Lnone/adq;
 	FIELD c defaultState Lnone/asj;
 	FIELD d unlocalizedName Ljava/lang/String;
-	FIELD h registry Lnone/cs;
-	FIELD i blockstateList Lnone/cx;
+	FIELD h REGISTRY Lnone/cs;
+	FIELD i BLOCKSTATE_IDLIST Lnone/cx;
```
asiekierka: BLOCKSTATE_ID_LIST

mappings/net/minecraft/block/impl/BlockStone.mapping
asiekierka: Why is your translationKey a color?

asiekierka:
Other than the two marked issues LGTM

mappings/net/minecraft/world/biome/BiomeGeneratorSettings.mapping
shadowfacts: I think height is fine, it describes the height something is at not the height of the thing itself.

shadowfacts:
Aside from what asie mentioned, LGTM.

mappings/net/minecraft/world/biome/BiomeGenerator.mapping
darkevilmac: I'd like more discussion on this if possible it just doesn't really seem right to me.

darkevilmac:
I think that's everything.

PR #35, opened 2016-08-24
Expand World/Save mappings

Darkhax:
This PR expands the mappings for world stuff and saving. Not sure if this is exactly what was asked, but it should give a decent starting point.

### DIFF
--- a/mappings/net/minecraft/world/World.mapping
+++ b/mappings/net/minecraft/world/World.mapping
@@ -1,14 +1,20 @@
 CLASS none/aiu net/minecraft/world/World
+	FIELD B lootTableHandler Lnone/bbj;
 	FIELD C profiler Lnone/os;
-	FIELD E isRemote Z
+	FIELD E isClient Z
+	FIELD L calendar Ljava/util/Calendar;
 	FIELD N border Lnone/atd;
+	FIELD a seaLevel I
 	FIELD i playerList Ljava/util/List;
 	FIELD r rand Ljava/util/Random;
 	FIELD u listeners Ljava/util/List;
 	FIELD v chunkProvider Lnone/ath;
 	FIELD x info Lnone/bar;
+	METHOD <init> (Lnone/bas;Lnone/bar;Lnone/auc;Lnone/os;Z)V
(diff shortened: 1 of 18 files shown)


mappings/net/minecraft/world/WorldProperties.mapping
```diff
@@ -0,0 +1,156 @@
...
+	METHOD l getLastPlayed ()J
+	METHOD m isThundering ()Z
+	METHOD n getThunderTime ()I
+	METHOD o isRaining ()Z
+	METHOD p getRainTime ()I
+	METHOD q getGameMode ()Lnone/ais;
+	METHOD r getMapFeatures ()Z
+	METHOD s isHardcore ()Z
+	METHOD t getWorldGenerator ()Lnone/aiz;
+	METHOD u commandsAllowed ()Z
```
asiekierka: areCommandsAllowed

mappings/none/bao.mapping
```diff
@@ -1,3 +1,66 @@
...
+	METHOD l getLastPlayed ()J
+	METHOD m isThundering ()Z
+	METHOD n getThunderTime ()I
+	METHOD o isRaining ()Z
+	METHOD p getRainTime ()I
+	METHOD q getGameMode ()Lnone/ais;
+	METHOD r getMapFeatures ()Z
+	METHOD s isHardcore ()Z
+	METHOD t getWorldGenerator ()Lnone/aiz;
+	METHOD u commandsAllowed ()Z
```
asiekierka: areCommandsAllowed

mappings/net/minecraft/world/World.mapping
```diff
@@ -1,14 +1,25 @@
 CLASS none/aiu net/minecraft/world/World
+	FIELD B lootTableHandler Lnone/bbj;
 	FIELD C profiler Lnone/os;
-	FIELD E isRemote Z
+	FIELD E isClient Z
```
asiekierka: isRemote is more correct (as to the intention rather than the practice?), IMHO. You could theoretically have a remote world which is not just an instance of WorldClient.

PR #36, opened 2016-08-24
Fix some namings as well as add some names to MinecraftServer and related classes.

darkevilmac:
Most of this should be alright, only thing I'd like feedback on is "currentServerJob", it's changed based on what the server is doing in relation to loading worlds for the most part. Only three calls I can find and these are the jobs: "menu.generatingTerrain", "menu.loadingLevel", and "menu.convertingLevel".

### DIFF
--- a/mappings/net/minecraft/server/MinecraftServer.mapping
+++ b/mappings/net/minecraft/server/MinecraftServer.mapping
@@ -8,9 +8,11 @@ CLASS net/minecraft/server/MinecraftServer net/minecraft/server/MinecraftServer
 	FIELD H keyPair Ljava/security/KeyPair;
 	FIELD L demo Z
 	FIELD M bonusChest Z
+	FIELD R currentTaskName Ljava/lang/String;
 	FIELD T forceGameMode Z
 	FIELD Z serverThread Ljava/lang/Thread;
-	FIELD a userCacheFile Ljava/io/File;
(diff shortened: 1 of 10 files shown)


mappings/net/minecraft/server/MinecraftServer.mapping
```diff
@@ -8,9 +8,10 @@ CLASS net/minecraft/server/MinecraftServer net/minecraft/server/MinecraftServer
 	FIELD H keyPair Ljava/security/KeyPair;
 	FIELD L demo Z
 	FIELD M bonusChest Z
+	FIELD R currentServerJob Ljava/lang/String;
```
asiekierka: currentTaskName? Don't forget to change the getters/setters. setCurrentTaskName too vs setTaskName

darkevilmac:
PR updated. Changed a few things.

mappings/net/minecraft/sortme/IMultiThreadedObject.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS none/qp net/minecraft/sortme/IGameThread
+CLASS none/qp net/minecraft/sortme/IMultiThreadedObject
```
darkevilmac, line 1: For reference this is an interface that lets you check if you're on the main thread as well as submit tasks to the main thread.

PR #42, opened 2016-08-25
Map nd -> ConsoleCommandSender

maxpowa:
`net.minecraft.server.dedicated.ConsoleCommandSender`

Not really confident that `net.minecraft.server.dedicated` is the best package for this, but it's all I could find that really worked.

Maps two things from ICommandSender
`y_` -> `shouldOutput`
`a(Lnone/fa;)` -> `append(Lnone/fa;)`

### DIFF
--- a/mappings/net/minecraft/block/tile/impl/TileEntitySign.mapping
+++ b/mappings/net/minecraft/block/tile/impl/TileEntitySign.mapping
@@ -1,6 +1,6 @@
 CLASS none/art net/minecraft/block/tile/impl/TileEntitySign
 	CLASS none/art$1
-		METHOD a append (Lnone/fa;)V
+		METHOD a appendCommandFeedback (Lnone/fa;)V
 			ARG 0 component
 		METHOD c getBlockPos ()Lnone/cn;
 		METHOD d getPos ()Lnone/bcs;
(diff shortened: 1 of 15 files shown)


asiekierka:
I'm not entirely sure if the names are good. Thoughts?

shadowfacts:
Is it used for the console or just rcon?

maxpowa:
Console & Rcon, I'm open to name suggestions. I was questioning the package name in #fabric but nobody was around to provide input...

maxpowa:
I'm on the fence about `append` now. `append` makes sense when something is only an ICommandSender impl, but when it's an ICommandSender impl and a bunch of other things it loses its meaning. Not sure what the best name is, but it definitely just appends the text component to the command buffer.

asiekierka:
appendToCommandBuffer. We don't want conflicts with obvious names...

maxpowa:
Actually wait, I think `appendToSenderBuffer` makes more sense than `appendToCommandBuffer`, because it's not actually appending to the command message itself, but the buffer which holds the responses. Going to change `clear` to `clearSenderBuffer` as well.

asiekierka:
Sure

maxpowa:
There we go, should be good now.

asiekierka:
What is shouldOutput, by the way? I feel it might be something along the lines of shouldEchoSenderBuffer, or something.

PR #45, opened 2016-08-25
Finished off Item's mappings. Also mapped some itemblock stuff. 

Darkhax:
This PR finishes off all mappings in Item

### DIFF
--- a/mappings/net/minecraft/item/Item.mapping
+++ b/mappings/net/minecraft/item/Item.mapping
@@ -1,4 +1,25 @@
 CLASS none/aeo net/minecraft/item/Item
+	CLASS none/aeo$1
+		METHOD a getProperty (Lnone/aeq;Lnone/aiu;Lnone/sn;)F
+			ARG 0 stack
+			ARG 1 world
+			ARG 2 user
+	CLASS none/aeo$2
+		METHOD a apply (Lnone/aeq;)Ljava/lang/String;
+	CLASS none/aeo$3
+		METHOD a apply (Lnone/aeq;)Ljava/lang/String;
+	CLASS none/aeo$4
+		METHOD a apply (Lnone/aeq;)Ljava/lang/String;
+	CLASS none/aeo$5
+		METHOD a apply (Lnone/aeq;)Ljava/lang/String;
+	CLASS none/aeo$6
+		METHOD a apply (Lnone/aeq;)Ljava/lang/String;
+	CLASS none/aeo$7
+		METHOD a apply (Lnone/aeq;)Ljava/lang/String;
+	CLASS none/aeo$8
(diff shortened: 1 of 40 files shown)


mappings/net/minecraft/item/Item.mapping
```diff
@@ -10,24 +39,81 @@ CLASS none/aeo net/minecraft/item/Item
...
+		METHOD a apply (Lnone/aeq;)Ljava/lang/String;
+			ARG 0 stack
+	CLASS none/aeo$20
+		METHOD a apply (Lnone/aeq;)Ljava/lang/String;
+			ARG 0 stack
+	CLASS none/aeo$21
+		METHOD a apply (Lnone/aeq;)Ljava/lang/String;
+			ARG 0 stack
 	FIELD a BLOCK_ITEM_MAP Ljava/util/Map;
+	FIELD b GETTER_DAMAGED Lnone/aer;
```
shadowfacts, line 97: `GETTER_X` or `X_GETTER`? I personally prefer `X_GETTER`

mappings/net/minecraft/item/Item.mapping
```diff
@@ -98,24 +186,32 @@ CLASS none/aeo net/minecraft/item/Item
...
 		ARG 0 block
 		ARG 1 itemBlock
 	METHOD a canHarvestBlock (Lnone/asj;)Z
 		ARG 0 state
-	METHOD a (Lnone/dt;)Z
+	METHOD a readTag (Lnone/dt;)Z
+		ARG 0 tag
+	METHOD a getProperty (Lnone/kp;)Lnone/aer;
+		ARG 0 id
+	METHOD a addProperty (Lnone/kp;Lnone/aer;)V
```
shadowfacts, line 204: Shouldn't this be `setProperty`?

shadowfacts:
Other than that, LGTM

mappings/net/minecraft/item/Item.mapping
darkevilmac: Prefixing with getter makes things a bit easier as far as usage in an ide goes.

mappings/net/minecraft/item/Item.mapping
Darkhax: I went with addProperty, because the method adds it to a list rather than setting a field.

shadowfacts: Aha, that's good then

asiekierka:
LGTM

mappings/net/minecraft/item/Item.mapping
Darkhax: I went with `GETTER_X` for the reason mentioned above. That has been our naming scheme for a lot of things, like `MODIFIER_X` and `ItemX`

PR #49, opened 2016-08-26
Mapped some more sound stuff

Darkhax:
Mapped out a few positioned audio things.


mappings/net/minecraft/client/audio/Sound.mapping
```diff
@@ -1 +1,4 @@
 CLASS none/cab net/minecraft/client/audio/Sound
+	METHOD e getWeight ()I
+	METHOD f getCopy ()Lnone/cab;
```
asiekierka, line 3: getCopy()? Not getClone() or something?

Darkhax: Both words are interchangeable so I went with the shorter one. Clone is also used by Java, so it may cause confusion.

asiekierka:
LGTM

shadowfacts:
LGTM

mappings/net/minecraft/client/audio/Sound.mapping
copygirl: Does this literally `get` some preexisting copy/clone or does it create one when the method is called? If the latter, it shouldn't be prefixed with `get` imo.

PR #52, opened 2016-08-28
[WIP] World related mappings.

darkevilmac:
Please feel free to leave feedback on any of this, I'd like to map a bit more and confirm things are consistently being named across a few classes before this could be merged.

Most of these mappings should be correct. The only mapping I'm not 100% sure on is the ChunkBatcher, this mapping is simply based on the initialization of it's thread factory and logger. If more research is to be done on the class prior to just defaulting to the Mojang name I'll remove the mapping of the class name.


mappings/net/minecraft/world/WorldProperties.mapping
```diff
@@ -123,8 +123,8 @@ CLASS none/bar net/minecraft/world/WorldProperties
 		ARG 0 damage
 	METHOD f setThunderTime (I)V
 		ARG 0 thunderTime
-	METHOD f setMapFeatures (Z)V
-		ARG 0 mapFeatures
+	METHOD f setStructures (Z)V
```
copygirl, line 126: `setStructuresEnabled` (and `hasStructures` to `getStructuresEnabled`)?

asiekierka: areStructuresEnabled, but sounds nice.

mappings/net/minecraft/world/World.mapping
```diff
@@ -61,6 +61,7 @@ CLASS none/aiu net/minecraft/world/World
 		ARG 3 category
 		ARG 4 volume
 		ARG 5 pitch
+	METHOD a initializeWorld (Lnone/aix;)V
```
asiekierka: Why not just initialize or init?

mappings/net/minecraft/world/chunk/ChunkBatcher.mapping
```diff
@@ -0,0 +1,4 @@
+CLASS none/brk net/minecraft/world/chunk/ChunkBatcher
```
asiekierka: Be more careful! This is a client-side class involved in rendering!

mappings/net/minecraft/world/WorldProperties.mapping
copygirl: If it's a getter/setter pair, shouldn't it be mirrored?

mappings/net/minecraft/world/chunk/ChunkBatcher.mapping
darkevilmac: The main reason I had marked this as WIP was to get feedback about this class because I wasn't 100% sure of it's function. Not sure how I missed that Sided annotation.

asiekierka: If unsure, just get rid of the mapping for now. I've mapped some rendering stuff just now. Might help.

mappings/net/minecraft/world/WorldProperties.mapping
darkevilmac: I had gone over this one in IRC before mapping, should it be left as is or changed?

PR #56, opened 2016-08-28
Mappings around world storage

magik6k:


asiekierka:
Will fix up with author, don't worry about me!

PR #60, opened 2016-08-31
LootTable, and KeyBinding class name.

darkevilmac:


mappings/net/minecraft/reference/Sounds.mapping
```diff
@@ -1,30 +1,4 @@
 CLASS none/nl net/minecraft/reference/Sounds
```
modmuss50, line 1: This file is broken, I had the same issue, @shadowfacts can you take a look?

mappings/net/minecraft/block/tile/TileEntityLootTable.mapping
```diff
@@ -0,0 +1,9 @@
+CLASS none/ars net/minecraft/block/tile/TileEntityLootTable
```
copygirl: This isn't a tile entity class, right? It should be named differently to not confuse anyone. Additionally, since the class is already a "loot table", maybe the methods don't all need the `LootTable` suffix?

mappings/net/minecraft/sortme/ILootTable.mapping
```diff
@@ -0,0 +1,2 @@
+CLASS none/rh net/minecraft/sortme/ILootTable
```
copygirl: `IHasLootTable`? The interface doesn't implement a loot table but signalizes that the implementation _provides_ one... maybe even `ILootTableProvider`? I prefer the former though. This should be decided upon and put into the guidelines.

asiekierka: I prefer xProvider to Hasx

mappings/net/minecraft/block/tile/TileEntityLootTable.mapping
asiekierka: TileEntityWithLootTable

copygirl: If it is a tile entity class, maybe `TileEntityLootBase`?

darkevilmac:
Gonna map something in sounds, dont merge just yet.

darkevilmac:
Alright I'm done, feel free to leave feedback.

asiekierka:
Merging because it's SNAPSHOT TIME!

PR #62, opened 2016-09-04
Map Explosion related things

copygirl:


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -53,6 +53,17 @@ CLASS none/se net/minecraft/entity/Entity
 		ARG 1 commandName
 	METHOD a setUuid (Ljava/util/UUID;)V
 		ARG 0 uuid
+	METHOD a getExplosionResistanceTowards (Lnone/aip;Lnone/aiu;Lnone/cn;Lnone/asj;)F
```
copygirl: This and `canExplosionDestroyBlock` are some weird methods on the base entity class that only seem to be for anything that can cause explosions.

The name for this one in particular is probably not quite correct. Looking for suggestions. Essentially this method is called for each block that would be affected and allows modifying its effective explosion resistance (for example, TNT minecarts could increase the resistance of rails).

mappings/net/minecraft/sortme/Explosion.mapping
```diff
@@ -1 +1,46 @@
...
+		ARG 0 world
+		ARG 1 entity
+		ARG 2 x
+		ARG 3 y
+		ARG 4 z
+		ARG 5 power
+		ARG 6 createFire
+		ARG 7 destroyBlocks
+		ARG 8 affectedBlocks
+	METHOD a collectBlocksAndDamageEntities ()V
```
copygirl, line 40: This is one weird name but it's basically what this function does. It collects all blocks that would be destroyed and then damages and pushes around entities. `affectWorld` then actually destroys these blocks, creates particles and fire.

shadowfacts:
Aside from your request for suggestions (not sure about what would be good for that), LGTM

mappings/net/minecraft/block/impl/BlockTNT.mapping
```diff
@@ -1,6 +1,10 @@
 CLASS none/aqk net/minecraft/block/impl/BlockTNT
 	FIELD a EXPLODE Lnone/asv;
 	METHOD a deserializeState (I)Lnone/asj;
+	METHOD a doesDropItemsOnExplosion (Lnone/aip;)Z
```
asiekierka: shouldDropItemsOnExplosion would be slightly more in tune with the set standard

copygirl:
`getExplosionResistanceTowards` => `getEffectiveExplosionResistance`
`doesDropItemsOnExplosion`=> `shouldDropItemsOnExplosion`

Still couldn't think of a better name for `collectBlocksAndDamageEntities`.

asiekierka:
Anyone got any other issues?

PR #63, opened 2016-09-04
Move BlockState related stuff into its own package

copygirl:
I thought this would make sense, so I thought I'd make a PR for it in case everyone else thinks so as well. Otherwise I won't mind if you reject it.


shadowfacts:
LGTM

asiekierka:
LGTM

PR #72, opened 2016-09-13
Fixed RCON mappings

magik6k:


asiekierka:
Let's Go To Mars

mappings/net/minecraft/server/rcon/BufferHelper.mapping
```diff
@@ -0,0 +1,17 @@
+CLASS none/nc net/minecraft/server/rcon/BufferHelper
+	FIELD a HEX_CHARS [C
```
copygirl: `HEX_CHAR_LOOKUP`? Because name things after what they're used for, not what they are?
(I don't mean to talk down or anything. Sorry if I do make that impression.)

mappings/net/minecraft/util/math/MathUtils.mapping
```diff
@@ -50,17 +64,25 @@ CLASS none/or net/minecraft/util/math/MathUtils
 		ARG 0 d
 	METHOD c sqrt (F)F
 		ARG 0 f
+	METHOD c hsvToRgb (FFF)I
+		ARG 0 h
```
copygirl: Imo these should be expanded to the full names, along with other parameters in here: `hue`, `saturation`, `value`, and also `value` instead of `f`, `d`, `i` and such.

mappings/net/minecraft/server/rcon/RconBase.mapping
```diff
@@ -0,0 +1,34 @@
...
+	FIELD f sockets Ljava/util/List;
+	FIELD g serverSockets Ljava/util/List;
+	METHOD <init> (Lnone/la;Ljava/lang/String;)V
+		ARG 0 server
+		ARG 1 description
+	METHOD a start ()V
+	METHOD a log (Ljava/lang/String;)V
+		ARG 0 msg
+	METHOD a registerSocket (Ljava/net/DatagramSocket;)V
+		ARG 0 secket
```
maxpowa: secket

mappings/net/minecraft/util/math/MathUtils.mapping
asiekierka: Yes.

asiekierka:
Let's Not Go To Mars yet

magik6k:
Can We Go To Mars Now?

PR #73, opened 2016-09-13
Add mapping for Rotation and improve armor stand

maxpowa:
I'm also pretty sure at this point that `Entity.setHeadPitch` is named incorrectly. As such, getHeadPitch is also probably incorrect.


asiekierka:
LGTM

shadowfacts:
LGTM

PR #74, opened 2016-09-13
Handlers/Filters for handshake

maxpowa:
Not really sure if Filter or Handler is better. Also not sure where the common interface should go.

### DIFF
--- a/mappings/net/minecraft/server/DummyFilter.mapping
+++ b/mappings/net/minecraft/server/DummyFilter.mapping
@@ -1,0 +1,10 @@
+CLASS none/me net/minecraft/server/DummyFilter
+	FIELD a server Lnet/minecraft/server/MinecraftServer;
+	FIELD b connection Lnone/eq;
+	METHOD <init> (Lnet/minecraft/server/MinecraftServer;Lnone/eq;)V
+		ARG 0 server
+		ARG 1 connection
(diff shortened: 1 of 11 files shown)


asiekierka:
LGTM

mappings/net/minecraft/server/DummyFilter.mapping
```diff
@@ -0,0 +1,10 @@
+CLASS none/me net/minecraft/server/DummyFilter
```
asiekierka: Just Filter? Not ConnectionFilter?

maxpowa: I can change it, should I change the interface to IConnectionFilter as well then?

asiekierka (approved):

mappings/net/minecraft/sortme/IConnectionFilter.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS none/jp net/minecraft/sortme/IConnectionFilter
```
shadowfacts: Maybe put this in the `network` package?

maxpowa: Done

shadowfacts (approved):

PR #80, opened 2016-09-15
Scrollable entry lists + LAN server querying

iamgreaser:
The latter lead to the former. The former is more important, but it would be a waste of time to revert everything I did on the latter. So yeah, you get two things for the price of one.

Currently need names for these things:
- `bfb`: Abstract base class for all vertically-scrollable lists w/ entries of uniform height.
- `bew` extends `bfb`: Less abstract base class for lists where each entry can have its own draw and mouse functions.
  - `bew.a`: Interface for each entry in `bew`.
- `bid` extends `bew.a`: The `o o O` spoopy entry which tells you that it's scanning for LAN servers.
- `bie` extends `bew.a`: Visible entry for a LAN server.
- `bif` extends `bew.a`: Visible entry for a regular serverlist server.
- `big` extends `bew`: The server selection scrollbox itself.

### DIFF
--- a/mappings/net/minecraft/client/gui/menu/ListEntryRemoteServer.mapping
+++ b/mappings/net/minecraft/client/gui/menu/ListEntryRemoteServer.mapping
@@ -1,6 +1,6 @@
 CLASS none/bif net/minecraft/client/gui/menu/ListEntryRemoteServer
 	FIELD a LOGGER Lorg/apache/logging/log4j/Logger;
-	FIELD b SERVER_PINGER Ljava/util/concurrent/ThreadPoolExecutor;
+	FIELD b PING_THREAD_POOL Ljava/util/concurrent/ThreadPoolExecutor;
 	FIELD c UNKNOWN_SERVER Lnone/kp;
 	FIELD d SERVER_SELECTION Lnone/kp;
 	FIELD e guiMultiplayer Lnone/bic;
--- a/mappings/net/minecraft/client/gui/menu/ListEntryRemoteServer.mapping
+++ b/mappings/net/minecraft/client/gui/menu/ListEntryRemoteServer.mapping
@@ -1,0 +1,31 @@
+CLASS none/bif net/minecraft/client/gui/menu/ListEntryRemoteServer
+	FIELD a LOGGER Lorg/apache/logging/log4j/Logger;
+	FIELD b PING_THREAD_POOL Ljava/util/concurrent/ThreadPoolExecutor;
+	FIELD c UNKNOWN_SERVER Lnone/kp;
+	FIELD d SERVER_SELECTION Lnone/kp;
+	FIELD e guiMultiplayer Lnone/bic;
+	FIELD f mc Lnone/bdq;
+	FIELD g serverEntry Lnone/bmj;
+	METHOD a (III)V
+		ARG 0 index
+		ARG 1 x
+		ARG 2 y
+	METHOD a handleMouseClick (IIIIII)Z
+		ARG 0 index
+		ARG 1 mouseX
+		ARG 2 mouseY
+		ARG 3 mouseButton
+		ARG 4 relMouseX
+		ARG 5 relMouseY
+	METHOD a drawEntry (IIIIIIIZ)V
+		ARG 0 index
+		ARG 1 x
+		ARG 2 y
+		ARG 3 width
+		ARG 4 height
+		ARG 5 mouseX
+		ARG 6 mouseY
+		ARG 7 mouseIsOver
+	METHOD a drawIcon (IILnone/kp;)V
+	METHOD b (IIIIII)V
+		ARG 0 index
--- a/mappings/net/minecraft/client/gui/menu/ListEntryRemoteServer.mapping
+++ b/mappings/net/minecraft/client/gui/menu/ListEntryRemoteServer.mapping
@@ -1,4 +1,4 @@
-CLASS none/bif
+CLASS none/bif net/minecraft/client/gui/menu/ListEntryRemoteServer
 	FIELD a LOGGER Lorg/apache/logging/log4j/Logger;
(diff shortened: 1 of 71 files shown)


asiekierka:
- bfb: Abstract base class for all vertically-scrollable lists w/ entries of uniform height.

net.minecraft.client.gui.widget.WidgetListBase (or WidgetListAbstract?)
- bew extends bfb: Less abstract base class for lists where each entry can have its own draw and mouse functions.

net.minecraft.client.gui.widget.WidgetList
- bew.a: Interface for each entry in bew.

net.minecraft.client.gui.widget.WidgetList.Entry
- bid extends bew.a: The o o O spoopy entry which tells you that it's scanning for LAN servers.

net.minecraft.client.gui.menu.ListEntryLocalScanProgress
- bie extends bew.a: Visible entry for a LAN server.

net.minecraft.client.gui.menu.ListEntryLocalServer
- bif extends bew.a: Visible entry for a regular serverlist server.

net.minecraft.client.gui.menu.ListEntryRemoteServer
- big extends bew: The server selection scrollbox itself.

net.minecraft.client.gui.menu.WidgetServerList

iamgreaser:
Only real problem with going with WidgetList is that it's incompatible with the other things that call themselves Widget\* (that is, these classes are not `Drawable`s). If we _do_ take this approach, I'd put it in `net.minecraft.client.gui.scroll` or something like that.

`bew` is also an abstract class, and there are also classes which depend on `bfb` but not `bew`.

asiekierka:
@iamgreaser Yes, but the Base one is more abstract. All you need to implement in bew is the entry getter, which is fair.

As for the Widget thing, it might not be, but it has roughly the same methods involved and I can't think of a better way to call it. Perhaps @copygirl?

copygirl:
These classes are only used in the menu, right? What about `MenuWidget`? It's also worth considering renaming our current `Widget` classes to something else.

mappings/net/minecraft/client/gui/menu/WidgetListBase.mapping
```diff
@@ -0,0 +1,79 @@
...
+		ARG 1 x
+		ARG 2 y
+		ARG 3 height
+		ARG 4 mouseX
+		ARG 5 mouseY
+	METHOD a (IZII)V
+		ARG 2 mouseX
+		ARG 3 mouseY
+	METHOD a handleButtonPressed (Lnone/bek;)V
+	METHOD b entryCount ()I
```
shadowfacts: `entryCount` or `getEntryCount`?

mappings/net/minecraft/client/gui/menu/WidgetListBase.mapping
```diff
@@ -0,0 +1,79 @@
...
+	METHOD b drawEntries (IIII)V
+		ARG 0 x
+		ARG 1 y
+		ARG 2 mouseX
+		ARG 3 mouseY
+	METHOD c entryWidth ()I
```
shadowfacts: `entryWidth` or `getEntryWidth`?

mappings/net/minecraft/client/gui/menu/WidgetListBase.mapping
```diff
@@ -0,0 +1,79 @@
...
+		ARG 0 up
+		ARG 1 down
+	METHOD d (Z)V
+	METHOD g mouseInRange (I)Z
+		ARG 0 y
+	METHOD h scrollByY (I)V
+		ARG 0 dy
+	METHOD i setLeftPos (I)V
+		ARG 0 x1
+	METHOD k contentHeight ()I
```
shadowfacts: `contentHeight` or `getContentHeight`?

mappings/net/minecraft/server/LanServerPinger.mapping
```diff
@@ -1,7 +1,19 @@
...
+	FIELD c motd Ljava/lang/String;
 	FIELD d socket Ljava/net/DatagramSocket;
 	FIELD e isRunning Z
+	FIELD f addressPort Ljava/lang/String;
+	METHOD <init> (Ljava/lang/String;Ljava/lang/String;)V
+		ARG 0 motd
+		ARG 1 addressPort
+	METHOD a parseAnnouncementMotd (Ljava/lang/String;)Ljava/lang/String;
+		ARG 0 announcement
+	METHOD a makeAnnouncement (Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String;
```
shadowfacts: `make` or `create`?

mappings/net/minecraft/client/gui/menu/ListEntryRemoteServer.mapping
```diff
@@ -0,0 +1,31 @@
+CLASS none/bif net/minecraft/client/gui/menu/ListEntryRemoteServer
+	FIELD a LOGGER Lorg/apache/logging/log4j/Logger;
+	FIELD b SERVER_PINGER Ljava/util/concurrent/ThreadPoolExecutor;
```
shadowfacts: `PING_THREAD_POOL` or `PING_POOL` or something to indicate it's a thread pool?

shadowfacts (changes requested):

mappings/net/minecraft/client/gui/menu/WidgetListBase.mapping
copygirl: `get` for sure, same with the other 2.

mappings/net/minecraft/server/LanServerPinger.mapping
iamgreaser: ~~Or `construct` perhaps?~~ EDIT: actually looks and sounds bad. Going with `create`.

shadowfacts (approved):

copygirl:
Haha, my first merge! :tada:

PR #83, opened 2016-09-18
Map Level related things and a bunch more (WIP?)

copygirl:
This renames some stuff related to levels (collections of worlds) appropriately. This is just a start. I'm not sure what else is missing.


mappings/net/minecraft/world/GameMode.mapping
```diff
@@ -21,4 +21,6 @@ CLASS none/ait net/minecraft/world/GameMode
 	METHOD a setAbilitites (Lnone/aaj;)V
 		ARG 0 abilities
 	METHOD b getName ()Ljava/lang/String;
+	METHOD c isAdventureOrSpectator ()Z
```
asiekierka, line 24: I refrained from mapping those methods as I believe they have a more special context.

mappings/net/minecraft/world/WorldType.mapping
```diff
@@ -1,19 +1,20 @@
...
 		ARG 3 providerClass
 	METHOD a getID ()I
-	METHOD a getDimensionByID (I)Lnone/auf;
+	METHOD a getById (I)Lnone/auf;
 		ARG 0 id
 	METHOD b getName ()Ljava/lang/String;
-	METHOD c getWorldName ()Ljava/lang/String;
-	METHOD d constructDimension ()Lnone/aue;
+	METHOD c getSuffix ()Ljava/lang/String;
+	METHOD d construct ()Lnone/aue;
```
asiekierka: Not create?

mappings/net/minecraft/world/chunk/Chunk.mapping
```diff
@@ -18,7 +20,7 @@ CLASS none/ato net/minecraft/world/chunk/Chunk
 		ARG 2 z
 	METHOD <init> (Lnone/aiv;Lnone/auo;II)V
 		ARG 0 world
-	METHOD a matches (II)Z
+	METHOD a matchesPos (II)Z
```
asiekierka, line 23: ~~Perhaps contains?~~ I'm a derp.

asiekierka (changes requested):

mappings/net/minecraft/world/WorldType.mapping
copygirl: It creates a new instance of the `WorldProvider` assigned to that `WorldType` through reflection. What is that usually called? Should it just be `createProvider` ?

mappings/net/minecraft/world/GameMode.mapping
copygirl: I was thinking of leaving them, but I believe mapping them like this might be better since it'll help mapping other things easier. Would be nice if we could mark something as "might need a better name" (yellow) and "request changes" (orange) ..?
**edit:** Created an issue: FabricMC/Enigma#27

mappings/net/minecraft/world/chunk/Chunk.mapping
copygirl: It's a `ChunkPos`.

**edit:** ... oh yeah, in that case, that should definitely be changed.
`matches(chunkX, chunkY)`, `matchesChunkPos(x, y)` or any better idea?

mappings/net/minecraft/world/GameMode.mapping
asiekierka: That might get confusing. Also, it would break the format.

My policy is to not map names unless I'm sure... and you can always poke me to look at something if you need it.

mappings/net/minecraft/world/WorldType.mapping
asiekierka: I think it should be just createProvider, yes.

mappings/net/minecraft/world/chunk/Chunk.mapping
asiekierka: matches was my original idea.

copygirl: I wanted to avoid anyone thinking anything beyond the position is checked to match, which is what it may look like when you first see the method being used (without looking at its code).

asiekierka: I suppose. Also, x,z and not x,y.

mappings/net/minecraft/world/WorldType.mapping
copygirl: :white_check_mark: Fixed with 570274f

PR #86, opened 2016-09-20
Textures, Banners, Skins, and Maths

iamgreaser:
Also shields but they're treated very similarly to banners.

Took a bit of time to fill in some gaps in MathUtils too. Even found the double-precision version of What The Fuck which was then fed into some not-quite-as-bad almost-black-magic implementation of atan2.

`byk` may be worth looking into afterwards, although it does seem to be somewhat small. It does some skin and cape handling.


mappings/net/minecraft/client/texture/BannerCache.mapping
```diff
@@ -0,0 +1,22 @@
...
+			ARG 0 id
+			ARG 1 filename
+			ARG 2 baseDir
+		METHOD a removeOldEntries ()Z
+		METHOD a get (Ljava/lang/String;Ljava/util/List;Ljava/util/List;)Lnone/kp;
+			ARG 0 variant
+			ARG 1 patterns
+			ARG 2 dyes
+	CLASS none/bpk$b Entry
+		FIELD a lastRequestedMillis J
```
asiekierka: lastRequestMillis? lastRequestTimeMillis?

mappings/net/minecraft/client/util/DefaultSkinUtils.mapping
```diff
@@ -0,0 +1,10 @@
+CLASS none/bxs net/minecraft/client/util/DefaultSkinUtils
+	FIELD a STEVE Lnone/kp;
+	FIELD b ALEX Lnone/kp;
+	METHOD a getSteveIdentifier ()Lnone/kp;
+	METHOD a getIdentifier (Ljava/util/UUID;)Lnone/kp;
+		ARG 0 uuid
+	METHOD b getModel (Ljava/util/UUID;)Ljava/lang/String;
+		ARG 0 uuid
+	METHOD c useAlex (Ljava/util/UUID;)Z
```
asiekierka: shouldUseAlex, but that sounds kinda wrong still, in a different sense

asiekierka (changes requested):

mappings/net/minecraft/client/texture/BannerCache.mapping
copygirl: Imo the explicit "millis" isn't needed.

mappings/net/minecraft/client/texture/TexturePlayerSkin.mapping
```diff
@@ -0,0 +1,23 @@
+CLASS none/bwu net/minecraft/client/texture/TexturePlayerSkin
+	CLASS none/bwu$1
+		METHOD run run ()V
+	FIELD g LOGGER Lorg/apache/logging/log4j/Logger;
+	FIELD h DOWNLOAD_THREAD_COUNTER Ljava/util/concurrent/atomic/AtomicInteger;
```
copygirl, line 5: If it's mutable then it shouldn't be written in `ALL_CAPS` in my opinion (exceptions being things set on game initialization)..?

mappings/net/minecraft/client/texture/TexturePlayerSkin.mapping
```diff
@@ -0,0 +1,23 @@
...
+	FIELD i cacheFile Ljava/io/File;
+	FIELD j url Ljava/lang/String;
+	FIELD k filter Lnone/bov;
+	FIELD l image Ljava/awt/image/BufferedImage;
+	FIELD m DOWNLOADER Ljava/lang/Thread;
```
copygirl, line 10: `DOWNLOAD_TASK` maybe?

mappings/net/minecraft/client/util/DefaultSkinUtils.mapping
```diff
@@ -0,0 +1,10 @@
+CLASS none/bxs net/minecraft/client/util/DefaultSkinUtils
+	FIELD a STEVE Lnone/kp;
+	FIELD b ALEX Lnone/kp;
```
copygirl, line 3: `STEVE_SKIN` and `ALEX_SKIN` - imo it's best to name all skin fields consistently.
Similar to naming other textures `_TEX`.

mappings/net/minecraft/client/util/DefaultSkinUtils.mapping
```diff
@@ -0,0 +1,10 @@
...
+	METHOD a getSteveIdentifier ()Lnone/kp;
+	METHOD a getIdentifier (Ljava/util/UUID;)Lnone/kp;
+		ARG 0 uuid
+	METHOD b getModel (Ljava/util/UUID;)Ljava/lang/String;
+		ARG 0 uuid
+	METHOD c shouldUseAlex (Ljava/util/UUID;)Z
```
copygirl, line 9: Alternative suggestions: `useAlexModel` or `useThinArmsModel`

mappings/net/minecraft/util/math/MathUtils.mapping
```diff
@@ -60,14 +66,40 @@ CLASS none/or net/minecraft/util/math/MathUtils
 		ARG 0 values
 	METHOD b fastFloor (D)I
 		ARG 0 value
+	METHOD b modFloor (DD)D
+		ARG 0 num
+		ARG 1 denom
 	METHOD b clampedLerp (DDD)D
+		ARG 0 v0
+		ARG 1 v1
+		ARG 2 time
```
copygirl, line 75: I'd prefer "alpha" myself. Same for the `lerp` function.

mappings/net/minecraft/util/math/MathUtils.mapping
```diff
@@ -60,14 +66,40 @@ CLASS none/or net/minecraft/util/math/MathUtils
...
+		ARG 0 r
+		ARG 1 g
+		ARG 2 b
 	METHOD b wrapDegrees (I)I
 		ARG 0 value
-	METHOD b (II)I
+	METHOD b modFloor (II)I
+		ARG 0 num
+		ARG 1 denom
+	METHOD b packRgbInts (III)I
```
copygirl, line 90: Both `packRgbFloats` and this could be named `packRgb` - let's make use of overloading!

copygirl (changes requested):

mappings/net/minecraft/util/math/MathUtils.mapping
iamgreaser: Very bad idea in this case. If someone accidentally drops in a float when they mean to call the int version, all values will be 255x what they should be.

copygirl: Then signalize that the int version is 0-255 and the float version is 0-1 somehow, rather than calling them "int" and "float" variants?

PR #88, opened 2016-09-20
Vec3d fully mapped; Polar2f identified

iamgreaser:
While we're at it there's a couple of already-mapped names I'd like remapped:
- `getDistanceTo` should be `distanceTo`
- `getSquaredDistanceTo` should be `squaredDistanceTo`

If there's an agreement that this change should be made I'll add it to this PR. Otherwise, review away.

### DIFF
--- a/mappings/net/minecraft/util/math/BlockPos.mapping
+++ b/mappings/net/minecraft/util/math/BlockPos.mapping
@@ -24,6 +24,8 @@ CLASS none/cn net/minecraft/util/math/BlockPos
 		METHOD c setOffset (Lnone/cu;I)Lnone/cn$a;
 			ARG 0 facing
 			ARG 1 distance
+		METHOD d crossProduct (Lnone/dk;)Lnone/dk;
+			ARG 0 vec
 		METHOD g set (Lnone/dk;)Lnone/cn$a;
 			ARG 0 vec
 		METHOD h toImmutable ()Lnone/cn;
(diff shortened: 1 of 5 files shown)


iamgreaser:
OK, I've realised I've made some goofs pertaining to polars.
- The two-angle version of `coerce` makes me think they should both be `fromPolar`.
- Should it be `Polar2f` instead?

asiekierka:
If it's floats it should be Polar2f. Also, I agree with the distance change, as it's not merely a getter.

mappings/net/minecraft/util/math/BlockPos.mapping
```diff
@@ -107,17 +126,21 @@ CLASS none/cn net/minecraft/util/math/BlockPos
...
 		ARG 0 vec
 	METHOD c north ()Lnone/cn;
 	METHOD c down (I)Lnone/cn;
 		ARG 0 distance
+	METHOD c crossProduct (Lnone/dk;)Lnone/cn;
+		ARG 0 vec
 	METHOD d south ()Lnone/cn;
 	METHOD d north (I)Lnone/cn;
 		ARG 0 distance
+	METHOD d crossProduct (Lnone/dk;)Lnone/dk;
```
copygirl, line 142: Are there two methods with the same name and parameters?

copygirl (approved):

mappings/net/minecraft/util/math/BlockPos.mapping
iamgreaser: ...that's odd. I can check this in the morning but at least one of those does a cross product.

PR #90, opened 2016-09-22
Map ambientDarkness and isDaylight

copygirl:
Unsure about the name `ambientDarkness`. It appears to be the number of full light levels that blocks that receive skylight will be considered lower as (bad wording, hope you understand regardless). It's used for mob AI (skeletons hiding from light, endermen teleporting) and `isDaylight`, which is also used to check if the player can sleep in a bed.

### DIFF
--- a/mappings/net/minecraft/world/World.mapping
+++ b/mappings/net/minecraft/world/World.mapping
@@ -2,6 +2,7 @@ CLASS none/aiw net/minecraft/world/World
 	FIELD B lootTableHandler Lnone/bbm;
 	FIELD C profiler Lnone/os;
 	FIELD E isRemote Z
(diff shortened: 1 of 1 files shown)


iamgreaser:
Typically you'd refer to ambient **light**, although this would be in the wrong direction here, but I guess if we refer to ambient light **attenuation**, while it would be more correct, there's fewer people who know what attenuation is than what darkness is.

For everyone else's benefit:

This value is an int and appears to range from 0 to 11 (not sure whether the last is inclusive or exclusive as floating point numbers are weird), and one of the checks for a skeleton catching fire is that this value must be less than 4.

It is calculated using the cosine of the time of day and several other things.

shadowfacts (approved):

PR #91, opened 2016-09-23
Particle system

iamgreaser:
Will still need a lot of work. Posting this PR right now mostly because I'm trying to suss out where everything is supposed to even go.

### DIFF
--- a/mappings/net/minecraft/client/particle/Particle.mapping
+++ b/mappings/net/minecraft/client/particle/Particle.mapping
@@ -1,4 +1,4 @@
-CLASS none/bnl net/minecraft/particle/Particle
+CLASS none/bnl net/minecraft/client/particle/Particle
 	FIELD A colorRed F
 	FIELD B colorGreen F
 	FIELD C colorBlue F
--- a/mappings/net/minecraft/client/particle/Particle.mapping
+++ b/mappings/net/minecraft/client/particle/Particle.mapping
@@ -25,7 +25,7 @@ CLASS none/bnl net/minecraft/client/particle/Particle
 	FIELD o dead Z
 	FIELD p spacingXZ F
 	FIELD q spacingY F
-	FIELD r random Ljava/util/Random;
+	FIELD r rand Ljava/util/Random;
 	FIELD s tileU I
 	FIELD t tileV I
 	FIELD w age I
@@ -44,7 +44,7 @@ CLASS none/bnl net/minecraft/client/particle/Particle
 		ARG 4 velocityX
 		ARG 5 velocityY
 		ARG 6 velocityZ
-	METHOD a tick ()V
+	METHOD a update ()V
 	METHOD a addPos (DDD)V
 		ARG 0 dx
 		ARG 1 dy
--- a/mappings/net/minecraft/client/particle/Particle.mapping
+++ b/mappings/net/minecraft/client/particle/Particle.mapping
@@ -7,6 +7,7 @@ CLASS none/bnl net/minecraft/client/particle/Particle
 	FIELD H lerpX D
 	FIELD I lerpY D
 	FIELD J lerpZ D
+	FIELD K billboardNormal Lnone/bcv;
 	FIELD a EMPTY_BOUNDING_BOX Lnone/bcs;
 	FIELD b boundingBox Lnone/bcs;
 	FIELD c world Lnone/aiw;
@@ -19,8 +20,11 @@ CLASS none/bnl net/minecraft/client/particle/Particle
 	FIELD j velocityX D
 	FIELD k velocityY D
 	FIELD l velocityZ D
+	FIELD m onGround Z
 	FIELD n collidesWithWorld Z
 	FIELD o dead Z
+	FIELD p spacingXZ F
+	FIELD q spacingY F
 	FIELD r random Ljava/util/Random;
 	FIELD s tileU I
 	FIELD t tileV I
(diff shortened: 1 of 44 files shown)


mappings/net/minecraft/client/particle/Particle.mapping
```diff
@@ -0,0 +1,81 @@
+CLASS none/bnl net/minecraft/client/particle/Particle
+	FIELD A colorRed F
+	FIELD B colorGreen F
+	FIELD C colorBlue F
+	FIELD D colorAlpha F
+	FIELD E sprite Lnone/bxe;
```
copygirl, line 6: We haven't used sprite before, why start now?

mappings/net/minecraft/client/particle/ParticleManager.mapping
```diff
@@ -0,0 +1,37 @@
+CLASS none/bnm net/minecraft/client/particle/ParticleManager
+	FIELD a world Lnone/aiw;
+	FIELD b PARTICLE_TEX Lnone/kp;
+	FIELD c particleQueues [[Ljava/util/ArrayDeque;
+	FIELD e textureManager Lnone/bxf;
+	FIELD f random Ljava/util/Random;
```
copygirl: `rand` (This isn't static? Odd.)

mappings/net/minecraft/client/particle/ParticleManager.mapping
```diff
@@ -0,0 +1,37 @@
...
+		ARG 2 y
+		ARG 3 z
+		ARG 4 velocityX
+		ARG 5 velocityY
+		ARG 6 velocityZ
+		ARG 7 extra
+	METHOD a registerFactory (ILnone/bnn;)V
+		ARG 0 particleId
+		ARG 1 factory
+	METHOD a tickParticleQueue (Ljava/util/Queue;)V
```
copygirl: `update`?

mappings/net/minecraft/client/particle/ParticleManager.mapping
```diff
@@ -0,0 +1,37 @@
...
+		ARG 0 particleQueue
+	METHOD a setWorld (Lnone/aiw;)V
+		ARG 0 world
+	METHOD a addParticle (Lnone/bnl;)V
+		ARG 0 particle
+	METHOD b getParticleCountAsString ()Ljava/lang/String;
+	METHOD b tickParticle (Lnone/bnl;)V
```
copygirl: `update`?

mappings/net/minecraft/client/particle/Particle.mapping
```diff
@@ -0,0 +1,81 @@
...
+		ARG 3 z
+	METHOD <init> (Lnone/aiw;DDDDDD)V
+		ARG 0 world
+		ARG 1 x
+		ARG 2 y
+		ARG 3 z
+		ARG 4 velocityX
+		ARG 5 velocityY
+		ARG 6 velocityZ
+	METHOD a tick ()V
```
copygirl: `update`? We should keep things consistent and I believe we've been using `update` for everything.

copygirl (changes requested):

mappings/net/minecraft/client/particle/Particle.mapping
iamgreaser: `tick` is used in other places.

mappings/net/minecraft/client/particle/Particle.mapping
iamgreaser: There's a class called `Sprite`, what are these instances _usually_ called?

mappings/net/minecraft/client/particle/Particle.mapping
iamgreaser: Also, `tick` is also less vague. If you insist though I can rename it to update.

mappings/net/minecraft/client/particle/ParticleManager.mapping
iamgreaser: Comment as per `tick`->`update` but I will NOT overload these into all being called `update`.

exaskye: I think they mean `updateParticleQueue`, not just `update`.

mappings/net/minecraft/client/particle/Particle.mapping
exaskye: none/bxe _is_ `Sprite`, so this seems fine to me.

mappings/net/minecraft/client/particle/Particle.mapping
copygirl: I agree. Then we should update all other mentions of update, too.

iamgreaser:
OK, how should all the different particles be named?

Should we go e.g. `ParticleFire` or `FireParticle`? The former is hell from a tab-completion perspective, and they're all going to be in `net.minecraft.client.particle.*` anyway.

darkevilmac:
> OK, how should all the different particles be named?
> 
> Should we go e.g. ParticleFire or FireParticle? The former is hell from a tab-completion perspective, and they're all going to be in net.minecraft.client.particle.\* anyway.

ParticleFire matches better to the current naming of other things.

iamgreaser:
I guess from that perspective you won't be using the classes directly, so I guess `ParticleFire` could still work fine.

copygirl (approved):

exaskye (approved):

iamgreaser:
Would you rather I dropped the WIP tag and had this pulled now, or would you rather I get names for the particle classes first?

copygirl:
This PR looks good to me so I can merge it and have another one for the particle class names.

iamgreaser:
@copygirl Go for it.

PR #98, opened 2016-09-28
Map world generation related stuff

copygirl:


mappings/net/minecraft/world/IWorldListener.mapping
```diff
@@ -1,8 +1,8 @@
 CLASS none/aiy net/minecraft/world/IWorldListener
-	METHOD a onSpawnParticle (IZDDDDDD[I)V
+	METHOD a onParticle (IZDDDDDD[I)V
 		ARG 0 typeId
 		ARG 1 ignoreRange
-	METHOD a onPlaySound (Lnone/aam;Lnone/nk;Lnone/nm;DDDFF)V
+	METHOD a onSound (Lnone/aam;Lnone/nk;Lnone/nm;DDDFF)V
```
asiekierka, line 5: Are you sure? I named them after the respective methods in World.

mappings/net/minecraft/world/chunk/Chunk.mapping
```diff
@@ -69,17 +72,19 @@ CLASS none/atp net/minecraft/world/chunk/Chunk
...
 		ARG 0 entity
 	METHOD b update (Z)V
-	METHOD c addEntities ()V
+	METHOD c addEntitiesToWorld ()V
 	METHOD c setInhabitedTime (J)V
 		ARG 0 inhabitedTime
 	METHOD c setEntitiesLoaded (Z)V
 		ARG 0 value
-	METHOD d dropEntities ()V
+	METHOD d dropEntitiesFromWorld ()V
```
asiekierka, line 87: drop or remove? If we have add...

mappings/net/minecraft/world/gen/chunk/ChunkGeneratorFlat.mapping
```diff
@@ -0,0 +1,33 @@
+CLASS none/aus net/minecraft/world/gen/chunk/ChunkGeneratorFlat
+	FIELD a world Lnone/aiw;
+	FIELD b rand Ljava/util/Random;
+	FIELD c pallette [Lnone/asm;
```
asiekierka: palette

asiekierka (changes requested):

mappings/net/minecraft/world/chunk/Chunk.mapping
copygirl: Add instantly adds entities to the world, drop adds them to a list to be removed next tick.

mappings/net/minecraft/world/IWorldListener.mapping
copygirl: It felt more appropriate with `onParticle` and `onEntityAdded/Removed`. I was thinking maybe `onSoundPlayed` but that sounded weird, and `onSound` doesn't feel wrong to me. Also it doesn't necessarily "play" a sound - it is called when a sound "happens" in the world, both on client and server, right?

mappings/net/minecraft/world/gen/chunk/ChunkGeneratorFlat.mapping
copygirl: I didn't name that so it's not my fault! :stuck_out_tongue:

PR #109, opened 2016-10-13
Map world event

copygirl:
Renamed the packet previously called `Effect` to `WorldEvent` too. Mojang calls this "level event" it appears. There's global and normal world events and they basically seem to play sounds and spawn particles when received by the client (ender dragon / wither sounds being global events).

Should I keep it named as `Effect`? What are your opinions?

### DIFF
--- a/mappings/net/minecraft/network/packet/client/CPacketWorldEvent.mapping
+++ b/mappings/net/minecraft/network/packet/client/CPacketWorldEvent.mapping
@@ -1,21 +1,21 @@
-CLASS none/gw net/minecraft/network/packet/client/CPacketEffect
-	FIELD a effectId I
+CLASS none/gw net/minecraft/network/packet/client/CPacketWorldEvent
+	FIELD a eventId I
 	FIELD b pos Lnone/cn;
-	FIELD c effectData I
-	FIELD d disableRelativeVolume Z
+	FIELD c data I
+	FIELD d global Z
 	METHOD <init> (ILnone/cn;IZ)V
-		ARG 0 effectId
+		ARG 0 eventId
 		ARG 1 pos
-		ARG 2 effectData
-		ARG 3 disableRelativeVolume
-	METHOD a disableRelativeVolume ()Z
+		ARG 2 data
+		ARG 3 global
+	METHOD a isGlobal ()Z
 	METHOD a readPacket (Lnone/es;)V
 		ARG 0 packet
 	METHOD a applyPacket (Lnone/ev;)V
 		ARG 0 handler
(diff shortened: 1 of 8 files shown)


mappings/net/minecraft/network/packet/client/CPacketWorldEvent.mapping
```diff
@@ -1,21 +1,21 @@
-CLASS none/gw net/minecraft/network/packet/client/CPacketEffect
-	FIELD a effectId I
+CLASS none/gw net/minecraft/network/packet/client/CPacketWorldEvent
+	FIELD a eventId I
```
maxpowa, line 2: Why not just `id`? Wouldn't it be implied to be an `eventId` since this a `CPacketWorldEvent` class?

mappings/net/minecraft/world/IWorldListener.mapping
```diff
@@ -1,7 +1,16 @@
 CLASS none/aiy net/minecraft/world/IWorldListener
+	METHOD a onGlobalWorldEvent (ILnone/cn;I)V
+		ARG 0 eventId
```
maxpowa, line 3: `eventId` vs `id`

mappings/net/minecraft/world/World.mapping
```diff
@@ -83,6 +83,10 @@ CLASS none/aiw net/minecraft/world/World
 	METHOD a isChunkLoaded (IIZ)Z
 		ARG 0 chunkX
 		ARG 1 chunkZ
+	METHOD a fireGlobalWorldEvent (ILnone/cn;I)V
+		ARG 0 eventId
```
maxpowa, line 87: `eventId` vs `id`

mappings/net/minecraft/network/packet/client/CPacketWorldEvent.mapping
copygirl: It's still a bit ambiguous if `id` is referring to the event ID or something else, like world ID. I was more thinking of renaming `data` to `eventData`, actually.

maxpowa: Unfortunately we can't really rely on precedent for this because it's all over the place in the other `CPacket...`s. We should probably make an attempt to make all the `CPacket...` use the same field name for event ids, another one that appears to have the same idea behind it is `CPacketBlockEntityUpdate.actionId`, thought it might just be a bad name.

asiekierka: Sure, open an issue for that. eventId is probably for the better.

PR #111, opened 2016-10-13
Map / move DispenserBehavior

copygirl:


mappings/net/minecraft/block/BlockDispenser.mapping
```diff
@@ -50,6 +50,8 @@ CLASS none/ama net/minecraft/block/BlockDispenser
 	METHOD a applyRotation (Lnone/asm;Lnone/apl;)Lnone/asm;
 		ARG 0 state
 		ARG 1 rotation
+	METHOD a getDispenseOffset (Lnone/co;)Lnone/de;
```
asiekierka: getDispensionOffset? getDispensingOffset? Maybe...

asiekierka (changes requested):

mappings/net/minecraft/block/BlockDispenser.mapping
copygirl: Both sound weirder than `Dispense` to me.

maxpowa: If I'm reading this correctly, isn't this more the position where dispensed items will initially spawn? This appears to return a full `Position` with the world x,y,z coords, not an offset `Position`.

mappings/net/minecraft/block/dispenser/DispenserBehaviorProjectile.mapping
```diff
@@ -0,0 +1,12 @@
+CLASS none/kr net/minecraft/block/dispenser/DispenserBehaviorProjectile
+	METHOD a getProjectileForce ()F
+	METHOD a createProjectile (Lnone/aiw;Lnone/de;Lnone/aes;)Lnone/aav;
+		ARG 0 world
+		ARG 1 pos
+		ARG 2 stack
+	METHOD a playSound (Lnone/co;)V
+		ARG 0 block
+	METHOD b getProjectileVariation ()F
+	METHOD b dispenseInternal (Lnone/co;Lnone/aes;)Lnone/aes;
```
maxpowa: Is this really the best name for this? Seems like it could definitely be better. Maybe something like `doDispense` or just `dispenseProjectile`. Looks like it's extending from `DispenserBehaviorItem` though so generic `performDispense` might be be better?

mappings/net/minecraft/block/BlockDispenser.mapping
copygirl: `getDispenseLocation` then? `getDispenseSpawnPosition` is something else that comes to mind.

asiekierka: getDispenseLocation.

PR #113, opened 2016-10-13
Map GuiSettingsVideo and related classes

maxpowa:
Also changed misspelled `SENSETIVITY` to `SENSITIVITY` in GameSettings & related places.


mappings/net/minecraft/client/gui/menu/GuiSettingsVideo.mapping
```diff
@@ -2,6 +2,8 @@ CLASS none/bgv net/minecraft/client/gui/menu/GuiSettingsVideo
 	FIELD a title Ljava/lang/String;
 	FIELD f parent Lnone/bgn;
 	FIELD g settings Lnone/bdt;
+	FIELD h widgetList Lnone/bex;
```
copygirl, line 5: Don't GUIs already have a collection of widgets? Is this special in some way? If so the name should reflect that.

mappings/net/minecraft/client/settings/GameSettings.mapping
```diff
@@ -13,9 +13,14 @@ CLASS none/bdt net/minecraft/client/settings/GameSettings
 		FIELD K SHOW_SUBTITLES Lnone/bdt$a;
 		FIELD L REALMS_NOTIFICATIONS Lnone/bdt$a;
 		FIELD M AUTO_JUMP Lnone/bdt$a;
+		FIELD N isSlider Z
+		FIELD O isBoolean Z
```
copygirl: Fields should not have a `is` prefix or similar. `slider` and `switch` ?

mappings/net/minecraft/client/settings/GameSettings.mapping
```diff
@@ -40,7 +45,36 @@ CLASS none/bdt net/minecraft/client/settings/GameSettings
...
+			ARG 2 isBoolean
+			ARG 3 sliderMin
+			ARG 4 sliderMax
+			ARG 5 sliderStep
+		METHOD a isSlider ()Z
+		METHOD a setMaximum (F)V
+			ARG 0 value
+		METHOD a getOption (I)Lnone/bdt$a;
+			ARG 0 id
+		METHOD b isBoolean ()Z
```
copygirl: `isSwitch` ?

asiekierka: Or isBoolean? Depends on the usage of it...

copygirl: Well if this corresponds to the field of the same name ... the field can't be named `boolean`. And I think switch is the closest thing I can think of that represents an on/off state (other than checkbox but that's usually rendered in a specific way).

mappings/net/minecraft/client/gui/menu/GuiSettingsVideo.mapping
maxpowa: This is a `WidgetListMulti` instance, it is the container for the 2 column button view.

mappings/net/minecraft/client/settings/GameSettings.mapping
maxpowa: I was considering `toggle` but it didn't feel quite right. I guess I ended up prefixing these because what they indicate is somewhat confusing if they are just `slider` and `switch`. Being able to doc this would be nice :)

copygirl: Oh, `toggle` is the one I couldn't think of! I'd actually prefer that tbh.

mappings/net/minecraft/client/settings/GameSettings.mapping
maxpowa: I've changed it to `isToggle` but it's still very confusing without documentation, both of the values (`toggle` and `slider`) are just booleans. They indicate whether or not the option is a numeric slider value option or a boolean value option.

copygirl (approved):

Darkhax (approved):

PR #118, opened 2016-10-15
More misc mappings

Darkhax:
I would like feedback on the first person renderer class. Also WIP, but I don't have perms to add the label.

### DIFF
--- a/mappings/net/minecraft/client/render/FirstPersonRenderer.mapping
+++ b/mappings/net/minecraft/client/render/FirstPersonRenderer.mapping
@@ -1,0 +1,51 @@
+CLASS none/box net/minecraft/client/render/FirstPersonRenderer
+	FIELD a TEXTURE_MAP_BACKGROUND Lnone/kp;
+	FIELD b TEXTURE_UNDERWATER Lnone/kp;
+	FIELD c minecraft Lnone/bdr;
+	FIELD d mainhand Lnone/aes;
+	FIELD e offhand Lnone/aes;
+	FIELD f equipProgressMainhand F
+	FIELD g previousEquipProgressMainhand F
+	FIELD h equipProgressOffhand F
+	FIELD i previousEquipProgressOffhand F
+	FIELD j renderManager Lnone/btd;
+	FIELD k itemRenderer Lnone/btr;
+	METHOD <init> (Lnone/bdr;)V
+		ARG 0 minecraft
+	METHOD a updateHeldItems ()V
+	METHOD a renderFirstPersonItem (F)V
+		ARG 0 deltaTicks
+	METHOD a rotate (FF)V
+		ARG 0 degreeX
+		ARG 1 degreeY
+	METHOD a renderFirstPersonMap (FFF)V
+	METHOD a renderBlock (FLnone/bxe;)V
+		ARG 0 deltaTicks
+		ARG 1 atlas
+	METHOD a renderFirstPersonMap (Lnone/aes;)V
+		ARG 0 map
+	METHOD a isTranslucent (Lnone/aky;)Z
+		ARG 0 block
+	METHOD a renderFirstPersonItem (Lnone/boe;FFLnone/rb;FLnone/aes;F)V
+		ARG 0 player
+		ARG 3 hand
+		ARG 5 item
+	METHOD a resetEquipProgress (Lnone/rb;)V
+		ARG 0 hand
+	METHOD a renderArm (Lnone/so;)V
+		ARG 0 hand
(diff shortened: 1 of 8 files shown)


mappings/net/minecraft/client/render/FirstPersonRenderer.mapping
```diff
@@ -0,0 +1,51 @@
+CLASS none/box net/minecraft/client/render/FirstPersonRenderer
+	FIELD a TEXTURE_MAP_BACKGROUND Lnone/kp;
+	FIELD b TEXTURE_UNDERWATER Lnone/kp;
```
copygirl: `MAP_BACKGROUND_TEX` and `UNDERWATER_TEX`

mappings/net/minecraft/client/render/FirstPersonRenderer.mapping
```diff
@@ -0,0 +1,51 @@
...
+	FIELD c minecraft Lnone/bdr;
```
copygirl: `game`

mappings/net/minecraft/client/render/FirstPersonRenderer.mapping
```diff
@@ -0,0 +1,51 @@
...
+	FIELD d mainhand Lnone/aes;
+	FIELD e offhand Lnone/aes;
+	FIELD f equipProgressMainhand F
+	FIELD g previousEquipProgressMainhand F
+	FIELD h equipProgressOffhand F
+	FIELD i previousEquipProgressOffhand F
```
copygirl: `mainHand` and `offHand`
Possibly, `prev` instead of `previous` ?

mappings/net/minecraft/client/render/FirstPersonRenderer.mapping
```diff
@@ -0,0 +1,51 @@
...
+	FIELD j renderManager Lnone/btd;
+	FIELD k itemRenderer Lnone/btr;
+	METHOD <init> (Lnone/bdr;)V
+		ARG 0 minecraft
+	METHOD a updateHeldItems ()V
+	METHOD a renderFirstPersonItem (F)V
+		ARG 0 deltaTicks
+	METHOD a rotate (FF)V
+		ARG 0 degreeX
+		ARG 1 degreeY
```
copygirl, line 20: `pitch` / `yaw` ? Unsure what this actually is and which is which if it would apply.

copygirl (changes requested):

mappings/net/minecraft/client/render/FirstPersonRenderer.mapping
Darkhax: It's simply a wrapper for the GL rotate. Each parameters specifies the first argument of rotate call. I don't believe pitch/yaw would fit here.

asiekierka: Yes, that.

mappings/net/minecraft/client/render/FirstPersonRenderer.mapping
```diff
@@ -0,0 +1,51 @@
+CLASS none/box net/minecraft/client/render/FirstPersonRenderer
+	FIELD a MAP_BACKGROUND_TEX Lnone/kp;
+	FIELD b UNDERWATER_TEX Lnone/kp;
+	FIELD c game Lnone/bdr;
+	FIELD d mainhand Lnone/aes;
+	FIELD e offhand Lnone/aes;
+	FIELD f equipProgressMainHand F
+	FIELD g previEquipProgressMainHand F
```
lclc98: Probably should be prev no previ

copygirl: Also it's still `mainhand` and `offhand`.

Darkhax: Isn't that what I had before copy?

mappings/net/minecraft/client/render/FirstPersonRenderer.mapping
Darkhax: I am hesitant to use pitch/yaw in this situation. There seems to be some confusion about how the 3 terms relate to the various axis. I have always been taught that Yaw is around Z, Roll is around X and Pitch is around Y, however this does not seem to match with the way things are done in Minecraft. If you both think that it's a good idea I won't disagree, however I would like to know how the two of you would assign xyz to the terms, to ensure that I don't mess it up here or in the future.

copygirl (approved):

PR #122, opened 2016-10-16
Add some entity mappings, more general name for EntityParts

mezz:


Darkhax (approved):

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -322,6 +329,8 @@ CLASS none/sg net/minecraft/entity/Entity
 		ARG 0 entity
 	METHOD r isTeammate (Lnone/sg;)Z
 		ARG 0 entity
+	METHOD s isEqualEntity (Lnone/sg;)Z
```
asiekierka: Why not just isEqual? Or isEqualType? Not sure what it does...

mappings/net/minecraft/entity/decoration/EntityDecorationBase.mapping
```diff
@@ -23,6 +23,7 @@ CLASS none/yo net/minecraft/entity/decoration/EntityDecorationBase
 		ARG 0 entity
 	METHOD a onStruckByLightning (Lnone/yx;)V
 		ARG 0 lightning
+	METHOD aq hasCollision ()Z
```
asiekierka: doesCollide? canCollide?

mappings/net/minecraft/entity/Entity.mapping
mezz: The param type here is `Entity`, so I don't want to override `isEqual`.
It returns true if the param == this, or if this is an entity part and the param is the entity that owns that part.

mappings/net/minecraft/entity/decoration/EntityDecorationBase.mapping
mezz: hmm I agree it's not quite right but I don't know what's best.

mappings/net/minecraft/entity/Entity.mapping
asiekierka: hasEqualParts?

mezz: sounds good

mezz: well, not quite actually.
`dragonPart.isEqualEntity(dragon)`

mappings/net/minecraft/entity/decoration/EntityDecorationBase.mapping
mezz: I'll go with doesCollide

mappings/net/minecraft/entity/Entity.mapping
mezz: Going with `isPartOf`

asiekierka: isOwnerOf?

mezz: isOwnerOf would be the opposite logic

PR #124, opened 2016-10-16
Add/fix some chunk mappings

mezz:


mappings/net/minecraft/world/chunk/Chunk.mapping
```diff
@@ -57,18 +67,24 @@ CLASS none/atp net/minecraft/world/chunk/Chunk
...
 		ARG 0 pos
+		ARG 1 mode
 	METHOD a (Lnone/es;IZ)V
 		ARG 0 buffer
 	METHOD a addEntity (Lnone/sg;)V
 		ARG 0 entity
 	METHOD a removeEntity (Lnone/sg;I)V
 		ARG 0 entity
 		ARG 1 chunkY
+	METHOD a getEntitiesAndPartsInAreaExcludingOne (Lnone/sg;Lnone/bcs;Ljava/util/List;Lcom/google/common/base/Predicate;)V
```
Darkhax: This seems like a very long name. Would something like `getEntitiesInAreaExcluding` be an accurate name for it?

mezz: It's an unusually bad function so I feel like it requires the long name.
This gives people an idea of what it's used for without having to puzzle it out by reading the code.

asiekierka: I don't think we need a name this long. Parameters can also have clues.

mezz: changed to `getEntitiesInArea`

mappings/net/minecraft/world/chunk/Chunk.mapping
```diff
@@ -57,18 +67,24 @@ CLASS none/atp net/minecraft/world/chunk/Chunk
...
+		ARG 1 mode
 	METHOD a (Lnone/es;IZ)V
 		ARG 0 buffer
 	METHOD a addEntity (Lnone/sg;)V
 		ARG 0 entity
 	METHOD a removeEntity (Lnone/sg;I)V
 		ARG 0 entity
 		ARG 1 chunkY
+	METHOD a getEntitiesInArea (Lnone/sg;Lnone/bcs;Ljava/util/List;Lcom/google/common/base/Predicate;)V
+		ARG 0 excludedEntity
```
copygirl, line 84: There's a `sendToAllPlayers` that sends packets as well where I named the "excluded player" parameter `except`. For consistency's sake, should I rename this one to `except` or the packet one to `excludedPlayer` ?

PR #127, opened 2016-10-16
Add some mappings for mob pathfinding

mezz:
This code is so confusing in places, hopefully easier after this starting point.

### DIFF
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -19,6 +19,7 @@ CLASS none/sg net/minecraft/entity/Entity
 	FIELD M prevRenderX D
 	FIELD N prevRenderY D
 	FIELD O prevRenderZ D
+	FIELD P stepHeight F
 	FIELD S rand Ljava/util/Random;
 	FIELD T age I
 	FIELD V swimming Z
@@ -191,6 +192,7 @@ CLASS none/sg net/minecraft/entity/Entity
 	METHOD aj isSwimming ()Z
 	METHOD ak updateSwimming ()Z
 	METHOD al onSwimmingStart ()V
+	METHOD ao isTouchingLava ()Z
 	METHOD aq doesCollide ()Z
 	METHOD au isValid ()Z
 	METHOD av isInsideWall ()Z
--- a/mappings/net/minecraft/entity/ai/pathing/PathNodeMaker.mapping
+++ b/mappings/net/minecraft/entity/ai/pathing/PathNodeMaker.mapping
@@ -1,0 +1,53 @@
+CLASS none/bab net/minecraft/entity/ai/pathing/PathNodeMaker
+	FIELD a blockView Lnone/aja;
+	FIELD b entity Lnone/sq;
+	FIELD c pathNodeCache Lnone/om;
+	FIELD d xSize I
+	FIELD e ySize I
+	FIELD f zSize I
+	FIELD g canEnterOpenDoors Z
+	FIELD h canPathThroughDoors Z
+	FIELD i canSwim Z
+	METHOD a clear ()V
+	METHOD a getPathNode (DDD)Lnone/baa;
+		ARG 0 x
+		ARG 1 y
+		ARG 2 z
+	METHOD a getPathNode (III)Lnone/baa;
+		ARG 0 x
+		ARG 1 y
+		ARG 2 z
+	METHOD a getPathNodeType (Lnone/aja;III)Lnone/azz;
+		ARG 0 blockView
+		ARG 1 x
+		ARG 2 y
+		ARG 3 z
+	METHOD a getPathNodeType (Lnone/aja;IIILnone/sq;IIIZZ)Lnone/azz;
+		ARG 0 blockView
+		ARG 1 x
+		ARG 2 y
+		ARG 3 z
+		ARG 4 entity
+		ARG 5 xSize
+		ARG 6 ySize
+		ARG 7 zSize
+		ARG 8 canPathThroughDoors
+		ARG 9 canEnterOpenDoors
+	METHOD a init (Lnone/aja;Lnone/sq;)V
+		ARG 0 blockView
+		ARG 1 entityMob
+	METHOD a setCanEnterOpenDoors (Z)V
+		ARG 0 value
+	METHOD a getPathNodes ([Lnone/baa;Lnone/baa;Lnone/baa;F)I
+		ARG 0 nodes
+		ARG 1 startNode
+		ARG 2 endNode
+		ARG 3 maxDistance
+	METHOD b getStart ()Lnone/baa;
+	METHOD b setCanPathThroughDoors (Z)V
+		ARG 0 value
+	METHOD c canEnterOpenDoors ()Z
+	METHOD c setCanSwim (Z)V
+		ARG 0 value
+	METHOD d canPathThroughDoors ()Z
+	METHOD e canSwim ()Z
--- a/mappings/net/minecraft/entity/ai/pathing/PathNodeMaker.mapping
+++ b/mappings/net/minecraft/entity/ai/pathing/PathNodeMaker.mapping
@@ -5,9 +5,9 @@ CLASS none/bab net/minecraft/entity/ai/pathing/PathNodeMaker
 	FIELD d xSize I
 	FIELD e ySize I
 	FIELD f zSize I
-	FIELD g canEnterOpenDoors Z
-	FIELD h canPathThroughDoors Z
-	FIELD i canSwim Z
+	FIELD g entersOpenDoors Z
+	FIELD h pathsThroughDoors Z
+	FIELD i swims Z
 	METHOD a clear ()V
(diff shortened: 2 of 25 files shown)


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -191,6 +192,7 @@ CLASS none/sg net/minecraft/entity/Entity
 	METHOD aj isSwimming ()Z
 	METHOD ak updateSwimming ()Z
 	METHOD al onSwimmingStart ()V
+	METHOD ao isTouchingLava ()Z
```
copygirl, line 195: I'd prefer `isInLava` even though this is technically more correct. ( :+1: / :-1: ?)

mappings/net/minecraft/entity/ai/pathing/PathNodeMaker.mapping
```diff
@@ -0,0 +1,53 @@
+CLASS none/bab net/minecraft/entity/ai/pathing/PathNodeMaker
+	FIELD a blockView Lnone/aja;
+	FIELD b entity Lnone/sq;
+	FIELD c pathNodeCache Lnone/om;
+	FIELD d xSize I
+	FIELD e ySize I
+	FIELD f zSize I
+	FIELD g canEnterOpenDoors Z
+	FIELD h canPathThroughDoors Z
+	FIELD i canSwim Z
```
copygirl: These shouldn't be prefixed with `can` if possible, that's reserved for methods,

mappings/net/minecraft/entity/ai/pathing/PathNodeNavigator.mapping
```diff
@@ -0,0 +1,20 @@
+CLASS none/bad net/minecraft/entity/ai/pathing/PathNodeNavigator
+	FIELD a minHeap Lnone/azy;
+	FIELD b unusedSet Ljava/util/Set;
```
copygirl: If it's unused then don't map it? Unless this is "unused" as in unused nodes or something, doubt it though because then it should probably be `unusedNodes` or something.

"unused" could be another type of marking: FabricMC/Enigma#27 @Thog

mappings/net/minecraft/entity/ai/pathing/PathNodeType.mapping
```diff
@@ -0,0 +1,20 @@
...
+	FIELD f LAVA Lnone/azz;
+	FIELD g WATER Lnone/azz;
+	FIELD h RAIL Lnone/azz;
+	FIELD i FIRE_NEAR Lnone/azz;
+	FIELD j FIRE Lnone/azz;
+	FIELD k CACTUS_NEAR Lnone/azz;
+	FIELD l CACTUS Lnone/azz;
+	FIELD o DOOR_OPEN Lnone/azz;
+	FIELD p DOOR_WOOD Lnone/azz;
+	FIELD q DOOR_METAL Lnone/azz;
```
copygirl, line 16: `DOOR_BREAKABLE` and `DOOR_UNBREAKABLE` ?

mappings/net/minecraft/entity/ai/pathing/PathingCoordinator.mapping
```diff
@@ -20,6 +28,17 @@ CLASS none/vx
 		ARG 5 z
 		ARG 6 volume
 		ARG 7 pitch
+	METHOD a didBoundingBoxChange (Lnone/aiw;Lnone/cn;Lnone/asm;Lnone/asm;)Z
```
copygirl, line 31: `hasBoundingBoxChanged` ? (Even though we use `has` prefix for boolean stuff usually `^^`)

mappings/net/minecraft/entity/mob/EntityMob.mapping
```diff
@@ -2,18 +2,25 @@ CLASS none/sq net/minecraft/entity/mob/EntityMob
...
 	FIELD bt handDropChances [F
 	FIELD bu armorDropChances [F
 	FIELD bv target Lnone/sp;
 	FIELD bw visibilityCache Lnone/wb;
 	FIELD bx handItems Lnone/dc;
 	FIELD by armorItems Lnone/dc;
 	FIELD bz pickUpLoot Z
+	FIELD f move Lnone/to;
+	FIELD g jump Lnone/tm;
+	FIELD h navigation Lnone/vw;
```
copygirl, line 23: Are these (and `look`) AI tasks or something? I feel like prefixing them would make sense. `taskLook` etc?

copygirl (changes requested):

mappings/net/minecraft/entity/Entity.mapping
asiekierka: The technically more correct variant is the better one here.

mappings/net/minecraft/entity/ai/pathing/PathNodeMaker.mapping
asiekierka: Yes.

mappings/net/minecraft/entity/ai/pathing/PathNodeType.mapping
asiekierka: Depends.

mappings/net/minecraft/entity/ai/pathing/PathNodeMaker.mapping
mezz: I disagree, what's the alternative? "swimmer"? "doorPathThrougher"?

mappings/net/minecraft/entity/ai/pathing/PathNodeNavigator.mapping
mezz: It is created an cleared, but never added to. I can remove the mapping.

mappings/net/minecraft/entity/ai/pathing/PathNodeType.mapping
mezz: Minecraft checks for `material == WOOD` and `material == METAL` for these, unfortunately.

mappings/net/minecraft/entity/ai/pathing/PathingCoordinator.mapping
mezz: I do not understand your desired has/is conventions :/. Can you write up something or point me at a source for the standard?

mappings/net/minecraft/entity/mob/EntityMob.mapping
mezz: they're actions, I can change them to `actionLook` etc

mappings/net/minecraft/entity/ai/pathing/PathNodeMaker.mapping
mezz: ah got it, "swims" "pathsThroughDoors", "entersOpenDoors"

mappings/net/minecraft/entity/ai/pathing/PathNodeType.mapping
mezz: I think breakable and unbreakable is better, based on how they're used.

mezz: Actually no, villagers can't open the metal doors. So it's unopenable and unbreakable.

mappings/net/minecraft/entity/ai/pathing/PathingCoordinator.mapping
copygirl: I think the most important thing is that fields should not use prefixes and `is` is generally used for more simple methods like `isSolid(side)` and `can` is more for stuff like `canPlaceOn(block,side)`.

mappings/net/minecraft/entity/ai/pathing/PathNodeType.mapping
copygirl: Aww. Well, keep them as is, then.

mappings/net/minecraft/entity/ai/pathing/PathingCoordinator.mapping
mezz: Ok that makes sense, what about this case though?

asiekierka: isBoundingBoxDirty?

mezz: Here's what it looks like in context:

``` java
protected boolean didBoundingBoxChange(final World world, final BlockPos pos, final IBlockState oldState, final IBlockState newState) {
    final BoundingBox v1 = oldState.getCollisionBox(world, pos);
    final BoundingBox v2 = newState.getCollisionBox(world, pos);
    return v1 != v2 && (v1 == null || !v1.equals(v2));
}
```

mezz: Dirty implies that it needs an update. In this case we do not update the bounding box, we just change the mob's path to account for the changed bounding box.

copygirl (approved):

PR #130, opened 2016-10-16
Update Predicates, BlockPattern, and some loot entries.

ryantheleach:


mappings/net/minecraft/world/World.mapping
```diff
@@ -11,7 +11,7 @@ CLASS none/aiw net/minecraft/world/World
 	FIELD B lootTableHandler Lnone/bbm;
 	FIELD C profiler Lnone/os;
 	FIELD D scoreboard Lnone/bdb;
-	FIELD E isRemote Z
+	FIELD E isProxy Z
```
ryantheleach: Probably controversial, but is there a reason why it's called isRemote aside from the fact it's the name everyone is used to?

mappings/none/BlockInWorld.mapping
```diff
@@ -0,0 +1,19 @@
+CLASS none/asp none/BlockInWorld
```
ryantheleach: There is probably a better name, but from what I can tell, BlockInWorld can represent either a cached block or a snapshot depending on the boolean liveState that is passed in.

mappings/net/minecraft/world/World.mapping
asiekierka: Because it says whether the world is a remote proxy. IMHO, isRemote is more descriptive than isProxy.

mappings/none/BlockPattern.mapping
```diff
@@ -0,0 +1,58 @@
...
+		METHOD <init> (Lnone/aiw;Z)V
+			ARG 0 world
+		METHOD a load (Lnone/cn;)Lnone/asp;
+			ARG 0 pos
+	CLASS none/asq$b Result
+		FIELD a frontTopLeft Lnone/cn;
+		FIELD b forwards Lnone/cu;
+		FIELD c up Lnone/cu;
+		FIELD d cache Lcom/google/common/cache/LoadingCache;
+		FIELD e leftSize I
```
ryantheleach: It would be possible to change the name of leftSize / downSize / forwardSize to width, height, and depth? Wasn't too sure if forwards should be depth? or Breadth? however, so left it as it is for the moment while better names could be considered.

mappings/none/BlockPattern.mapping
```diff
@@ -0,0 +1,58 @@
...
+		METHOD d getLeftSize ()I
+		METHOD e getDownSize ()I
+	FIELD a tests [[[Lcom/google/common/base/Predicate;
+	FIELD b forwardsSize I
+	FIELD c downSize I
+	FIELD d leftSize I
+	METHOD <init> ([[[Lcom/google/common/base/Predicate;)V
+		ARG 0 tests
+	METHOD a getForwardsSize ()I
+	METHOD a searchAroundBuggy (Lnone/aiw;Lnone/cn;)Lnone/asq$b;
```
ryantheleach: This method is actually very broken. 
https://bugs.mojang.com/browse/MC-74249

asiekierka: Don't name the method Buggy. A mixin can do the fixin'!

mappings/net/minecraft/server/dedicated/DedicatedServer.mapping
```diff
@@ -58,7 +60,7 @@ CLASS none/lf net/minecraft/server/dedicated/DedicatedServer
 	METHOD a_ executeCommand (Ljava/lang/String;)Ljava/lang/String;
 		ARG 0 command
 	METHOD aa isDedicated ()Z
-	METHOD af isUsingNativeTransport ()Z
+	METHOD af useNativeTransport ()Z
```
ryantheleach: Didn't realize I accidentally renamed this, must have happened during a rebase.

mappings/net/minecraft/reference/LootTables.mapping
```diff
@@ -1,6 +1,81 @@
 CLASS none/bbe net/minecraft/reference/LootTables
+	FIELD A ENTITY_SNOWMAN Lnone/kp;
```
ryantheleach, line 2: Is this format ok for static final public fields?

mappings/net/minecraft/block/pattern/BlockPatternBuilder.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS none/asr net/minecraft/block/pattern/BlockPatternBuilder
+	METHOD a start ()Lnone/asr;
+	METHOD a where (CLcom/google/common/base/Predicate;)Lnone/asr;
+	METHOD a aisle ([Ljava/lang/String;)Lnone/asr;
```
ryantheleach, line 11: Should probably have a better name, taken from https://twitter.com/Dinnerbone/status/477358451115720704

mappings/net/minecraft/reference/LootTables.mapping
asiekierka: Sure.

mappings/net/minecraft/server/dedicated/DedicatedServer.mapping
asiekierka: isUsing.

mappings/net/minecraft/world/World.mapping
liach: `client`?

mappings/none/BlockInWorld.mapping
asiekierka: BlockSnapshot? Snapshots are fine for a cache...

mappings/net/minecraft/world/World.mapping
ryantheleach: Client doesn't represent it properly though, the variables should be named in context to the object they are on, in this case world.

world.isClient being true for worlds representing a server, but false for the embedded server on the client would be just wrong.

mappings/none/BlockInWorld.mapping
ryantheleach: I don't think I explained this properly.

Depending on liveState it can either be set to pass-through to the world showing the current state, or show the cached state.

If it was just lazily setting it to a cached state I could be happy with calling it a blockSnapshot, but it acts differently.

asiekierka: BlockPointer. Wait, I think we're using that name.

We must think of a better one.

mappings/net/minecraft/world/World.mapping
liach: `forClient`. We probably should not use `isXx` for fields but only for methods.

mappings/net/minecraft/block/pattern/BlockPattern.mapping
```diff
@@ -0,0 +1,61 @@
...
+		METHOD a load (Lnone/cn;)Lnone/asp;
+			ARG 0 pos
+	CLASS none/asq$b Result
+		FIELD a frontTopLeft Lnone/cn;
+		FIELD b forwards Lnone/cu;
+		FIELD c up Lnone/cu;
+		FIELD d cache Lcom/google/common/cache/LoadingCache;
+		FIELD e leftSize I
+		FIELD f downSize I
+		FIELD g forwardSize I
```
copygirl: What's wrong with `x/y/z` and `width`/`height`/`depth` or `x/y/zSize` or `sizeX/Y/Z` ?

mappings/none/BlockInWorld.mapping
liach: `BlockPosProxy`?

mappings/net/minecraft/world/loot/entry/LootTableEntryEmpty.mapping
```diff
@@ -1,5 +1,6 @@
 CLASS none/bbf net/minecraft/world/loot/entry/LootTableEntryEmpty
 	METHOD a deserialize (Lcom/google/gson/JsonObject;Lcom/google/gson/JsonDeserializationContext;II[Lnone/bcc;)Lnone/bbf;
+		ARG 0 jsObject
```
copygirl: Should be `json` I think.

mappings/none/BlockInWorld.mapping
```diff
@@ -0,0 +1,19 @@
...
+	CLASS none/asp$1
+		METHOD a apply (Lnone/asp;)Z
+			ARG 0 blockInWorld
+	FIELD a world Lnone/aiw;
+	FIELD b pos Lnone/cn;
+	FIELD c liveState Z
+	FIELD d blockstate Lnone/asm;
```
copygirl: `blockState`

mappings/none/BlockInWorld.mapping
asiekierka: BlockProxy!

mappings/none/BlockInWorld.mapping
```diff
@@ -0,0 +1,19 @@
+CLASS none/asp none/BlockInWorld
+	CLASS none/asp$1
+		METHOD a apply (Lnone/asp;)Z
+			ARG 0 blockInWorld
+	FIELD a world Lnone/aiw;
+	FIELD b pos Lnone/cn;
+	FIELD c liveState Z
```
copygirl: Just `live` ?

copygirl (changes requested):

mappings/none/BlockInWorld.mapping
asiekierka: Or just state, if the class context is Block.

mappings/net/minecraft/block/pattern/BlockPattern.mapping
ryantheleach: It's not xyz-world co-ordinates, it's local (ego-centric) to the blockpattern, as described by the facing directions. But I have renamed them to better fit width/height/depth, I may have missed some fields.

ryantheleach: Is there a way to add comments?

mappings/none/BlockInWorld.mapping
ryantheleach: Hm, maybe it's not caching but force chunk loading?

mappings/none/BlockInWorld.mapping
ryantheleach: forceLoad?

mappings/net/minecraft/block/pattern/BlockPattern.mapping
copygirl: > Is there a way to add comments?

Not quite yet, it's on the [todo list](https://github.com/FabricMC/Enigma/issues/28).

copygirl (approved):

PR #132, opened 2016-10-17
BIomes, biome generation

mezz:
This stuff is not easy to name because the functions are very strange. 
For review on this one, I think it's best to check this out in a branch and follow along in the code to see what things are doing, or else you might think I'm just crazy.


mappings/net/minecraft/sortme/BiomeProviderSingle.mapping
```diff
@@ -1 +1,18 @@
...
+		ARG 0 x
+		ARG 1 z
+		ARG 2 radius
+		ARG 3 allowedBiomes
+	METHOD a getValidPosition (IIILjava/util/List;Ljava/util/Random;)Lnone/cn;
+		ARG 0 x
+		ARG 1 z
+		ARG 2 radius
+		ARG 3 allowedBiomes
+		ARG 4 worldRand
```
copygirl: Does the provider have its own `rand` field or why is it called `worldRand`?

mappings/net/minecraft/world/biome/Biome.mapping
```diff
@@ -13,13 +15,19 @@ CLASS none/ajj net/minecraft/world/biome/Biome
 		METHOD a getName (Lnone/ajj$a;)Ljava/lang/String;
 		METHOD b setRainfall (F)Lnone/ajj$a;
 			ARG 0 rainfall
+		METHOD c setHeight (F)Lnone/ajj$a;
+			ARG 0 value
+		METHOD d setScale (F)Lnone/ajj$a;
+			ARG 0 scale
```
copygirl: `value`
Also I'm seeing `setRainfall` doesn't use `value` either. Fix that too pls :smile:

mappings/net/minecraft/world/biome/layer/BiomeLayerMapToValidBiomes.mapping
```diff
@@ -0,0 +1,16 @@
...
+	FIELD c biomesHot [Lnone/ajj;
+	FIELD d biomesTrees [Lnone/ajj;
+	FIELD e biomesCold [Lnone/ajj;
+	FIELD f biomesFrozen [Lnone/ajj;
+	FIELD g biomeGeneratorSettings Lnone/auq;
+	METHOD <init> (JLnone/azi;Lnone/ajb;Ljava/lang/String;)V
+		ARG 0 seed
+		ARG 1 previous
+		ARG 2 levelGeneratorType
+		ARG 3 generatorOptions
```
copygirl, line 11: Is this assigned to `biomeGeneratorSettings` ? Should probably either be `level` or `biomeGeneratorSettings`. Let's not call them `options` ..?

mappings/net/minecraft/world/biome/layer/BiomeLayerRareVariantsAndMakeIdsValid.mapping
```diff
@@ -0,0 +1,14 @@
+CLASS none/azo net/minecraft/world/biome/layer/BiomeLayerRareVariantsAndMakeIdsValid
```
copygirl, line 1: Why `MakeIdsValid` ?

mappings/none/bei.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS none/bei
+	FIELD a MAP_ICONS Lnone/kp;
```
copygirl: If this is a texture it should have `_TEX` suffix.
(Though maybe we should use a different term for texture maps that contain multiple sprites/icons.)

copygirl (changes requested):

PR #135, opened 2016-10-17
Tonights Work for review.

ryantheleach:


mappings/net/minecraft/entity/EntityLiving.mapping
```diff
@@ -14,6 +14,8 @@ CLASS none/sp net/minecraft/entity/EntityLiving
 	FIELD f POTION_SWIRLS_COLOR Lnone/kj;
 	FIELD g POTION_SWIRLS_AMBIENT Lnone/kj;
 	FIELD h STUCK_ARROWS Lnone/kj;
+	METHOD B (Lnone/sg;)Z
+		ARG 0 entity
```
asiekierka: Please do not map argument names for unnamed methods.

mappings/net/minecraft/entity/mob/EntityHusk.mapping
```diff
@@ -1,7 +1,11 @@
...
+	METHOD B (Lnone/sg;)Z
+		ARG 0 entity
 	METHOD G getSoundAmbient ()Lnone/nk;
 	METHOD J getLootTableId ()Lnone/kp;
 	METHOD b registerTagUpgraders (Lnone/pf;)V
 		ARG 0 handler
 	METHOD bV getSoundHurt ()Lnone/nk;
 	METHOD bW getSoundDeath ()Lnone/nk;
+	METHOD df getSoundStep ()Lnone/nk;
+	METHOD dg getSkull ()Lnone/aes;
```
asiekierka, line 9: Just Skull?　Sure?

mappings/net/minecraft/entity/mob/EntityShulker.mapping
```diff
@@ -32,6 +32,8 @@ CLASS none/zu net/minecraft/entity/mob/EntityShulker
 		ARG 0 tag
 	METHOD a onTrackedDataSet (Lnone/kj;)V
 		ARG 0 data
+	METHOD a (Lnone/ra;Lnone/ss;)Lnone/ss;
+		ARG 0 localDifficulty
```
asiekierka: Please do not map argnames for unmapped methods.

mappings/none/bmg.mapping
```diff
@@ -5,9 +5,18 @@ CLASS none/bmg
...
 		ARG 0 game
 		ARG 1 networkHandler
+	METHOD a isSpectator ()Z
+	METHOD a copyAbilities (Lnone/aam;)V
+		ARG 0 playerEntity
 	METHOD a setCurrentGameMode (Lnone/aiu;)V
 		ARG 0 value
+	METHOD a (Lnone/cn;)Z
+		ARG 0 pos
+	METHOD b isGameModeSurvivalLike ()Z
```
asiekierka: Are you sure this method is used in this manner? Same for all the other ones...

asiekierka (changes requested):

mappings/net/minecraft/entity/mob/EntityHusk.mapping
ryantheleach: The only place this method is used, is to give differing results between the husk, zombie, pigman and villager inside an if statement to check if it's being killed by a charged creeper.

It's not even used for Skeleton / Stray, as they don't inherit onDeath from SkeletonBase.

Sure this could change, but then it would end up being chargedCreeperDrop , and considering it's only returning a single itemstack, I doubt it.

If the method changed, If it returned more then 1 item, If it was inherited higher then EntityZombie I would agree that getSkull would be a bad name.

asiekierka: It might very well be the drop! Doesn't have to be a list... I think getChargedCreeperDrop, or getDroppedSkull, might work better.

mappings/net/minecraft/entity/thrown/EntityThrownPotion.mapping
```diff
@@ -15,11 +15,16 @@ CLASS none/abf net/minecraft/entity/thrown/EntityThrownPotion
 		ARG 0 value
 	METHOD a onCollision (Lnone/bct;)V
 		ARG 0 hit
+	METHOD a extinguishFire (Lnone/cn;)V
+		ARG 0 pos
 	METHOD a deserializeCustomData (Lnone/dt;)V
 		ARG 0 tag
+	METHOD a waterHurts (Lnone/sp;)Z
```
asiekierka: doesWaterHurt

mappings/net/minecraft/entity/player/InventoryPlayer.mapping
```diff
@@ -41,14 +41,21 @@ CLASS none/aal net/minecraft/entity/player/InventoryPlayer
 		ARG 0 stack
 	METHOD c_ getInvProperty (I)I
 		ARG 0 pos
+	METHOD d removeOne (Lnone/aes;)V
+		ARG 0 item
+	METHOD e validHotbarIndex (I)Z
```
asiekierka: isValidHotbarIndex.

(as a future note, please try to use is/does/can/has for boolean methods and _NOT_ use it for boolean fields)

asiekierka (changes requested):

PR #141, opened 2016-10-18
TargetSelectorParser done.

ryantheleach:


mappings/net/minecraft/util/Identifier.mapping
```diff
@@ -1,5 +1,5 @@
 CLASS kp net/minecraft/util/Identifier
-	FIELD a owner Ljava/lang/String;
+	FIELD a namespace Ljava/lang/String;
```
asiekierka: Is this a good change?

mappings/net/minecraft/world/World.mapping
```diff
@@ -215,6 +215,9 @@ CLASS aiw net/minecraft/world/World
 		ARG 0 packet
 	METHOD a getEntitiesIn (Ljava/lang/Class;Lbcs;)Ljava/util/List;
 	METHOD a getEntitiesIn (Ljava/lang/Class;Lbcs;Lcom/google/common/base/Predicate;)Ljava/util/List;
+	METHOD a filterEntities (Ljava/lang/Class;Lcom/google/common/base/Predicate;)Ljava/util/List;
```
asiekierka: This doesn't seem to accept a collection, so is it a filter? Also, clazz and not entityClass?...

mappings/net/minecraft/util/Identifier.mapping
modmuss50: I don't think it is. 

Also if you decide to change this the getter needs updating to accommodate the rename.

mappings/net/minecraft/world/World.mapping
ryantheleach: If world was considered a collection of entities it would match 

https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#filter-java.util.function.Predicate-

Except for the class-check & returning a List.

mappings/net/minecraft/command/CommandBase.mapping
```diff
@@ -55,9 +55,11 @@ CLASS j net/minecraft/command/CommandBase
 		ARG 0 args
 		ARG 1 values
 	METHOD b getAliases ()Ljava/util/List;
+	METHOD b (DLjava/lang/String;IIZ)D
```
ryantheleach: Not sure why this appeared in the diff without it removing / adding something

mappings/net/minecraft/command/selector/TargetSelectorParser.mapping
```diff
@@ -0,0 +1,131 @@
...
+	METHOD a canReturnMultiple (Ljava/lang/String;)Z
+		ARG 0 selector
+	METHOD a sortAndLimit (Ljava/util/List;Ljava/util/Map;Ln;Ljava/lang/Class;Ljava/lang/String;Lbcv;)Ljava/util/List;
+		ARG 0 targets
+		ARG 1 rawArguments
+		ARG 2 source
+		ARG 3 clazz
+		ARG 4 variable
+		ARG 5 origin
+	METHOD a extractScoreArguments (Ljava/util/Map;)Ljava/util/Map;
```
ryantheleach, line 60: Do you prefer extract or parse?

mappings/net/minecraft/util/Identifier.mapping
maxpowa: As I said in IRC before, I don't really like `owner` because it (in my opinion) implies that there's something backing ownership. `namespace` is more commonly just a key, which is true in `Identifier`.

mappings/net/minecraft/command/selector/TargetSelectorParser.mapping
```diff
@@ -0,0 +1,131 @@
...
+		ARG 0 rawArguments
+		ARG 1 clazz
+		ARG 2 argumentPredicates
+		ARG 3 variable
+		ARG 4 world
+		ARG 5 sourceBlockPos
+	METHOD a extractEntityTypePredicate (Ljava/util/Map;Ljava/lang/String;)Ljava/util/List;
+		ARG 0 rawArguments
+		ARG 1 variable
+	METHOD a extractAxis (Ljava/util/Map;Ljava/lang/String;DZ)D
```
ryantheleach, line 78: This is the Axis vs Component problem I discussed in IRC. 

It fetches the value from rawArgs based on the key.

If it's not in rawArgs it defaults to the default.
If the number is unparseable, it defaults, but optionally add's .5 to center the block, if the boolean is true.

Why the difference of behaviour between the two defaulting options I'm not 100% sure.

mappings/net/minecraft/command/selector/TargetSelectorParser.mapping
```diff
@@ -0,0 +1,131 @@
...
+	METHOD b extractCoordinate (Ljava/util/Map;Lbcv;)Lbcv;
+		ARG 0 rawArguments
+		ARG 1 defaultPos
+	METHOD b getString (Ljava/util/Map;Ljava/lang/String;)Ljava/lang/String;
+		ARG 0 rawArguments
+		ARG 1 key
+	METHOD b targetsAsText (Ln;Ljava/lang/String;)Lfa;
+		ARG 0 source
+		ARG 1 selector
+	METHOD b calculateTargets (Ln;Ljava/lang/String;Ljava/lang/Class;)Ljava/util/List;
```
ryantheleach, line 108: This is the main entry point and thus is relatively expensive to call, thus the verbiage calculate.

mappings/net/minecraft/command/selector/TargetSelectorParser.mapping
```diff
@@ -0,0 +1,131 @@
...
+		ARG 5 origin
+	METHOD a extractScoreArguments (Ljava/util/Map;)Ljava/util/Map;
+		ARG 0 rawArguments
+	METHOD a extractRadiusPredicate (Ljava/util/Map;Lbcv;)Ljava/util/List;
+		ARG 0 rawArguments
+		ARG 1 origin
+	METHOD a extractCoordinate (Ljava/util/Map;Lcn;)Lcn;
+		ARG 0 rawArguments
+		ARG 1 defaultPos
+	METHOD a getTargetsInWorld (Ljava/util/Map;Ljava/lang/Class;Ljava/util/List;Ljava/lang/String;Laiw;Lcn;)Ljava/util/List;
```
ryantheleach: This is but one step to the process, and if called directly may cause unexpected partial parses.

mappings/net/minecraft/command/selector/TargetSelectorParser.mapping
```diff
@@ -0,0 +1,131 @@
...
+	METHOD a extractAxis (Ljava/util/Map;Ljava/lang/String;DZ)D
+		ARG 0 rawArguments
+		ARG 1 key
+		ARG 2 defaultValue
+		ARG 3 centerBlock
+	METHOD a getInt (Ljava/util/Map;Ljava/lang/String;I)I
+		ARG 0 rawArguments
+		ARG 1 key
+		ARG 2 defaultValue
+	METHOD a onePlayerOrNull (Ln;Ljava/lang/String;)Llw;
```
ryantheleach, line 87: alternate entry point.

mappings/net/minecraft/command/selector/TargetSelectorParser.mapping
```diff
@@ -0,0 +1,131 @@
...
+		ARG 0 source
+		ARG 1 selector
+	METHOD a oneOrNull (Ln;Ljava/lang/String;Ljava/lang/Class;)Lsg;
```
ryantheleach, line 90: alternate entry point

mappings/net/minecraft/command/selector/TargetSelectorParser.mapping
```diff
@@ -0,0 +1,131 @@
...
+		ARG 0 rawArguments
+		ARG 1 origin
+	METHOD a extractCoordinate (Ljava/util/Map;Lcn;)Lcn;
+		ARG 0 rawArguments
+		ARG 1 defaultPos
+	METHOD a getTargetsInWorld (Ljava/util/Map;Ljava/lang/Class;Ljava/util/List;Ljava/lang/String;Laiw;Lcn;)Ljava/util/List;
+		ARG 0 rawArguments
+		ARG 1 clazz
+		ARG 2 argumentPredicates
+		ARG 3 variable
```
ryantheleach, line 72: Sounds like a bad name, but it's the usage from the minecraft wiki.

mappings/net/minecraft/command/selector/TargetSelectorParser.mapping
```diff
@@ -0,0 +1,131 @@
...
+	FIELD c SPLIT_EQUALS_PAIR Lcom/google/common/base/Splitter;
+	FIELD d POSITIONAL_ARGUMENTS Ljava/util/Set;
+	METHOD a createBoundingBox (Lcn;III)Lbcs;
+		ARG 0 pos
+		ARG 1 xOffset
+		ARG 2 yOffset
+		ARG 3 zOffset
+	METHOD a canReturnMultiple (Ljava/lang/String;)Z
+		ARG 0 selector
+	METHOD a sortAndLimit (Ljava/util/List;Ljava/util/Map;Ln;Ljava/lang/Class;Ljava/lang/String;Lbcv;)Ljava/util/List;
```
ryantheleach: sort and limit is a required step, without it parsing doesn't act as it should.

ryantheleach: It's private you dummy.

mappings/net/minecraft/command/selector/TargetSelectorParser.mapping
ryantheleach: It's private you dummy.

mappings/net/minecraft/command/CommandBase.mapping
copygirl: Naming something, then clearing the name again will do that I think. @Thog Fix that!

mappings/net/minecraft/world/World.mapping
asiekierka: That's a Predicate, though. It has more functional semantics...

mappings/net/minecraft/util/Identifier.mapping
asiekierka: I'm still not sure about it. I recommend you remove it from here and open up an issue.

asiekierka: As for ownership, that's the general intention - the only key in vanilla is "minecraft".

mappings/net/minecraft/world/World.mapping
ryantheleach: Lcom/google/common/base/Predicate; ?

asiekierka: That's also intended to be functional-style.

Don't apply names just because it matches some class used in a massively different way. Predicates filter their input, this doesn't filter its input.

copygirl:
Made some changes to the entity matching in World and Chunk. Will take a look at the other things later, hence added the WIP label. @ryantheleach I hope it's okay if I take over the PR from here? `:)`

copygirl:
Actually, I don't have much to add regarding `TargetSelectorParser`. Not sure about the entry methods, but I don't have any better idea.

asiekierka:
If you're okay with how it looks, go ahead and merge.

PR #145, opened 2016-10-21
Misc. Mappings, again

Darkhax:


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -20,6 +20,7 @@ CLASS sg net/minecraft/entity/Entity
 	FIELD N prevRenderY D
 	FIELD O prevRenderZ D
 	FIELD P stepHeight F
+	FIELD Q noClip Z
```
exaskye: Possibly `ignoreCollision` to avoid "no" in a boolean field name?

I'm aware noClip is a common term but I've gotten the impression negatives like "no" should be avoided in boolean field names, and in the event someone isn't familiar with the term "no clip" it would be more descriptive.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -78,13 +79,15 @@ CLASS sg net/minecraft/entity/Entity
 	METHOD T invalidate ()V
 	METHOD U updateLogic ()V
 	METHOD X extinguish ()V
+	METHOD a inRenderRange (D)Z
```
exaskye: `isInRenderRange`?

mappings/net/minecraft/entity/Entity.mapping
Darkhax: Does MC's no clipping affect entity collision as well? If so, I think `ignoreBlockCollision` may be a better name.

exaskye: Not sure. If it doesn't, then I'd agree.

mappings/net/minecraft/client/particle/ParticleAnimated.mapping
```diff
@@ -1,14 +1,20 @@
...
 	METHOD <init> (Laiw;DDDIIF)V
 		ARG 0 world
 		ARG 1 x
 		ARG 2 y
 		ARG 3 z
+		ARG 4 textureId
+		ARG 5 frameCount
 	METHOD a update ()V
+	METHOD a getColorMultiplier (F)I
+	METHOD c hasTransparency ()Z
```
maxpowa: `isTransparent` may be better if transparency is a binary thing, `hasTransparency` sounds like transparency could be a value between 0 and 1.

mappings/net/minecraft/client/player/EntityPlayerOther.mapping
```diff
@@ -0,0 +1,29 @@
+CLASS boi net/minecraft/client/player/EntityPlayerOther
```
maxpowa, line 1: Other?

mappings/net/minecraft/client/render/entity/model/EntityModelSheepArmor.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS bkw
+CLASS bkx net/minecraft/client/render/entity/model/EntityModelSheepArmor
```
maxpowa: `SheepArmor`? not `SheepWool`?

mappings/net/minecraft/client/player/EntityPlayerOther.mapping
exaskye: As in another player on the server besides the client, I think.

mappings/net/minecraft/client/render/entity/model/EntityModelSheepArmor.mapping
Darkhax: Lol, good catch.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -20,7 +20,7 @@ CLASS sg net/minecraft/entity/Entity
 	FIELD N prevRenderY D
 	FIELD O prevRenderZ D
 	FIELD P stepHeight F
-	FIELD Q noClip Z
+	FIELD Q ignoreCollision Z
```
copygirl, line 23: Why not `ignoreBlockCollision` ?

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -79,7 +79,7 @@ CLASS sg net/minecraft/entity/Entity
 	METHOD T invalidate ()V
 	METHOD U updateLogic ()V
 	METHOD X extinguish ()V
-	METHOD a inRenderRange (D)Z
+	METHOD a isInRenderRange (D)Z
 		ARG 0 range
```
copygirl, line 83: `distance` ? Not sure, but the argument you pass to the function isn't the range itself, but the distance that has to be checked against the range, right?

mappings/net/minecraft/client/particle/Particle.mapping
```diff
@@ -75,7 +75,7 @@ CLASS bnl net/minecraft/client/particle/Particle
 		ARG 2 z
 	METHOD b setSpriteIndex (I)V
 		ARG 0 value
-	METHOD c hasTransparency ()Z
+	METHOD c isTransparent ()Z
```
copygirl, line 78: Does this mean the particle has some fully transparent pixels or partial transparency?

copygirl:
*silently shouts\* **GiiiitHuuuub!**

mappings/net/minecraft/client/player/EntityPlayerOther.mapping
mezz: Maybe `EntityPlayerPeer`?

mappings/net/minecraft/entity/Entity.mapping
mezz: `noClip` is pretty widely understood.
Does this ignore collision with Entities as well as Blocks?

copygirl: > EntityPlayer: Collision off if spectator.
> EntityPlayerOther: Initially off (reset by `EntityPlayer.update`).
> EntityItem: Collision off if stuck in block?
> EntityEnderDragon: Always has collision off.
> EntityArmorStand: Collision off if `isUnaffectedByGravity`.
> EntityShulkerBullet: Always off.
> EntityAreaEffectCloud: Always off.
> 
> `Entity.pushAwayFrom` doesn't do anything if either entity's collision is off.
> `Entity.isInsideWall` always returns false if collision is off.
> `Entity.move` shortcuts, ignoring block collision if off.

It doesn't _completely_ turn off collision, since these entities can still be hit by arrows, unless prevented by other means, which in my opinion is a form of collision.

copygirl:
Meant to mention the PR in the `ignoreCollision => noClip` commit as well.

> Rename Entity.ignoreCollision => noClip
> 
> As this doesn't disable all kinds of collision (such as being hit by arrows), only movement related ones: Block collision and entities pushing each other (both ways).

PR #147, opened 2016-10-22
More Village and Entity mappings

mezz:


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -266,6 +266,7 @@ CLASS sg net/minecraft/entity/Entity
 	METHOD e handleFallDamage (FF)V
 		ARG 0 fallDistance
 	METHOD e (Laam;)Z
+		ARG 0 player
```
exaskye: Parameters to an unmapped method shouldn't be mapped

mappings/net/minecraft/entity/player/InventoryPlayer.mapping
```diff
@@ -19,8 +19,21 @@ CLASS aal net/minecraft/entity/player/InventoryPlayer
...
+		ARG 0 item
+		ARG 1 meta
+		ARG 2 amount
+		ARG 3 tag
+	METHOD a addPickBlock (Laes;)V
+		ARG 0 stack
+	METHOD a canStackAddMore (Laes;Laes;)Z
+		ARG 0 existingStack
+		ARG 1 stackToAdd
+	METHOD a getCurrentMiningStrength (Lasm;)F
```
exaskye: `getMiningStrength`? Is there a reason "Current" is added on here?

mezz: It varies depending on what you're holding

exaskye: I suppose that makes sense, as `getMiningStrength` could imply it would only change on a call to a setter.

At the same time, such a setter could be considered "invisible" (only called by vanilla code).

Not a big deal either way.

exaskye (approved):

mappings/net/minecraft/entity/player/InventoryPlayer.mapping
copygirl: Avoid "mining/digging/harvesting". Use `breaking`. `blockBreaking` if ambiguous.

mappings/net/minecraft/entity/EntityLiving.mapping
```diff
@@ -8,9 +8,13 @@ CLASS sp net/minecraft/entity/EntityLiving
 	FIELD au LIVING_FLAGS Lkj;
 	FIELD aw preferredHand Lrb;
 	FIELD ay stuckArrowTimer I
+	FIELD az hurtTime I
+	FIELD bD absorptionAmount F
 	FIELD b ATTR_SPRINTING_SPEED_BOOST Ltd;
 	FIELD br attributeContainer Ltf;
 	FIELD bt activePotionEffects Ljava/util/Map;
+	FIELD bx attacker Lsp;
+	FIELD by lastAttackedTimestamp I
```
copygirl: `time` instead of `timestamp` would be enough.

mappings/net/minecraft/entity/EntityLiving.mapping
```diff
@@ -8,9 +8,13 @@ CLASS sp net/minecraft/entity/EntityLiving
 	FIELD au LIVING_FLAGS Lkj;
 	FIELD aw preferredHand Lrb;
 	FIELD ay stuckArrowTimer I
+	FIELD az hurtTime I
```
copygirl: What is "hurt time"? I think this needs a better name. If this is the thing that counts down / makes entities briefly invulnerable it should be called `timer`, as it counts down..? Or is it just the red effect on an entity?

mappings/net/minecraft/villager/VillageDoor.mapping
```diff
@@ -1,19 +1,37 @@
 CLASS wf net/minecraft/villager/VillageDoor
 	FIELD a position Lcn;
 	FIELD c facing Lcu;
+	FIELD d lastTickSeenByVillager I
+	FIELD f entityInsideTicks I
```
copygirl, line 5: In my opinion, `tick` doesn't make sense in this case. Use `time`?

mappings/net/minecraft/world/chunk/Chunk.mapping
```diff
@@ -107,6 +107,8 @@ CLASS atp net/minecraft/world/chunk/Chunk
 		ARG 1 maxY
 	METHOD c setInhabitedTime (J)V
 		ARG 0 inhabitedTime
+	METHOD c isSkyVisible (Lcn;)Z
```
copygirl: `isSkyUnobstructed` ?

copygirl (changes requested):

mappings/net/minecraft/villager/VillageDoor.mapping
mezz: This one is difficult to name. Each entity "inside" the door will tick it, so if there are two villagers inside the door they will tick that door at twice the rate of one villager.

mappings/net/minecraft/entity/EntityLiving.mapping
mezz: Sounds good

mappings/net/minecraft/entity/EntityLiving.mapping
mezz: I got this name from the EntityLiving serialize/deserialize tags. I will remove the mapping since I don't understand it.

mappings/net/minecraft/villager/VillageDoor.mapping
copygirl: What about `lastTickSeenByVillager` ?

mappings/net/minecraft/world/chunk/Chunk.mapping
mezz: That's a better name, I agree.

mappings/net/minecraft/villager/VillageDoor.mapping
mezz: I can change that one, yes.

mappings/net/minecraft/entity/player/InventoryPlayer.mapping
mezz: This is based on another mapping that uses "mining", so I will update them all to use "blockBreaking" instead.

copygirl (approved):

PR #150, opened 2016-11-04
Reorganize block classes into packages

copygirl:
This also changes the access of block constructors to public so we don't run into package boundary issues due to them being protected. (@Thog ♥)

Looking for opinions on this PR, suggestions on how to structure this PR and such. Open questions:

- [ ] Should blocks and their entities (`BlockChest`, `BlockEntityChest`) be in the same package?
- [ ] And items? Should content classes (except base ones (which?)) share packages based on categories?
- [ ] Should block classes go into a separate common package to not share a one with `block.material`, `block.property`, `block.state` etc? (A content package would solve this.)
- [ ] How to actually structure packages?
- [ ] Should we adopt the naming scheme `NoteBlock`, `NoteBlockEntity` - `FrameItem`, `FrameEntity` ?


magik6k:
So maybe apply this to the whole game structure? Like:
`net.minecraft.redstone`
`net.minecraft.end`
`net.minecraft.brewing`
`net.minecraft.village`
`net.minecraft.menu`
With packages containing all logic about things they do. It would be a huge refactor, but it'd make it easier for modders to see how things work IMO.

asiekierka:
That sounds like a horrible idea. Maybe net.minecraft.content.\* would work,
but... eh. I'm not sure about it.

2016-11-04 16:02 GMT+01:00 Łukasz Magiera notifications@github.com:

modmuss50:
Blocks and tiles in the same package doesnt seem a horrible idea, but im not sure.
if this was to happen I say something like:

`net.minecraft.content` contains all the tiles and blocks that do something, and `net.minecraft.block` should still have the base classes.

I think content classes and base classes should be split.
So
`net.minecraft.block` contains all the base things
and then `net.minecraft.block.content` could have the other classes or something.

Thats my opinion on this. Sorry for layout/spelling issues due to phone keyboard.

copygirl:
Added another question:

> Should we adopt the naming scheme `NoteBlock`, `NoteBlockEntity` - `FrameItem`, `FrameEntity` ?

darkevilmac:
Personally I think blocks and block entities should be in separate packages as there's a distinct difference between the two.

copygirl:
@darkevilmac So, say you wanted to check out how item frames or chests are implemented, you'd rather have to look in two different packages?

I think we should set a precedence for how mods could structure their packages, and I can't shake the feeling that structuring them by content is more useful than by whether it's a block/block entity/entity/item.

darkevilmac:
Structuring by content could quickly become a mess. I may have misunderstood your suggestion though. I think it's perfectly reasonable to have net.minecraft.block.entity I think it's a bad idea to have net.minecraft.block store entities. Having a content/object package for in game stuff is also fine, my only concern there is stupidly long package names.

exaskye:
I feel overly long package names aren't really important, if it helps clarify what is what. Imports are generally hidden and there are vanishingly few cases that you need fully-qualified names.

In the event a .content package is adopted without a .block subpackage, then I suggest we keep prefix naming for better organization when using alphabetical sort. If it becomes .content._x_.block, postfix naming should be fine, but could lead to confusing cases like `RedstoneBlockBlock`.

Darkhax:
> Should blocks and their entities (BlockChest, BlockEntityChest) be in the same package?
>  And items? Should content classes (except base ones (which?)) share packages based on categories?

I am on the fence about this one. In my content heavy projects I tend to group all related content in to their own packages (items, blocks, event handlers, renderers and so on). Although my project is structured in such a way that it makes sense. I don't think MC's base code is structured in an appropriate way for that to work though. 

> Should block classes go into a separate common package to not share a one with block.material, block.property, block.state etc? (A content package would solve this.)

I don't really care that they share a package. 

> Should we adopt the naming scheme NoteBlock, NoteBlockEntity - FrameItem, FrameEntity ?

Didn't we already debate that the other naming scheme is better?

modmuss50:
Should I wait for this to be merged before I update to 1.11? or just leave it behind, and I can redo it for 1.11?

copygirl:
This should not be merged as is. Perhaps the access portion, but nothing beyond that. Feel free to keep the PR up for discussion, as I'm procrastinating over this due to lack of ideas.

asiekierka:
Well, that's not getting merged any time soon...

PR #155, opened 2017-05-03
Add mappings for tracking recent damage to entities

mezz:
Each entity has a `DamageTracker` for keeping track of fall damage and attacks received, so it can create a descriptive message when the entity dies.


asiekierka (approved):

modmuss50 (approved):
Looks good.

PR #157, opened 2017-05-03
net.minecraft.world mapping additions

CalmBit:
I'm really unsure how to name pull requests, so forgive the name - this is mostly just assorted changes within that package, unless changes required some extensive searching _outside_ that package...

(I'm really liking that zero deletions - thanks enigma <3)

### DIFF
--- a/mappings/net/minecraft/client/world/WorldClient.mapping
+++ b/mappings/net/minecraft/client/world/WorldClient.mapping
@@ -7,6 +7,9 @@ CLASS bqk net/minecraft/client/world/WorldClient
 	METHOD a playSound (DDDLpb;Lpd;FFZ)V
 	METHOD a getEntityById (I)Lub;
 	METHOD a isChunkLoaded (IIZ)Z
+		ARG 0 chunkX
+		ARG 1 chunkY
+		ARG 2 consultRemoteProvider
 	METHOD a playSound (Lacu;DDDLpb;Lpd;FF)V
 	METHOD a setScoreboardClient (Lbfv;)V
 	METHOD a playSoundClient (Ldx;Lpb;Lpd;FFZ)V
--- a/mappings/net/minecraft/client/world/WorldClient.mapping
+++ b/mappings/net/minecraft/client/world/WorldClient.mapping
@@ -9,7 +9,7 @@ CLASS bqk net/minecraft/client/world/WorldClient
 	METHOD a isChunkLoaded (IIZ)Z
 		ARG 0 chunkX
 		ARG 1 chunkZ
-		ARG 2 consultRemoteProvider
+		ARG 2 ignoreRemote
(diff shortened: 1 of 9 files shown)


mappings/net/minecraft/client/world/WorldClient.mapping
```diff
@@ -7,12 +7,16 @@ CLASS bqk net/minecraft/client/world/WorldClient
 	METHOD a playSound (DDDLpb;Lpd;FFZ)V
 	METHOD a getEntityById (I)Lub;
 	METHOD a isChunkLoaded (IIZ)Z
+		ARG 0 chunkX
+		ARG 1 chunkY
```
asiekierka: chunkZ

asiekierka (changes requested):

mappings/net/minecraft/client/world/WorldClient.mapping
CalmBit: Whoops! Completely slipped by me.

modmuss50 (approved):

mappings/net/minecraft/client/world/WorldClient.mapping
```diff
@@ -7,12 +7,16 @@ CLASS bqk net/minecraft/client/world/WorldClient
 	METHOD a playSound (DDDLpb;Lpd;FFZ)V
 	METHOD a getEntityById (I)Lub;
 	METHOD a isChunkLoaded (IIZ)Z
+		ARG 0 chunkX
+		ARG 1 chunkZ
+		ARG 2 consultRemoteProvider
```
exaskye: `allowRemote`? `consultRemoteProvider` is a bit lengthy

CalmBit: Yeah, me and @asiekierka were discussing this one - it's a difficult parameter, because it's essentially an override - if the variable is true, the ChunkProvider isn't consulted on the clientside, with no bearing on the serverside. In hindsight, it probably needs to be renamed anyway, as it's sorta the opposite of what I named it...

CalmBit: I guess `ignoreRemote` could work?

exaskye:
ignoreRemote sounds like the exact opposite of consultRemoteProvider to me. Why was the name inverted?

CalmBit:
`consultRemoteProvider` is the inverse of what it does, it was my initial stupidity >3<

CalmBit:
To elaborate, the basic idea behind the boolean (as far as I can see) is that if it's true, a clientside chunkprovider check will essentially be bypassed, figuring it `||`s the boolean and the chunkprovider call together and returns that value (on the clientside).

asiekierka (approved):

PR #158, opened 2017-05-04
Add mappings for Evoker, Vex, their rendering and some AI

mezz:
Some parameters I added seem to have spanned across many files so this looks bigger than it is.
The main focus was on the Evoker, things spread out from there.


mappings/net/minecraft/entity/ai/AiGoalTrackTarget.mapping
```diff
@@ -0,0 +1,32 @@
+CLASS xu net/minecraft/entity/ai/AiGoalTrackTarget
+	FIELD a checkCanNavigate Z
+	FIELD b canNavigateFlag I
```
CalmBit, line 3: Sort of a nitpick, but is there any reason for this to have the suffix `Flag`? I might also be missing something, because I don't have the mappings applied in front of me...

mappings/net/minecraft/entity/ai/AiGoalTrackTarget.mapping
```diff
@@ -0,0 +1,32 @@
...
+	FIELD c checkCanNavigateCooldown I
+	FIELD d timeWithoutVisibility I
+	FIELD e entity Lus;
+	FIELD f checkVisibility Z
+	FIELD g target Luk;
+	FIELD h maxTimeWithoutVisibility I
```
CalmBit, line 9: A little verbose - ~`maxTimeNotVisible` or something similar?~

**Edit:** Probably not the best example, but I still think it's a little verbose, along with `timeWithoutVisibility` and `checkCanNavigateCooldown`.

mappings/net/minecraft/entity/mob/EntityCaveSpider.mapping
```diff
@@ -1,5 +1,7 @@
 CLASS abk net/minecraft/entity/mob/EntityCaveSpider
 	METHOD J getLootTableId ()Lmd;
 	METHOD a registerTagUpgraders (Lqv;)V
+	METHOD a (Lsv;Lun;)Lun;
```
CalmBit, line 4: Naming a parameter without naming the function was outlined in the style guide as something not to do, although I'm not sure if it's current...

CalmBit (commented):
It's entirely possible I'm slightly out of my league here, but definitely just things to consider in a friendly manner :smile:

mappings/net/minecraft/entity/ai/AiGoalJumpOutOfLiquid.mapping
```diff
@@ -0,0 +1,6 @@
+CLASS vu net/minecraft/entity/ai/AiGoalJumpOutOfLiquid
```
exaskye: Would `AiGoalSwim` be better? Looks like you also mapped a method `canSwim` which is probably related to this goal.

mappings/net/minecraft/entity/EntityWithAiGoals.mapping
```diff
@@ -1,4 +1,17 @@
 CLASS us net/minecraft/entity/EntityWithAiGoals
-	FIELD a pos Ldx;
+	FIELD a aiCenter Ldx;
```
exaskye: Generally I believe this is referred to as a "home", assuming this is what I think it is.

<*insert tirade about Scuttlebugs*>

`aiCenter` → `home`?

mappings/net/minecraft/entity/EntityWithAiGoals.mapping
```diff
@@ -1,4 +1,17 @@
...
+	FIELD b aiRadius F
```
exaskye: `aiRadius` → `homeRadius`?

mappings/net/minecraft/entity/EntityWithAiGoals.mapping
```diff
@@ -1,4 +1,17 @@
...
 	FIELD bv ID Ljava/util/UUID;
+	METHOD <init> (Lali;)V
+		ARG 0 world
+	METHOD a setAiRange (Ldx;I)V
```
exaskye: `setAiRange` → `setHomeRadius`?

mappings/net/minecraft/entity/EntityWithAiGoals.mapping
```diff
@@ -1,4 +1,17 @@
...
+		ARG 0 center
+		ARG 1 radius
 	METHOD cO canSpawn ()Z
+	METHOD dd isInAiRange ()Z
```
exaskye: `isInAiRange` → `isInHomeRadius`? `isInHome`? `isHome`?

mappings/net/minecraft/entity/EntityWithAiGoals.mapping
```diff
@@ -1,4 +1,17 @@
...
+	METHOD de getAiCenter ()Ldx;
```
exaskye: `getAiCenter` → `getHome`?

mappings/net/minecraft/entity/EntityWithAiGoals.mapping
```diff
@@ -1,4 +1,17 @@
...
+	METHOD df getAiRadius ()F
```
exaskye: `getAiRadius` → `getHomeRadius`?

mappings/net/minecraft/entity/EntityWithAiGoals.mapping
```diff
@@ -1,4 +1,17 @@
...
+	METHOD dg setAiRangeUnlimited ()V
```
exaskye: `setAiRangeUnlimited` → `setHomeIgnored`?

mappings/net/minecraft/entity/EntityWithAiGoals.mapping
```diff
@@ -1,4 +1,17 @@
...
+	METHOD dh hasLimitedAiRange ()Z
```
exaskye: `hasLimitedAiRange` → `hasHome`?

mappings/net/minecraft/entity/EntityWithAiGoals.mapping
```diff
@@ -1,4 +1,17 @@
...
+	METHOD f isInAiRange (Ldx;)Z
```
unascribed: `isInAiRange` → `isInHomeRadius`? See other similarly named method for other suggestions

exaskye (commented):
Most of my comments are a "maybe", and you by no means need to listen to me. Looks good :+1:

mappings/net/minecraft/entity/ai/AiGoalTrackTarget.mapping
liach: it is int.

CalmBit: Ah, okay - is it being treated like a mask, then?

marysaka:
Noooooooo I was doing it time to remove it from my side and check the differences

Edit: More seriously, "EntityEvoker" or "EntityEvocation" ? Because actually, apparently entity based classes seems to be named using the EntityRegistry name in pomf mappings

mappings/net/minecraft/entity/ai/AiGoalTrackTarget.mapping
copygirl: Shouldn't it be plural `flags` then?

mezz: It is a value that can be 0, 1, or 2.  It's a strange way to do things, so naming it isn't terribly straightforward. I think `flag` is the right term for this kind of value.

mappings/net/minecraft/entity/ai/AiGoalTrackTarget.mapping
mezz: I tend to make things more verbose when in doubt. I'd rather have a long name than a confusing one. I'm open to ideas for shorter names though.

mappings/net/minecraft/entity/mob/EntityCaveSpider.mapping
mezz: This was one of the parameter names I had difficulty reverting, I must have missed this one. If it's important to remove, I'll fix up this PR when I get a chance later.

mappings/net/minecraft/entity/ai/AiGoalJumpOutOfLiquid.mapping
mezz: I think that makes sense. In the logic the only thing it tries to do is jump so I figured it was trying to get out out the liquid. It's basically the same action as swimming though.

mappings/net/minecraft/entity/EntityWithAiGoals.mapping
mezz: I'll have to look at the code more closely to decide if it's a real home or not. My first interpretation was that it's not, but it's worth looking into.

asiekierka:
@Thog I named them based on that, I welcome changes

mappings/net/minecraft/entity/ai/AiGoalTrackTarget.mapping
asiekierka: I agree with it being "flag" if it's just one multi-state flag as opposed to mutliple flags

modmuss50: `flag` seems good to me

mappings/net/minecraft/entity/ai/AiGoalTrackTarget.mapping
modmuss50: Could be `maxTimeBlind`? but the current one seems fine tbh.

mappings/net/minecraft/entity/ai/AiGoalTrackTarget.mapping
Darkhax: Flag would be correct here, yeah.

modmuss50:
This PR need the `range` to `home` looking at (Even just some input), other than that it should be good.

mezz:
Sorry haven't had time to re-review, will try to get to this tomorrow.

mappings/net/minecraft/entity/EntityWithAiGoals.mapping
mezz: I think `getAiHome` would be best. I want to keep `Ai` because this class is extended by lots of entities. Any time they reference the variable, I think its purpose would be more clear with `Ai` in it.

mappings/net/minecraft/entity/EntityWithAiGoals.mapping
mezz: Similar to `getAiHome`, I want to keep `Ai` in the name. I think this existing name is good.
I reviewed the scuttlebug transportation section of the 0.5 A Press video and "home" is always a single point. The area around the home is separate.

mappings/net/minecraft/entity/mob/EntityCaveSpider.mapping
mezz: This has been fixed in the updated PR

modmuss50 (approved):

exaskye (approved):

Darkhax:
🌔

PR #164, opened 2017-05-27
Render tick calculations and some advancement mappings

mezz:
Started to dig into the advancement mappings but got a bit sidetracked with "delta" ticks.


Darkhax:
Looks good to me

PR #165, opened 2017-06-08
Rebuild method names with enigma "fix"

modmuss50:


asiekierka:
We don't have server mappings in this branch. No point in this PR until Player regenerates mappings for the 1.12 merged JAR.

PR #166, opened 2017-06-21
Map all criterion triggers and some predicates

3TUSK:
All mappings used here are based on my own thinking; [one page on minecraft.gamepedia.com](https://minecraft.gamepedia.com/Advancements#List_of_triggers) is used for references. Comments are welcome.

One side note: one method in `BlockEntityWithLootTable` was removed, and I couldn't find it in the original `1.12.jar`either. Please double check to make sure that this is not a user error, as I didn't touch that class while doing mapping.


modmuss50:
Quickly scan read over it and it seems to be good. I will take a full look later.

exaskye (approved):

modmuss50 (approved):

PR #167, opened 2017-06-28
Map new 1.12 sounds

modmuss50:


asiekierka:
I don't think there's anything worth cross-verifying here, so I'll just merge it. Hope nobody minds!

PR #181, opened 2018-10-31
Map NBT classes

maxpowa:
Also mapped a couple small things in TextComponent/TextFormat since they're used directly by NBT pretty printing.

CompoundWrapper -> PersistedState
baseTag -> key,  it's the key that the tag should be stored as in the parent NBTCompound

### DIFF
--- a/mappings/net/minecraft/nbt/CompoundWrapper.mapping
+++ b/mappings/net/minecraft/nbt/CompoundWrapper.mapping
@@ -11,3 +11,4 @@ CLASS cee net/minecraft/nbt/CompoundWrapper
 		ARG 0 tag
 	METHOD c markDirty ()V
 	METHOD d getDirty ()Z
+	METHOD e getBaseTag ()Ljava/lang/String;
--- a/mappings/net/minecraft/nbt/TagString.mapping
+++ b/mappings/net/minecraft/nbt/TagString.mapping
@@ -9,7 +9,7 @@ CLASS ib net/minecraft/nbt/TagString
 		ARG 2 tracker
 	METHOD a write (Ljava/io/DataOutput;)V
 		ARG 0 output
-	METHOD a prettyPrint (Ljava/lang/String;I)Lix;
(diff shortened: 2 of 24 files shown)


mappings/net/minecraft/nbt/TagString.mapping
```diff
@@ -9,6 +9,12 @@ CLASS ib net/minecraft/nbt/TagString
 		ARG 2 tracker
 	METHOD a write (Ljava/io/DataOutput;)V
 		ARG 0 output
+	METHOD a prettyPrint (Ljava/lang/String;I)Lix;
```
asiekierka: Is prettyPrint a good conventional name to use?

asiekierka: It's not. I'd like to propose toComponent or toTextComponent, with keeping of the latest discussins in mind.

maxpowa: toTextComponent felt kinda weird at the time but prettyPrint does too, I'll go with toTextComponent.

maxpowa:
Comments done

mappings/net/minecraft/nbt/CompoundWrapper.mapping
```diff
@@ -11,3 +11,4 @@ CLASS cee net/minecraft/nbt/CompoundWrapper
 		ARG 0 tag
 	METHOD c markDirty ()V
 	METHOD d getDirty ()Z
+	METHOD e getBaseTag ()Ljava/lang/String;
```
asiekierka: I'm not sure if CompoundWrapper is the best name here. Not sure about getBaseTag either. Could someone look into it?

maxpowa:
We still need to figure out if we want `serialize`/`deserialize` | `toNbt`/`fromNbt` | `toCompound`/`fromCompound` etc...

asiekierka:
https://github.com/FabricMC/pomf/issues/172

asiekierka:
But that's for another day. Let's merge.

PR #183, opened 2018-11-01
Basic tutorial steps mapping

maxpowa:
Would like some input on TutorialManager, it's used in a manager style way but I'm not sure if the name makes sense. The main code that makes me question the name is in the Tutorial Step enums, they're defined like: `(Function<TutorialManager, T>)TutorialStepMovement::new`

### DIFF
--- a/mappings/net/minecraft/client/tutorial/ITutorialStep.mapping
+++ b/mappings/net/minecraft/client/tutorial/ITutorialStep.mapping
@@ -1,0 +1,1 @@
+CLASS dmq net/minecraft/client/tutorial/ITutorialStep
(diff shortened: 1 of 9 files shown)


asiekierka:
I'm pretty sure TutorialManager is the right name within our conventions. If you look at "Show Calls", you'll find that it's only created as a singleton instance in MinecraftGame - this plus the setTutorialStep behaviour is clear enough. This is probably something that we lost in the hacked up 1.12 -> 18w43b move.

setTutorialStep may benefit from being remapped to setStep - we're already operating on a TutorialManager, after all.

I'm concerned about ITutorialStep. That should be named TutorialStep as we're moving away from I-prefixes (or at least I think we are?), but then TutorialStep the enum needs to be called something else - and it does anyway. TutorialSteps? TutorialStepDefinition? TutorialStepFactory? With .a() becoming .create().

asiekierka:
As the commit is small and I want to change most of it anyway, I'll just finish it myself. Thank you for the input, though!

PR #191, opened 2018-11-07
Some structure generation and random stuff

Matrix89:
I don't think this should get merged now, I still need to deobf some names.


mappings/byy.mapping
```diff
@@ -0,0 +1,7 @@
+CLASS byy
+	FIELD a x0 I
```
asiekierka: minX/maxX/... is the convention

mappings/bzn.mapping
```diff
@@ -0,0 +1,6 @@
+CLASS bzn
+	FIELD a logger Lorg/apache/logging/log4j/Logger;
+	FIELD d structurePart Ljava/util/Map;
+	FIELD e partStructure Ljava/util/Map;
+	METHOD a addPart (Ljava/lang/Class;Ljava/lang/String;)V
+	METHOD b registerStructure (Ljava/lang/Class;Ljava/lang/String;)V
```
asiekierka, line 6: Structure or Feature? (I'm not sure)

mappings/net/minecraft/sortme/structures/DesertTemple.mapping
```diff
@@ -0,0 +1,6 @@
+CLASS bza net/minecraft/sortme/structures/DesertTemple
```
asiekierka: DesertTempleStructure or StructureDesertTemple (again, no clear standard yet). not DesertTemple alone

asiekierka (changes requested):

mappings/bzn.mapping
Matrix89: These classes don't inherit form Feature only generate it, so more like FeatureGenerator

mappings/net/minecraft/sortme/structures/DesertTemple.mapping
Matrix89: Yep, named it only to keep track of it

mappings/bzn.mapping
asiekierka: FeatureFactory?

mappings/net/minecraft/sortme/structures/DesertTemple.mapping
asiekierka: OK

mappings/bzn.mapping
```diff
@@ -0,0 +1,6 @@
+CLASS bzn
+	FIELD a logger Lorg/apache/logging/log4j/Logger;
```
jamierocks, line 2: These fields are static, they should be upper-case.

```java
    private static final Logger a = LogManager.getLogger();
    private static final Map<String, Class<? extends bzr>> b = Maps.newHashMap();
    private static final Map<Class<? extends bzr>, String> c = Maps.newHashMap();
    private static final Map<String, Class<? extends bzq>> d = Maps.newHashMap();
    private static final Map<Class<? extends bzq>, String> e = Maps.newHashMap();
```

jamierocks (changes requested):

modmuss50:
Looks good, just need to fix that logger name.

I have a feeling this has a bit of overlap with #194

asiekierka:
-> 18w46a

PR #192, opened 2018-11-08
more random mappings and other changes

NikkyAI:
CommandAdvamcement,
onPlayerConnect,
onChatMessage


jamierocks (approved):
Looks fine to me, though I'm thinking `net/minecraft/command` should be changed to `net/minecraft/server/command`.

NikkyAI:
maybe for some commands where `isDedicated` is checked, but the advancement command is registered both sides

mappings/net/minecraft/server/config/ServerConfigurationManager.mapping
```diff
@@ -45,8 +45,13 @@ CLASS wg net/minecraft/server/config/ServerConfigurationManager
 	FIELD v latencyUpdateTimer I
 	METHOD a reloadWhitelist ()V
 	METHOD a setViewDistance (I)V
+	METHOD a sendToAllPlayers (Lapv;Liz;)V
 	METHOD a setGameMode (Lazv;)V
 	METHOD a addToOperators (Lcom/mojang/authlib/GameProfile;)V
+	METHOD a onPlayerConnect (Lin;Luf;)V
+		ARG 1 connection
+		ARG 2 player
+	METHOD a onChatmessage (Liz;Z)V
```
asiekierka: onChatMessage, and I assume you've mapped the relevant packets if any?

NikkyAI: the packets were mapped already, thats how i traversed this and found this..
i am trying to only map stuff i am 100% certain about

to complete all necessary hooks

jamierocks:
> maybe for some commands where isDedicated is checked, but the advancement command is registered both sides

I'm not sure you understood me, I mean move the whole `net/minecraft/command` package under the `net/minecraft/server/command` package - because commands are tied to the server, integrated or otherwise.

Though that's not within the scope of this PR, just a point to think about.

asiekierka:
@jamierocks I already opened an issue for this. However, the command classes are also used on the client side now - for suggestions!

jamierocks:
> @jamierocks I already opened an issue for this. However, the command classes are also used on the client side now - for suggestions!

The client contains a server too, commands are only present with a server present. The `net/minecraft/server` package is certainly not exclusive to the dedicated server (e.g. `MInecraftServer`).

asiekierka:
@jamierocks Commands themselves yes, but net.minecraft.client.network.handler.ClientGameNetworkHandler contains a CommandDispatcher, which is populated with a command tree downloaded from the server side, and presumably some of the classes in command/ may be involved, such as the CommandSource interface.

NikkyAI:
should i split this into two PRs ? one for just mappings and one for everything else buildscript related ?

.editorconfig
```diff
@@ -0,0 +1,5 @@
+[*.gradle]
+indent_style = tab
+
+[*.java]
```
modmuss50, line 4: There are no java files in this repo, not an issue having this here, but no reason to have it

modmuss50 (commented):
Looks good, the minor java editor config thing could go.

.editorconfig
ProfessorProspector: Interesting...no idea where that came from

PR #200, opened 2018-11-22
[WIP] Move prefixes to suffixes

asiekierka:
TODO:

* Small classes in .world
* .network.packet
* .client.render.model

PR #201, opened 2018-11-22
Resolve #199, and Ocelot/Cat

### LINKED ISSUE #199: EntityFactory -> EntityType; EntityType -> EntityCategory
EntityFactory just makes little sense...a factory should build entities not be the actual type of entity. Also mojang calls it EntityType.

EntityType is ok, but I think EntityCategory is a more fitting name. This is the class that defines an entity as either MONSTER, CREATURE, AMBIENT, or WATER_CREATURE.

jamierocks:
Just a minor fix for the Ocelot class :)


asiekierka:
Resolving #199 has already been done in #200 which is probably a bit of a conflict.

jamierocks:
That's fine, I'll rebase this pr when #200 is merged :)

PR #202, opened 2018-11-22
A couple of minor changes to Entity

SquidDev:
- Rename `.getCommandStats` to `.getWorld`
 - Add .getPos

`EntityVillager` define both these methods already (with the same definition, and the `@Override` annotation), so I believe these are acceptable choices.


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -197,7 +197,8 @@ CLASS age net/minecraft/entity/Entity
 	METHOD bE getComponentHoverEvent ()Ljc;
 	METHOD bF getBoundingBox ()Lckc;
 	METHOD bH getEyeHeight ()F
-	METHOD bL getCommandStats ()Lbam;
+	METHOD bJ getPos ()Les;
```
asiekierka, line 200: getBlockPos, or maybe getPosVec on the one below. Do look into how bK interacts with this.

PR #203, opened 2018-11-23
Add some mappings for GUIs and input handling

SquidDev:
- Add keyboard, mouse and clipboard classes. The names and locations are terrible. It may be worth having a `client.input` package or something.
 - Flesh out the GuiEventListener mappings.

I'm happy to remove some things - this was just the end result of a mod-porting session, so not everything may be entirely sane.

### DIFF
--- a/mappings/net/minecraft/client/gui/widget/WidgetListVideoSettings.mapping
+++ b/mappings/net/minecraft/client/gui/widget/WidgetListVideoSettings.mapping
@@ -6,11 +6,15 @@ CLASS cox net/minecraft/client/gui/widget/WidgetListVideoSettings
 		METHOD a drawEntry (IIIIZF)V
 		METHOD mouseClicked (DDI)Z
 			ARG 0 mouseX
-			ARG 1 mouseY
+			ARG 1 mouseX
 			ARG 2 keyCode
+			ARG 3 mouseY
+			ARG 5 modifiers
 		METHOD mouseReleased (DDI)Z
 			ARG 0 mouseX
-			ARG 1 mouseY
+			ARG 1 mouseX
 			ARG 2 keyCode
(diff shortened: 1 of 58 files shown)


mappings/net/minecraft/client/gui/widget/WidgetListVideoSettings.mapping
```diff
@@ -6,11 +6,15 @@ CLASS cox net/minecraft/client/gui/widget/WidgetListVideoSettings
 		METHOD a drawEntry (IIIIZF)V
 		METHOD mouseClicked (DDI)Z
 			ARG 0 mouseX
-			ARG 1 mouseY
```
SquidDev, line 9: I'm really not sure what's going on here - it's like one tool was started at 0, and another at 1. Similarly why there's a `keyCode` argument here at all.

I'm happy to clean this up, just would like someone else to look over it first.

mappings/net/minecraft/util/ChatCharacterFilter.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS k
+CLASS k net/minecraft/util/ChatCharacterFilter
```
jamierocks: Judging from realm's `RealmsSharedConstants`, this is more likely `net/minecraft/SharedConstants`. It'll be the obfuscator removing fields that are inlined, as to why it doesn't fell much like a typically constants class.

jamierocks (commented):
Looks good :+1:

jamierocks (approved):

jamierocks:
Just need to look into the duplicate `mouseX` arguments.

asiekierka:
That's likely to be Enigma bugfix artifacts, don't worry about it.

PR #204, opened 2018-11-24
Add mappings  for most of the datafixer code

SquidDev:
This was partly a way for me to dig into some of the guts of the data-fixing code, but there are some useful transformations here too.

Note that the `DataFix` class names are mostly copied verbatim from the source code, so may not be the most consistent with the rest of the codebase. However, if #200 is merged, I don't think this'll be an issue.


Prospector:
No worries about suffixing. A lot of stuff is already suffixed

jamierocks (approved):
:+1:

Issue #193, opened 2018-11-08
net/minecraft/command should be changed to net/minecraft/server/command


asiekierka:

asiekierka:
https://github.com/FabricMC/pomf/tree/18w47b

PR #206, opened 2018-11-27
Map argument types

jamierocks:
*This is somewhat of a follow-up to #193*.

The arguments types are present on both the client and server, therefore they do not need to be in the `net/minecraft/server` package.


modmuss50 (approved):
Looks good

PR #208, opened 2018-11-27
Flesh out several "list" classes

SquidDev:
I realise this is very low-hanging fruit, but it was bugging me and it sorta makes some deobfuscated sources easier to understand.

 - Some fields of `Material`
 - `Blocks`
 - `Items` (and some item classes)
 - Some fields in `BlockSoundGroup`
 - `Stats`
 - `RecipeSerializers`
 - `StatusEffects`


Prospector:
hey, someone had to do the lowhanging fruit. However note that the status effects and stats should be mapped by asie's thing automatically in dev envs, just not in pomf yet

jamierocks (approved):
Wow!

Though you should've mentioned this in IRC - there is WIP tooling to automatically do this sort of work.

Prospector (approved):

SquidDev:
Ahhh, neat! I remember asie mentioning something, but wasn't sure how far along it was.

asiekierka:
``` However note that the status effects and stats should be mapped by asie's thing automatically in dev envs, just not in pomf yet```

Yes, we're generally not merging that and it should be dropped from the PR. particularly Blocks and Items and the other ones with string-based registries.

PR #213, opened 2018-11-28
A random assortment of things

SquidDev:
I'm sorry for the random assortment of changes: this was mostly me going through any unmapped methods/classes I was using and trying to find something for them.

I've tried to keep each commit as atomic as possible, so if you'd prefer, you can cherry-pick individual ones in. There's definitely a few here I'm less sure about, so suggestions are welcome.

### DIFF
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -209,6 +209,7 @@ CLASS bfo net/minecraft/block/Block
 	METHOD d isSimpleFullBlock (Lboh;Lazz;Let;)Z
 	METHOD e isFullBoundsCubeForCulling (Lboh;)Z
 	METHOD e getMapColor (Lboh;Lazz;Let;)Lcde;
+	METHOD f getFluidState (Lboh;)Lccz;
 	METHOD g hasBlockEntity ()Z
 	METHOD g emitsRedstonePower (Lboh;)Z
 	METHOD g canSuffocate (Lboh;Lazz;Let;)Z
--- a/mappings/net/minecraft/block/Waterlogged.mapping
+++ b/mappings/net/minecraft/block/Waterlogged.mapping
@@ -1,1 +1,0 @@
-CLASS bkg net/minecraft/block/Waterlogged
--- a/mappings/net/minecraft/block/Waterlogged.mapping
+++ b/mappings/net/minecraft/block/Waterlogged.mapping
@@ -1,0 +1,1 @@
+CLASS bkg net/minecraft/block/Waterlogged
--- a/mappings/net/minecraft/client/gui/hud/ChatHud.mapping
+++ b/mappings/net/minecraft/client/gui/hud/ChatHud.mapping
@@ -5,7 +5,9 @@ CLASS cob net/minecraft/client/gui/hud/ChatHud
 	METHOD a addMessage (Lja;)V
 	METHOD a addMessage (Lja;I)V
 		ARG 1 message
+	METHOD a addMessage (Lja;IIZ)V
 	METHOD a clear (Z)V
+	METHOD b removeChatMessage (I)V
 	METHOD d isChatFocused ()Z
 	METHOD e getWidth ()I
 	METHOD f getHeight ()I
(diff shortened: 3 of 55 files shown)


mappings/net/minecraft/util/math/Polar2f.mapping
```diff
@@ -1,5 +1,7 @@
 CLASS cjx net/minecraft/util/math/Polar2f
+	FIELD a FORWARDS Lcjx;
```
asiekierka: > I'm really unsure about this name, so comments welcome!

mappings/net/minecraft/client/gui/hud/InGameHud.mapping
```diff
@@ -17,8 +21,13 @@ CLASS cnr net/minecraft/client/gui/hud/InGameHud
 	FIELD w hudBossBar Lcnz;
 	METHOD a draw (F)V
 	METHOD a addChatMessage (Liy;Lja;)V
+	METHOD a setOverlayMessage (Lja;Z)V
+	METHOD a setPlayingOverlay (Ljava/lang/String;)V
```
asiekierka: Is this the best name?

mappings/net/minecraft/block/Block.mapping
```diff
@@ -209,6 +209,7 @@ CLASS bfo net/minecraft/block/Block
 	METHOD d isSimpleFullBlock (Lboh;Lazz;Let;)Z
 	METHOD e isFullBoundsCubeForCulling (Lboh;)Z
 	METHOD e getMapColor (Lboh;Lazz;Let;)Lcde;
+	METHOD f getFluidState (Lboh;)Lccz;
```
asiekierka, line 212: I'm not too sure on it. I've been trying to think of a better name for a long time.

mappings/net/minecraft/block/Waterlogged.mapping
```diff
@@ -0,0 +1 @@
+CLASS bkg net/minecraft/block/Waterlogged
```
jamierocks: `WaterloggedBlock`?

mappings/net/minecraft/client/gui/hud/ChatHud.mapping
```diff
@@ -5,7 +5,9 @@ CLASS cob net/minecraft/client/gui/hud/ChatHud
 	METHOD a addMessage (Lja;)V
 	METHOD a addMessage (Lja;I)V
 		ARG 1 message
+	METHOD a addMessage (Lja;IIZ)V
 	METHOD a clear (Z)V
+	METHOD b removeChatMessage (I)V
```
jamierocks: To be consistent with `addMessage`, `removeMessage`?

mappings/net/minecraft/block/Waterlogged.mapping
asiekierka: This is an interface, *not* a class.

asiekierka: I say Waterloggable.

mappings/net/minecraft/client/gui/hud/InGameHud.mapping
asiekierka: Specifically, what does it refer to? Jukeboxes?

SquidDev: Yep. It's the "now playing: 13" label you get above the hotbar.

asiekierka: setPlayingOverlay sounds wrong to me... it's either too vague or insufficiently so.

mappings/net/minecraft/util/math/Polar2f.mapping
kashike: This class is a generic two-dimensional float vector - this field should be `ZERO`, instance fields should be `x` and `y`.

liach:
i recall that playing overlay method's hotbar overlay has rainbow color while regular overlays from chat packets are white

SquidDev:
@liach Yep! `setRecordPlayingOverlay` just calls `setOverlayMessage` with the second parameter set to true (which then will apply a tint).

Prospector (approved):

PR #215, opened 2018-11-28
Improve renderer mappings

jamierocks:
*Dispatcher, DefaultEntityRenderer:
- See https://notch.tumblr.com/post/581187056/evolution-of-the-entityrenderdispatcher

Cube, Polygon:
- See the very old (RubyDung) versions pre-obfuscation
- Cube is used in the Bedrock entity model JSONs, see https://minecraft.gamepedia.com/Tutorials/Changing_Minecraft_entity_models

### DIFF
--- a/mappings/cvs.mapping
+++ b/mappings/cvs.mapping
@@ -2,6 +2,7 @@ CLASS cvs
 	METHOD <init> (FFII)V
 		ARG 1 scale
 		ARG 3 textureWidth
+		ARG 4 textureHeight
 	METHOD a setRotationAngles (FFFFFFLagf;)V
 		ARG 1 limbMoveAngle
 		ARG 2 limbMoveAmount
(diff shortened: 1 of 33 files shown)


mappings/net/minecraft/client/render/entity/EntityRendererDispatcher.mapping
```diff
@@ -1,18 +1,21 @@
-CLASS dfy net/minecraft/client/render/entity/EntityRenderManager
+CLASS dfy net/minecraft/client/render/entity/EntityRendererDispatcher
```
asiekierka: Should we thus rename BlockEntityRenderManager to BlockEntityRenderDispatcher likewise?

asiekierka: Oh. Haha.

Prospector (approved):

PR #216, opened 2018-11-28
switch around some predicate stuff

kashike:


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -64,7 +64,7 @@ CLASS agf net/minecraft/entity/Entity
 	METHOD B_ getPistonBehavior ()Lcdf;
 	METHOD C_ shouldBroadcastConsoleToOps ()Z
 	METHOD E updatePortalCooldown ()V
-	METHOD Q getDisplayedName ()Lja;
+	METHOD Q getDisplayName ()Lja;
```
kashike, line 67: `Display` over `Displayed` as it's what will be be used as the display name assuming nothing changes it, not the currently displayed name.

jamierocks (approved):
:+1:

mappings/net/minecraft/world/biome/Biome.mapping
```diff
@@ -58,7 +69,7 @@ CLASS bbm net/minecraft/world/biome/Biome
 		FIELD d NAME_MAP Ljava/util/Map;
 		FIELD e name Ljava/lang/String;
 		METHOD a getName ()Ljava/lang/String;
-	CLASS bbm$e SpawnListEntry
+	CLASS bbm$e SpawnEntry
```
Prospector, line 72: Personally I like the name EntitySpawn...Entry seems a bit unnecessary

mappings/net/minecraft/world/biome/Biome.mapping
```diff
@@ -82,15 +93,17 @@ CLASS bbm net/minecraft/world/biome/Biome
...
 	FIELD t structureFeatures Ljava/util/Map;
-	FIELD u entitySpawnMap Ljava/util/Map;
+	FIELD u spawns Ljava/util/Map;
+	METHOD <init> (Lbbm$a;)V
+		ARG 1 builder
 	METHOD a addDefaultFeatures ()V
 	METHOD a getSkyColor (F)I
 	METHOD a getEntitySpawnList (Lagq;)Ljava/util/List;
-	METHOD a addEntitySpawnEntry (Lagq;Lbbm$e;)V
+	METHOD a addSpawn (Lagq;Lbbm$e;)V
```
Prospector, line 106: addEntitySpawn IMO

mappings/net/minecraft/world/biome/Biome.mapping
asiekierka: If it's on a List, I think Entry is reasonable.

mappings/net/minecraft/world/biome/Biome.mapping
asiekierka: I used EntitySpawnEntry because:

* Just "Spawn" would be a bit vague for my tastes.
* "SpawnEntry" alone could imply spawning something else than entities (block structures? flowers? trees?)

Prospector: Yeah, that why I suggested addEntitySpawn. To me, 'add' already implies a list so Entry is redundant

mappings/net/minecraft/world/biome/Biome.mapping
Prospector: Sure, then EntitySpawnEntry

Prospector: However, I'm not certain it would necessarily have to be on a list. You could maybe use it in other contexts... though I'm not sure what.

PR #222, opened 2018-11-29
Rename several methods to be suffixed

SquidDev:
- `*TagList` -> `*ListTag`
 - `*TagCompound` -> `*CompoundTag`
 - Several mappings within the `ServerCommandSource`.

Some things to discuss:

 - Do we need `Tag` on methods within the `TagList` classes?
 - Do we need `Server` on methods within `ServerCommandSource`?

### DIFF
--- a/mappings/net/minecraft/client/hotbar/HotbarEntry.mapping
+++ b/mappings/net/minecraft/client/hotbar/HotbarEntry.mapping
@@ -1,4 +1,4 @@
 CLASS dbp net/minecraft/client/hotbar/HotbarEntry
 	FIELD a delegate Lfh;
-	METHOD a toTagList ()Lhv;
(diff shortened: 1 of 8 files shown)


Prospector:
I'd say no to both questions

Prospector:
Wait... I change my mind. Yes to `Do we need Tag on methods within the TagList classes?` and no to `Do we need Server on methods within ServerCommandSource?`

asiekierka:
No to both. nbt.ListTag is already in the "NBT" context, so we can assume future entries being "NBT"-context is implied.

Prospector:
IMO having nbt.getInt and the others right alongside nbt.getList could be a little confusing, that's why I said it should have the tag suffix

asiekierka:
I sort-of disagree, but I'll get more opinions in.

PR #229, opened 2018-11-30
Assorted mappings for commands

SquidDev:
Changes:
 - `ArgumentType` serialisers.
 - Exception fields and helper methods within `ArgumentType`s
 - Entity selectors and it's associated helper classes. I'm not sure if the current package (`net.minecraft.commands`) is the correct place.
 - Command source methods
 - `net.minecraft.text.TextFormatter` - a helper class for `TextComponents`. I'm really unhappy of the names in here, so please do suggest alternatives!


mappings/net/minecraft/command/EntitySelector.mapping
```diff
@@ -0,0 +1,22 @@
...
+	FIELD j playerName Ljava/lang/String;
+	FIELD k entityId Ljava/util/UUID;
+	FIELD l type Ljava/lang/Class;
+	FIELD m checkPermissions Z
+	METHOD a getLimit ()I
+	METHOD a getEntity (Lca;)Lagi;
+	METHOD a getPositionPredicate (Lckp;)Ljava/util/function/Predicate;
+	METHOD a getEntities (Lckp;Ljava/util/List;)Ljava/util/List;
+	METHOD a getNames (Ljava/util/List;)Lja;
+	METHOD b getIncludesEntities ()Z
```
asiekierka: We try to use is/should/can for booleans.

SquidDev: Is `doesIncludeEntities` OK? I had a fiddle with `has` and `is`, but wasn't sure either worked.

mappings/net/minecraft/command/arguments/OperationArgumentType.mapping
```diff
@@ -1,3 +1,10 @@
 CLASS cs net/minecraft/command/arguments/OperationArgumentType
+	CLASS cs$a Operator
+	CLASS cs$b IntOperator
 	FIELD a EXAMPLES Ljava/util/Collection;
+	FIELD b INVALID_OPERATION Lcom/mojang/brigadier/exceptions/SimpleCommandExceptionType;
+	FIELD c DIV_0_EXCEPTION Lcom/mojang/brigadier/exceptions/SimpleCommandExceptionType;
```
asiekierka: DIVISION_ZERO_EXCEPTION?

mappings/net/minecraft/server/command/ServerCommandSource.mapping
```diff
@@ -4,23 +4,47 @@ CLASS ca net/minecraft/server/command/ServerCommandSource
...
 	FIELD i minecraftServer Lnet/minecraft/server/MinecraftServer;
+	FIELD j silent Z
 	FIELD k entity Lagi;
 	FIELD l resultConsumer Lcom/mojang/brigadier/ResultConsumer;
+	FIELD m entityAnchor Lch$a;
+	FIELD n rotation Lcko;
+	METHOD a withSilent ()Lca;
+	METHOD a withLevel (I)Lca;
 	METHOD a withEntity (Lagi;)Lca;
+	METHOD a withLookingAt (Lagi;Lch$a;)Lca;
```
asiekierka, line 19: withLookingAt sounds a bit weird... Is it a good name?

mappings/net/minecraft/command/EntitySelector.mapping
asiekierka: What about simply includesEntities?

mappings/net/minecraft/server/command/ServerCommandSource.mapping
SquidDev: I wasn't sure how important the `with` prefix was here. It pretty returns a new command source with the rotation vector set so you're looking at the provided entity or position.

asiekierka: withLookingAt it is then

modmuss50 (approved):
Looks great. 👍

PR #230, opened 2018-12-01
map baked model loader-related classes

asiekierka:


SquidDev:
I wonder if `getTextureDependencies` and `getModelDependencies` would be a set of names? They're rather long, but `getDependencies` feels a little a little ambiguous as it just returns `Identifier`s.

asiekierka:
Yes, that's probably a better idea.

modmuss50 (approved):
Looks fine to me, I dont know much about the model system so it might be best to another option on the names. But like I said nothing jumps out at me.

PR #235, opened 2018-12-03
various mappings

kashike:


asiekierka:
This is almost entirely schemefixes or extensions of existing conventions, so I think it wouldn't be a stretch for me to just merge it.

PR #240, opened 2018-12-04
Lots of loot table work

SquidDev:
I just wanted to drop items, and then ended up a lot further down the rabbit hole than I meant to.


liach:
I suggest you wait a bit on this. In loot tables there are complex inheritance that some classes inherit a method from both superclasses and interfaces with the same name, which enigma currently fails to handle.

asiekierka:
"wait a bit on this"

First of all, a quick run of tiny-remapper should detect naming conflicts, if any exist. Just run "gradle build" and use the resulting .tiny.gz to map a merged JAR from "official" to "named".

Second, we can't really wait.

liach:
Well, I was not aware of the tiny-remapper. Now I am.

asiekierka:
It's fine.

Our plan in the longer term is to add automatic tests to the buildscript which do exactly that.

SquidDev:
liach was right in there being conflicts. I've manually patched it, but don't know if this is OK, or if it'd be better to remove that mapping for now. tiny-remapper is no longer complaining, but I don't know if there are any other verification steps I can take.

asiekierka:
This is the best step. There's also findMappingErrors in Weave which finds class visibility issues - I need to go through *that* at some point as to ditch the PublicAccessTransformer hack.

PR #246, opened 2018-12-07
Bit of work

Prospector:
This ended up being much less than I thought it was but whatever. Also woooo PR 200


mappings/net/minecraft/entity/vehicle/MinecartBaseEntity.mapping
```diff
@@ -1,5 +1,12 @@
 CLASS arp net/minecraft/entity/vehicle/MinecartBaseEntity
 	CLASS arp$a Type
+		FIELD a RIDEABLE Larp$a;
```
asiekierka: Get rid of these - as I said, I'll add a way to hook up enum data soon-ish.

asiekierka:
Other than that, LGTM <3

modmuss50 (approved):

PR #251, opened 2018-12-10
singular names

kashike:
Fields match the registry names, as they do/will in `Blocks`, `Items`, etc.
When you read it: "block registry", "item registry", etc


mappings/net/minecraft/util/registry/Registry.mapping
```diff
@@ -1,33 +1,33 @@
...
-	FIELD r DECORATORS Lfk;
-	FIELD s BIOMES Lfk;
-	FIELD t PARTICLE_TYPES Lfk;
-	FIELD u BIOME_SOURCE_TYPES Lfk;
-	FIELD v BLOCK_ENTITIES Lfk;
-	FIELD w CHUNK_GENERATOR_TYPES Lfk;
-	FIELD x DIMENSIONS Lfk;
-	FIELD y MOTIVES Lex;
-	FIELD z CUSTOM_STATS Lfk;
+	FIELD g SOUND Lfk;
```
kashike: Unsure about this. The registry name is `sound_event`, but we use `Sound` for the class name - this should change, since it isn't an actual sound, it's the event that plays *a* sound.

asiekierka:
I have to say, I'm not a fan. But I'll let a discussion take place.

mappings/net/minecraft/util/registry/Registry.mapping
```diff
@@ -1,33 +1,33 @@
...
+	FIELD h FLUID Lex;
+	FIELD i STATUS_EFFECT Lfk;
```
kashike, line 13: Unsure about this. The registry name is `mob_effect`, but we use `StatusEffect` for the class name.

Prospector:
I suggested the name change to status effect because it makes more sense than mob effect and enough people agreed for it to go thru

mappings/net/minecraft/util/registry/Registry.mapping
Prospector: It should change to SoundEvent then

mappings/net/minecraft/util/registry/Registry.mapping
liach: StatusEffect is definitely better, since mobs mean entities with ai while effects are applicable to all livings

kashike: I'd have suggested `EntityEffect`.

Player3324:
AFAIK the correct (English) term would be "block registry", not "blocks registry".

"block registry" -> "BlockRegistry" -> "RegistryBlock" -> "Registry.BLOCK" is imo a reasonable transformation, so the change makes sense.

mappings/net/minecraft/util/registry/Registry.mapping
modmuss50: I agree change the class + this field

modmuss50:
Overall I like the changes, just a few that could do with better class/field names.

mappings/net/minecraft/util/registry/Registry.mapping
asiekierka: Yes.

mappings/net/minecraft/util/registry/Registry.mapping
asiekierka: I have no strong opinion one way or the other, *but* StatusEffect would be the Minecraft wiki term among others.

jamierocks (approved):

liach:
Beware, the sound instances in yarn is called "SoundEvent". Consider changing that.

Prospector:
I guess it depends on how you read it. You could read it as Block Registry or Registry of Blocks. Block Registry is probably the more proper english way... so I guess singular makes sense. I could also see Registry of Blocks just as well though.

modmuss50 (approved):

PR #252, opened 2018-12-10
Facing -> Direction

kashike:
This is not just used to get the "facing", it's generic direction.

### DIFF
--- a/mappings/net/minecraft/util/math/Direction.mapping
+++ b/mappings/net/minecraft/util/math/Direction.mapping
@@ -1,4 +1,4 @@
-CLASS ey net/minecraft/util/math/Facing
+CLASS ey net/minecraft/util/math/Direction
 	CLASS ey$a Axis
 		CLASS ey$a$1
 			METHOD <init> (Ljava/lang/String;ILjava/lang/String;)V
@@ -58,7 +58,7 @@ CLASS ey net/minecraft/util/math/Facing
 		METHOD b isVertical ()Z
 		METHOD c isHorizontal ()Z
 		METHOD m asString ()Ljava/lang/String;
-	CLASS ey$b Direction
+	CLASS ey$b AxisDirection
 		FIELD a POSITIVE Ley$b;
 		FIELD b NEGATIVE Ley$b;
 		FIELD c offset I
--- a/mappings/net/minecraft/state/property/DirectionProperty.mapping
+++ b/mappings/net/minecraft/state/property/DirectionProperty.mapping
@@ -1,12 +1,13 @@
-CLASS bpg net/minecraft/state/property/FacingProperty
+CLASS bpg net/minecraft/state/property/DirectionProperty
 	METHOD <init> (Ljava/lang/String;Ljava/util/Collection;)V
 		ARG 1 name
+		ARG 2 values
 	METHOD a create (Ljava/lang/String;Ljava/util/Collection;)Lbpg;
 		ARG 0 name
-		ARG 1 validValues
+		ARG 1 values
 	METHOD a create (Ljava/lang/String;Ljava/util/function/Predicate;)Lbpg;
 		ARG 0 name
-		ARG 1 valuePredicate
+		ARG 1 predicate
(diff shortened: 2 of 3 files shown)


Player3324:
While facing may also be interpreted as "facing of a direction vector", making the existing name somewhat legitimate, direction is clearly the more direct/descriptive name.

"Side" just became available and should also be considered.

MC doesn't use "facings" uniformly, they could mean a "direction" offset to apply to e.g. a coordinate, a "side" of a block to interact with, the "originating side"/"direction" or the "target side"/"inverse direction". Note how some are inverses of others depending on a specific method's use (inverse sign).

With the many "at facing" uses around blocks, I currently tend to slightly favor Side over Direction. Using args like "Side toSide" when the facing is actually a direction appears nicer than "Direction side" or "Direction dirTo[wards]" when the facing is actually a side.

asiekierka:
Side is used by Terasology for similar purposes, by the way.

modmuss50:
Both Side and Direction are better than Facing imo. Now that side isn't taken up with the classes I think it could be a good contender.

Either way I would be be happy with this change, might need some more input from others.

asiekierka:
Side > Facing = Direction IMO.

kashike:
I can't say I am a fan of `Side`, especially side it has *such* a different meaning in the Forge community.

It is worth noting that:
- "direction" is a well-known term
  - north, east, south, and west are cardinal directions, not cardinal sides
  - you don't generally say to someone "hey, look in that side!", you say "hey, look in that direction!"
- one of the methods in this class throw an exception with message `No such direction: ...`.

jamierocks (approved):

Prospector (approved):

Prospector:
I agree with Direction. You face a direction. A side of a block faces a direction. A direction is not a face or a side.

asiekierka:
I'm not sure if "up/top" or "down/bottom" really make me think of a Direction.

I'll wait for a few more comments with this, but I guess Direction really is better...

kashike:
When I ask someone where I should look, I ask "Which direction should I be looking?", not "Which side should I be looking?"
`UP` is not a side, nor is `DOWN`.

asiekierka:
Okay. Fair.

I'll delay merging this to just before 18w50a, though. Same with #251, which is also accepted FWIW.

mappings/net/minecraft/util/math/Direction.mapping
```diff
@@ -58,7 +58,7 @@ CLASS ey net/minecraft/util/math/Facing
 		METHOD b isVertical ()Z
 		METHOD c isHorizontal ()Z
 		METHOD m asString ()Ljava/lang/String;
-	CLASS ey$b Direction
+	CLASS ey$b AxisDirection
```
liach, line 61: Any better name for this? I initially called this direction because I don't know if there's such a word that expresses if a scalar/number is positive or negative.

asiekierka: In Polish, we have "kierunek" and "zwrot" for the two. English does not really have the concept - some people use "magnitude" to refer to the latter, but it's kind of artificial in a binary context.

Sign?

jamierocks: > I don't know if there's such a word that expresses if a scalar/number is positive or negative

I would have said Polarity, but apparently that's a Chemistry word :S 

Anyhow should probably be `Direction.Axis` not `Direction.AxisDirection`.

Vexatos: In Mathematics, the word for this would be "orientation". It would definitely be the best word, but `AxisOrientation` might be a bit long.

Vexatos:
Semantically, "direction" makes more sense. "Side" doesn't fit given the names of the fields.

modmuss50 (approved):

JamiesWhiteShirt:
I think Direction is a good fit. I feel Side is too bound to blocks whereas Direction is more universal.

PR #253, opened 2018-12-10
Map various obfuscated names

dblsaiko:


mappings/net/minecraft/client/gui/hud/DebugHud.mapping
```diff
@@ -1,4 +1,42 @@
 CLASS cph net/minecraft/client/gui/hud/DebugHud
 	FIELD a client Lcof;
 	FIELD f fontRenderer Lcou;
+	FIELD g blockHit Lckw;
+	FIELD h fluidHit Lckw;
+	FIELD i pos Lbag;
+	FIELD j chunk Lbql;
+	METHOD <init> (Lcof;)V
+		ARG 1 client
+	METHOD a interpColor (IIF)I
```
asiekierka: interpolateColor

mappings/net/minecraft/world/chunk/IChunk.mapping
```diff
@@ -1,7 +1,9 @@
-CLASS bpz
+CLASS bpz net/minecraft/world/chunk/IChunk
```
asiekierka: We don't use I-prefixes... but then there's IWorld for lack of a better name.

mappings/net/minecraft/world/chunk/ChunkReadOnly.mapping
```diff
@@ -1,13 +1,19 @@
-CLASS bqk
+CLASS bqk net/minecraft/world/chunk/ChunkReadOnly
```
asiekierka: ReadOnlyChunk

mappings/net/minecraft/client/resource/ResourceIndexDirect.mapping
```diff
@@ -0,0 +1,14 @@
+CLASS dmh net/minecraft/client/resource/ResourceIndexDirect
```
asiekierka: DirectResourceIndex. Suffixes!

asiekierka (changes requested):

mappings/net/minecraft/world/chunk/IChunk.mapping
dblsaiko: Yeah, I saw.. and I really do not like I prefixes either, but since Chunk was already taken, I didn't know of a better name to call it. ChunkBase, maybe?

asiekierka: If it's getter-only, ChunkView. I think we have another ChunkView but it can be renamed?

dblsaiko: It's not read only, that's ReadOnlyChunk :P

This is the base interface for chunks, which the other classes like EmptyChunk, Chunk, ReadOnlyChunk implement

asiekierka: Rename bpz to Chunk and Chunk to something else. Maybe WorldChunk?

modmuss50:
Everything else apart from that chunk class looks fine.

dblsaiko:
~~How should I call it now, then?~~

ah, nevermind, missed asie's comment. That'll work

modmuss50:
@asiekierka 's suggestion was to rename bpz to Chunk and then rename the current chunk to WorldChunk. I think this is fine as bpz provides most of the common methods people using when accessing chunks.

PR #254, opened 2018-12-10
fix ParticleManager, map LookupManager and related

asiekierka:


mappings/net/minecraft/client/lookup/IdentifierSearchableContainer.mapping
```diff
@@ -0,0 +1,13 @@
+CLASS doj net/minecraft/client/lookup/IdentifierSearchableContainer
```
ChloeDawn: This name is pretty odd. Something like `IndexedSearchableContainer` or `MappedSearchableContainer` would be more appropriate imo.

ChloeDawn (changes requested):

mappings/net/minecraft/client/lookup/IdentifierSearchableContainer.mapping
modmuss50: I think Identifier is fine imo

PR #267, opened 2018-12-11
Mapping a lot of block related classes

dblsaiko:
Most of Block + BlockState is now mapped, and some other stuff.

Also I need a better name for `VerticalEntityPosition`/`VerticalEntityPositionImpl`, but I can't come up with one. It consists of the player's Y position and if the player is sneaking or not.


mappings/net/minecraft/block/BlockState.mapping
```diff
@@ -1,54 +1,205 @@
 CLASS bom net/minecraft/block/BlockState
-	CLASS bom$a
+	CLASS bom$a ShapeCache
```
asiekierka, line 2: I don't think it's so simple.

mappings/net/minecraft/block/BlockState.mapping
```diff
@@ -1,54 +1,205 @@
...
+		ARG 2 pos
 	METHOD a canPlaceAtSide (Lbae;Let;Lcgh;)Z
+		ARG 1 view
+		ARG 2 pos
+		ARG 3 env
+	METHOD a (Lbae;Let;Lcle;)Lclt;
+		ARG 1 view
+		ARG 2 pos
+		ARG 3 ePos
+	METHOD a getCullShape (Lbae;Let;Ley;)Lclt;
```
asiekierka, line 36: Too vague.

mappings/net/minecraft/fluid/LavaFluid.mapping
```diff
@@ -12,5 +12,6 @@ CLASS cec net/minecraft/fluid/LavaFluid
 	METHOD d getBlastResistance ()F
 	METHOD e getFlowing ()Lcdy;
 	METHOD f getStill ()Lcdy;
+	METHOD g isWater ()Z
```
asiekierka: Are you sure that's what the method does?

asiekierka (changes requested):
I will need *many* sign-offs on this PR. It touches Blocks, and the methods are touchy and often not immediately obvious.

mappings/net/minecraft/fluid/LavaFluid.mapping
dblsaiko: The only place where this is used is `FluidState BaseFluid#a(ViewableWorld, BlockPos, BlockState)`, as a condition for an if block. That if block checks something about the block/fluid below this block. Missing some names though, so can't exactly say what it does.

I called it isWater because it only returns true for water in vanilla. I agree, it isn't exactly the best name choice, there's definitely a better one.

asiekierka: A name unmapped > a name unsure.

mappings/net/minecraft/block/BlockState.mapping
dblsaiko: This is used for determining whether to cull or draw the face of a block, based on the results of this method for this block and the neighbor block. Maybe getFaceShapeForCulling?

asiekierka: I'm just curious what the method actually returns, what it looks like that is.

dblsaiko: `public VoxelShapeContainer getCullShape(final BlockView view, final BlockPos pos, final Facing facing)`

asiekierka: Nonono. What's the shape of the VoxelShapeContainer? I'm curious if it's just an outline of a face or something.

dblsaiko: Oh, right.

So it calls `return VoxelShapes.a(this.m(view, pos), facing);`. For the default implementation, `this.m(view, pos)` calls getBoundingShape which returns VoxelShapes.fullCube() by default. From what it looks like, `VoxelShapes.a` takes that shape and offsets it in some way (returns an instance of `clr`, which extends `VoxelShapeContainer`), or returns an empty shape if the shape doesn't touch the edge of the block on that face

Gegy: Unrelated to implementation, but should this name maybe mention occlusion?

mappings/net/minecraft/block/AbstractPressurePlateBlock.mapping
```diff
@@ -1,8 +1,29 @@
 CLASS bfm net/minecraft/block/AbstractPressurePlateBlock
+	FIELD a DEPRESSED_SHAPE Lclt;
+	FIELD b DEFAULT_SHAPE Lclt;
+	FIELD c box Lckv;
```
shadowfacts:
```suggestion
	FIELD c BOX Lckv;
```

mappings/net/minecraft/block/AbstractPressurePlateBlock.mapping
```diff
@@ -1,8 +1,29 @@
...
 	METHOD a canMobSpawnInside ()Z
+	METHOD a (Lbas;Let;)V
+		ARG 1 world
+		ARG 2 pos
+	METHOD a updatePlateState (Lbas;Let;Lbom;I)V
+		ARG 1 world
+		ARG 2 pos
+		ARG 3 state
+		ARG 4 rsOut
+		ARG 5 prevRsOut
```
shadowfacts: Local variables should be left unmapped for now.

mappings/net/minecraft/block/AbstractPressurePlateBlock.mapping
```diff
@@ -11,29 +32,51 @@ CLASS bfm net/minecraft/block/AbstractPressurePlateBlock
...
 		ARG 3 pos
-		ARG 4 random
+		ARG 4 rnd
+		ARG 5 rsOut
 	METHOD a canPlaceAt (Lbom;Lbaw;Let;)Z
-		ARG 1 world
+		ARG 1 state
+		ARG 2 world
+		ARG 3 pos
+		ARG 4 groundPos
```
shadowfacts: Local vars

mappings/net/minecraft/block/BlockState.mapping
```diff
@@ -1,54 +1,205 @@
...
+		ARG 3 stack
 	METHOD a neighborUpdate (Lbas;Let;Lbft;Let;)V
+		ARG 1 world
+		ARG 2 pos
+		ARG 3 neighborBlock
+		ARG 4 neighborPos
 	METHOD a onBlockAdded (Lbas;Let;Lbom;)V
+		ARG 1 world
+		ARG 2 pos
+		ARG 3 prevState
```
kashike: `oldState` instead, to go with `newState`.

kashike (changes requested):

mappings/net/minecraft/block/TallSeagrassBlock.mapping
```diff
@@ -1,16 +1,38 @@
 CLASS blm net/minecraft/block/TallSeagrassBlock
+	FIELD c PROPERTY_HALF Lbpj;
+	FIELD d shape Lclt;
```
shadowfacts:
```suggestion
	FIELD d SHAPE Lclt;
```

mappings/net/minecraft/item/ItemPlacementContext.mapping
```diff
@@ -1,5 +1,25 @@
...
+		ARG 1 world
+		ARG 2 player
+		ARG 3 stack
+		ARG 4 pos
+		ARG 5 facing
+		ARG 6 hitX
+		ARG 7 hitY
+		ARG 8 hitZ
 	METHOD a getPos ()Let;
+	METHOD a (Laty;Let$a;Ley;)Laty;
```
shadowfacts: `create`

mappings/net/minecraft/server/world/ServerTickScheduler.mapping
```diff
@@ -1,16 +1,72 @@
 CLASS bbh net/minecraft/server/world/ServerTickScheduler
+	FIELD a filter Ljava/util/function/Predicate;
```
shadowfacts: This predicate is used to exclude things that shouldn't be ticked (see `ServerTickScheduler.schedule`). It's used in `ServerWorld` to exclude null and air blocks as well as null and empty fluids.

Filter isn't a great name for this as it filters what _shouldn't_ be ticked as opposed to what should. Something more descriptive would be better. `shouldExclude`, perhaps?

mappings/net/minecraft/server/world/ServerTickScheduler.mapping
```diff
@@ -1,16 +1,72 @@
...
 	FIELD g ticksCurrent Ljava/util/List;
 	FIELD h tickConsumer Ljava/util/function/Consumer;
+	METHOD <init> (Luk;Ljava/util/function/Predicate;Ljava/util/function/Function;Ljava/util/function/Function;Ljava/util/function/Consumer;)V
+		ARG 1 world
+		ARG 2 filter
+		ARG 3 idToName
+		ARG 4 nameToId
+		ARG 5 tickConsumer
 	METHOD a tick ()V
+		ARG 1 tasksCount
```
shadowfacts: Local vars.

mappings/net/minecraft/server/world/ServerTickScheduler.mapping
```diff
@@ -1,16 +1,72 @@
...
+		ARG 3 tick
+		ARG 6 report
+		ARG 7 el
 	METHOD a toTag (Lbag;)Lhw;
+		ARG 1 pos
+		ARG 2 ticks
+		ARG 3 worldTime
+		ARG 5 tags
+		ARG 8 nbt
+	METHOD a (Lcap;Let;)V
```
shadowfacts, line 19: Maybe `scheduleForcedAllInBox`? Sounds a bit unwieldy, any better options?

mappings/net/minecraft/server/world/ServerTickScheduler.mapping
```diff
@@ -1,16 +1,72 @@
...
+	METHOD a getScheduledTicksInBox (Lcap;Z)Ljava/util/List;
+		ARG 1 box
+		ARG 2 remove
+		ARG 3 ticks
+		ARG 5 iter
+		ARG 6 tick
+		ARG 7 pos
 	METHOD a isScheduled (Let;Ljava/lang/Object;)Z
 		ARG 1 pos
+		ARG 2 data
```
shadowfacts: Not sure about `data`. `item` or `object`? (Same goes for the other `T` params)

mappings/net/minecraft/server/world/ServerTickScheduler.mapping
```diff
@@ -1,16 +1,72 @@
...
+	METHOD b isTicking (Let;Ljava/lang/Object;)Z
+		ARG 1 pos
+		ARG 2 data
 	METHOD b scheduleForced (Let;Ljava/lang/Object;ILbbl;)V
-	METHOD c scheduleInternal (Let;Ljava/lang/Object;ILbbl;)V
+		ARG 1 pos
+		ARG 2 data
+		ARG 3 delay
+		ARG 4 priority
+	METHOD c (Let;Ljava/lang/Object;ILbbl;)V
```
shadowfacts: `addScheduledTick`

mappings/net/minecraft/util/SystemUtil.mapping
```diff
@@ -22,6 +22,7 @@ CLASS m net/minecraft/util/SystemUtil
 		METHOD b getURLOpenCommand (Ljava/net/URL;)[Ljava/lang/String;
 	FIELD a NANO_TIME_SUPPLIER Ljava/util/function/LongSupplier;
 	FIELD b LOGGER Lorg/apache/logging/log4j/Logger;
+	METHOD a toMap ()Ljava/util/stream/Collector;
```
shadowfacts, line 25: `toMapCollector` to `entriesToMapCollector`? Something with collector in it would be better.

mappings/net/minecraft/world/ChunkTickScheduler.mapping
```diff
@@ -0,0 +1,34 @@
+CLASS bqu net/minecraft/world/ChunkTickScheduler
```
shadowfacts, line 1: Same comments as `ServerTickScheduler` above.

mappings/net/minecraft/world/TickScheduler.mapping
```diff
@@ -1,5 +1,16 @@
 CLASS bbj net/minecraft/world/TickScheduler
 	METHOD a isScheduled (Let;Ljava/lang/Object;)Z
 		ARG 1 pos
+		ARG 2 data
```
shadowfacts: Same `data` -> `item`/`object` comment.

shadowfacts (changes requested):
Overall pretty good. Biggest thing is that local variable names shouldn't be mapped (I think) until they're properly supported.

mappings/net/minecraft/util/SystemUtil.mapping
kashike: Disagree, added noise for no real benefit. The return type makes it clear already.

mappings/net/minecraft/server/world/ServerTickScheduler.mapping
Gegy: What about just `schedule`?

mappings/net/minecraft/server/world/ServerTickScheduler.mapping
dblsaiko: As far as I understand, the T parameter contains additional data that should get passed to the scheduled tick when it's run, so that's why I called it that. `object` would work too, I guess

mappings/net/minecraft/server/world/ServerTickScheduler.mapping
shadowfacts: `schedule` is used elsewhere in the tick scheduler,
So I though it would be better have something different. This method is what actually does the work of creating a `ScheduledTick` instance and adding it to the queue.

dblsaiko: > What about just `schedule`?

A method with that name/signature already exists, I'll do addScheduledTick

modmuss50:
Due to the size of the PR I think I'm going to merge it before we update versions, and then open an issue with the points that have been brought up.

This is mostly looking good, local vars will get dropped so no point in worrying about those.

dblsaiko:
Alright, sounds good.

modmuss50:
Moving ongoing conversation to #276 Can improve on this next snapshot. Nice work so far.

PR #272, opened 2018-12-12
minor input mappings

kashike:


asiekierka:
Input or PlayerInput?

kashike:
It isn't really *from* a "player", it's just the current input being provided by something on the client.

asiekierka:
LGTM

SquidDev:
There's a bit of me which would wonder if `Input.update` is more accurate in this case? It recalculates the entire state, rather than progressing it based on a previous one.

Dunno if it's also worth mapping the other fields/methods in `Input` too?

kashike:
Except it does use old state in `KeyboardInput` - it modifies fields.

modmuss50 (approved):

PR #277, opened 2018-12-12
More facing -> direction

modmuss50:
Auto rename facing to direction


jamierocks (approved):

asiekierka:
Shouldn't we keep it as Facing on Blocks, at least?

Prospector:
yeah not all directions are facings but some still are.

Gegy:
In the context of blocks, it may make more sense with `face` or `side` parameters instead of `facing`.

PR #286, opened 2018-12-13
Several mappings for chunk/block rerendering

SquidDev:
Also fleshes out a couple of other missing mappings within `World` - most of these are existing methods which just forward to another.


mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -69,7 +69,7 @@ CLASS yw net/minecraft/util/math/MathHelper
 	METHOD b cos (F)F
 	METHOD b floorMod (FF)F
 	METHOD b wrapDegrees (I)I
-	METHOD b floorMOd (II)I
+	METHOD b floorMod (II)I
```
asiekierka, line 72: Oops, sorry about that!

asiekierka (approved):

shadowfacts (approved):

PR #287, opened 2018-12-13
Several changes to blocks

SquidDev:
Apologies for PR spam, just this one has a couple of changes I was unsure of, so it felt cleaner to split it into something separate:

 - I'm a bit conflicted on `.getCollisionShape` - it appears to be used both
for entity collision and determining if a block's side is a "full cube" (and so can have torches placed against or w/e)
 - I've renamed `getRenderingState` to `getStateForNeighborUpdate`, though am not 100% it's a good name - it may also perform side effects (mostly scheduling this block to be ticked next thread), so isn't quite a pure getter.

### DIFF
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -24,7 +24,7 @@ CLASS bgs net/minecraft/block/Block
 		FIELD h randomTicks Z
 		FIELD i friction F
 		FIELD j dropTableId Lqc;
-		FIELD k pistonExtension Z
+		FIELD k dynamicBounds Z
 		METHOD <init> (Lcfh;Lcfi;)V
 			ARG 1 material
 			ARG 2 materialColor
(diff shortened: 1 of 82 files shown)


Gegy:
`getCollisionShape` could possibly be something along the lines of `getSolidShape`?

mappings/net/minecraft/block/Block.mapping
```diff
@@ -450,7 +450,7 @@ CLASS bgs net/minecraft/block/Block
 	METHOD n getFrictionCoefficient ()F
 	METHOD n hasRandomTicks (Lbpm;)Z
 		ARG 1 state
-	METHOD n (Lbpm;Lbbb;Let;)Z
+	METHOD n isFullCube (Lbpm;Lbbb;Let;)Z
```
shadowfacts, line 453: Can you clarify the difference between `isFullCube` and the two `isShapeFullCube` methods?

shadowfacts (changes requested):

mappings/net/minecraft/block/Block.mapping
SquidDev: `isShapeFullCube` operates on a `VoxelShape` - it basically checks if it's equivalent to `VoxelShapes.fullCube()`. `isFullCube` accepts a world, block state and block position, gets the collision box and then uses `isFullCube`.

IMO it's fairly self-explanatory when you have the mapped class names too, but happy to rename it to something more specific if people prefer :).

asiekierka: I think we should go with this for now and see what happens. I'm all for naming Block methods based on *uses*, not *behavior*, but that's tricky.

asiekierka (approved):

SquidDev:
Sorry for augmenting the PR: the above methods was pretty linked to the other mapping done here with bounding and collision boxes, so it felt it made sense to do it in here. Happy to cherry pick it out if preferred though.

shadowfacts (approved):

SquidDev:
🎉 ✨ Thank you all for reviewing!

PR #288, opened 2018-12-13
Updated some entity fields and class mappings

strikerrocker:


mappings/net/minecraft/entity/mob/IllagerEntity.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS apb net/minecraft/entity/mob/IllagerEntity
+	CLASS apb$a IllagerState
```
asiekierka: It's an inner class; why not State?

mappings/net/minecraft/entity/mob/SlimeEntity.mapping
```diff
@@ -17,6 +17,7 @@ CLASS aqe net/minecraft/entity/mob/SlimeEntity
 		METHOD a canStart ()Z
 		METHOD e tick ()V
 	FIELD bD SLIME_SIZE Lpr;
+	FIELD bE wasOnGround Z
```
asiekierka: When is this true?

asiekierka (changes requested):

mappings/net/minecraft/entity/mob/SlimeEntity.mapping
strikerrocker: What do ya mean?

strikerrocker: Anwyays this is true when onGround was true.

asiekierka: And it's never set to false, ever?

strikerrocker:
Did the requested changes

asiekierka (approved):

SquidDev (approved):
I'm slightly worried by the `isJohnny` field, but who am I to question Mojang?

shadowfacts (approved):

PR #290, opened 2018-12-13
Partial lighting mappings

Niko-sk2x:
Some of these names may not be completely correct, as this is not an easy part of code to work with. 

The class names aren't perfect, but I really ran out of ideas here, if anyone has any better ideas - comment here.

There are still large parts of this code I don't fully understand, do full mappings of this area will take some more time. I also mapped a lot of local variables to be able to actually understand what it's doing.

### DIFF
--- a/mappings/net/minecraft/client/world/ClientChunkManager.mapping
+++ b/mappings/net/minecraft/client/world/ClientChunkManager.mapping
@@ -9,7 +9,9 @@ CLASS dbd net/minecraft/client/world/ClientChunkManager
 		ARG 2 z
 		ARG 3 status
 		ARG 4 require
+	METHOD a onLightUpdate (Lbca;III)V
 	METHOD a tick (Ljava/util/function/BooleanSupplier;)V
 	METHOD f getStatus ()Ljava/lang/String;
 	METHOD g getChunkGenerator ()Lbrb;
 	METHOD i getLightingProvider ()Lceu;
+	METHOD j getWorldAsView ()Lbbb;
--- a/mappings/net/minecraft/server/world/chunk/light/ServerLightingProvider.mapping
+++ b/mappings/net/minecraft/server/world/chunk/light/ServerLightingProvider.mapping
@@ -1,2 +1,12 @@
 CLASS uu net/minecraft/server/world/chunk/light/ServerLightingProvider
-	METHOD a queueLightCheck (Let;)V
+	METHOD a scheduleChunkLightUpdate (IIIZ)V
+		ARG 1 chunkX
+		ARG 2 chunkY
+		ARG 3 chunkZ
+		ARG 4 isEmpty
+	METHOD a doLightUpdates (IZZ)I
+		ARG 1 maxUpdateCount
+	METHOD a enqueueLightUpdate (Let;)V
+		ARG 1 pos
+	METHOD a (Let;I)V
+		ARG 1 pos
--- a/mappings/net/minecraft/sortme/LevelIndexedProcessor.mapping
+++ b/mappings/net/minecraft/sortme/LevelIndexedProcessor.mapping
@@ -1,0 +1,80 @@
+CLASS cep net/minecraft/sortme/LevelIndexedProcessor
+	FIELD a maxLevelCount I
+	FIELD b toProcess [Lit/unimi/dsi/fastutil/longs/LongLinkedOpenHashSet;
+	FIELD c levelMap Lit/unimi/dsi/fastutil/longs/Long2ByteFunction;
+	FIELD d minUnprocessedLevel I
+	FIELD e hasUpdates Z
+	METHOD <init> (III)V
+		ARG 1 levelCount
+		ARG 2 initLevelCapacity
+		ARG 3 initTotalCapacity
+	METHOD a updateMinUnprocessed (I)V
+		ARG 1 maxLevel
+		ARG 2 previousMinUnprocessed
+	METHOD a minLevel (II)I
+		ARG 1 level
+		ARG 2 max
+		ARG 3 ret
+	METHOD a setLevelFor (JI)V
+		ARG 1 index
+		ARG 3 level
+	METHOD a addWithLevel (JII)V
+		ARG 1 id
+		ARG 3 level
+		ARG 4 targetLevel
+	METHOD a removeFromLevel (JIIZ)V
+		ARG 1 index
+		ARG 3 level
+		ARG 4 maxLevel
+		ARG 5 removeFromLevelMap
+	METHOD a processLevelAt (JIZ)V
+		ARG 1 index
+		ARG 3 level
+	METHOD a getMergedLevel (JJI)I
+		ARG 1 position
+		ARG 3 source
+		ARG 5 newLevel
+	METHOD a scheduleUpdate (JJIIIZ)V
+		ARG 1 sourceIndex
+		ARG 3 updateIndex
+		ARG 5 newBaseLevelAtUpdate
+		ARG 6 currentLevelAtUpdate
+		ARG 7 oldLevelAtUpdate
+		ARG 8 modeFlag
+		ARG 9 isOrigin
+		ARG 10 levelAfterUpdate
+	METHOD a scheduleNewLevelUpdate (JJIZ)V
+		ARG 1 sourceIndex
+		ARG 3 updateIndex
+		ARG 5 lastLevel
+	METHOD b updateLevels (I)I
+		ARG 1 maxSteps
+		ARG 2 level
+		ARG 3 nextIndex
+		ARG 5 positionStatus
+		ARG 6 targetLevel
+	METHOD b getBaseLevelFor (JJI)I
+		ARG 1 source
+		ARG 3 position
+		ARG 5 previousLevel
+	METHOD b scheduleUpdateRecursively (JJIZ)V
+		ARG 1 sourceIndex
+		ARG 3 updateIndex
+		ARG 5 sourceLevel
+		ARG 6 modeFlag
+		ARG 7 oldLevel
+		ARG 8 newBaseLevel
+		ARG 9 levelAtUpdateIndex
+		ARG 10 isUpdateSource
+	METHOD c isInvalidIndex (J)Z
+		ARG 1 index
+	METHOD e hasLevelUpdates ()Z
+	METHOD e getCurrentLevelFor (J)I
+		ARG 1 index
+	METHOD g remove (J)V
+		ARG 1 index
+		ARG 3 currentLevel
+		ARG 4 newLevel
+		ARG 5 minLevel
+	METHOD h scheduleNewUpdate (J)V
+		ARG 1 idx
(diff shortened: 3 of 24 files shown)


mappings/uk.mapping
```diff
@@ -0,0 +1,8 @@
+CLASS uk
+	CLASS uk$a
+		METHOD a process2 (JI)V
+		METHOD e getCurrentLevelFor (J)I
+			ARG 1 index
+	METHOD a process2 (JI)V
```
SquidDev: Just wondering why `process2`? I couldn't see a `process` function so I'm assuming this isn't to avoid shadowing?

Niko-sk2x: Ah, initially when I named them, there was `process1`, which is now `processLevelAt`.

The entire class is a bit weird and at first appears to be designed specifically for processing light propagation updates, but then it's used for other things I wasn't able to understand so far (some kind of chunk status transitions?). And the way I think it's used for lighting is that the light levels are "inverted", meaning light level 0 is level 15 in this class, and light level 15 is 0 in this class.

Niko-sk2x: This class is one of very good reasons why javadoc comments would be VERY useful for enigma to support. There is no way to explain such details with just names.

mappings/net/minecraft/server/world/chunk/light/ServerLightingProvider.mapping
```diff
@@ -1,2 +1,5 @@
 CLASS uu net/minecraft/server/world/chunk/light/ServerLightingProvider
 	METHOD a queueLightCheck (Let;)V
+		ARG 1 blockPos
```
Gegy: Probably should be just `pos` -- most other code seems to follow that.

mappings/net/minecraft/sortme/LevelIndexedProcessor.mapping
```diff
@@ -0,0 +1,73 @@
...
+	METHOD a minLevel (II)I
+		ARG 1 level
+		ARG 2 max
+		ARG 3 ret
+	METHOD a process2 (JI)V
+	METHOD a addWithlevel (JII)V
+		ARG 1 id
+		ARG 3 level
+		ARG 4 targetLevel
+	METHOD a removeFromlevel (JIIZ)V
```
Gegy: Extremely minor thing again, the casing here and a few other places is wrong. 😄

mappings/net/minecraft/server/world/chunk/light/ServerLightingProvider.mapping
Niko-sk2x: I will change that together with the next time I push my changes

mappings/net/minecraft/sortme/LevelIndexedProcessor.mapping
Niko-sk2x: Ah, I must have missed that. I initially didn't name it level, and then used `sed` to search and replace it, and unintentionally used option to ignore case.

Niko-sk2x:
Pushed some more changes, most of the hard parts are already done but I'm still not quite happy with a few names, so I would like to see some alternatives:

 * `LevelIndexedProcessor` and the whole name `level` here. From the constructor it appears to be what mojang calls it, and it does seem to fit nicely for lighting (but the implementation is a bit confusing, with inverted light levels), but it doesn't really fit for the way this class is used by `LightStorage` and `ug` and `uk` classes.
 * `LevelIndexedProcessor.scheduleUpdateRecursively` - this method only appears to be called from implementations of `processLevelAt`. It's not *real* recursion but that's the best name I could come up with to distinguish it and avoid a name collision.
 * `LevelIndexedProcessor.getMergedLevel` - this one I really don't have much of an idea for how to name on this abstraction level. For lighting this would be `getLightFromNeighbors` but this is used for more than just lighting.
 * `getCurrentLevelFor` and `getBaseLevelFor` - those are again with a heavy hint for use in lighting, because I didn't have a better idea. For lighting, the "base level" refers to light source "luminosity", and "current level" refers to the currently stored light value.
 * all the `hasUpdates*` and `update*` methods are up to discussion

mappings/net/minecraft/server/world/chunk/light/ServerLightingProvider.mapping
```diff
@@ -1,2 +1,7 @@
 CLASS uu net/minecraft/server/world/chunk/light/ServerLightingProvider
+	METHOD a doLightUpdates (IZZ)I
```
RedstoneParadox, line 7: I would say updateLighting would make slightly more sense.

Niko-sk2x: I think this would make more sense if this was updating lighting at specified position. This is processing all the scheduled light updates. Also this is actually a method from superclass.

asiekierka: Barteks2x is correct here.

mappings/net/minecraft/client/world/ClientChunkManager.mapping
```diff
@@ -9,7 +9,9 @@ CLASS dbd net/minecraft/client/world/ClientChunkManager
 		ARG 2 z
 		ARG 3 status
 		ARG 4 require
+	METHOD a onLightUpdate (Lbca;III)V
 	METHOD a tick (Ljava/util/function/BooleanSupplier;)V
 	METHOD f getStatus ()Ljava/lang/String;
 	METHOD g getChunkGenerator ()Lbrb;
 	METHOD i getLightingProvider ()Lceu;
+	METHOD j worldBlockView ()Lbbb;
```
asiekierka: Is this not a getter?

mappings/net/minecraft/world/chunk/light/LightStorage.mapping
```diff
@@ -1,9 +1,74 @@
...
+	METHOD b hasChunk (J)Z
+		ARG 1 chunkPos
+	METHOD b set (JI)V
+		ARG 1 blockPos
+		ARG 3 value
+	METHOD b getBaseLevelFor (JJI)I
+		ARG 1 source
+		ARG 3 position
+		ARG 5 previousLevel
+	METHOD c updateNotifyChunkProvider ()V
```
asiekierka: Why not just update or just notify?

mappings/net/minecraft/world/chunk/light/SkyLightStorage.mapping
```diff
@@ -0,0 +1,30 @@
...
+		ARG 3 sectionOrigin
+		ARG 5 sectionOriginY
+		ARG 6 uncachedStorage
+		ARG 7 height
+		ARG 8 lightData
+	METHOD a processUpdates (Lcer;ZZ)V
+		ARG 1 lightProvider
+		ARG 2 doSkylight
+		ARG 3 doEdgeLightPropagation
+	METHOD d refreshHasUpdates ()V
```
asiekierka: checkForUpdates, maybe?

mappings/net/minecraft/world/chunk/light/SkyLightStorage.mapping
```diff
@@ -0,0 +1,30 @@
+CLASS cex net/minecraft/world/chunk/light/SkyLightStorage
+	CLASS cex$a SkyLightData
```
asiekierka: SkyLightStorage$Data?

asiekierka (changes requested):

mappings/net/minecraft/client/world/ClientChunkManager.mapping
Niko-sk2x: It is but I didn't have a good idea for a name. `getBlockView` sounds really unhelpful, `getWorld` says what it does, but doesn't match return type, and `getWorldBlockView` is even longer than this.

asiekierka: getWorldAsView?

mappings/net/minecraft/server/world/chunk/light/ServerLightingProvider.mapping
```diff
@@ -1,2 +1,7 @@
...
+		ARG 1 maxUpdateCount
 	METHOD a queueLightCheck (Let;)V
```
Gegy: `enqueueLightUpdate`? `Update` seems to be used everywhere else.

mappings/net/minecraft/sortme/LevelIndexedProcessor.mapping
```diff
@@ -0,0 +1,80 @@
+CLASS cep net/minecraft/sortme/LevelIndexedProcessor
```
Gegy, line 1: `Level` could possibly become confusing due to being used for world saves.

mappings/net/minecraft/world/chunk/light/ChunkLightProvider.mapping
```diff
@@ -0,0 +1,47 @@
+CLASS cer net/minecraft/world/chunk/light/ChunkLightProvider
+	FIELD a chunkProvider Lbro;
+	FIELD b type Lbca;
+	FIELD c lightStorage Lcet;
+	FIELD d DIRECTIONS [Ley;
+	FIELD f srcMutablePos Let$a;
+	FIELD g destMutablePos Let$a;
+	FIELD h mutablePosGetLightBlockedBetween Let$a;
```
Gegy, line 8: This may be a bit excessively verbose. 😄

mappings/net/minecraft/sortme/LevelIndexedProcessor.mapping
Niko-sk2x: As explained before, the reason I named it like that was that there is evidence in the form of exception message that mojang names them `levels`. I don't really agree with that name, but if you have any better idea - I would like to see it.

mappings/net/minecraft/world/chunk/light/ChunkLightProvider.mapping
Niko-sk2x: Any idea for a shorter name? I wanted to avoid naming them `mutablePos1/2/3`

Niko-sk2x:
Changed everything as requested, except `LevelIndexedProcessor` and `mutablePosGetLightBlockedBetween` as I didn't get any other suggestions.

Prospector (approved):

mappings/net/minecraft/world/chunk/light/ChunkLightProvider.mapping
Gegy: While maybe not specific enough, could possibly be just `mutablePos`? It is only used in `getLightBlockedBetween`, but it could be used more generically.

Niko-sk2x: Looking at the way it's done, mojang isn't reusing them between methods at all. The 2 fields above that one are restricted to use in a single method too.

Gegy (approved):

PR #280, opened 2018-12-13
Many villager-related mappings, some trade related ones and some new …

liach:
…block related ones (two furnaces)

Also mapped some lost entity methods.

Signed-off-by: liach <liach@users.noreply.github.com>

PR #294, opened 2018-12-14
18w50a: Update registries

jamierocks:


liach:
Uh oh jamie, you killed my pr with all those village mappings! please let us discuss a bit before merging next time! #280

asiekierka:
@liach "killed"

It's just conflict resolution; it's not difficult in Enigma and for now you can leave that to me.

PR #295, opened 2018-12-14
Item-related name tweaks

asiekierka:
* ItemContainer is also used on Blocks, which don't really *contain* items, so ItemProvider is a more idiomatic name.
* Container items have nothing to do with containers bar the fact they're occasionally a bucket or bottle; their real purpose is to be used as recipe replacements, so they have been named accordingly.
* addInformation is a bad name. The only method which calls it is getTooltipText.

~~Let's defeat Lord Voldemap together~~

### DIFF
--- a/mappings/net/minecraft/item/ArmorMaterial.mapping
+++ b/mappings/net/minecraft/item/ArmorMaterial.mapping
@@ -1,0 +1,7 @@
+CLASS auj net/minecraft/item/ArmorMaterial
+	METHOD a getDurability (Laha;)I
+	METHOD b getEquipSound ()Lxm;
+	METHOD b getProtectionAmount (Laha;)I
+	METHOD c getRepairIngredient ()Layt;
+	METHOD d getName ()Ljava/lang/String;
+	METHOD e getToughness ()F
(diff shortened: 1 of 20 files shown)


ChloeDawn:
I suggest `buildTooltip` over `getTooltipText`. The method itself does not "get" anything, it is used as a builder for the passed collection.

A "replacement" implies it is a substitution for something. It is more of a remainder from a crafting task. It is what remains once the input is consumed.

3TUSK:
May I suggest to use "lore" rather than "tooltip"? 
According to [Minecraft Wiki](https://minecraft.gamepedia.com/Player.dat_format#Display_Properties), this seems to be called "lore" internally. It is also the name used by Bukkit IIRC. Also, it is shorter than "tooltip". The only downside is that "lore" is a bit more ambiguous if context is not clear than that of "tooltip".

asiekierka:
Ew, no. All "Lore" means is text which can *additionally* go on the tooltip; so it's a subset of the tooltip, not a complete set.

liach:
Lore is a part of the item nbt. It does not include the builtin description on an item. E.g. If you add a lore to a music disc in the game, the tooltip will be displayed above the lore. Moreover, the tooltip cannot be removed with nbt settings while the lore can be. The lore is itatlic and purple by default as well.

MasterEric:
+1 on changing addInformation. Should it just be addTooltipText?

modmuss50 (approved):
👍

PR #296, opened 2018-12-14
Map ItemColorMap

SquidDev:
Absurdly small PR, sorry. It appears that cpw causes Enigma to crash, so this one was done manually -  I've confirmed that it builds and tiny-remapper handles it correctly though.

It just uses the same names as `BlockColorMap`, as that seemed the most sane.


asiekierka:
>cpw causes Enigma to crash

I laughed harder than I should have.

PR #302, opened 2018-12-14
Multiple Item Related Mappings

RedstoneParadox:
Mappings related to several item classes and a rename of auj to ArmorMaterial.

p.s. Ignore the villager-related classes; I was merely deleting redundant files from my fork.


mappings/net/minecraft/item/ArmorMaterials.mapping
```diff
@@ -1,11 +1,26 @@
 CLASS auk net/minecraft/item/ArmorMaterials
```
shadowfacts, line 1: Should all the armor classes be grouped together in their own package? `net/minecraft/item/armor`?

shadowfacts (changes requested):
Local variables should not be mapped yet.

Also, please resolve merge conflicts.

RedstoneParadox:
What do you mean local variables should not be mapped yet?

shadowfacts:
Variables inside of methods should not be mapped yet as they're not properly supported across all of Fabric's toolchain.

mappings/net/minecraft/item/ArmorMaterials.mapping
RedstoneParadox: In that case, so should any tool classes.

shadowfacts: I'm not certain about this. Leave them in `item` for now, this can be dealt with later.

RedstoneParadox:
That would have been nice to know earlier.

asiekierka (approved):

mappings/net/minecraft/item/BannerPatternItem.mapping
```diff
@@ -1,5 +1,12 @@
 CLASS aup net/minecraft/item/BannerPatternItem
+	FIELD a patterns Lbnk;
```
SquidDev: Maybe just `pattern`, as the getter is just `getPattern()`?

SquidDev (changes requested):

mappings/net/minecraft/item/BannerPatternItem.mapping
asiekierka: Definitely. It's not a collection. Whoops!

asiekierka (approved):

SquidDev (approved):

shadowfacts (approved):

PR #303, opened 2018-12-15
Many names following from UnlockRecipesClientPacket

codewarrior0:


mappings/net/minecraft/container/Container.mapping
```diff
@@ -29,6 +29,11 @@ CLASS asw net/minecraft/container/Container
 		ARG 1 stack
 		ARG 2 startIndex
 		ARG 3 endIndex
+	METHOD a onCraftingContentChanged (Lbbp;Larb;Lafi;Latz;)V
```
asiekierka, line 32: Is this only used for crafting tables still, or also other Recipe-based inventories?

asiekierka (changes requested):

mappings/net/minecraft/container/Container.mapping
codewarrior0: It's only used for crafting tables and in-hand crafting.

shadowfacts (approved):

codewarrior0:
Rebased onto branch head, added a couple new class names.

PR #304, opened 2018-12-15
map Material.breakByHand

asiekierka:


SquidDev (approved):

mappings/net/minecraft/item/MiningToolItem.mapping
```diff
@@ -19,3 +19,4 @@ CLASS avj net/minecraft/item/MiningToolItem
 		ARG 4 pos
 	METHOD a getBlockBreakingSpeed (Lawo;Lbpm;)F
 		ARG 1 stack
+		ARG 2 vBlockState3
```
SquidDev: Is this intentional? It's a wee bit of a weird name, so wasn't sure if an artefact from Enigma or not. Either way, not a blocker.

asiekierka: Enigma artifact...

shadowfacts (approved):

PR #306, opened 2018-12-15
Server Properties, and minor fixes

jamierocks:


mappings/gz.mapping
```diff
@@ -0,0 +1,8 @@
+CLASS gz
+	CLASS gz$a
+		FIELD b group Ljava/lang/String;
+		FIELD c ingrediant Layt;
```
asiekierka: ingredient

asiekierka (changes requested):

shadowfacts (approved):

PR #278, opened 2018-12-12
Biome and feature work, a little bit of villager stuff

Prospector:

PR #307, opened 2018-12-15
Standard feature configurators

Gegy:


Gegy:
Oops, conflicts with #278.

PR #312, opened 2018-12-15
LightMapTextureManager and various ambient light things

JamiesWhiteShirt:


dblsaiko:
iirc `partialTicks` is called `delta` in other mappings.

mappings/net/minecraft/world/IWorld.mapping
```diff
@@ -15,5 +15,7 @@ CLASS bbq net/minecraft/world/IWorld
 	METHOD g getLevelProperties ()Lcih;
 	METHOD g getLocalDifficulty (Let;)Lafn;
 	METHOD i getSaveHandler ()Lcii;
+	METHOD k getSkyAngle (F)F
```
JamiesWhiteShirt, line 25: There was some discussion on this. It could be "pitch" instead of "angle". We could also use "celestial" instead of "sky".

JamiesWhiteShirt:
Could have sworn I saw the voldemapping elsewhere... Will change to "delta".

dblsaiko:
Quick search only turned up [this](https://github.com/FabricMC/yarn/blob/e9ccf701a7caddeac277b25998a917970ec698ff/mappings/net/minecraft/client/render/entity/model/DragonEntityModel.mapping)

SquidDev (approved):

mappings/net/minecraft/world/IWorld.mapping
```diff
@@ -15,5 +15,7 @@ CLASS bbq net/minecraft/world/IWorld
...
+		ARG 1 deltaTicks
```
asiekierka: just delta if that's the new standard

mappings/net/minecraft/world/IWorld.mapping
asiekierka: What does the value returned mean?

asiekierka (changes requested):

mappings/net/minecraft/world/IWorld.mapping
JamiesWhiteShirt: [0, 1), used to rotate the sky and the clock item among other things.

JamiesWhiteShirt:
Converted `deltaTicks` to `delta`. Converted all other instances of `deltaTicks` in the process.

mappings/net/minecraft/item/ItemPropertyGetter.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS awn net/minecraft/item/ItemPropertyGetter
 	METHOD call (Lawo;Lbbp;Lahe;)F
-		ARG 1 world
-		ARG 2 user
+		ARG 1 stack
+		ARG 2 world
+		ARG 3 user
```
Gegy, line 5: Not particularly relevant to this PR, but what is `user`?

JamiesWhiteShirt: It seems for clocks and compasses it is used to get additional information like the world and the orientation of the user.

mappings/net/minecraft/world/IWorld.mapping
asiekierka: If it,s [0, 1), is it really an angle? Does it map like a scaled angle or like a height?

JamiesWhiteShirt: Turns is a valid angle unit, as is radians and degrees.

mappings/net/minecraft/client/render/WorldRenderer.mapping
```diff
@@ -6,11 +6,16 @@ CLASS dea net/minecraft/client/render/WorldRenderer
...
 	METHOD <init> (Lcpk;Lwj;)V
 		ARG 1 client
 	METHOD a onSetCameraEntity (Lagv;)V
 	METHOD a loadShader (Lqc;)V
 	METHOD a onResourceReload (Lwj;)V
 		ARG 1 manager
+	METHOD d (F)F
+		ARG 1 delta
 	METHOD e tick ()V
+	METHOD h disableLightMap ()V
```
asiekierka: Do we want to capitalize LightMap or use Lightmap?

mappings/net/minecraft/world/IWorld.mapping
asiekierka: Fair.

asiekierka (changes requested):

mappings/net/minecraft/client/render/WorldRenderer.mapping
JamiesWhiteShirt: I believe it is referred to as "light_map" somewhere which suggests it could be a "light map" instead of a "lightmap". I'm down with either, really.

asiekierka: It is the texture name. I'd still like a 2nd opinion, though

dblsaiko: I vote for `Lightmap`

grondag: Light maps are implemented as textures. 

Personally, my brain likes them as separate words but common usage seems to be “lightmap” and not “light map” so if you’re trying to be consistent disableLightmap seems best.

Vexatos: I also agree with calling it `Lightmap` instead.

mappings/net/minecraft/client/render/LightMapTextureManager.mapping
```diff
@@ -0,0 +1,16 @@
...
+	FIELD b image Lcob;
+	FIELD c textureIdentifier Lqc;
+	FIELD d isDirty Z
+	FIELD e prevFlicker F
+	FIELD f flicker F
+	FIELD g worldRenderer Ldea;
+	FIELD h client Lcpk;
+	METHOD <init> (Ldea;)V
+		ARG 1 worldRenderer
+	METHOD a flicker ()V
```
asiekierka: tick(), actually

asiekierka (changes requested):

mappings/net/minecraft/client/render/WorldRenderer.mapping
Player3324: Google and Wikipedia indicate that "lightmap" is the common technical term/word.

JamiesWhiteShirt: Renamed to "Lightmap" as that seems to be the consensus.

JamiesWhiteShirt:
Merge conflicts have been handled.

PR #316, opened 2018-12-15
isRemote -> isClient

asiekierka:
Vote with your reactions.


Prospector (approved):

modmuss50 (approved):
Makes more sense imo.

Remote has been ingrained into most of us, but I feel like we can push past it.

NikkyAI (approved):

kashike (approved):

PR #318, opened 2018-12-16
Further RecipeBookGui-related names

codewarrior0:
Curiously, a method previously named "isClickInContainerBounds" will in fact return true only when the click is *outside* the bounds...


mappings/net/minecraft/container/LoomContainer.mapping
```diff
@@ -33,3 +33,4 @@ CLASS ats net/minecraft/container/LoomContainer
 	METHOD e getCraftingResultSlotIndex ()I
 	METHOD f getCraftingWidth ()I
 	METHOD g getCraftingHeight ()I
+	METHOD h craftingSlotCount ()I
```
asiekierka: getCraftingSlotCount?

asiekierka (changes requested):

mappings/net/minecraft/container/LoomContainer.mapping
codewarrior0: It's used by RecipeBookGui to detect when any of the crafting or crafting result slots is clicked on so it can hide the ghost item overlay that it displays whenever you select a recipe from the book that you lack the ingredients for.

codewarrior0: If CraftingContainer had documentation, it would say that the first N slots of the container are considered "crafting slots" that can have said overlay applied to them, where N is the number returned by getCraftingSlotCount.

asiekierka: From code observation, clearly a getter to me...

codewarrior0: ....oh. I didn't put a "get" because there's no corresponding setter. One sec.

shadowfacts (approved):

asiekierka (approved):

PR #322, opened 2018-12-16
Some Rendering and Model names

Tschipp:
Mapped `LivingEntityRenderer#method_4054` and its overrides to `render`
Mapped `Model#method_17080` to `setAngles`

Added/Changed names of both Tropical fish models


shadowfacts (approved):

asiekierka:
setAngles? Is there a better potential name?

Tschipp:
I‘ve thought about naming it `setModelRotation` but that implies that the entire model gets rotated... `setPartRotation` maybe? Or what about `setModelAngles`?

PR #324, opened 2018-12-17
ActionTypeSlot names match enum values

codewarrior0:
image


quat1024:
Not a huge fan of mojang names here but it does make sense to map the enum items accordingly

liach:
Instead of adding these lines, the correct way to fix it is to remove these mappings straight.

PR #326, opened 2018-12-17
Mapped some persistence data stuff

liach:
I renamed all persisted to persistent, though vanilla fix types enum calls them saved data.

Mapped out the persistent data manager that is bound to each dimension and the persistent data updater for stuctures.


mappings/net/minecraft/nbt/ChunkUpdateState.mapping
```diff
@@ -0,0 +1,12 @@
+CLASS ccl net/minecraft/nbt/ChunkUpdateState
```
asiekierka: nbt?

asiekierka: Should PersistentState even be in nbt?

mappings/net/minecraft/world/PersistentStateContainer.mapping
```diff
@@ -0,0 +1,8 @@
+CLASS bbx net/minecraft/world/PersistentStateContainer
+	METHOD a getNextAvailableIdForPersistentState (Lbsg;Ljava/lang/String;)I
```
asiekierka: (note: bbx is an interface, the verbosity might be warranted)

asiekierka (changes requested):

mappings/net/minecraft/world/PersistentStateContainer.mapping
liach: This interface is implemented by `World` and `IWorld`. Dunno if this degree of clarity is needed in a class with a ton of stuff (maybe something else can also supply available ids?)

mappings/net/minecraft/nbt/ChunkUpdateState.mapping
liach: Agree. The whole thing should go to world package.

liach:
@asiekierka Can you review again?

asiekierka (approved):

shadowfacts (approved):

PR #329, opened 2018-12-17
Cleanup ModelTransformation/s class names.

covers1624:
`ModelTransformation` -> `ModelTransformationData`
`ModelTransformations` -> `ModelTransformation`

These names were extremely confusing, Open to second opinions on `ModelTransformationData`.
Also names a bunch of things in both of these.


dblsaiko (commented):
Still not so sure on the names for these. I suggested `ModelTransforms` (dfj) -> `ModelTransformationMap`, since it's basically a map of `ModelTransformations.Type` to `ModelTransform`, but as covers said, it also has that applyGl method so it's not only used as a map. Not sure what others think about this

mappings/net/minecraft/client/render/model/json/ModelTransformationData.mapping
```diff
@@ -0,0 +1,17 @@
+CLASS dfj net/minecraft/client/render/model/json/ModelTransformationData
```
asiekierka: Why not just... Transformation?

asiekierka (changes requested):

mappings/net/minecraft/client/render/model/json/ModelTransformationData.mapping
covers1624: Yeah that works, honestly slipped my mind..

Prospector (approved):

shadowfacts (approved):

PR #330, opened 2018-12-17
Drop verbose 'Inv' wording from Inventrory methods.

covers1624:
Drops 'Inv' from all Inventory methods. What else am i getting a stack on? A carrot?
Named `afg`(`class_3829`) -> `PasteCallback`, its only ever called by things that 'paste' a BlockEntity in world and its not specific to Inventories, although its only implemented by `Inventory` inheritors and `JukeboxBlockEntity`, still not 100% on the name imo, feedback would be nice.

### DIFF
--- a/mappings/afg.mapping
+++ b/mappings/afg.mapping
@@ -1,2 +1,0 @@
-CLASS afg
-	METHOD U_ clearInv ()V
--- a/mappings/net/minecraft/block/entity/AbstractFurnaceBlockEntity.mapping
+++ b/mappings/net/minecraft/block/entity/AbstractFurnaceBlockEntity.mapping
@@ -8,17 +8,22 @@ CLASS bni net/minecraft/block/entity/AbstractFurnaceBlockEntity
 	FIELD m recipesUsed Ljava/util/Map;
 	METHOD B isBurning ()Z
 	METHOD R_ getName ()Ljd;
-	METHOD U_ clearInv ()V
-	METHOD Z_ getInvSize ()I
-	METHOD a getInvStack (I)Lawo;
-	METHOD a takeInvStack (II)Lawo;
+	METHOD U_ onPasted ()V
+	METHOD Z_ getSize ()I
+	METHOD a getStack (I)Lawo;
 		ARG 1 slot
-	METHOD a setInvStack (ILawo;)V
+	METHOD a takeStack (II)Lawo;
 		ARG 1 slot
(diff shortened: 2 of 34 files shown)


asiekierka:
I'm not sure about dropping the verbose Inv-wording due to the ubiquitousness of that interface on BlockEntities. Thoughts from others?

dblsaiko (approved):

dblsaiko:
Eh, not sure about that. The only thing where this might be problematic would be isEmpty, I think.

asiekierka:
isEmpty is definitely a problem. as are the property setters/getters

shadowfacts:
I'm in favor of keeping `Inv`, especially in the cases where it's not immediately clear that the method operates on the inventory specifically (`getSize` being an obvious example).

covers1624:
Ill admit, it might be a good idea to keep the 'Inv'-Wording on a few of these methods, specifically as pointed out `getSize` and `isEmpty`. As for the other methods, it really feels out of place for me.

EDIT: Forgot about the property setters, Those are weird ones, they really only exist for Container syncing, they don't actually have anything to do with the operation of the Inventory.

Prospector:
I'd like to keep Inv personally

Prospector:
How many votes are needed to overturn a PR? We have 3 or 4 against this right now. So far I only see covers for it.

dblsaiko:
I'm also for it. (Of course excluding the names that would clash, like getSize)

modmuss50:
Im for it, but I suggest keeping it on 

 `getInvSize` 
`getInvWidth`
 `canPlayerUseInv`
`setInvProperty`
`onInvOpen`
`isInvEmpty`
`getInvProperty`
`onInvClose`
`getInvPropertyCount`
`getInvHeight`

This would leave us with the following changes

`getInvStack` -> `getStack`
`takeInvStack` -> `takeStack`
`setInvStack` -> `setStack`
`getInvMaxStackAmount` -> `getMaxStackAmount`
`removeInvStack` -> `removeStack`
`isValidInvStack` -> `isValidStack`

asiekierka:
OTOH I think it can feel somewhat inconsistent that way... But maybe...

asiekierka:
I'd say this needs further discussion. I'll keep it open until 19w hits, though, in case we reach a consensus early

modmuss50:
Closing this PR and moving it to an issue. #682

PR #335, opened 2018-12-18
 Added mappings for the Model class and fixed name for old Model class

sekwah41:
What was originally called model was not the base model class but rather the EntityModel class and the base Model class had no mappings

Note: Ignore the comment note I had forgotten for a sec that it also has isChild and isRiding so it is actually the proper EntityModel class


kashike (approved):

sekwah41:
Sorry didnt mean to push to the same pull request ill revert this

kashike (approved):

PR #336, opened 2018-12-18
Identifier#path -> Identifier#value

ChloeDawn:
With the desire to generify what Mojang considers a location as an `Identifier`, it makes sense to treat the value component of the identifier pair as a `value` rather than a `path`.


kashike:
The exception message calls this `path`. :-1: for renaming.

ChloeDawn:
The exception also describes the class as a location but Fabric is trying to generify that. ¯\_(ツ)_/¯

shadowfacts:
Value, IMO, is too generic. It's not clear whether it returns the whole thing (e.g. `"minecraft:stone"`) or just the second part (`"stone"`). Path makes that distinction.

PR #338, opened 2018-12-18
Updated a bunch of entity and world collision names

sekwah41:
There were many collision names missing. There are some that are still wrong and I will go through more and add more to this as I am writing my physics mod.

(The commits before were because id accidentally got code from the other pull request in this branch, just getting used to pull requests on GitHub :P)


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -155,7 +161,9 @@ CLASS agv net/minecraft/entity/Entity
 	METHOD a toListTag ([D)Lhy;
 	METHOD a toListTag ([F)Lhy;
 	METHOD aA getLightmapCoordinates ()I
+	METHOD aC setVelocityModified ()V
 	METHOD aD doesCollide ()Z
+	METHOD aE pushable ()Z
```
asiekierka: isPushable

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -155,7 +161,9 @@ CLASS agv net/minecraft/entity/Entity
 	METHOD a toListTag ([D)Lhy;
 	METHOD a toListTag ([F)Lhy;
 	METHOD aA getLightmapCoordinates ()I
+	METHOD aC setVelocityModified ()V
```
asiekierka: setVelocityDirty? Would we rather "dirty" or "modified"?

mappings/net/minecraft/world/ViewableWorld.mapping
```diff
@@ -10,6 +10,14 @@ CLASS bbt net/minecraft/world/ViewableWorld
 		ARG 3 minZ
 		ARG 4 maxX
 	METHOD a getClosestPlayer (Lagv;D)Larb;
+	METHOD a getCollisionVoxelShapes (Lagv;Lclz;DDD)Ljava/util/stream/Stream;
+		ARG 3 x
+		ARG 5 y
+		ARG 7 z
+	METHOD a getCollosionVoxelShapes (Lagv;Lclz;Ljava/util/Set;DDD)Ljava/util/stream/Stream;
```
asiekierka: getCollisionVoxelShapes

asiekierka (changes requested):

mappings/net/minecraft/entity/Entity.mapping
sekwah41: Well by the looks of a lot of these names it seems they are trying to avoid mcp names majorly and also Upcraft made it kinda vague to me as to if there can be any crossover at all :P

I think mcp may have setVelocityDirty. Though tbh it does describe it. If its fine ill be less worried about what mcp may have and just set what i think they should be.

If its fine then ill change it to Dirty because its what i wanted in the first place

mappings/net/minecraft/world/ViewableWorld.mapping
sekwah41: Oh my bad xD ill fix it now

mappings/net/minecraft/util/StreamHandler.mapping
```diff
@@ -0,0 +1,2 @@
+CLASS yy net/minecraft/util/StreamHandler
```
Gegy: This is a class that allows a `Stream` to be iterated more than once. I propose `CyclingStream` or `LoopingStream`.

sekwah41: Oh good idea didn't notice that part ill do LoopingStream

Prospector (approved):

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -155,7 +162,9 @@ CLASS agv net/minecraft/entity/Entity
 	METHOD a toListTag ([D)Lhy;
 	METHOD a toListTag ([F)Lhy;
 	METHOD aA getLightmapCoordinates ()I
+	METHOD aC setVelocityDirty ()V
```
asiekierka: Since the community didn't chime in...

This field (and method) actually pertains SPECIFICALLY to packet updates. For this, I propose: scheduleVelocityUpdate?

mappings/net/minecraft/world/ViewableWorld.mapping
```diff
@@ -10,6 +10,14 @@ CLASS bbt net/minecraft/world/ViewableWorld
 		ARG 3 minZ
 		ARG 4 maxX
 	METHOD a getClosestPlayer (Lagv;D)Larb;
+	METHOD a getCollisionVoxelShapes (Lagv;Lclz;DDD)Ljava/util/stream/Stream;
+		ARG 3 x
+		ARG 5 y
+		ARG 7 z
+	METHOD a getCollsionVoxelShapes (Lagv;Lclz;Ljava/util/Set;DDD)Ljava/util/stream/Stream;
```
asiekierka: ... getCollsion?

asiekierka (changes requested):

mappings/net/minecraft/world/ViewableWorld.mapping
sekwah41: No it gets the VoxelShape objects within a surrounding area for you to then iterate through and check collisions. So its all the solid collision boxes including entities if they have the flag like the boat does.

sekwah41: The classes dont have names because ive named them in here but i use it this way
`class_3538<VoxelShape> voxels = new class_3538(entity.world.method_8609(entity, boundingBox, moveX, moveY, moveZ));`
`moveY = VoxelShapes.method_1085(Direction.Axis.Y, boundingBox, voxels.method_15418(), moveY);`

Its the line from the first snippet method_8609 And class_3538 is the suggested LoopingStream

asiekierka: No, I mean, there's a typo

sekwah41: OHHH i thought you ment you wanted it to be getCollision overall. I was thinking wait that aint right

mappings/net/minecraft/entity/Entity.mapping
sekwah41: I thought things were marked as dirty if a change was flagged and needed to be synced? As in its dirty the next time it does the sync checks it sends data if it's dirty. I know that was the case with some of the other parts in vanilla MC.

Though scheduleVelocityUpdate is technically a better overall explanation of it. Though it sounds like if you called it 50 times in 1 tick it would send 50 updates because it's scheduled 50 times

asiekierka: Can we get more people to chime in here?

shadowfacts: +1 for `scheduleVelocityUpdate`

sekwah41: Tbh it is a better overall explanation of what ends up happening so yea +1 despite my reasons why I was unsure

shadowfacts (approved):

PR #339, opened 2018-12-18
Entitity rendering + Feature rendering

JamiesWhiteShirt:
NOTE: Currently, due to issues with synthetic bridge methods in Enigma, a lot of override methods are not mapped properly, so I don't believe this is ready for merge.

This maps all entity renderers and adds new semantics for what seems to have been called "layers", now what I call a feature renderer. A feature renderer is attached to a living entity renderer and renders various things like capes, armor, elytra wings, held items and much more.


Prospector (approved):

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -107,10 +108,12 @@ CLASS ahe net/minecraft/entity/LivingEntity
 	METHOD d getCurrentExperience (Larb;)I
 	METHOD da getActiveItem ()Lawo;
 	METHOD dg isFallFlying ()Z
+	METHOD e turnHead (FF)F
```
asiekierka: setHeadRotation? Or what does this do?

asiekierka (changes requested):

mappings/net/minecraft/entity/LivingEntity.mapping
JamiesWhiteShirt: I couldn't really tell completely. It seemed to have something to do with turning the head (and the body within a certain angle). I have simply removed the mapping as it could be inaccurate.

shadowfacts (approved):

shadowfacts:
@asiekierka can you re-review?

PR #341, opened 2018-12-19
Mapped some stuff about fish, especially fish group and tropical fish

liach:
also include some mob spawning/despawning mappings

### DIFF
--- a/mappings/net/minecraft/entity/mob/MobEntity.mapping
+++ b/mappings/net/minecraft/entity/mob/MobEntity.mapping
@@ -27,7 +27,7 @@ CLASS ahf net/minecraft/entity/mob/MobEntity
 	METHOD A playAmbientSound ()V
 	METHOD D getAmbientSound ()Lxm;
 	METHOD F getLootTableId ()Lqc;
-	METHOD G ()Z
+	METHOD G cannotDespawn ()Z
 	METHOD H checkDespawn ()V
 	METHOD I mobTick ()V
 	METHOD N getLimitPerChunk ()I
@@ -47,6 +47,7 @@ CLASS ahf net/minecraft/entity/mob/MobEntity
 		ARG 1 player
 	METHOD a (Lbbq;Lafn;Lahh;Lahp;Lhs;)Lahp;
 		ARG 3 difficulty
+	METHOD a canSpawn (Lbbq;Lahh;)Z
 	METHOD a getPathNodeTypeWeight (Lchg;)F
 	METHOD a setPathNodeTypeWeight (Lchg;F)V
 		ARG 1 type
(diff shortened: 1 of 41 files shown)


mappings/net/minecraft/entity/passive/GroupingFishEntity.mapping
```diff
@@ -0,0 +1,23 @@
...
+		ARG 1 type
+	METHOD N getLimitPerChunk ()I
+	METHOD W_ update ()V
+	METHOD a joinGroupOf (Lalz;)Lalz;
+	METHOD a (Lbbq;Lafn;Lahh;Lahp;Lhs;)Lahp;
+	METHOD a pullInOtherFish (Ljava/util/stream/Stream;)V
+	METHOD dB getMaxGroupSize ()I
+	METHOD dC hasLeader ()Z
+	METHOD dD leaveGroup ()V
+	METHOD dE canHaveMoreFishesInGroup ()Z
```
asiekierka: Fishes? :thinking:

asiekierka (changes requested):

mappings/net/minecraft/entity/passive/GroupingFishEntity.mapping
```diff
@@ -0,0 +1,23 @@
+CLASS alz net/minecraft/entity/passive/GroupingFishEntity
```
Gegy: I think `school` would be more correct within the context of fish.

mappings/net/minecraft/entity/passive/GroupingFishEntity.mapping
sciguyryan: https://grammarist.com/usage/fish-fishes/

Apparently it's valid as long as it's multiple species of fish. Who knew?

mappings/net/minecraft/entity/mob/MobEntity.mapping
```diff
@@ -27,7 +27,7 @@ CLASS ahf net/minecraft/entity/mob/MobEntity
 	METHOD A playAmbientSound ()V
 	METHOD D getAmbientSound ()Lxm;
 	METHOD F getLootTableId ()Lqc;
-	METHOD G ()Z
+	METHOD G cannotDespawn ()Z
```
Prospector, line 30: Does this get if it cannot despawn or set it so it cannot despawn?

mappings/net/minecraft/entity/passive/GroupingFishEntity.mapping
liach: Fixed.

mappings/net/minecraft/entity/passive/GroupingFishEntity.mapping
liach: Done.

mappings/net/minecraft/entity/mob/MobEntity.mapping
liach: You definitely know this is a getter. A setter should be named `setNoDespawn` here instead.
`canXxx` `isXxx` are regular getter method name patterns, while `setXxx` is the setter.
`canXxx` is a verb phrase, so it is a good method name.

liach:
@asiekierka Can you review again?

mappings/net/minecraft/entity/passive/GroupingFishEntity.mapping
Gegy: Maybe should apply to other contexts where `group` is used?

PR #344, opened 2018-12-20
A handful of Block mappings

gudenau:
Probably made a mess of things, first try after all.


asiekierka (approved):

mappings/net/minecraft/block/Block.mapping
```diff
@@ -113,6 +117,7 @@ CLASS bgs net/minecraft/block/Block
 		ARG 3 pos
 		ARG 4 state
 		ARG 5 blockEntity
+		ARG 6 itemStack
```
shadowfacts: `stack` is used as the parameter name for `ItemStack` parameters in `Item` methods. Should be the same here for consistency.

shadowfacts (changes requested):
Local variables shouldn't be mapped until they're properly supported across all of Fabric's toolchain.

Other than that, looks good.

mappings/net/minecraft/block/Block.mapping
Prospector: agree with shadowfacts; this should be `stack`

Prospector (changes requested):
Other than that one `itemStack` parameter, looks good 👍

gudenau:
There are conflicts on this and I have no idea how one would fix that.

gudenau:
I also don't know how to pull the suggested changes.

shadowfacts:
Make the changes locally and then commit & push to your fork. I can deal with resolving the merge conflicts.

gudenau:
Changed the itemStack param to stack.

shadowfacts (approved):

PR #348, opened 2018-12-21
Change BucketInteractable to less specific FluidDrainable

Virtuoel:
Since Sponges also make use of `class_2263`, it seems better suited for it to be named something that is applicable to more than just Buckets. 

The `FluidDrainable` name felt appropriate since in `Waterloggable` the `class_2263` implementation essentially performs the opposite task of the `FluidFillable` implementation.

### DIFF
--- a/mappings/net/minecraft/block/BubbleColumnBlock.mapping
+++ b/mappings/net/minecraft/block/BubbleColumnBlock.mapping
@@ -31,7 +31,7 @@ CLASS bgw net/minecraft/block/BubbleColumnBlock
 		ARG 6 neighborPos
 	METHOD a appendProperties (Lbpn$a;)V
 		ARG 1 builder
-	METHOD b getBucketFluid (Lbbq;Let;Lbpm;)Lcfc;
+	METHOD b tryDrainFluid (Lbbq;Let;Lbpm;)Lcfc;
 		ARG 1 world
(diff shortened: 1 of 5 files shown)


shadowfacts (approved):

mappings/net/minecraft/block/BubbleColumnBlock.mapping
```diff
@@ -31,7 +31,7 @@ CLASS bgw net/minecraft/block/BubbleColumnBlock
 		ARG 6 neighborPos
 	METHOD a appendProperties (Lbpn$a;)V
 		ARG 1 builder
-	METHOD b getBucketFluid (Lbbq;Let;Lbpm;)Lcfc;
+	METHOD b tryDrainFluid (Lbbq;Let;Lbpm;)Lcfc;
```
SquidDev, line 34: Why not just `drainFluid`? I realise it's an action which can fail, but the `try` prefix feels a little superfluous.

Virtuoel: `FluidFillable` initially had `fillWithFluid` which then changed to `tryFillWithFluid`, so with the `try` prefix, the naming is consistent between the interfaces.

SquidDev (approved):

PR #355, opened 2018-12-23
Mapped out some stuff in enchantments.

Darkhax:
This PR mainly focuses on mapping some of the parameters and methods in enchantment related classes. This PR also relates some misc things that got in the way while working on the enchantment stuff. The main one is the new EntityGroup class.

Initially I tried to name EntityGroup as EntityType, but that is used for the registry entry. I then tried to use EntityCategory, but that is mapped for spawn related logic. This class is primarily used to group entities together for things like the enchantment effects.

### DIFF
--- a/mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
+++ b/mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
@@ -6,10 +6,17 @@ CLASS azu net/minecraft/enchantment/EnchantmentHelper
 	METHOD a getHighestApplicableEnchantmentsAtPower (ILawo;Z)Ljava/util/List;
 		ARG 0 power
 		ARG 1 stack
+		ARG 3 applicableEnchantments
+		ARG 4 item
+		ARG 5 isBook
 	METHOD a getSweepingMultiplier (Lahe;)F
 		ARG 0 entity
 		ARG 1 sweepingLevel
 	METHOD a onUserDamaged (Lahe;Lagv;)V
+		ARG 0 user
+		ARG 1 attacker
+		ARG 2 userDamageHandler
+		ARG 3 level
 	METHOD a getEnchantments (Lawo;)Ljava/util/Map;
 		ARG 0 stack
 		ARG 1 enchantments
@@ -17,6 +24,10 @@ CLASS azu net/minecraft/enchantment/EnchantmentHelper
 		ARG 4 enchantmentTag
 		ARG 5 enchantment
 	METHOD a getAttackDamage (Lawo;Lahi;)F
+		ARG 0 stack
+		ARG 1 group
+		ARG 2 damage
+		ARG 3 level
 	METHOD a getEquipmentLevel (Lazs;Lahe;)I
 		ARG 0 ench
 		ARG 1 entity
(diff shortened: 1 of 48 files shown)


mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
```diff
@@ -1,37 +1,82 @@
 CLASS azu net/minecraft/enchantment/EnchantmentHelper
 	CLASS azu$a Consumer
 		METHOD accept handle (Lazs;I)V
-			ARG 1 level
+			ARG 1 enchantment
+			ARG 2 level
 	METHOD a getHighestApplicableEnchantmentsAtPower (ILawo;Z)Ljava/util/List;
 		ARG 0 power
 		ARG 1 stack
```
JamiesWhiteShirt, line 8: There's something weird about all the argument mappings below. They don't match the signatures at all. How did this happen?

JamiesWhiteShirt (changes requested):

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
Darkhax: The first argument (level) was incorrectly mapped, and is an Enchantment object type. I believe this error may have initially been my fault to begin with, as there was an error where some parameters would be mapped incorrectly ages ago.

se7kn8: I think @JamiesWhiteShirt mean the methods below his comment because they do not match the original methods. For example, onUserDamaged() has only two parameters in the signature but you've mapped four arguments.

Darkhax: Oh I see. Those mappings are for local variables. Here is what it looks like in Enigma. I did notice that 3 is shown as `integer` instead of `level` though which is odd. @asiekierka Did I mess up?
image

se7kn8: [Here](https://github.com/FabricMC/yarn/pull/302#issuecomment-447517879) is a comment from @shadowfacts about local variables.

JamiesWhiteShirt (approved):

mappings/net/minecraft/entity/EntityGroup.mapping
```diff
@@ -0,0 +1,6 @@
+CLASS ahi net/minecraft/entity/EntityGroup
```
asiekierka, line 1: Other names to consider: EntityKind? EntityFamily?

asiekierka (approved):
Other than potentially musing on EntityGroup, I like this

mappings/net/minecraft/entity/EntityGroup.mapping
```diff
@@ -0,0 +1,6 @@
...
+	FIELD a DEFAULT Lahi;
```
liach, line 2: `NONE`?

liach (approved):

mappings/net/minecraft/entity/EntityGroup.mapping
Darkhax: I don't think none would be appropriate here, since it's definitely a group still, it's just very broad.

mappings/net/minecraft/entity/EntityGroup.mapping
Darkhax: Those could work, although both of those names would imply something else beyond what it actually is. These are currently used as somewhat broad classifications for mobs based on lore/flavor of the game. Honestly it may be better to think of this class as hardcoded entity tags, would be cool if Mojang just removed this in favor of them tbh.

PR #363, opened 2018-12-25
Some mappings for falling blocks, food items and ParticleArgumentType

Juuxel:
(I have no idea why those previous commits are there...)


mappings/net/minecraft/block/AnvilBlock.mapping
```diff
@@ -11,8 +11,15 @@ CLASS bfz net/minecraft/block/AnvilBlock
 	METHOD <init> (Lbgs$c;)V
 		ARG 1 settings
 		ARG 2 stateBuilder
+	METHOD a configureEntity (Laox;)V
```
SquidDev: I wonder if something like `configureFallingBlock` would be better? `Entity` is a little vague, though I realise it's obvious once you see the method signature.

Juuxel: `configureFallingBlock` could be confusing as the class itself is called FallingBlock.

shadowfacts: Maybe not `configureFallingBlock`, but `configureEntity` is definitely too vague.

mappings/net/minecraft/block/AnvilBlock.mapping
```diff
@@ -11,8 +11,15 @@ CLASS bfz net/minecraft/block/AnvilBlock
...
+		ARG 1 entity
 	METHOD a getPlacementState (Laus;)Lbpm;
 		ARG 1 ctx
+	METHOD a (Lbbp;Let;)V
+		ARG 2 pos
+	METHOD a (Lbbp;Let;Lbpm;Lbpm;)V
+		ARG 2 pos
+		ARG 3 state
```
shadowfacts: There are two state parameters here, what's the difference? The names should reflect that

shadowfacts (changes requested):

mappings/net/minecraft/block/AnvilBlock.mapping
Juuxel: I've accidentally mapped that state parameter, it's supposed to be unmapped for this PR.

mappings/net/minecraft/block/AnvilBlock.mapping
Juuxel: Maybe `configureFallingBlockEntity`?

asiekierka:
If nothing changes, I'm going to close this pull request by next snapshot due to inactivity.

mappings/net/minecraft/block/AnvilBlock.mapping
Juuxel: Fixed.

mappings/net/minecraft/block/AnvilBlock.mapping
Juuxel: I've changed it to `configureFallingBlockEntity`.

Juuxel:
@shadowfacts Could you review again?

0f-0b:
> (I have no idea why those previous commits are there...)

Create a [branch](https://git-scm.com/docs/git-branch) for every PR :)

Juuxel:
> Create a branch for every PR :)

Yeah, I really should.

shadowfacts (approved):

PR #364, opened 2018-12-26
fix mapping package errors

asiekierka:
Also mapped RailPlacementHelper and a few oddball fields, and renamed GLFWUtil to GlfwUtil.

This does not fix all package errors, just some. We need more complete class mappings before we fix the rest.


shadowfacts (approved):

mappings/net/minecraft/world/gen/decorator/CountExtraChanceDecoratorConfig.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS cap net/minecraft/world/gen/config/decorator/CountExtraChanceDecoratorConfig
+CLASS cap net/minecraft/world/gen/ydecorator/CountExtraChanceDecoratorConfig
```
covers1624: `ydecorator`?

asiekierka: Fixed

PR #365, opened 2018-12-28
18w50a exploration 2

Darkhax:
This PR is a few mappings I noticed while reverse engineering some of the vanilla code, not much of a common or logical thread between what was mapped. 

- Bell model and rederer class names
- Class names for where village structure/component data is being registered
- Mapped some of the missing enums and fields in the block states properties class.
- Mapped some methods in entity related to checking spawn light levels, and checking if they're in contact with different water sources.


mappings/net/minecraft/block/enums/Attatchment.mapping
```diff
@@ -0,0 +1,2 @@
+CLASS bqc net/minecraft/block/enums/Attatchment
```
Gegy: `Attachment`?

mappings/net/minecraft/state/property/Properties.mapping
```diff
@@ -12,11 +12,13 @@ CLASS bqd net/minecraft/state/property/Properties
 	FIELD K FACING_HOPPER Lbqh;
 	FIELD L FACING_HORIZONTAL Lbqh;
 	FIELD M WALL_MOUNT_LOCAITON Lbqk;
+	FIELD N ATTATCHMENT Lbqk;
```
Gegy: `ATTACHMENT` here too

mappings/net/minecraft/block/enums/Attatchment.mapping
Darkhax: One day I will learn how to spell lol

mappings/net/minecraft/block/enums/Half.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS bql
+CLASS bql net/minecraft/block/enums/Half
```
asiekierka: Half or a more specific Half?

asiekierka (changes requested):

mappings/net/minecraft/block/enums/Half.mapping
Darkhax: It's just a generic top/bottom half of a block. I was initially going to name this calls BlockHalf, however another class has that name. The other class is used for describing blocks that exist as two separate blocks, like doors and tall grass. The class I have named is used to describe the half of a specific block, like trap doors and stairs.

shadowfacts: Then both should probably be renamed. Having both BlockHalf and Half is confusing if you don't remember which one is which.

Darkhax: Okay, how do these names sound?

BlockHalf -> DoubleBlockHalf
Half -> BlockHalf

Prospector: better than any alternatives I've heard. Do double plants use the former as well?

Darkhax: @Prospector The "TallPlantBlock" class uses it, I think it's all the same class there.

asiekierka (approved):

shadowfacts (approved):

PR #369, opened 2018-12-29
VoxelShape stuff

shadowfacts:
# ~~Do not merge, this is a work in progress.~~

This is my initial attempt at mapping out more of the VoxelShape-related stuff.

Reviews/double checks would be greatly appreciated, there are a lot of names here that I'm not entirely certain about.

### DIFF
--- a/mappings/cmv.mapping
+++ b/mappings/cmv.mapping
@@ -8,5 +8,5 @@ CLASS cmv
 		ARG 0 shape
 		ARG 1 axis
 		ARG 2 amount
-	METHOD a getIncludedCoords (Ley$a;)Lit/unimi/dsi/fastutil/doubles/DoubleList;
+	METHOD a getIncludedPoints (Ley$a;)Lit/unimi/dsi/fastutil/doubles/DoubleList;
 		ARG 1 axis
--- a/mappings/cmv.mapping
+++ b/mappings/cmv.mapping
@@ -2,9 +2,11 @@ CLASS cmv
 	FIELD b parent Lcmx;
 	FIELD c axis Ley$a;
 	FIELD d axisList Lit/unimi/dsi/fastutil/doubles/DoubleList;
-	METHOD a (Lcmm;Ley$a;I)Lcmm;
+	METHOD <init> (Lcmx;Ley$a;I)V
+		ARG 1 shape
+	METHOD a offset (Lcmm;Ley$a;I)Lcmm;
 		ARG 0 shape
 		ARG 1 axis
 		ARG 2 amount
-	METHOD a (Ley$a;)Lit/unimi/dsi/fastutil/doubles/DoubleList;
+	METHOD a getIncludedCoords (Ley$a;)Lit/unimi/dsi/fastutil/doubles/DoubleList;
 		ARG 1 axis
--- a/mappings/net/minecraft/util/shape/AbstractVoxelShapeContainer.mapping
+++ b/mappings/net/minecraft/util/shape/AbstractVoxelShapeContainer.mapping
@@ -1,0 +1,30 @@
+CLASS cmm net/minecraft/util/shape/AbstractVoxelShapeContainer
+	CLASS cmm$b
+	FIELD a xSize I
+	FIELD b ySize I
+	FIELD c zSize I
+	FIELD d AXES [Ley$a;
+	METHOD <init> (III)V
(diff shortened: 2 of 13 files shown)


shadowfacts:
This is pretty much all I'm gonna get. This can be reviewed/merged now.

mappings/cmv.mapping
```diff
@@ -2,9 +2,11 @@ CLASS cmv
...
 	FIELD d axisList Lit/unimi/dsi/fastutil/doubles/DoubleList;
-	METHOD a (Lcmm;Ley$a;I)Lcmm;
+	METHOD <init> (Lcmx;Ley$a;I)V
+		ARG 1 shape
+	METHOD a offset (Lcmm;Ley$a;I)Lcmm;
 		ARG 0 shape
 		ARG 1 axis
 		ARG 2 amount
-	METHOD a (Ley$a;)Lit/unimi/dsi/fastutil/doubles/DoubleList;
+	METHOD a getIncludedCoords (Ley$a;)Lit/unimi/dsi/fastutil/doubles/DoubleList;
```
asiekierka: Or getIncludedPoints?

mappings/net/minecraft/util/shape/AbstractVoxelShapeContainer.mapping
```diff
@@ -0,0 +1,30 @@
...
+		ARG 2 ySize
+		ARG 3 zSize
+	METHOD a isEmpty ()Z
+	METHOD a modify (IIIZZ)V
+		ARG 1 x
+		ARG 2 y
+		ARG 3 z
+		ARG 4 resize
+		ARG 5 included
+	METHOD a getAxisMin (Ley$a;)I
```
asiekierka: getMinAxis/getXSize? getAxisMin/getSizeX? :thinking:

asiekierka (changes requested):

mappings/net/minecraft/util/shape/AbstractVoxelShapeContainer.mapping
shadowfacts: I'm not sure about this. It should definitely be getXSize, as that's how you'd order the words in english, but getMinAxis sounds wrong. What the method does is get the minimum/maximum coordinate in the shape on the given axis. getAxisMin makes sense because, if you didn't have a parameter and instead had individual methods, they'd be getXMin, getXMax, etc.

asiekierka: Why not just getMin/getMax, then? Axis is already an argument.

shadowfacts: That works. Updated.

mappings/net/minecraft/util/shape/AbstractVoxelShapeContainer.mapping
```diff
@@ -0,0 +1,30 @@
...
+		ARG 1 axis
+	METHOD b getXSize ()I
+	METHOD b contains (III)Z
+		ARG 1 x
+		ARG 2 y
+		ARG 3 z
+	METHOD b getMax (Ley$a;)I
+		ARG 1 axis
+	METHOD c getYSize ()I
+	METHOD c getAxisSize (Ley$a;)I
```
asiekierka: if we have getMin and getMax we can prob also have getSize... or maybe getLength? Min... Max; is Size a good term?

asiekierka (changes requested):

mappings/net/minecraft/util/shape/AbstractVoxelShapeContainer.mapping
shadowfacts: getSize is better, IMO. I'd expect length to be used in conjunction with width & height, but it's not here.

asiekierka: Right. IMO don't use width/height in 3D space; it's a tad confusing.

shadowfacts: Fixed.

asiekierka (approved):

modmuss50 (approved):
Looks fine to me, names can always be changed later if people find some better names.

PR #374, opened 2019-01-01
Map a shader effect class and an exception

Pyrofab:
Note: I did not really know where to put those so I put them in the same package as `net.minecraft.client.gl.PostProcessShader`. However there are a few other classes related to post process shader stuff, so I would suggest adding a dedicated `shader` package. If the sentiment is shared, I can do it in this PR.

### DIFF
--- a/mappings/net/minecraft/client/gl/PostProcessShader.mapping
+++ b/mappings/net/minecraft/client/gl/PostProcessShader.mapping
@@ -12,4 +12,6 @@ CLASS dek net/minecraft/client/gl/PostProcessShader
 		ARG 1 time
 	METHOD a setProjectionMatrix (Lcop;)V
 		ARG 1 projectionMatrix
+	METHOD a addAuxTarget (Ljava/lang/String;Ljava/lang/Object;II)V
 	METHOD b getProgram ()Lddw;
+	METHOD close ()V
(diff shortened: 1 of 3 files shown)


Gegy (approved):

mappings/net/minecraft/client/gl/PostProcessShader.mapping
```diff
@@ -12,4 +12,6 @@ CLASS dek net/minecraft/client/gl/PostProcessShader
 		ARG 1 time
 	METHOD a setProjectionMatrix (Lcop;)V
 		ARG 1 projectionMatrix
+	METHOD a addAuxTarget (Ljava/lang/String;Ljava/lang/Object;II)V
```
asiekierka, line 15: addAuxillaryTarget?

asiekierka (approved):

mappings/net/minecraft/client/gl/PostProcessShader.mapping
Pyrofab: I went with the abbreviated version as that's what the json files use, but I guess we can go for something more verbose, it's not used much anyway.

asiekierka: No. If that's what the JSON files use, then it's probably better to stick to that.

PR #375, opened 2019-01-01
Fix draw methods being back-to-front

ChloeDawn:
I derped my last PR, here's the fix for it, finally.


asiekierka:
Can anyone else confirm that this is the case?

ChloeDawn:
I had mixed up the GL modes in my head when writing the original PR. It's harder to keep track when you're glancing at raw ints and not field constants.

`drawBox` and `drawBoxOutline` use `GL_TRIANGLE_STRIP` and `GL_LINE_STRIP`, respectively.

**Reference code:**

drawBox
drawBoxOutline
GL modes

PR #376, opened 2019-01-01
drawBoxOutline -> buildBoxOutline

ChloeDawn:
Another oversight in my original PR. This method does not draw the box, it merely appends the data to the builder. `buildBoxOutline` is equivalent semantics to `buildBox` in this class.


asiekierka:
Screenshot proof: https://github.com/FabricMC/yarn/pull/375

PR #378, opened 2019-01-05
Some advancement mappings

liach:


asiekierka (approved):

mappings/net/minecraft/predicate/entity/EntityEffectPredicate.mapping
```diff
@@ -15,6 +15,7 @@ CLASS bg net/minecraft/predicate/entity/EntityEffectPredicate
 	FIELD a EMPTY Lbg;
 	FIELD b effects Ljava/util/Map;
 	METHOD a create ()Lbg;
+	METHOD a effect (Lagn;)Lbg;
```
Gegy: Haven't checked the method, but should this be `getEffect`?

liach: Notice the return type is bg, which is the predicate itself. It is a setter named in builder style.

asiekierka: If it's a static method, then perhaps create(arg) or createEffect?

liach: This is an instance method.
image
The subclass `a` is per-effect qualification, but enigma fails to rename the subclass and I thus did not rename it.

asiekierka: I'd say "addEffect" or "withEffect" here, if only because, even if builder style, this is *not* a builder - it's the class itself.

liach: `withPresence`? This method just makes the predicate require the presence of the status effect in the parameter.

asiekierka: Eh, let's keep it as-is for now. It's not a particularly important class anyway.

PR #382, opened 2019-01-08
Worldgen mapping

Gegy:
General world generator mappings. Possibly controversial change would be `ChunkGeneratorSettings` -> `ChunkGeneratorConfig`. This is based off of the generic type being `C`, and matching up with the term being used for feature placement.


modmuss50 (approved):
Im not seeing anything wrong with this. Might be best for someone else who knows their way around this to take a look as well.

👍

shadowfacts (approved):

asiekierka (approved):

Issue #300, opened 2018-12-14
Particle spawning method


Juuxel:
`World.method_8494` spawns particles.

Edit: There's already `World.addParticle`. Unlike it, `8494` sets a boolean flag to true.

SquidDev:
If you've time, we really recommend you create a PR for the changes you propose. The README details how to get everything set up, and it's surprisingly easy to start mapping!

Juuxel:
Hm... I think it's a problem that I know the Voldemap name for it?

asiekierka:
Yes. Lord Voldemap is watching.

PR #387, opened 2019-01-09
19w02a mappings

Runemoro:


mappings/net/minecraft/container/CartographyContainer.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS atl
+CLASS atl net/minecraft/container/CartographyContainer
```
asiekierka: CartographyTableContainer?

Runemoro: Done

liach:
what entity is aru?

liach:
btw why important particle in world class?

Runemoro:
> btw why important particle in world class?

It renders even when particles are disabled. They're used for things like bubble columns and campfire signals, which are important and not just decoration.

Juuxel:
Is it the one in #300?

Runemoro:
> Is it the one in #300?

No, but I named that one `addImportantParticle` too in a previous PR. This one does the same thing but with more arguments.

Juuxel:
That issue can be closed by this, then.

Juuxel:
Oh, it's already mapped.

mappings/net/minecraft/container/ContainerProvider.mapping
```diff
@@ -1 +1,2 @@
 CLASS auf net/minecraft/container/ContainerProvider
+	METHOD createMenu (ILarj;Lark;)Latf;
```
SquidDev, line 2: It feels a little odd to refer to it as `Menu`, at least until we've renamed all other `Container` classes.

mappings/net/minecraft/container/ContainerType.mapping
```diff
@@ -0,0 +1,32 @@
+CLASS aug net/minecraft/container/ContainerType
+	CLASS aug$a Factory
+		METHOD create (ILarj;)Latf;
+			ARG 1 syncId
+			ARG 2 playerInventory
+	FIELD a GENERIC_9X3 Laug;
```
SquidDev, line 6: These should automatically be mapped, as the class has a `register`-like method.

mappings/net/minecraft/util/registry/Registry.mapping
```diff
@@ -5,6 +5,7 @@ CLASS fk net/minecraft/util/registry/Registry
 	FIELD D RULE_TEST Lfk;
 	FIELD E STRUCTURE_PROCESSOR Lfk;
 	FIELD F STRUCTURE_POOL_ELEMENT Lfk;
+	FIELD G CONTAINER Lfk;
```
SquidDev, line 8: Again, this should be mapped automatically (though to `MENU`) - though changing this may be intentional?

mappings/net/minecraft/container/ContainerProvider.mapping
Runemoro: It's a vanilla name, it wasn't obfuscated! I just named the parameters but didn't change its name.

mappings/net/minecraft/util/registry/Registry.mapping
Runemoro: It isn't automatically mapped. Look at all the other manually-mapped names above it.

mappings/net/minecraft/container/ContainerProvider.mapping
SquidDev: Ahhh, that's fine then :).

mappings/net/minecraft/util/registry/Registry.mapping
SquidDev: I'm fairly sure most of those can also be mapped automatically - it looks like only the `DefaultMappedRegistry`s need to be explicitly mapped.

Runemoro: So then why are all the others manually mapped, and why doesn't the container field show the automatically mapped name in enigma?

SquidDev: I'm not sure, it's showing up as so for me (I unmapped a couple of other fields too just to see if it was working).

image

It's possible I've a different version of something in the dependency tree, which means they're being mapped when they shouldn't be though. I've no clue, sorry.

modmuss50 (approved):

mappings/net/minecraft/server/WorldGenerationProgressLogger.mapping
```diff
@@ -0,0 +1,9 @@
+CLASS vg net/minecraft/server/WorldGenerationProgressLogger
+	FIELD a LOGGER Lorg/apache/logging/log4j/Logger;
+	FIELD d startTime J
+	FIELD e nextMessageTime J
+	METHOD a start (Lbbt;)V
+		ARG 1 spawnPos
+	METHOD a setChunkStatus (Lbbt;Lbrx;)V
+	METHOD b stop ()V
+	METHOD c getPercentageDone ()I
```
ChloeDawn: `getPercentageDone` -> `getProgress` imo. No need to specify the percentage in the context, and "progress" typically refers to a percentage or completion out of a max amount anyway. The method body is self explanatory.

ChloeDawn (changes requested):

mappings/net/minecraft/server/WorldGenerationProgressLogger.mapping
Runemoro: Progress is usually 0-1. Percentage makes it clear it's 0-100.

mappings/net/minecraft/block/entity/CampfireBlockEntity.mapping
```diff
@@ -0,0 +1,24 @@
+CLASS bol net/minecraft/block/entity/CampfireBlockEntity
+	CLASS bol$a ItemBeingCookedInventory
```
asiekierka: Too verbose, IMO

mappings/net/minecraft/client/gui/WorldGenerationProgressTracker.mapping
```diff
@@ -0,0 +1,23 @@
...
+		ARG 2 z
+	METHOD a start (Lbbt;)V
+		ARG 1 spawnPos
+	METHOD a setChunkStatus (Lbbt;Lbrx;)V
+		ARG 1 pos
+		ARG 2 status
+	METHOD b stop ()V
+	METHOD c getCenterSize ()I
+	METHOD d getSize ()I
+	METHOD e getPercentageDone ()I
```
asiekierka: getProgressPercentage

mappings/net/minecraft/server/WorldGenerationProgressLogger.mapping
asiekierka: getProgressPercentage

liach:
finally!

PR #388, opened 2019-01-10
BlockState-related mappings

MisterPeModder:
Mapped class db and de. also renamed BlockProxy to CachedBlockPosition because the name was bad.


shadowfacts (approved):

mappings/net/minecraft/command/arguments/BlockStateArgumentParser.mapping
```diff
@@ -0,0 +1,47 @@
...
+	METHOD a parse (Z)Lde;
+	METHOD b getBlockState ()Lbqe;
+	METHOD b suggestBlockPropertiesOrEnd (Lcom/mojang/brigadier/suggestion/SuggestionsBuilder;)Ljava/util/concurrent/CompletableFuture;
+	METHOD c getNbtData ()Lhs;
+	METHOD c suggestTagPropertiesOrEnd (Lcom/mojang/brigadier/suggestion/SuggestionsBuilder;)Ljava/util/concurrent/CompletableFuture;
+	METHOD d suggestBlockProperties (Lcom/mojang/brigadier/suggestion/SuggestionsBuilder;)Ljava/util/concurrent/CompletableFuture;
+	METHOD e parseBlockId ()V
+	METHOD e suggestTagProperties (Lcom/mojang/brigadier/suggestion/SuggestionsBuilder;)Ljava/util/concurrent/CompletableFuture;
+	METHOD f parseTagId ()V
+	METHOD f suggestMojangson (Lcom/mojang/brigadier/suggestion/SuggestionsBuilder;)Ljava/util/concurrent/CompletableFuture;
```
asiekierka: Is "Mojangson" a term?

SquidDev: I'd assume it's a contraction of Mojang and Person. It might be better to make this a little clearer though. 

*Edit* And I was entirely wrong. Sorry.

Chocohead: It gets used within the vanilla command blocking community to describe NBT serialised to a JSON like string, contracting Mojang and JSON rather than Person I'd presume.

mappings/net/minecraft/item/ItemStack.mapping
```diff
@@ -61,7 +61,7 @@ CLASS axd net/minecraft/item/ItemStack
 		ARG 2 state
 		ARG 3 pos
 	METHOD a getBlockBreakingSpeed (Lbqe;)F
-	METHOD a areBlockProxiesEqual (Lbqi;Lbqi;)Z
+	METHOD a areBlockCachesEqual (Lbqi;Lbqi;)Z
```
ChloeDawn: This should just be `areBlocksEqual`. The block position and world are not compared in this method, only the block and block entity are compared for equality.

ChloeDawn (changes requested):

mappings/net/minecraft/command/arguments/BlockStateArgumentParser.mapping
ChloeDawn: Yes, Mojangson is a fairly well-known term for Mojang's variation on JSON.

liach:
is mojangson just nbt? if yes, we should call it nbt instead.

mappings/net/minecraft/command/arguments/BlockStateArgument.mapping
```diff
@@ -0,0 +1,7 @@
+CLASS db net/minecraft/command/arguments/BlockStateArgument
+	FIELD a state Lbqe;
+	FIELD b properties Ljava/util/Set;
+	FIELD c data Lhs;
+	METHOD a getBlockState ()Lbqe;
+	METHOD a test (Lbqi;)Z
+	METHOD a setBlockState (Lut;Let;I)Z
```
Juuxel: Some variation of `place` could be used here.

mappings/net/minecraft/command/arguments/BlockStateArgument.mapping
```diff
@@ -0,0 +1,7 @@
+CLASS db net/minecraft/command/arguments/BlockStateArgument
```
Juuxel: Should the name reflect that it also contains BE data?

Chocohead:
Mojangson is specifically the JSON like string form of NBT. Sometimes a Mojangson string is called a data tag, like on the [wiki](https://minecraft.gamepedia.com/Commands#Data_tags), which can be confusing when NBT is sometimes called data tag too.

mappings/net/minecraft/command/arguments/BlockStateArgument.mapping
ChloeDawn: Good point, I overlooked that. Ideally, yes, this could be a more generic `BlockArgument`, and the type would be renamed to `BlockArgumentType`.

mappings/net/minecraft/item/ItemStack.mapping
MisterPeModder: I agree. changing it right now.

mappings/net/minecraft/command/arguments/BlockStateArgument.mapping
MisterPeModder: This also means that `BlockStateArgumentParser` should be renamed to `BlockArgumentParser`

Juuxel (approved):

PR #390, opened 2019-01-10
Coordinates and positions

Juuxel:
The `toActual*` names mean (almost always) the contained position offset with the caller position. (The only exception is `DefaultPosArgument` when it's not relative).

### DIFF
--- a/mappings/dj.mapping
+++ b/mappings/dj.mapping
@@ -1,3 +1,0 @@
-CLASS dj
-	METHOD equals (Ljava/lang/Object;)Z
-		ARG 1 o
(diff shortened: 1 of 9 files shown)


asiekierka (approved):

asiekierka:
toActual versus toAbsolute?

Juuxel:
Both would work.

asiekierka:
toAbsolute is IMO better if you're already using Relative

Juuxel:
I've renamed the `toActual*` methods to `toAbsolute*`.

shadowfacts (approved):

PR #393, opened 2019-01-11
Add field / arg mappings for various things. Also map some some thing…

Blayyke:
…s on `aud` -> `LecternContainer`.


modmuss50 (approved):
Seems simple enough

Issue #386, opened 2019-01-09
Container -> Menu


Runemoro:
The 19w02a snapshot reveals that Mojang calls `Container`s `Menu`s:

net.minecraft.uti.Registry:
```java
public static final Registry<ContainerType<?>> CONTAINER = registerRegistry("menu", new IdRegistry(), () -> ContainerType.ANVIL);
```

`net.minecraft.container.ContainerProvider`:
```java
@FunctionalInterface
public interface ContainerProvider {
    @Nullable
    Container createMenu(int syncId, PlayerInventory playerInventory, PlayerEntity player);
}
```

It would be a good idea to rename Container (and all related things) to Menu, since:
 1. As https://github.com/FabricMC/yarn/issues/351 describes, the name "Container" is wrong
 2. Not using the same names as in vanilla strings and exception messages is confusing and inconsistent
 3. While other names might make sense for now, Mojang might make changes to the menu system in the future, such registering other GUIs that don't need syncing. By using "Menu", we can be sure that our name matches the real meaning of the class as Mojang sees it.
 4. If we named containers anything else, we'd need to remap an unobfuscated Mojang name.
 5. Yarn's goal isn't to fix bad Mojang names

LemmaEOF:
Yeah, there's a (soft) policy of using Mojang names where we know them, so we're probably ultimately gonna switch to `Menu`, and probably `Screen` for client guis.

Prospector:
How about a PR for this now that Screen is merged

PR #394, opened 2019-01-11
Various container mappings

SquidDev:
- Most unregistered GUIs and containers.
 - A couple of commonly-used missing fields.
 - Parts of the new property system, mostly the user-facing things.
 - A couple of random fields which I stumbled upon.

I'm steering well-clear of #386 for now. Just trying to map a couple of fields and methods which I ended up using a lot when trying to update to the latest snapshot.


Runemoro:
I already mapped ~~almost~~ all of these in https://github.com/FabricMC/yarn/pull/387

SquidDev:
Ahh, sorry. I went through the diff when starting this, but clearly not well enough.

PR #396, opened 2019-01-12
Some ServerWorld, GameRenderer, and MinecraftServer mappings

alothmanmoe:
Just some methods I thought to map after using them.

### DIFF
--- a/mappings/net/minecraft/server/MinecraftServer.mapping
+++ b/mappings/net/minecraft/server/MinecraftServer.mapping
@@ -95,6 +95,7 @@ CLASS net/minecraft/server/MinecraftServer
 	METHOD a setResourcePack (Ljava/lang/String;Ljava/lang/String;)V
 		ARG 1 url
 	METHOD a setKeyPair (Ljava/security/KeyPair;)V
+	METHOD a update (Ljava/util/function/BooleanSupplier;)V
 	METHOD a appendCommandFeedback (Ljd;)V
 	METHOD a registerTickable (Lqv;)V
 	METHOD a getSpawnRadius (Lut;)I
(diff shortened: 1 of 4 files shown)


mappings/net/minecraft/server/MinecraftServer.mapping
```diff
@@ -95,6 +95,7 @@ CLASS net/minecraft/server/MinecraftServer
 	METHOD a setResourcePack (Ljava/lang/String;Ljava/lang/String;)V
 		ARG 1 url
 	METHOD a setKeyPair (Ljava/security/KeyPair;)V
+	METHOD a update (Ljava/util/function/BooleanSupplier;)V
```
Runemoro: How is this different from the `tick` method also in MinecraftServer?

alothmanmoe: Whoops, nice catch, yeah it basically is just tick, I mapped it to update on accident while looking at another method I was going to map.

mappings/net/minecraft/server/MinecraftServer.mapping
```diff
@@ -95,6 +95,7 @@ CLASS net/minecraft/server/MinecraftServer
 	METHOD a setResourcePack (Ljava/lang/String;Ljava/lang/String;)V
 		ARG 1 url
 	METHOD a setKeyPair (Ljava/security/KeyPair;)V
+	METHOD a tick (Ljava/util/function/BooleanSupplier;)V
```
liach, line 98: can you name the parameter?

asiekierka (approved):

mappings/net/minecraft/server/MinecraftServer.mapping
Runemoro: Wait, that's not what I meant... Now there's two methods with the same arguments in the same class named the same... 
Look below: `METHOD b tick (Ljava/util/function/BooleanSupplier;)V`

liach:
Runemoro is right. These erroneous tick names definitely have issues.

alothmanmoe:
Yup looking into it now, the method I just mapped with tick is being called by run() every loop, so honestly I'd think the name tick for that would be fine, however, the one currently mapped to tick is called by method_3748 (the one i mapped to tick) which by the looks of it is just updating time, entities, lighting. Maybe remap it to a more fitting name other than tick? Opinions on this?

liach:
rename to describe the boolean suppliers.

asiekierka:
The canAddEntity() one is not exactly right, and we'll probably get to the two other ones later.

PR #406, opened 2019-01-16
Some player related mappings

fewizz:


mappings/net/minecraft/server/network/ServerPlayerEntity.mapping
```diff
@@ -8,12 +8,18 @@ CLASS ut net/minecraft/server/network/ServerPlayerEntity
 	FIELD cb clientLanguage Ljava/lang/String;
 	FIELD cd advancementManager Lqm;
 	FIELD cq clientChatVisibility Larj$b;
+	FIELD cs lastActionTime J
+	FIELD cu inTeleportationState Z
 	FIELD cv seenCredits Z
 	FIELD cw recipeBook Lxw;
 	FIELD d interactionManager Luu;
+	FIELD g seesCredits Z
```
Runemoro: Maybe call it `notInAnyWorld`? It's used to skip certain things such as handling movement when the player is connected but not in any world. Seeing the vanilla credits screen is an example of this, but a mod may want to add their own "not in any world" state.

fewizz: Good idea.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -86,6 +86,10 @@ CLASS ahc net/minecraft/entity/Entity
 	METHOD a setPositionAnglesAndUpdate (DDDFF)V
 		ARG 1 x
 		ARG 3 y
+		ARG 5 z
+		ARG 7 yaw
+		ARG 8 pitch
+		ARG 9 yawDiff
```
shadowfacts: Diff or delta?

PR #404, opened 2019-01-16
19w03a mappings

Runemoro:

PR #408, opened 2019-01-17
Misc recipe mappings + Some HitResult mappings.

Blayyke:


Runemoro:
I already mapped all of these in https://github.com/FabricMC/yarn/pull/404

Blayyke:
Ah okay, didn't see. My bad

PR #409, opened 2019-01-17
Some End fight and BossBar mappings

Noojuno:
Not 100% on where to put the `BossBarManager` and `CommandBossBar` or what a better name for `CommandBossBar` would be that isn't potentially confused with `BossBarCommand`

### DIFF
--- a/mappings/net/minecraft/entity/boss/BossBarManager.mapping
+++ b/mappings/net/minecraft/entity/boss/BossBarManager.mapping
@@ -1,5 +1,5 @@
 CLASS qv net/minecraft/entity/boss/BossBarManager
-	FIELD a minecraftServer Lnet/minecraft/server/MinecraftServer;
+	FIELD a server Lnet/minecraft/server/MinecraftServer;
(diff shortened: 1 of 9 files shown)


mappings/net/minecraft/entity/boss/BossBarManager.mapping
```diff
@@ -0,0 +1,8 @@
+CLASS qv net/minecraft/entity/boss/BossBarManager
+	FIELD a minecraftServer Lnet/minecraft/server/MinecraftServer;
```
asiekierka: server?

asiekierka (changes requested):

mappings/net/minecraft/entity/boss/BossBarManager.mapping
Noojuno: good point, the minecraft was redundant

asiekierka (approved):

shadowfacts (approved):

PR #415, opened 2019-01-20
Some door mappings

YanisBft:


mappings/net/minecraft/block/DoorBlock.mapping
```diff
@@ -65,5 +70,7 @@ CLASS biw net/minecraft/block/DoorBlock
 	METHOD a appendProperties (Lbqj$a;)V
 		ARG 1 builder
 	METHOD c getRenderLayer ()Lbbu;
+	METHOD d getClosingSound ()I;
```
asiekierka: That's not a sound, that's an int

asiekierka (changes requested):

shadowfacts (approved):

mappings/net/minecraft/block/DoorBlock.mapping
```diff
@@ -70,7 +70,7 @@ CLASS biw net/minecraft/block/DoorBlock
 	METHOD a appendProperties (Lbqj$a;)V
 		ARG 1 builder
 	METHOD c getRenderLayer ()Lbbu;
-	METHOD d getClosingSound ()I;
-	METHOD e getOpeningSound ()I;
+	METHOD d getClosingSound ()Lxr;
+	METHOD e getOpeningSound ()Lxr;
```
modmuss50, line 74: Are these methods actaully returning an int or `xr`?

Can these mappings be checked to see if they are still valid.

modmuss50 (changes requested):

mappings/net/minecraft/block/DoorBlock.mapping
Chocohead: They get passed into `World#fireWorldEvent` so should be `()I`

YanisBft: Oh my bad I didn't get what asie requested first

Juuxel: `getClosingWorldEvent`?

PR #417, opened 2019-01-21
Block entities and blocks

Runemoro:
With this PR, there are only 10 names left to be mapped in `net.minecraft.block.entity`


se7kn8 (commented):
Looks good, but maybe there is a better name for `dispenseSimple`?

Runemoro:
> Looks good, but maybe there is a better name for `dispenseSimple`?

Do you have any suggestions? It's a method that dispenses the item without sound or particles.

asiekierka (changes requested):
* dispenseSimple -> dispenseStack
* getTopMaterialColor? Is there a non-top material color?
* CoralParentBlock.checkIfInWater -> checkInWater? Maybe? Not sure...
* InfestedBlock.INFESTED_BLOCK_TO_REGULAR_BLOCK -> INFESTED_TO_REGULAR_MAP? INFESTED_REGULAR_MAP?
* Not sure how I feel about Infested/Regular as terminology
* LecternBlock.tryPuttingBook -> tryPutBook
* PoweredRailBlock.isPoweredFromOtherRails -> isPoweredByOtherRails or even isRailPowered depending on what it's used for
* RedstoneWireBlock.getWireConnections -> getWireConnectionMask
* StructureBlock.STRUCTURE_MODE -> MODE, you don't need to prefix it twice IMO in this case
* TrappedChestBlock.getOpenChestStat -> getOpeningStat? getOpenedStat? getOpenStat?
* VineBlock has shouldConnectTo and cannotConnectTo. I think it's fine, but worth pointing out in case anyone has a better idea
* spawnExperienceOrbs -> dropExperience? Unless it *literally* handles the spawning *only*...
* BarrelBlockEntity.isOpen field -> open? opened?
* lastBeamTextureProgressUpdateTime and beamTextureProgress? Can these be shorter?
* A list called "beam"? Not "beams" or anything?
* BeaconBlockEntity.EFFECTS -> ALL_EFFECTS?
* givePlayersEffects -> applyPlayerEffects?
* CommandBlockBlockEntity.updateAndGetConditionMet also seems verbose... just getConditionMet IMO
* CommandBlockBlockEntity -> CommandBlockEntity in general?
* ConduitBlockEntity.getAttackableZone -> getAttackZone? (Remember, most of these have question marks for a reason! You know better than I do - just suggestions!)
* EndGatewayBlockEntity.getRecentlyGeneratedBeamHeight also seems a bit verbose...
* HopperBlockEntity.tryTransfering -> tryTransfer? tryInsert? tryExtract?
* waterImmediatelyVaporizes/etc. in Dimension -> vaporizeWater?

The rest is good (or, rarely, is not worth bringing up and can be fixed later if someone complains). Amazing work - but could you please make *slightly* smaller PRs next time?

Runemoro:
> getTopMaterialColor? Is there a non-top material color?

Yes, there is. For logs, the return value depends on the axis. Maybe the method should actually be named `getMaaterialColor`/`getMapColor`?

> Not sure how I feel about Infested/Regular as terminology

The block names are "infested cobblestone", "infested stone", etc.

> A list called "beam"? Not "beams" or anything?

A list of beam segments forms the complete beam. Maybe it should be named `beamSegments`?

asiekierka:
>The block names are "infested cobblestone", "infested stone", etc.

Particularly "Regular". I guess it's the best, as "normal" could get confusing...

>A list of beam segments forms the complete beam. Maybe it should be named beamSegments?

Yes, please.

>Yes, there is. For logs, the return value depends on the axis. Maybe the method should actually be named getMaterialColor/getMapColor?

getMapColor is fine.

Runemoro:
> CommandBlockBlockEntity.updateAndGetConditionMet also seems verbose... just getConditionMet IMO

The problem is that there's already a `isConditionMet` that returns `conditionMet` without updating it. This method both updates and returns it. Maybe it could be just named updateConditionMet, since it's clear from the return type it returns the updated value?

asiekierka:
OK

mappings/net/minecraft/block/ChorusFlowerBlock.mapping
```diff
@@ -1,6 +1,23 @@
...
+		ARG 1 pos
+		ARG 2 random
+		ARG 3 size
+	METHOD a generate (Lbcj;Let;Ljava/util/Random;Let;II)V
+		ARG 0 world
+		ARG 1 pos
+		ARG 2 random
+		ARG 4 size
+		ARG 5 layer
+	METHOD a isSurroundedByAir (Lbcm;Let;Ley;)Z
```
LemmaEOF, line 17: maybe `isIsolated` would be better? `isSurroundedByAir` just sounds kinda clunky, imo.

LemmaEOF (approved):
I think I'm generally ok with these, but I'm not a Yarn Expert™.

PR #424, opened 2019-01-25
Move some classes to a more specific package

kashike:


ChloeDawn:
The name `phys` is a known Mojang package name, correct? I'm personally not a huge fan of it. I can see it as being rather confusing to those who don't speak English as their first language, and even then it's a rather uncommon acronym I feel.

mappings/net/minecraft/phys/BoundingBox.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS cmm net/minecraft/util/math/BoundingBox
+CLASS cmm net/minecraft/phys/BoundingBox
```
ChloeDawn: This should be kept to a general `math` package along with any other common vector or math data classes, imo.

ChloeDawn (changes requested):

mappings/net/minecraft/phys/BoundingBox.mapping
kashike: It really shouldn't. It's the axis-aligned bounding box class, used pretty much entirely for physics.

kashike:
`phys` is pretty common for `physics` (`physic` is wrong sounding, and `physics` is plural so no).

mappings/net/minecraft/phys/BoundingBox.mapping
ChloeDawn: It's a data class for holding and transforming two vectors (min, max), there's no need for it to be associated with physics. Maybe this should be brought up for proper discussion

kashike: It's not a generic data class, it's specialised.

Runemoro:
> `phys` is pretty common for `physics` (`physic` is wrong sounding, and `physics` is plural so no).

"physics" is much better. I've never seen it abbreviated as "phys".  Also, "physics" is not plural despite ending with an "s". It's a [mass noun](https://en.wikipedia.org/wiki/Mass_noun), and doesn't have a plural form (see the last sentence in [this section](https://en.wikipedia.org/wiki/Mass_noun#Conflation_of_collective_noun_and_mass_noun)). "physic" is wrong too because [it's archaic](https://en.wiktionary.org/wiki/physic#Noun).

mappings/net/minecraft/phys/BoundingBox.mapping
asiekierka: It is, yes. BoundingBox includes raytracing logic, as far as I see.

asiekierka:
Since 19w04b is *de facto* 19w04a, I'll let this one last a bit longer.

Why physics/ and not util/physics/?

EduardoRFS:
About non native speakers, physics will be a lot easier, i know phys because of PhysX but still needs to think about it, when i read physics it's more direct.
+1 utils/physics

modmuss50:
I think it should be `physics` over `phys`.

Also why not in the `utils `package?

kashike:
It is in `physics` already.

The same reason we don't have everything else in `util`.

asiekierka:
But we do have things like util.math, util.registry, etc. - should we move these out as well, then?

PR #425, opened 2019-01-25
Some block fields

YanisBft:
I think almost all the block fields are now done

### DIFF
--- a/mappings/net/minecraft/block/TurtleEggBlock.mapping
+++ b/mappings/net/minecraft/block/TurtleEggBlock.mapping
@@ -1,6 +1,6 @@
 CLASS bnp net/minecraft/block/TurtleEggBlock
-	FIELD a ONE_EGG_SHAPE Lcnl;
-	FIELD b MULTIPLE_EGGS_SHAPE Lcnl;
+	FIELD a SMALL_SHAPE Lcnl;
+	FIELD b BIG_SHAPE Lcnl;
 	FIELD c HATCH Lbrk;
 	FIELD d EGGS Lbrk;
 	METHOD <init> (Lbhn$c;)V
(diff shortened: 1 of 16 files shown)


mappings/net/minecraft/block/TurtleEggBlock.mapping
```diff
@@ -1,3 +1,7 @@
 CLASS bnp net/minecraft/block/TurtleEggBlock
+	FIELD a ONE_EGG_SHAPE Lcnl;
+	FIELD b MULTIPLE_EGGS_SHAPE Lcnl;
```
asiekierka: ONE_EGG_SHAPE/MULTIPLE_EGG_SHAPE?

asiekierka (approved):

mappings/net/minecraft/block/TurtleEggBlock.mapping
YanisBft: The block has a small shape when EGGS blockstate value is 1, and a bigger shape when it is 2, 3 or 4

asiekierka: Why not just SMALL_SHAPE and BIG_SHAPE?

YanisBft: To be more accurate, but could be good too

mappings/net/minecraft/world/GameMode.mapping
```diff
@@ -1,5 +1,9 @@
 CLASS bci net/minecraft/world/GameMode
 	FIELD a INVALID Lbci;
+	FIELD b SURVIVAL Lbci;
```
asiekierka: Didn't these map themselves (in gray)?

se7kn8 (approved):

mappings/net/minecraft/world/GameMode.mapping
YanisBft: They did, I just added them to better see they are mapped

asiekierka: Don't. If they're gray, DON'T touch them.

modmuss50 (approved):

PR #427, opened 2019-01-25
Crossbows and some entity mappings

Juuxel:


mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -84,6 +84,7 @@ CLASS ahl net/minecraft/entity/LivingEntity
 	METHOD d removeStatusEffect (Lagu;)Z
 	METHOD d isPotionEffective (Lagv;)Z
 	METHOD d getCurrentExperience (Larj;)I
+	METHOD da isPullingItem ()Z
```
Juuxel: This means right-click-holding an item (like pulling bows or using shields). There's probably a better name for this.

asiekierka: Hmm. Right-click-holding has actions... I would consider "charging" as a term, but I wonder if there's a more generic game design-y term for this?

asiekierka (approved):

mappings/net/minecraft/entity/LivingEntity.mapping
Juuxel: *Charging* would work well IMO.

JamiesWhiteShirt: This name also applies to food and potions so I think "charging" doesn't work that well.

Gegy: `isUsingItem`? The term `use` is already used in `ItemStack#getMaxUseTime`

asiekierka: Use sounds a tad generic to me, but might work...

Juuxel: I changed it to `isUsingItem`.

Gegy (approved):

Issue #395, opened 2019-01-12
Gui -> Screen


Runemoro:
Mojang's name for `Gui` is actually `Screen`, as we can see from many vanilla strings. This name makes a lot more sense than `Gui`:
 - Some GUIs don't extend `Gui` and several of those can be open at the same time (chat hud, debug hud, etc.)
 - The GUIs that extend `Gui` fill the whole *screen* and only one can be open at once. The name "screen" makes this clear

asiekierka:
Should we move to *GuiScreen for overlaid screens? (container screens etc)

Prospector:
No, I think just Screen (and even if we did go with Gui and Screen, it would be ScreenGui as a screen would be a type of gui)

Prospector:
This can be closed

PR #432, opened 2019-01-27
Gui -> Screen

Runemoro:
Since https://github.com/FabricMC/yarn/issues/395 has had only positive reactions, I made a pull request.

This only renames screen GUIs (GUIs which fill the entire screen and of which only one can be open at once) to "screen". Other kinds of GUIs such as HUDs and recipe book GUI, and things used by both screen and non-screen GUIs have not been renamed.

Screens have been kept in the `gui.ingame`, `gui.menu`, and `gui.container` package so that they're grouped with other GUI stuff such as HUDs.


EduardoRFS:
That's really good, not only because of Mojang choices because GUI isn't a good name for start. 

About non screen GUI perhaps a better name also would makes sense. Even HUD it's more meaningful.

Also something like Component can be a lot meaningful to it. Specially because it's a popular busy word, perhaps ScreenComponent. And interface can be really well used but I fear that it can be mistaken as Java interfaces

modmuss50 (approved):
I agree with this change, the new names look good to me.

SquidDev (approved):

PR #434, opened 2019-01-27
Added/improved obfuscation of some chunk handling classes

Siphalor:


mappings/net/minecraft/world/chunk/ChunkPos.mapping
```diff
@@ -19,3 +19,5 @@ CLASS bby net/minecraft/world/chunk/ChunkPos
 		ARG 1 o
 	METHOD f getXEnd ()I
 	METHOD g getZEnd ()I
+	METHOD h getRegionX ()I
+	METHOD i getRegionY ()I
```
asiekierka: ... Y?

mappings/net/minecraft/world/chunk/storage/RegionFileCacheVersioned.mapping
```diff
@@ -0,0 +1,10 @@
+CLASS bsx net/minecraft/world/chunk/storage/RegionFileCacheVersioned
```
asiekierka: That's the wrong prefixing order

asiekierka (changes requested):

mappings/net/minecraft/world/chunk/ChunkPos.mapping
Siphalor: Whoops XD

mappings/net/minecraft/world/chunk/storage/RegionFileCache.mapping
```diff
@@ -1,2 +1,17 @@
 CLASS bta net/minecraft/world/chunk/storage/RegionFileCache
+	FIELD d regionFiles Lit/unimi/dsi/fastutil/longs/Long2ObjectLinkedOpenHashMap;
 	METHOD a getRegionFile (Lbby;)Lbsz;
+		ARG 2 regionId
+		ARG 4 regionFile
+		ARG 5 regionFolder
+		ARG 6 mcaFile
+		ARG 7 regionFile2
```
Gegy: What is this parameter?

Siphalor: This is a local variable for a new RegionFile if the first one doesn't exist.
That's due to Mojang making everything final.

Siphalor:
I'm actually not quite sure wether the ChunkSaveHandlerImpl should be renamed.
I think it's a bit misleading because it also handles loading of chunks but I don't know a better name for it.

asiekierka:
Unfortunately, a different PR conflicted with most of this. I tried to re-integrate the interesting things, though.

PR #437, opened 2019-01-30
Various things from 19w05a

YanisBft:
Just little fixes


mappings/net/minecraft/entity/passive/AbstractVillagerEntity.mapping
```diff
@@ -0,0 +1,2 @@
+CLASS aqz net/minecraft/entity/passive/AbstractVillagerEntity
+	FIELD e player Laro;
```
liach: Is this a customer or what?

YanisBft: Yeah it is, will rename it

liach:
lgtm, though optionally you can map more members for that new traveller

PR #438, opened 2019-01-31
DimensionalPersistentStateManager -> PersistentStateManager

JamiesWhiteShirt:
PersistentStateManager used to be a class managing shared persistent data between worlds, and DimensionalPersistentStateManager was for persistent data belonging to a single world.

As of 19w04a the class we knew as PersistentStateManager was removed in favor of using the DimensionalPersistentStateManager of the overworld as a shared persistent data manager.

At this point the mouthful "dimensional persistent state manager" is no longer necessary to distinguish the class from something else.

### DIFF
--- a/mappings/net/minecraft/server/world/ServerChunkManager.mapping
+++ b/mappings/net/minecraft/server/world/ServerChunkManager.mapping
@@ -10,7 +10,7 @@ CLASS uq net/minecraft/server/world/ServerChunkManager
 	FIELD i genQueue Ljava/util/Queue;
 	FIELD j players Luo;
 	FIELD k threadedAnvilChunkStorage Luf;
-	FIELD l dimensionalPersistentStateManager Lcjb;
+	FIELD l persistentStateManager Lcjb;
 	FIELD m maxWatchDistance I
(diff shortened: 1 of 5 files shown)


asiekierka (approved):

mappings/net/minecraft/entity/passive/AbstractVillagerEntity.mapping
```diff
@@ -1,2 +1 @@
 CLASS aqz net/minecraft/entity/passive/AbstractVillagerEntity
-	FIELD e customer Laro;
```
JamiesWhiteShirt: Enigma did this automatically, seems the field no longer exists.

Gegy: Interesting, the obf name seems to be `bD` now.

modmuss50 (approved):
Looks fine, is this ok to merge with that removed field? or do you want to exclude that change before merging?

JamiesWhiteShirt:
I have resolved the issue with unrelated changes. Another PR should be made to fix the other issue(s).

PR #441, opened 2019-02-02
AI mapping

Gegy:


liach:
can we rename pathing to.navigation

mappings/net/minecraft/entity/mob/MobEntity.mapping
```diff
@@ -68,6 +68,7 @@ CLASS aho net/minecraft/entity/mob/MobEntity
 		ARG 0 stack
 	METHOD f setTarget (Lahn;)V
 	METHOD n setCanPickUpLoot (Z)V
+	METHOD o addGoals ()V
```
Pyrofab: Why not `initGoals` ?

Gegy: They both work the same, imo. Any reason for choosing one over the other?

Pyrofab: We already have `initAttributes` and `initDataTracker`, so in my opinion that method should be named similarly.

Gegy: Good point, will change. 👍

modmuss50 (approved):

modmuss50:
Updating, merging, open an issue if you have further comments.

PR #447, opened 2019-02-06
Add more chunkloading-related mappings

Niko-sk2x:
I finally updated this from 18w50a.

There may be some mappings that shouldn't be there that I missed, because enigma didn't detect duplicate mappings and merged some of the untracked 18w50a files.

### DIFF
--- a/mappings/net/minecraft/server/world/ServerChunkManager.mapping
+++ b/mappings/net/minecraft/server/world/ServerChunkManager.mapping
@@ -27,6 +27,11 @@ CLASS uq net/minecraft/server/world/ServerChunkManager
 	METHOD a getWatchDistance (Lbcd;Lut;Z)I
 		ARG 0 chunkPos
 		ARG 1 player
+	METHOD a (Lbda;III)V
+		ARG 1 arg1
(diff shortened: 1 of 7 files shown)


mappings/net/minecraft/server/world/ServerChunkManager.mapping
```diff
@@ -27,6 +27,11 @@ CLASS uq net/minecraft/server/world/ServerChunkManager
 	METHOD a getWatchDistance (Lbcd;Lut;Z)I
 		ARG 0 chunkPos
 		ARG 1 player
+	METHOD a (Lbda;III)V
+		ARG 1 arg1
```
modmuss50, line 31: any reason they are named as `arg1`? Is this a meaningful name?

Gegy: Enigma bug, should be resolved now though. Also should remove itself after load & save.

modmuss50:
We are updating to 19w06a, this PR hasnt had a lot of time for review, so if you have any issues with it, please open an issue, or a PR. It mostly seems fine.

PR #450, opened 2019-02-08
Client/Server -> Clientbound/Serverbound

asiekierka:


Prospector:
Yessssssssss very good mr asie

mappings/net/minecraft/client/network/ClientboundLoginNetworkHandler.mapping
```diff
@@ -0,0 +1,10 @@
+CLASS dci net/minecraft/client/network/ClientboundLoginNetworkHandler
```
LemmaEOF, line 1: I dunno if Clientbound fits here. Same with ClientPlayNetworkHandler. Those are both client-*side*, instead of client-*bound*.

LemmaEOF (changes requested):

Niko-sk2x:
I don't think either is very descriptive. Does `Clientbound` mean "sent by client" or "received by client"?
It's still far from obvious which packets are which just by looking at that part of the name.

If anything is to be changed, I would vote for `StoC` and `CtoS` or some variation of it.

LemmaEOF:
Client bound means by definition "headed toward client", like with homebound.

Prospector:
Bound means going to. If you're homebound then you're on your way home.

Also StoC and CtoS are super ugly and ew ew ew

Niko-sk2x:
Ok, maybe that's just my lack of understanding of english language. Ignore me in that case. I actually didn't know that's what it means.

liach:
just like northbound and southbound are NB and SB, we can name clientbound and serverbound CB and SB

LemmaEOF:
that'd just make things more confusing to people who don't know what it stands for.

Player3324:
From my understanding adding "bound" doesn't add any additional information. "Bound" does not imply any direction, just a context. With MC's unicast based communication every packet is eventually always bound to both a client and a server, so always both "clientbound" and "serverbound".

"Homebound" is the short form of "homeward bound", where the actual direction is encoded in "homeward", while "bound" only means "determined". Extrapolating a general understanding of "<x>bound" from this is quite a stretch, especially if the context is more ambiguous than "home".

The established terminology in networking is "inbound" and "outbound", further reinforcing that "bound" by itself has no directional association. I can't find any evidence that the words "clientbound" and "serverbound" are being used with their implied meaning outside wiki.vg and MCP.

LemmaEOF:
"Bound" does imply a direction. It means it's headed towards the place mentioned. Pedantics like this won't help, because "bound" has a connotational meaning due to "homebound", "inbound", and "outbound". Inbound and outbound do, in fact, have a directional association. Inbound things are headed in, and outbound things are headed out.

liach:
we can say Hostbound/Simulatorbound and Clientbound instead

modmuss50:
I think this change is a good idea. Yes it does not provide anymore detail over what was already known. But it does remove a common place of confusion.

Bound seems fine to me and makes sense, I can see there being some confusion for people who are not familiar with it, but I think there are most likely other cases where this could be more of an issue, so should be ignored.

> we can say Hostbound/Simulatorbound and Clientbound instead

Im not a fan of `Simulatorbound` or `Hostbound`, server makes the most sense.

JamiesWhiteShirt:
In client-server model terms "Host" can refer to both the client and the server. I do not see how Simulatorbound is supposed to make sense either.

If anything, I believe Clientbound/Serverbound is at least better than the current naming and I don't see any viable alternatives.

mappings/net/minecraft/client/audio/SoundLoader.mapping
```diff
@@ -19,8 +19,6 @@ CLASS dsj net/minecraft/client/audio/SoundLoader
 	METHOD a readSounds (Ljava/io/InputStream;)Ljava/util/Map;
 	METHOD a get (Lqd;)Ldsk;
 	METHOD a stopSounds (Lqd;Lxx;)V
-	METHOD a (Lwr;Lafj;)Ljava/util/concurrent/CompletableFuture;
-	METHOD a (Lwr;Ljava/lang/Object;Lafj;)V
```
ChloeDawn, line 23: Unrelated changes

mappings/net/minecraft/client/font/FontRendererManager.mapping
```diff
@@ -4,6 +4,4 @@ CLASS csp net/minecraft/client/font/FontRendererManager
 	FIELD c textureManager Ldpe;
 	FIELD d forceUnicodeFont Z
 	METHOD a getFontRenderer (Lqd;)Lcqu;
-	METHOD a (Lwr;Lafj;)Ljava/util/concurrent/CompletableFuture;
-	METHOD a (Lwr;Ljava/lang/Object;Lafj;)V
```
ChloeDawn, line 8: Unrelated changes

mappings/net/minecraft/client/network/ClientPlayerEntity.mapping
```diff
@@ -18,3 +18,4 @@ CLASS dex net/minecraft/client/network/ClientPlayerEntity
 	METHOD g setServerBrand (Ljava/lang/String;)V
 		ARG 1 serverBrand
 	METHOD s updateHealth (F)V
+	METHOD w closeLocalGui ()V
```
ChloeDawn, line 21: Unrelated changes

mappings/net/minecraft/client/particle/ParticleManager.mapping
```diff
@@ -37,8 +37,6 @@ CLASS ddv net/minecraft/client/particle/ParticleManager
 	METHOD a registerFactory (Lfu;Lddw;)V
 		ARG 1 type
 		ARG 2 factory
-	METHOD a (Lwr;Lafj;)Ljava/util/concurrent/CompletableFuture;
-	METHOD a (Lwr;Ljava/lang/Object;Lafj;)V
```
ChloeDawn, line 41: Unrelated changes

mappings/net/minecraft/client/render/WorldRenderer.mapping
```diff
@@ -127,6 +127,8 @@ CLASS dfo net/minecraft/client/render/WorldRenderer
 		ARG 8 velocityX
 		ARG 10 velocityY
 		ARG 12 velocityZ
+	METHOD a addParticle (Lft;ZDDDDDD)V
+	METHOD a addParticle (Lft;ZZDDDDDD)V
```
ChloeDawn, line 131: Unrelated changes

mappings/net/minecraft/client/render/model/BakedModelManager.mapping
```diff
@@ -8,6 +8,4 @@ CLASS dqs net/minecraft/client/render/model/BakedModelManager
 	METHOD a getMissingModel ()Ldqo;
 	METHOD a getModel (Ldqt;)Ldqo;
 		ARG 1 id
-	METHOD a (Lwr;Lafj;)Ljava/util/concurrent/CompletableFuture;
-	METHOD a (Lwr;Ljava/lang/Object;Lafj;)V
```
ChloeDawn, line 12: Unrelated changes

mappings/net/minecraft/client/resource/FoliageColormapResourceLoader.mapping
```diff
@@ -1,4 +1,2 @@
 CLASS dpp net/minecraft/client/resource/FoliageColormapResourceLoader
 	FIELD a FOLIAGE_COLORMAP_LOC Lqd;
-	METHOD a (Lwr;Lafj;)Ljava/util/concurrent/CompletableFuture;
-	METHOD a (Lwr;Ljava/lang/Object;Lafj;)V
```
ChloeDawn, line 4: Unrelated changes

mappings/net/minecraft/client/resource/GrassColormapResourceLoader.mapping
```diff
@@ -1,4 +1,2 @@
 CLASS dpq net/minecraft/client/resource/GrassColormapResourceLoader
 	FIELD a GRASS_COLORMAP_LOC Lqd;
-	METHOD a (Lwr;Lafj;)Ljava/util/concurrent/CompletableFuture;
-	METHOD a (Lwr;Ljava/lang/Object;Lafj;)V
```
ChloeDawn, line 4: Unrelated changes

mappings/net/minecraft/client/resource/SplashTextResourceLoader.mapping
```diff
@@ -3,5 +3,3 @@ CLASS dpu net/minecraft/client/resource/SplashTextResourceLoader
 	FIELD b RANDOM Ljava/util/Random;
 	FIELD c splashTexts Ljava/util/List;
 	METHOD a get ()Ljava/lang/String;
-	METHOD a (Lwr;Lafj;)Ljava/util/concurrent/CompletableFuture;
-	METHOD a (Lwr;Ljava/lang/Object;Lafj;)V
```
ChloeDawn, line 7: Unrelated changes

mappings/net/minecraft/client/texture/TextureManager.mapping
```diff
@@ -11,7 +11,5 @@ CLASS dpe net/minecraft/client/texture/TextureManager
 		ARG 1 id
 	METHOD a registerTextureUpdateable (Lqd;Ldph;)Z
 		ARG 1 id
-	METHOD a (Lwr;Lafj;)Ljava/util/concurrent/CompletableFuture;
-	METHOD a (Lwr;Ljava/lang/Object;Lafj;)V
```
ChloeDawn, line 15: Unrelated changes

mappings/net/minecraft/client/world/ClientWorld.mapping
```diff
@@ -1,7 +1,9 @@
 CLASS dcm net/minecraft/client/world/ClientWorld
```
ChloeDawn, line 1: Unrelated changes in file

mappings/net/minecraft/server/world/ServerWorld.mapping
```diff
@@ -15,6 +15,8 @@ CLASS ur net/minecraft/server/world/ServerWorld
 	METHOD E getForcedChunks ()Lit/unimi/dsi/fastutil/longs/LongSet;
 	METHOD a init (Lbcy;)V
 		ARG 1 levelInfo
+	METHOD a (Ljava/lang/String;)Lcja;
+		ARG 1 arg1
```
ChloeDawn, line 19: Unrelated changes

mappings/net/minecraft/world/World.mapping
```diff
@@ -96,6 +96,7 @@ CLASS bcu net/minecraft/world/World
 		ARG 1 player
```
ChloeDawn, line 96: Unrelated changes in file

mappings/net/minecraft/tag/TagManager.mapping
```diff
@@ -6,8 +6,6 @@ CLASS yp net/minecraft/tag/TagManager
 	METHOD a blocks ()Lym;
 	METHOD a toPacket (Lis;)V
 		ARG 1 buf
-	METHOD a (Lwr;Lafj;)Ljava/util/concurrent/CompletableFuture;
-	METHOD a (Lwr;Ljava/lang/Object;Lafj;)V
```
ChloeDawn, line 10: Unrelated changes

ChloeDawn (changes requested):

Niko-sk2x:
If there is enough confusion about the meaning of "*bound", I think it should not be used. 

Given how I've seen it used before, I would never think it specifies direction. And maybe some pedantic definition of it does. 

But if so many people don't know that it interpret it differently I still think it's a bad idea. 

Obvious is more important than not-ugly.

asiekierka:
So, I think I'm going to close this and rename them to C2SPacket/S2CPacket instead, no?

LemmaEOF:
Yeah, I think that's what we decided on.

Runemoro:
> So, I think I'm going to close this and rename them to C2SPacket/S2CPacket instead, no?

That's the worst possible name. 

> Yeah, I think that's what we decided on.

Who decided on that?

LemmaEOF:
> That's the worst possible name.

No, the worst possible name is `CToSPacket` and `SToCPacket` which someone actually proposed.

> Who decided on that?

At least a half dozen people arguing in the yarn channel on discord over the past few weeks.

Runemoro:
> No, the worst possible name is `CToSPacket` and `SToCPacket` which someone actually proposed.

That's slightly better since it doesn't use `2` as a word. Why not also rename `BlockState` to `BlockSt8` because it's shorter?

`ClientToServerPacket` is the clearest solution but also the longest, `Clientbound` is clear to 95% of people and short, and `Client` is the shortest but not as clear.

> At least a half dozen people arguing in the yarn channel on discord over the past few weeks.

Not true

asiekierka:
>Who decided on that?

Honestly? I have. "Clear to 95% of people" is IMO not enough, and I'd say it's a fair bit less than 95% from what I've talked to people about it. ClientToServerPacket is better than Clientbound, IMO, and I'm willing to consider that, CToS is just ugly - we're shortening "C" and "S" so we might as well shorten "To", not to mention C2S/S2C actually being used in, say, game protocol analysis. It's not something we made up.

Unless you can re-convince me about Clientbound, I don't think it's a sufficiently suitable name.

asiekierka:
The one thing in favor of Clientbound/Serverbound is that it's commonly used in unofficial Minecraft protocol documentation - wiki.vg, Bukkit nomenclature, and [a Rust library I just learned about](https://docs.rs/ozelot/0.5.0/ozelot/) all seem to use this, though probably by virtue of it leaking from one of these to all the others.

HOWEVER, I found that those terms are used PRIMARILY in Minecraft nomenclature, whereas C2S/S2C's usage is far more varied, including things like general TCP-based protocol discussions. As such (and as because neither of the three above sources are of major significance to the client-side modding scene), these are probably more widespread.

Runemoro:
Another option would be to just drop the `Client`/`Server` suffix. For almost all packets it's clear from the name which side it's meant to be sent to, and the class contains the packet listener that handles it right next to its name. It's also what Mojang uses, and it would make code less verbose.

kashike:
> Another option would be to just drop the `Client`/`Server` suffix. 

Be gone, satan.

asiekierka:
>Another option would be to just drop the `Client`/`Server` suffix.



I thought the point of this discussion was because we admitted the current scheme was already too confusing to users? Not to mention, the single packet 90% of modders will ever interact with - the CustomPayload packet - runs against "being clear which side it's meant to be sent to" already.

grondag:
IMO the names *must* unambiguously indicate direction. I have learned the hard way that otherwise leads to time wasted due to confusion. 

C2S and S2C work for me. “Bound” for me is less clear.

modmuss50:
There needs to be some form of direction included in the packet name imo.

`bound` is still my vote, however there is a majority that disagree. Quite a lot of documentation/code designed to work with MC does use this. I cannot see it used outside of minecraft at all. After having a quick look around I cannot see a replacement.

`C2S` is ugly, yes. But its very clear. Honestly I think this might be the best solution to this issue, its not perfect, but imo its better than what we have atm.

asiekierka:
We can always change it again :thinking:

modmuss50:
Exactly, names can and will be improved on over time. I think most people here agree that `C2S` isnt perfect, but is better.

If someone comes up with a better idea in a month, or 6 then it can easily be changed again.

Finding something that is reasonable and most people agree on is a good middle ground, we would be arguing back and forth indefinitely.

mappings/net/minecraft/client/network/ClientboundLoginNetworkHandler.mapping
phase: Same goes for `ServerLoginNetworkHandler` and friends. Probably a find-and-replace error.

phase:
The _majority_ of wiki.vg refers to packets as being either _Clientbound_ or _Serverbound_. _"bound"_ in this context means _"This is bound to go to the Client"_. The `JoinGamePacket` is _Clientbound_ because it is sent from the Server to the Client.

MCProtocolLib, a widely used library for standalone communication with clients & servers, refers to packets as either being _from_ the Server or _from_ the Client. To continue my example, the [`ServerJoinGamePacket`](https://github.com/Steveice10/MCProtocolLib/blob/dae57d10909a5b7f17db551f2f7bff1f291005bf/src/main/java/com/github/steveice10/mc/protocol/packet/ingame/server/ServerJoinGamePacket.java) is sent _from_ the Server _to_ the Client.

I believe MCProtocolLib's way of naming packets is the simplest. What side is the packet on? Server. What's it called? `ServerJoinGamePacket`. Since you guys aren't using prefixes, it would be named `JoinGameServerPacket`. In my view, this makes the most sense.

asiekierka:
>MCProtocolLib, a widely used library for standalone communication with clients & servers, refers to packets as either being _from_ the Server or _from_ the Client.

I always read those as being _to_ the Server or _to_ the Client. As did many other people. In other words, many of us took double-takes, prompting a discussion.

phase:
I think "the side the packet is on is in the name" is more intuitive than "the side the packet is going to is in the name". I think once you know the format of the class names it doesn't matter too much, you'll end up typing `ServerClass` and searching through the auto completion results.

asiekierka:
As I said, the discussion as a whole is a response to observation done by the community about quality of usage, not a mere whim. If we didn't see the problem with mere Server/Client names, there would be no discussion - and, what's worse, flipping Server and Client *now* would cause a lot, a lot more confusion.

phase:
You've convinced me #458 is the better option :)

modmuss50:
Sorted by #458

Prospector:
it is very sad that #458 won out over this

liach:
Pure `Clientbound` is indeed too long, guess `Cb` would be better

Prospector:
Length was never the issue, the only argument against Clientbound/Serverbound was that some people didn't really understand what the words meant? But to me, that is not a reason because the words themselves are not at all ambiguous.

PR #451, opened 2019-02-08
Decipher RecipeProvider-related mappings

3TUSK:
Did some research on `RecipeProvider` and it was out of my surprise.  
I feel like something is wrong with my current setup; specifically, some local var. aren't correctly renamed. If this PR breaks something, let me know ASAP.  

Other than those, there is still one question left: I am not sure what should interface `gu` and its implementations be called now. It does the following:

  - `gu.a()` -> `gu.toJson`: get a valid recipe json in form of `JsonObject` from GSON. Delegate to `gu.a(JsonObject)` below after writing `Identifier`.
  - `gu.a(JsonObject)` -> `gu.serialize`: write things to that `JsonObject`, see also `BlockEntity.toTag`, but this one returns `void`.
  - `gu.b` -> `gu.getRecipeId`: getter of the recipe id.
  - `gu.c` -> `gu.getSerializer`: getter of corresponding `RecipeSerializer<?>`.
  - `gu.d` -> `gu.toAdvancementJson`: get a valid advancement json in form of `JsonObject` from GSON. This advancement seems to be used for unlocking the recipe defined in `toJson`.
  - `gu.e` -> `gu.getAdvancementId`: getter of id of that advancement.

`gu` is implemented by:

  - `gw$a`, who is in charge of generating shaped recipes.
  - `gx$a`, who is in charge of generating shapeless recipes.
  - `gy$a`, who is in charge of generating smelting/blasting recipes.
  - `gz$a`, who is in charge of generating recipes with just one input.
  - `ha$a`, who is in charge of "generating" all "dynamic" recipes (armor dyeing, map cloning, book duplication, etc.).

For me I would like to call `gu` as `RecipeJsonBuilder`, and thus `ShapedRecipeBuilder$Instance`, `ShapelessRecipeBuilder$Instance`, etc. - however those names sound very strange (esp. `gw`, `gx`, `gy`, `gz` and `ha` are in fact (to a certain extent) builder of builder - `BuilderFactory`?), so I am open for suggestions.


liach:
Just commit your proposed names instead of asking in pr as if this is an issue.

All names sound great, except `serialize` method is better named `fromJson` instead.

liach (approved):

3TUSK:
> except `serialize` method is better named `fromJson` instead.

Actually no, that `serialize` *writes* things to JSON rather than reads things from JSON.

liach:
So `toJson`.

3TUSK:
Decide to call `gu` as `RecipeJsonProvider` and thus other classes are `XXXFactory` and `XXXFactory$XXXProvider` respectively. This concludes the PR. Comments wanted.

mappings/net/minecraft/data/server/recipe/ComplexRecipeJsonFactory.mapping
```diff
@@ -0,0 +1,15 @@
+CLASS ha net/minecraft/data/server/recipe/ComplexRecipeJsonFactory
+	CLASS ha$1
+		METHOD a (Lcom/google/gson/JsonObject;)V
+			ARG 1 arg1
```
asiekierka, line 4: These shouldn't be mapped. I wonder why they got mapped?

mappings/net/minecraft/data/server/recipe/CookingRecipeJsonFactory.mapping
```diff
@@ -0,0 +1,70 @@
...
+			ARG 1 arg1
+			ARG 2 var2
+			ARG 3 var4
+			ARG 4 var6
+	FIELD a output Laxh;
+	FIELD b input Lazt;
+	FIELD c exp F
+	FIELD d time I
+	FIELD e builder Ln$a;
+	FIELD f gruop Ljava/lang/String;
```
asiekierka: group

mappings/net/minecraft/data/server/recipe/ComplexRecipeJsonFactory.mapping
3TUSK: ~~Those are local vars. I heard about the local var. mapping are broken after I mapped those things. How to revert?~~

Edit: wait there aren't any local vars in that method...

asiekierka: Don't worry about it.

PR #456, opened 2019-02-10
Mapped all methods and fields in village package

liach:
Seems there is a broken method in resource listener or so. As a result, there is some removal diff, but it shouldn't break the merge.


liach:
In fact, those removals are from useless diffs in #455.

asiekierka (approved):
depletes and consider sound a bit weird, but I trust you on those I guess

liach:
Consider means ticking all and removing invalid ones. Depletes means to decrease the input stacks' size with the villager trade.

asiekierka:
>Consider means ticking all and removing invalid ones.

Isn't that just "ticking"?

>Depletes means to decrease the input stacks' size with the villager trade.

Isn't that commitTrade or decrease or something? Not sure here, just ideas

modmuss50 (approved):

PR #458, opened 2019-02-12
Client/Server -> S2C/C2S

asiekierka:


phase:
I feel like this is obfuscating intent. The shorthand is only useful for people that know the shorthand, which is the same issue you have with only having `Server` or `Client` in the name.

asiekierka:
But once you know the shorthand it's clear! "server-to-client" and "client-to-server". No place for confusion, both the source and destination are in it, you can't get confused.

modmuss50 (approved):
I think we should go with this, we have decided its not perfect, but its better than what we had, and most people seem not hate it.

PR #460, opened 2019-02-14
Class mappings

Runemoro:


mappings/net/minecraft/inventory/ClearableInventory.mapping
```diff
@@ -0,0 +1,4 @@
+CLASS afy net/minecraft/inventory/ClearableInventory
```
asiekierka: Inventory extends afy, not the other way around.

Runemoro: I know, it's not an inventory, but rather an interface implemented by things that have an inventory that can be cleared. It has a single `clear()` method. `ClearableInventoryOwner` maybe?

asiekierka: Uh. Does the clear() method *require* or *imply* that the cleared thing is an inventory?

Runemoro: The clear method is called:
 - On block entities when using the /clone, /fill, /setblock commands and from a few blocks
 - Directly on inventories from containers and a few other places

asiekierka: Yes I'd argue this is more like just _Clearable_ or _ClearableState_ or something

Runemoro: Renamed it to `Clearable`.

asiekierka (approved):

mappings/net/minecraft/structure/StructurePieceWithDimensions.mapping
```diff
@@ -0,0 +1,15 @@
+CLASS cdr net/minecraft/structure/StructurePieceWithDimensions
+	FIELD a width I
+	FIELD b height I
+	FIELD c depth I
+	FIELD d hPos I
```
Gegy, line 5: hPos?

Runemoro: It's the NBT name

asiekierka: I think it's fine if it's the NBT name, myself

modmuss50 (approved):
Looks good

PR #477, opened 2019-02-17
add contributing document, take two - original by @b0undarybreaker

asiekierka:


LemmaEOF (approved):
Sounds good to me.

CONTRIBUTING.md
```diff
@@ -0,0 +1,85 @@
...
+### 5. Make names explicit and efficient.
+
+When someone calls a method, they should know exactly what it will do and exactly what will result from it. While this can be
+found out by looking at the method code, it's a much better experience for a dev to be able to find out by seeing the name.
+Instead of using terms like `serialize` and `deserialize`, use something like `toJson` and `fromJson` or `toNbt` and
+`fromNbt`, depending on what the result of the method is. Methods should only have matching names if they do the same thing
+with the same structures.
+
+Despite this, there's such thing as a name that's *too* descriptive. Excessively long class or method names can be a major
+annoyance, and a hindrance to being able to program as well. We've recently had large amounts of discussion of the name
```
JamiesWhiteShirt: IMO the document should avoid having some attachment to time. "recently" may be true now, but when someone reads this in 6 months it will be false. I much prefer the "as of writing" style below.

If the packet naming scheme is expected to change soon we might want justify the use of "S2C" to avoid excessively long class names and and avoid the end result of the discussion.

CONTRIBUTING.md
```diff
@@ -0,0 +1,85 @@
...
+scheme for packets, trying to find the best balance between descriptiveness and efficiency. We have, for the time being,
+settled on a name scheme that involves abbreviations. Abbreviations should be avoided when possible, but when you have
+classes with names like `ScoreboardObjectiveUpdateS2CPacket`, the abbreviations are necessary to prevent the name being
+even more excessively long.
+
+As of writing, there is not yet a way to add comments to classes, fields or methods. However, we hope to eventually add it
+to move a bit of the burden off names, as well as give them their sometimes much-needed context.
+
+### 6. Don't be afraid to dispute existing names.
+
```
JamiesWhiteShirt, line 99: This paragraph is very good, but it depends on the state of the packet naming scheme discussion.

CONTRIBUTING.md
```diff
@@ -0,0 +1,85 @@
...
+get upset if someone just takes them and runs with them. Essentially - don't open up a copy of the game decompiled with another
+mapping set "just to check how something works" or "just to check how they named it so I can come up with my own name" - that
+is not the right thing to do! (Of course, some names are derived from official terminology - those are likely to be identical
+across all projects.)
+
+### 3. Stick to Java coding conventions.
+
+Mappings should, of course, stick to the [standard Java naming conventions](https://www.oracle.com/technetwork/java/javase/documentation/codeconventions-135099.html#367).
+Packages should be all lower case, class/enum/interface names should be `TitleCase`, fields and methods should be `camelCase`,
+and constants should be `UPPER_CASE`.
```
JamiesWhiteShirt, line 61: It might also be worth adding getter/setter conventions. `getP` (or `isP` if boolean) and `setP`. I believe we follow those? It appears to be part of JavaBeans convention. https://docstore.mik.ua/orelly/java-ent/jnut/ch06_02.htm

CONTRIBUTING.md
```diff
@@ -0,0 +1,85 @@
...
+to create a cohesive, simple naming scheme that avoids confusion or being too verbose. Having a lot of things named in
+different schemes would be a nightmare for usability, so we want to avoid that. This includes naming schemes not just within
+Yarn but within the greater Minecraft community. Despite that goal, there are some name sources that are inherently more
+"reliable" than others; if an official term is hinted at or can be guessed, you should generally use that - sticking closer to 
+official terminology helps everyone.
+
+## Guidelines
+
+To keep with the goal of consistency within the mappings and the greater Minecraft community, here are a set of guidelines to
+follow when naming unmapped methods.
```
JamiesWhiteShirt: But this applies to more than methods! The final section also applies to already mapped methods (and everything else mapped).

CONTRIBUTING.md
```diff
@@ -0,0 +1,85 @@
...
+`NamespacedString`. In these cases, we should make a community consensus on whether to use a better name, and what the name
+should be.
+
+### 2. Do not rip off other mapping projects.
+
+Other mapping projects are the result of other people's hard work. Coming up with names is not quite trivial, and people can
+get upset if someone just takes them and runs with them. Essentially - don't open up a copy of the game decompiled with another
+mapping set "just to check how something works" or "just to check how they named it so I can come up with my own name" - that
+is not the right thing to do! (Of course, some names are derived from official terminology - those are likely to be identical
+across all projects.)
```
JamiesWhiteShirt: _Why_ is this not the right thing to do? We should probably mention that naming will likely be (subconsciously) influenced by the other mapping project.

JamiesWhiteShirt (changes requested):
The references to "recent" events could just be kept up-to-date, but it's easy to forget. Having the maintenance burden is a bit unnecessary but not the end of the world.

CONTRIBUTING.md
asiekierka: Tried to improve the paragraph

CONTRIBUTING.md
asiekierka: I think that's good enough, for now.

JamiesWhiteShirt (approved):
2. should now prove quite useful when the topic arises every now and then.

CONTRIBUTING.md
```diff
@@ -0,0 +1,105 @@
...
+## Guidelines
+
+To keep with the goal of consistency within the mappings and the greater Minecraft community, here are a set of guidelines to
+follow when naming unmapped classes, fields, methods, method arguments and other aspects of the code.
+
+### 1. When known, stick to official terminology.
+
+This can be a contentious decision, due to many of Notch's original names for things (that he himself has stated were
+terrible choices for names) sticking around. While it could be nice to have names like `Gui` or the like, we have decided to
+use `Screen`, since that's the name for guis that the client-side registry uses. Official terms can be found in many places,
```
Runemoro, line 21: `Screen` is a bad example. In this case the Screen name is much better than Gui as it actually is a screen (it fills the whole screeb and only one can be open at once), and non-screen GUIs are called huds or something else.

LemmaEOF: Oh, that is a good point for differentiation between the two. It was just the only example I could think of where we changed a yarn name after learning the official name.

Runemoro: Although I agree with this rule, I don't think it's what we're currently doing. For example, it's known from an exception message that Mojang doesn't use an Entity suffix for mobs, and several other people think dropping the suffix would be good, yet we still keep it.

Runemoro: @Boundarybreaker A better example would be "World -> Level". But as I said in my last comment, it seems like this rule about using known Mojang terms isn't being applied, so adding it in the guidelines will just confuse people.

LemmaEOF: not the place to have these discussions, really.

Runemoro: How is it not? I'm saying this rule is inaccurate and should be removed.

asiekierka: >For example, it's known from an exception message that Mojang doesn't use an Entity suffix for mobs, and several other people think dropping the suffix would be good, yet we still keep it.

Firstly, I haven't said no; secondly, I have not seen that exception message.

>A better example would be "World -> Level".

That, likewise, is still open - for this exact reason.

>How is it not? I'm saying this rule is inaccurate and should be removed.

Edited to reflect reality, but not removed - notice things like enum names, field names, many method names and class names do actually follow it.

LemmaEOF: > I have not seen that exception message.

In the original draft for this, I said that we'll only move to known official names when we have a source for knowing the official names, and that people proposing a change like that should provide sources. I think that part should be put back in.

CONTRIBUTING.md
```diff
@@ -0,0 +1,105 @@
...
+While they can be useful for differentiation, we feel that in most cases there should be enough room to differentiate
+classes without requiring prefixes. In short: the ideal is for field accesses and method calls to feel like sentences, making
+it easier to reason about code.
+
+### 5. Make names explicit and efficient.
+
+When someone calls a method, they should know exactly what it will do and exactly what will result from it. While this can be
+found out by looking at the method code, it's a much better experience for a dev to be able to find out by seeing the name.
+Instead of using terms like `serialize` and `deserialize`, use something like `toJson` and `fromJson` or `toNbt` and
+`fromNbt`, depending on what the result of the method is. Methods should only have matching names if they do the same thing
```
Runemoro, line 85: "deserialize" has a different meaning than "fromNbt" and "readNbt". The first implies the object being read is not the class itself and the class with the method is just a serializer for that object. The second implies it's a static method creating instances of the owner class from NBT. The third implies it's a non static method loading data from NBT into the already created instance.

So we shouldn't say we should always avoid "serialize" and "deserialize".

Also, are you saying that we should always add the paraneter type (json/nbt in this case) to the method's name rather than overloading methods? That's what it sounds like from what I read.

LemmaEOF: We have decided to, so far, always avoid serialize and deserialize, even with serializer classes.

asiekierka: >The second implies it's a static method creating instances of the owner class from NBT.

Except we haven't really defined how we're using the three :thinking:

CONTRIBUTING.md
liach: Also note that use adjective (phrases) for boolean fields and noun (phrases) for other fields.

CONTRIBUTING.md
```diff
@@ -0,0 +1,105 @@
...
+In keeping with the natural English feel, we prefer to avoid prefixing interfaces and enums with `I` and `Enum` respectively.
+While they can be useful for differentiation, we feel that in most cases there should be enough room to differentiate
+classes without requiring prefixes. In short: the ideal is for field accesses and method calls to feel like sentences, making
+it easier to reason about code.
+
+### 5. Make names explicit and efficient.
+
+When someone calls a method, they should know exactly what it will do and exactly what will result from it. While this can be
+found out by looking at the method code, it's a much better experience for a dev to be able to find out by seeing the name.
+Instead of using terms like `serialize` and `deserialize`, use something like `toJson` and `fromJson` or `toNbt` and
```
liach, line 84: We still use these for dfu `Dynamic<?>`, just a side note.

liach (commented):
@asiekierka Please update the base branch this pr targets.

modmuss50:
Replaced by https://github.com/FabricMC/yarn/pull/733

PR #485, opened 2019-02-19
Add mappings for relative movement

LemmaEOF:
- I've named the forward and side velocities for a LivingEntity `velocityWalk` and `velocityStrafe` based off their names in the controls, but falling is just as common as jumping so I changed that to `velocityVert`. That seems a bit clunky and I'd like ideas.
- `relativeVelocityToGlobal` is also a clunky name but I can't think of a more succinct way to phrase what it does.
- I considered naming `moveRelative` as `moveSelf` because it's only used with `MovementType.SELF` in LivingEntity, but I haven't checked subclasses so it might be overridden elsewhere.

### DIFF
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -99,7 +99,11 @@ CLASS ahl net/minecraft/entity/Entity
 	METHOD a setSize (FF)V
 		ARG 1 width
 		ARG 2 height
-	METHOD a doRelativeMovement (FFFF)V
+	METHOD a applyVelocity (FFFF)V
+		ARG 1 sideVelocity
+		ARG 2 upVelocity
+		ARG 3 forwardVelocity
+		ARG 4 friction
 	METHOD a damage (Lagv;F)Z
 		ARG 1 source
 		ARG 2 amount
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -160,6 +160,7 @@ CLASS ahl net/minecraft/entity/Entity
 	METHOD a playSound (Lxx;FF)V
 		ARG 1 sound
 		ARG 2 volume
+	METHOD a isInFluid (Lyp;Z)Z
 	METHOD a toListTag ([D)Lib;
 	METHOD a toListTag ([F)Lib;
 	METHOD aA spawnSprintingParticles ()V
@@ -172,6 +173,7 @@ CLASS ahl net/minecraft/entity/Entity
 	METHOD aI getSavedEntityId ()Ljava/lang/String;
 	METHOD aJ isValid ()Z
 	METHOD aK isInsideWall ()Z
+	METHOD aL updateRiding ()V
 	METHOD aM getHeightOffset ()D
 	METHOD aN getMountedHeightOffset ()D
 	METHOD aP removeAllPassengers ()V
@@ -191,6 +193,7 @@ CLASS ahl net/minecraft/entity/Entity
 	METHOD ag moveToBoundingBoxCenter ()V
 	METHOD ah getSoundSwim ()Lxx;
 	METHOD ai getSoundSplash ()Lxx;
+	METHOD aj getSoundHighSpeedSplash ()Lxx;
 	METHOD ak checkBlockCollision ()V
 	METHOD am isSilent ()Z
 	METHOD an isUnaffectedByGravity ()Z
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -99,11 +99,10 @@ CLASS ahl net/minecraft/entity/Entity
 	METHOD a setSize (FF)V
 		ARG 1 width
 		ARG 2 height
-	METHOD a applyVelocity (FFFF)V
-		ARG 1 sideVelocity
-		ARG 2 upVelocity
-		ARG 3 forwardVelocity
-		ARG 4 friction
+	METHOD a addVelocity (FFFF)V
+		ARG 1 intendedMovementSideways
+		ARG 2 intendedMovementUp
+		ARG 3 intendedMovementForwards
 	METHOD a damage (Lagv;F)Z
 		ARG 1 source
 		ARG 2 amount
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -99,6 +99,7 @@ CLASS ahl net/minecraft/entity/Entity
 	METHOD a setSize (FF)V
 		ARG 1 width
 		ARG 2 height
+	METHOD a relativeVelocityToAxial (FFFF)V
 	METHOD a damage (Lagv;F)Z
 		ARG 1 source
 		ARG 2 amount
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -99,7 +99,7 @@ CLASS ahl net/minecraft/entity/Entity
 	METHOD a setSize (FF)V
 		ARG 1 width
 		ARG 2 height
-	METHOD a relativeVelocityToAxial (FFFF)V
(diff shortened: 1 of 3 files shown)


mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -17,6 +17,9 @@ CLASS ahu net/minecraft/entity/LivingEntity
 	FIELD bG lastAttackTime I
 	FIELD bJ absorptionAmount F
 	FIELD b LOGGER_LIVING Lorg/apache/logging/log4j/Logger;
+	FIELD bf velocityWalk F
+	FIELD bg velocityVert F
```
asiekierka: Vert? Do we need the shortcut?

LemmaEOF: hmm, I'll see how it feels with `velocityVertical`.

LemmaEOF:
oh shit, I re-read things and re-calculated stuff and I think I mixed a few names up.

mappings/net/minecraft/entity/LivingEntity.mapping
ChloeDawn: These should be using suffix naming.

`walkVelocity`
`verticalVelocity`

ChloeDawn: And shouldnt `walk` just be `horizontal`?

LemmaEOF: > These should be using suffix naming.

I'm going off the scheme of `velocityX`, `velocityY`, and `velocityZ` here. We might wanna change those too, though. Should we?

> And shouldnt `walk` just be `horizontal`?

Well, strafing is horizontal too, and arguably *more* horizontal than walking forward. I'm using the names from Controls for these.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -99,6 +99,7 @@ CLASS ahl net/minecraft/entity/Entity
 	METHOD a setSize (FF)V
 		ARG 1 width
 		ARG 2 height
+	METHOD a doRelativeMovement (FFFF)V
```
ChloeDawn: `moveRelative`?

LemmaEOF: ehh, that could cause trouble.

ChloeDawn: the `do` prefix is bad and `moveRelative` would fit in along with our pre-existing `move`. Maybe you should change `move` to `doMovement` then but I feel this needs discussing properly.

LemmaEOF: that's fair. This should def be talked over a bit more

squeek502:
'Move' and 'relative' are both bad descriptions of what's going on in the Entity `(FFFF)V` function. There's no movement happening at all: position never changes, only velocity. I don't have a good name for it, but the function is more based around updating velocity while applying friction/dampening effects. Maybe something like `updateVelocity` or `applyVelocity`.

Also, just throwing this out there: in [Quake 1, these variables were named `sidemove`, `forwardmove`, and `upmove`](https://github.com/id-Software/Quake/blob/bf4ac424ce754894ac8f1dae6a3981954bc9852d/WinQuake/client.h#L26-L29). I've always liked those names, and the forward/up names have the benefit of making it clear which direction is positive vs negative (i.e. a positive `up` is up, a negative `up` is down).

mappings/net/minecraft/entity/Entity.mapping
modmuss50: `moveRelative` seems ok and goes with the current `move`.

`doMovement` is just a verbose way of saying `move`

EDIT, these names may be wrong anyway, see bellow.

modmuss50:
I do like @squeek502's suggestion, 'move' means go to a specific position so in this case it would be wrong.

The field names make more sense as well.

LemmaEOF:
I'm gonna keep with the `velocity` scheme for the fields, so they're `sideVelocity`, `upVelocity`, and `forwardVelocity`. I've decided to make `doRelativeMovement -> applyVelocity` and `moveRelative -> updateVelocity`.

squeek502:
Did some more digging into this:

- The LivingEntity `(FFF)V` function could probably be named `travel` as there is a `this.world.getProfiler().push("travel");`/`this.world.getProfiler().pop();` surrounding its call in `LivingEntity.updateMovement`. Note that this function *does* affect position.
- The side/forward living entity fields are mostly direct analogs to the float fields in `net.minecraft.client.input.Input`, and I think calling them velocities is wrong, as they are treated as always being between -1 and 1, and are more about the direction the player intends to be moving (relative to their look direction, which is only applied later). Also worth noting that speed effects are applied to these fields (i.e. when sneaking they get multiplied by 0.3, etc) I'm having trouble coming up with a good name for them, but I don't think `velocity` should be it.
- `friction` is probably an incorrect name for the 4th argument of Entity `(FFFF)V` (I was wrong about that before). Throughout the code its treated as a movement speed argument from what I can tell. For example, in the LivingEntity `(FFF)V` function, this is how the Entity `(FFFF)V` is called for normal movement (not falling, not in liquids, etc); comments added by me:

```java
// get the friction modifier of the block we're standing on
if (this.onGround) {
    float_10 = this.world.getBlockState(blockPos$PooledMutable_1).getBlock().getFrictionCoefficient() * 0.91F;
}

// get friction and apply the block's friction modifier
float_11 = 0.16277137F / (float_10 * float_10 * float_10);
if (this.onGround) {
    // method_6029 is a getter for movement speed (the value is set in `updateMovement` 
    // using the return of `this.getAttributeInstance(EntityAttributes.MOVEMENT_SPEED)`)
    // and then friction is applied to that move speed
    float_13 = this.method_6029() * float_11;
} else {
    // seems to be fly speed or something, not sure when this is used
    float_13 = this.field_6281;
}

// update velocity based on intended movement direction and move speed with friction already applied
this.method_5724(float_1, float_2, float_3, float_13);
```

So, my recommendations for naming this stuff:
- LivingEntity `(FFF)V` should be called `travel` or something better that conveys that it takes normalized (-1 to 1) intended movement (a combo of direction+speed, see the note about speed effects above) and actually moves the entity based on that. Arguments could be named something like `intendedMovementSideways`, `intendedMovementUp`, `intendedMovementForward`?
- Entity `(FFFF)V` should be called `updateVelocity` or something better that conveys that it takes intended movement and movement speed and updates velocity accordingly (projects intended movement on the player's look direction and applies the movement speed). Argument names should match the `(FFF)V` function and the fourth should be named `movementSpeed`
- The LivingEntity fields should match the `(FFF)V` arguments
- `net.minecraft.client.input.Input` fields could also be given the same names as the side/forward fields.

---

<details>
<summary>some description of float_4 in (FFFF)V</summary>
Wrote this while trying to understand the 4th argument to the `(FFFF)V` function, not sure if it's of any use but figured I'd keep it somewhere

---

`(FFFF)V` does `float_5 = float_4 / length_of_intended_movement_directions` (where `length_of_intended_movement_directions` is set to 1 if its lower than 1), and then multiplies each intended movement direction by `float_5`, and then projects the intended movement directions on the look direction of the entity and modifies velocity based on that. The `float_4 / length_of_intended_movement_directions` means that the larger that `length_of_intended_movement_directions` is, the smaller `float_5` will be. My best guess is that this is to combat the 'pressing forward/strafe together makes you go faster' problem
</details>

LemmaEOF:
@squeek502 wait, you mean `EntityLiving (FFF)V -> travel` and `Entity (FFFF)V -> updateVelocity`, correct? I think you mixed those two up in the last block.
I wanted to avoid the name `travel` since I thought it was a voldemap name, but if there's a constant string in the code for it I'll name it that.

squeek502:
> @squeek502 wait, you mean `LivingEntity (FFF)V -> travel` and `Entity (FFFF)V -> updateVelocity`, correct? I think you mixed those two up in the last block.

Yes, fixed.

asiekierka:
Can this be closed now?

LemmaEOF:
I believe so, yes. We've got some really good names in Squeek's PR, so we're all good here.

PR #489, opened 2019-02-20
Remap registries

ChloeDawn:
```
IndexedIterable extends Iterable
| get(int) # indexed getter, used for raw id lookups

Registry implements IndexedIterable
|  register(registry, int, identifier, entry) # registers entry and id to raw id
|  register(registry, identifier, entry) # registers entry to id
|  register(registry, string, entry) # registers entry to id
|  create(string, registry, supplier) # creates a new registry (defaulted)
|  create(string, registry) # creates a new registry
|
|  getRawId(entry)
|  getRandom(rand)
|  get(identifier)
|  getIds()
|  getId(entry)
|  getOrEmpty(identifier)
|  containsId(identifier)
|  stream()

MutableRegistry extends Registry
|  set(int, identifier, entry)
|  set(identifier, entry)
|  isEmpty()

SimpleRegistry extends MutableRegistry

DefaultedRegistry extends SimpleRegistry
|  getDefaultId()
```

### DIFF
--- a/mappings/net/minecraft/util/IndexedIterable.mapping
+++ b/mappings/net/minecraft/util/IndexedIterable.mapping
@@ -1,0 +1,3 @@
+CLASS ey net/minecraft/util/IndexedIterable
+	METHOD a get (I)Ljava/lang/Object;
+		ARG 1 index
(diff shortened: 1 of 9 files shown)


modmuss50 (approved):
Looks good, makes it a bit more clear

modmuss50:
Merging now due to update, open an issue with any further comments.

PR #491, opened 2019-02-21
A few assorted entity related mappings

liach:

### DIFF
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -23,10 +23,13 @@ CLASS ahl net/minecraft/entity/Entity
 	FIELD aC NAME_VISIBLE Lpx;
 	FIELD aD SILENT Lpx;
 	FIELD aE NO_GRAVITY Lpx;
+	FIELD aF STATUS Lpx;
 	FIELD aG invulnerable Z
 	FIELD aH scoreboardTags Ljava/util/Set;
 	FIELD aJ pistonMovementDelta [D
 	FIELD aK pistonMovementTick J
+	FIELD aL size Lahm;
+	FIELD aM standingEyeHeight F
 	FIELD aa ENTITY_FLAGS Lpx;
 	FIELD ac chunkX I
 	FIELD ad chunkY I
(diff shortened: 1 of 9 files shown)


mappings/net/minecraft/entity/passive/FoxEntity.mapping
```diff
@@ -67,6 +67,7 @@ CLASS anb net/minecraft/entity/passive/FoxEntity
 	METHOD a setType (Lanb$u;)V
 	METHOD a canJumpChase (Lanb;Lahu;)Z
 	METHOD b setOwner (Ljava/util/UUID;)V
+	METHOD c isFriend (Ljava/util/UUID;)Z
```
ChloeDawn: `canTrust`?

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -233,7 +240,7 @@ CLASS ahl net/minecraft/entity/Entity
 	METHOD bE getMovementDirection ()Lex;
 	METHOD bF getComponentHoverEvent ()Ljl;
 	METHOD bG getBoundingBox ()Lcnf;
-	METHOD bI getEyeHeight ()F
+	METHOD bI getStandingEyeHeight ()F
```
ChloeDawn, line 243: What is the difference between `getStandingEyeHeight` and `LivingEntity#getActiveEyeHeight`?

liach: the active eye height does not handle when the entity is sleeping; the standing eye height is precomputed when the entity is constructed

mappings/net/minecraft/entity/EntityStatus.mapping
```diff
@@ -0,0 +1 @@
+CLASS aie net/minecraft/entity/EntityStatus
```
Pyrofab, line 1: Since this class is only used to compute the size of entities and rendering them, maybe call it `Pose` ?

mappings/net/minecraft/entity/EntitySize.mapping
```diff
@@ -0,0 +1,12 @@
+CLASS ahm net/minecraft/entity/EntitySize
+	FIELD a width F
+	FIELD b height F
+	FIELD c constant Z
+	METHOD a grow (F)Lahm;
```
Pyrofab, line 5: This is used just as often to shrink as to grow, and it returns a new object, so maybe `scaled` ?

Gegy: Imo `grow` also implies growth by extending the edges -- `scaled` would seem more accurate

PR #493, opened 2019-02-21
Use eol=lf on *.mapping files to help Windows users

squeek502:
If a Windows user uses core.autocrlf=true in their git config then the .mapping files get checked out with CRLF line endings, but Enigma saves its mappings with LF endings leading to line ending mismatches. This avoids that.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.mapping text eol=lf


squeek502:
Switched target branch to 19w08b.

modmuss50 (approved):

squeek502:
Updated target branch to 19w11b. Any reason why this isn't being merged?

modmuss50:
Ill pull it in, open an issue if it causes any odd things.

PR #494, opened 2019-02-21
New Gui interfaces for 19w08a

shedaniel:
- Added GuiInputListener (extends `InputListener`)
  An interface for `focusOn`, `setActive`, `getEntries`
  I don't know about this one but `InputListener` is implemented in `MinecraftClient`, so I thought this is Screen specifc
- Added DrawProvider
  Basically an interface for `draw`
- Added `suggestMessage` in `Screen`

Please tell me if there are better names


mappings/net/minecraft/client/gui/GuiInputListener.mapping
```diff
@@ -0,0 +1,6 @@
+CLASS csu net/minecraft/client/gui/GuiInputListener
```
Gegy, line 1: This doesn't purely handle input-related behaviour -- proposing something along the lines of `GuiComponent`.

PR #495, opened 2019-02-21
Add some GameOptions mappings. 

Blayyke:
- cqo -> CloudRenderMode
- cqp -> StringGameOption
- cqk -> AttackIndicator

..and a few field names for GameOption.


mappings/net/minecraft/client/options/CloudRenderMode.mapping
```diff
@@ -0,0 +1 @@
+CLASS cqo net/minecraft/client/options/CloudRenderMode
```
asiekierka, line 1: Is this used globally as an enum/constant set of sorts, or just as an option?

Blayyke: It does get called in WorldRenderer to see if the current setting is OPTIONS_CLOUDS_FANCY, i assume to do its fancy clouds rendering.

E: A field of the class is stored in WorldRenderer.

PR #497, opened 2019-02-21
map the new enum for status effects

LemmaEOF:
named it `StatusEffectType`, which I think fits guidelines.


liach (approved):
Well, I cannot find a better word to categorize `BENEFICIAL`, `HARMFUL`, and `NEUTRAL`.

LemmaEOF:
They're explicitly for status effects too, there aren't any uses outside of the StatusEffect-related classes.

asiekierka:
Well, we *could* call them StatusEffectAlignment, but come on - that's way too confusing.

PR #498, opened 2019-02-22
Updated movement mappings for 19w08b

squeek502:
FFF changed to Vec3d, this updates those functions to match the mappings that were added in https://github.com/FabricMC/yarn/pull/492 + a few other new mappings

### DIFF
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -228,7 +228,7 @@ CLASS ahm net/minecraft/entity/Entity
 	METHOD b onPlayerCollision (Larz;)V
 	METHOD b interact (Larz;Lagg;)Z
 		ARG 1 player
-	METHOD b vec3dSquaredHorizontalLength (Lcnl;)D
+	METHOD b squaredHorizontalLength (Lcnl;)D
 	METHOD b writeCustomDataToTag (Lhx;)V
 	METHOD b removeScoreboardTag (Ljava/lang/String;)Z
 	METHOD b setCustomName (Lji;)V
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -98,6 +98,8 @@ CLASS ahm net/minecraft/entity/Entity
 	METHOD a setRotation (FF)V
 		ARG 1 yaw
 	METHOD a updateVelocity (FLcnl;)V
+		ARG 1 movementSpeed
+		ARG 2 movementInput
 	METHOD a damage (Lagv;F)Z
 		ARG 1 source
 		ARG 2 amount
@@ -144,6 +146,10 @@ CLASS ahm net/minecraft/entity/Entity
 		ARG 1 anchor
 		ARG 2 target
 	METHOD a setBoundingBox (Lcng;)V
+	METHOD a movementInputToVelocity (Lcnl;FF)Lcnl;
+		ARG 0 movementInput
+		ARG 1 movementSpeed
+		ARG 2 yaw
 	METHOD a isTeamPlayer (Lcoo;)Z
 	METHOD a setPositionAndAngles (Les;FF)V
 		ARG 1 pos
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -98,6 +98,8 @@ CLASS ahm net/minecraft/entity/Entity
 	METHOD a setRotation (FF)V
 		ARG 1 yaw
 	METHOD a updateVelocity (FLcnl;)V
+		ARG 1 movementSpeed
```
liach: Can you just name it `speed`? Same for below.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -222,6 +228,7 @@ CLASS ahm net/minecraft/entity/Entity
 	METHOD b onPlayerCollision (Larz;)V
 	METHOD b interact (Larz;Lagg;)Z
 		ARG 1 player
+	METHOD b vec3dHorizontalLengthSq (Lcnl;)D
```
liach: Does horizontal length squared mean just `x*x + z*z`? Moreover, this method name sounds a bit weird

squeek502: > Does horizontal length squared mean just `x*x + z*z`?

Yes, `return vec3d.x * vec3d.x + vec3d.z * vec3d.z;` is the entirety of the method body. Also, it's a static method.

mappings/net/minecraft/entity/Entity.mapping
squeek502: Could do, but it's [sufficiently complicated](https://github.com/FabricMC/yarn/pull/485#issuecomment-465793034) that I don't think 'speed' is descriptive enough. The method is not as simple as `direction * speed` (i.e. the length of the calculated velocity vector is not guaranteed to `== speed`). `movementSpeed` is much more accurate to what's being passed to the function, as it's more like a base movement speed than a speed that is directly applied.

To give a real example, the movementSpeed parameter is the same when both walking and crouching, but the movementInput components are 0.3 of normal when crouching.

liach: How about `attributeSpeed` as this is sound like calculated with vanilla entity attributes?

squeek502: Callers do many different transformations to the value they pass to `updateVelocity`. Very rarely is it an unmodified `EntityAttributes.MOVEMENT_SPEED` (only instance I see of that is in `DolphinEntity.travel`, and even then it's actually `getMovementSpeed()`, which the field it gets is only sometimes set to `EntityAttributes.MOVEMENT_SPEED`). For example, in the [code snippet in my comment here](https://github.com/FabricMC/yarn/pull/485#issuecomment-465793034), friction gets applied to the value before passing it to the function. And a lot of the time it's called with constant values, i.e. here's a call inside FlyingEntity.travel:

```
if (this.isInsideWater()) {
    this.updateVelocity(0.02f, movementInput);
```

squeek502:
Switched `movementSpeed` params to `speed` after looking into things [here](https://github.com/FabricMC/yarn/issues/499#issuecomment-466363020) and realizing that it doesn't really map cleanly to movement speed either. Not sure there's really a 'good' term for this param, as it's a muddled mix of movement speed, acceleration, etc.

Also changed `Entity.vec3dHorizontalLengthSq` to `Entity.vec3dSquaredHorizontalLength` to match methods like `Vec3d.squaredDistanceTo`. Kept the `vec3d` prefix because it's a static method that is really more of a vec3d function (and I wouldn't be surprised to see it moved there in a future snapshot).

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -222,6 +228,7 @@ CLASS ahm net/minecraft/entity/Entity
 	METHOD b onPlayerCollision (Larz;)V
 	METHOD b interact (Larz;Lagg;)Z
 		ARG 1 player
+	METHOD b vec3dSquaredHorizontalLength (Lcnl;)D
```
ChloeDawn: `squaredHorizontalLength` should be fine here. The prefixing is redundant, as the context is implied by the argument. It is a simple private method, and a simple name would suffice.

ChloeDawn (changes requested):

mappings/net/minecraft/entity/Entity.mapping
squeek502: It's a public static method, but fair enough.

asiekierka (approved):

Gegy (approved):

ChloeDawn (approved):

PR #505, opened 2019-02-23
Screen naming

Siphalor:
Changes `Screen.width` and `Screen.height` to `Screen.windowWidth` and `Screen.windowHeight` to clear up their roles.

Changes `ContainerScreen.containerWidth` (and respective `height`) to `ContainerScreen.width` to remove redundancy.

### DIFF
--- a/mappings/net/minecraft/client/gui/ContainerScreen.mapping
+++ b/mappings/net/minecraft/client/gui/ContainerScreen.mapping
@@ -1,7 +1,7 @@
 CLASS cwe net/minecraft/client/gui/ContainerScreen
 	FIELD a BACKGROUND_TEXTURE Lqi;
-	FIELD b containerWidth I
-	FIELD g containerHeight I
+	FIELD b width I
+	FIELD g height I
 	FIELD h container Latu;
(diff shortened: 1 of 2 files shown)


copygirl:
Out of curiosity, is `windowWidth` the "scaled" size or the actual (client) window size?
May I suggest `screenWidth` if the former, and `windowWidth` or `clientWidth` if the latter?

ghost:
`Screen.width` and `Screen.height` are scaled. As far as I can tell they are equal to your window width or height / GUI scale. I would also prefer `Screen.screenWidth` rather than `Screen.windowWidth`. Or perhaps something that indicates that the value is scaled.

Siphalor:
> Out of curiosity, is `windowWidth` the "scaled" size or the actual (client) window size?
> May I suggest `screenWidth` if the former, and `windowWidth` or `clientWidth` if the latter?

My original proposal was `scaledWindow...`.
But `screen...` really sounds better - Changed it.  😉

Gegy (approved):

asiekierka (approved):

PR #507, opened 2019-02-24
Arbitrary Mapping

Gegy:

PR #508, opened 2019-02-25
fix resource mappings

asiekierka:
Please take care of this before #507 - essentially, this one is a lot easier to review (and names some things better, IMO) while *also* containing a lot less changes alongside Enigma's automatic upgrade-related fixes.


liach (approved):

modmuss50 (approved):

mappings/net/minecraft/resource/ObjectPassingResourceReloadListener.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS wv net/minecraft/resource/ObjectPassingResourceReloadListener
+	METHOD a apply (Ljava/lang/Object;Lwt;Lafm;)V
```
Gegy: Not sure about `apply` here, imo `accept` works better. Java stdlib has `Consumer#accept` and `Function#apply`, of which `accept` matches better.

mappings/net/minecraft/resource/ObjectPassingResourceReloadListener.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS wv net/minecraft/resource/ObjectPassingResourceReloadListener
```
Gegy: This name is quite verbose. I believe the term `supplier` fits pretty well in this context; could it be used in the name somehow?

asiekierka: I'm not sure. SupplyingResourceReloadListener?

mappings/net/minecraft/resource/ObjectPassingResourceReloadListener.mapping
asiekierka: I'm referring to the "load" and "apply" stages of resource reloading. The canon names are "prepare" and "apply", FWIW.

asiekierka: But we can discuss "load" -> "prepare" later.

PR #510, opened 2019-02-28
ItemStack cleanup + amount->size

ChloeDawn:
It was settled that `ItemStack` is a name that will remain unchanged in Yarn, and as such, I have updated method names accordingly. It makes no sense to refer to the size of a stack as a count or amount. A stack with a count or amount attribute implies that there are multiple stacks, not multiple objects in a single stack. The only reason Mojang refers to it as a `Count` in NBT is that it is not represented as a stack in their source. However, we are stuck with `ItemStack` and should adapt accordingly.

### DIFF
--- a/mappings/net/minecraft/item/ItemStack.mapping
+++ b/mappings/net/minecraft/item/ItemStack.mapping
@@ -2,7 +2,7 @@ CLASS axx net/minecraft/item/ItemStack
 	FIELD a EMPTY Laxx;
 	FIELD b MODIFIER_FORMAT Ljava/text/DecimalFormat;
 	FIELD c LOGGER Lorg/apache/logging/log4j/Logger;
-	FIELD d amount I
+	FIELD d size I
 	FIELD e updateCooldown I
 	FIELD f item Laxs;
 	FIELD g tag Lhx;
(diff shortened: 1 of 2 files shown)


JamiesWhiteShirt:
I don't see how "size" is an improvement over "amount". If anything, "amount" should be "count" as this matches its NBT key.

ChloeDawn:
The int value determines the size of the stack. You don't have a stack amount or count. Such naming implies multiple stacks, imo. It is a stack size, not a stack count. It is only `Count` within NBT because Mojang does not consider this data class to represent a stack.

mappings/net/minecraft/item/ItemStack.mapping
```diff
@@ -12,42 +12,61 @@ CLASS axx net/minecraft/item/ItemStack
...
+		ARG 1 provider
+		ARG 2 size
+	METHOD <init> (Lhx;)V
+		ARG 1 tag
 	METHOD A getRepairCost ()I
 	METHOD B toTextComponent ()Lji;
+		ARG 1 style
 	METHOD C getUpdateCooldown ()I
-	METHOD D getAmount ()I
+	METHOD D getSize ()I
```
asiekierka, line 26: "size" implies a maximum.

ChloeDawn: In general, collections in Java have a size and can be expanded and contracted as needed. It does not represent a fixed value.

UpcraftLP (approved):

sekwah41:
Personally I think leave it as amount as its still pretty obvious what it is or if it was going to change I agree with @JamiesWhiteShirt where it should be count to match the nbt data and keep consistency.

Runemoro:
Count is much better. Size implies it's some kind of collection, while actually a stack is just an item along with the count telling you *how many* you have of that item.

modmuss50:
I dont think size is right here, according to its definition `a thing's overall dimensions`. `maxSize` could possibly make sense on its own.

Count or amount are both fine imo, I dont see a benefit over either one.

mappings/net/minecraft/item/ItemStack.mapping
Prospector: And even so, stacks do have a size maximum?

ChloeDawn:
Will be reapproached at a later time targetting 1.14 branch

PR #513, opened 2019-03-02
(Nearly) complete MainMenuScreen and Container mappings

holly-hacker:
Completes most of MainMenuScreen and ContainerScreen, adds CubeMapRenderer and RotatingCubeMapRenderer (panorama background) and adds some minor missing mappings for arguments.

I don't usually program in Java, so please do tell me if I screwed up some naming schemes.

### DIFF
--- a/mappings/net/minecraft/client/gui/ContainerScreen.mapping
+++ b/mappings/net/minecraft/client/gui/ContainerScreen.mapping
@@ -1,4 +1,17 @@
 CLASS cwe net/minecraft/client/gui/ContainerScreen
+	FIELD A touchDragStack Laxx;
+	FIELD B touchDropPositionX I
+	FIELD C touchDropPositionY I
+	FIELD D touchDropOriginSlot Lavi;
+	FIELD E touchDropTime J
+	FIELD F touchDropReturningStack Laxx;
+	FIELD G touchHoveredSlot Lavi;
+	FIELD H touchDropTimer J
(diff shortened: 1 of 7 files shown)


mappings/net/minecraft/client/gui/ContainerScreen.mapping
```diff
@@ -1,4 +1,17 @@
 CLASS cwe net/minecraft/client/gui/ContainerScreen
+	FIELD A touchDragStack Laxx;
+	FIELD B touchDropPositionX I
```
asiekierka: Why not just touchDropX/touchDropY? There is such a thing as "too verbose".

mappings/net/minecraft/client/gui/RotatingCubeMapRenderer.mapping
```diff
@@ -0,0 +1,9 @@
+CLASS dgo net/minecraft/client/gui/RotatingCubeMapRenderer
+	FIELD a minecraftClient Lcqx;
```
asiekierka: client

asiekierka (changes requested):
Too verbose, in general!

holly-hacker:
I'd rather have the user instantly understand what they're reading than have them guess but read a few characters less, though I agree with you on the ones you mentioned. Fixed those.

Gegy (approved):

asiekierka (approved):

PR #519, opened 2019-03-10
Polygon -> Quad

JamiesWhiteShirt:
The class allows only 4 vertices, which is a quad.


grondag (approved):

asiekierka:
I fear confusion with BakedQuad

grondag:
A person confusing this with BakedQuad won't stay confused for very long. 

It's only used in Entity models as part of a Box so I guess we could call it ModelQuad or BoxQuad but IMO those names aren't as good.  Quad is more consistent with the other names in this package.

PR #522, opened 2019-03-14
19w11a Gui

shedaniel:


modmuss50 (approved):
Looks good, I agree with renaming the current ButtonWidget and moving it to this, was just about to make a PR for this my self ;)

PR #523, opened 2019-03-14
19w11b FoodItem replacements

Terrails:


Terrails:
Oops, messed this up while copying over from 19w11a.
I was also thinking about making the methods inside Builder be e.g `food(int amount)` instead of `setFood(int amount)`

mappings/net/minecraft/item/FoodItemSetting.mapping
```diff
@@ -0,0 +1,37 @@
...
+		METHOD a setSaturation (F)Lawp$a;
+			ARG 1 amount
+		METHOD a setFood (I)Lawp$a;
+			ARG 1 amount
+		METHOD a addStatusEffect (Lahq;F)Lawp$a;
+			ARG 2 chance
+		METHOD b alwaysEdible ()Lawp$a;
+		METHOD c eatenFast ()Lawp$a;
+		METHOD d build ()Lawp;
+	FIELD a food I
```
LemmaEOF: this and any other place it says "food" should probably say "hunger", like with the old FoodItem stuff.

Terrails: Would renaming the fields to "hunger" and getters/setters in `FoodItemSetting` to getHunger/setHunger do it? Since some methods that I made contain "food" make sense to be called like that, for example LivingEntity#eatFood and LivingEntity#applyFoodEffects. And calling the class `FoodItemSetting` makes sense.

LemmaEOF: Yeah, that's what I mean. Just anywhere where it's talking about hunger restoration.

LemmaEOF (approved):

mappings/net/minecraft/item/FoodItemSetting.mapping
```diff
@@ -0,0 +1,37 @@
...
+			ARG 2 chance
+		METHOD b alwaysEdible ()Lawp$a;
+		METHOD c eatenFast ()Lawp$a;
+		METHOD d build ()Lawp;
+	FIELD a hunger I
+	FIELD b saturation F
+	FIELD c wolfFood Z
+	FIELD d alwaysEdible Z
+	FIELD e eatenFast Z
+	FIELD f statusEffectsChances Ljava/util/List;
```
YanisBft: statusEffectChances, with singular "effect"

Prospector:
Do these specifically apply to food or drinks too? I'd say alwaysConsumable and consumedQuickly and not use the word eat (unless it's *explicity* only used for food)

LemmaEOF:
PotionItem does not use any of the FoodItemSettings, and use a different UseAction (DRINK instead of EAT). While modded drinks *would* use alwaysEdible or eatenFast, it's not something in vanilla.

Prospector:
Ah ok, I retract what I said in that case.

mappings/net/minecraft/item/FoodItemSetting.mapping
```diff
@@ -0,0 +1,37 @@
+CLASS awp net/minecraft/item/FoodItemSetting
+	CLASS awp$a Builder
+		FIELD a hunger I
+		FIELD b saturation F
+		FIELD c wolfFood Z
```
Gegy, line 5: Is this specifically wolf food or could it be considered to be more generally 'meat'?

LemmaEOF: We called it "wolfFood" on the old FoodItem, and its only use in vanilla is whether wolves can eat it. I don't remember if cooked fishes count as wolf food, but then again whether fish is considered a type of meat is already super vague culinarily.

raphydaphy: Wolves can only be fed chicken, pork, beef, mutton, rabbit and rotten flesh according to the [Minecraft Wiki](https://minecraft.gamepedia.com/Wolf#Taming.2C_health_and_feeding)

LemmaEOF: Some background on "is fish meat": various branches of Christianity have at various times considered fish as non-meat, so that it could be eaten during lent to prevent starvation. there's also a distinction between "meat" and "fish" in cooking due to the vastly different types of flavors domesticated animal meat and fish can have, but there's also a distinction between "meat" and "game meat", or hunted animal meat, for the same reason. However, fish is still structurally a type of meat, so I don't feel "is this meat" is a good descriptor for its usage in Vanilla.

mappings/net/minecraft/item/FoodItemSetting.mapping
Terrails: ~~Its not a single effect, its a `List<Pair<StatusEffectInstance, Float>>`~~

Ah never mind, will fix that

mappings/net/minecraft/item/FoodItemSetting.mapping
```diff
@@ -0,0 +1,37 @@
...
+		FIELD d alwaysEdible Z
+		FIELD e eatenFast Z
+		FIELD f statusEffectChances Ljava/util/List;
+		METHOD a wolfFood ()Lawp$a;
+		METHOD a setSaturation (F)Lawp$a;
+			ARG 1 amount
+		METHOD a setHunger (I)Lawp$a;
```
Terrails: Would renaming this to `hunger` and `saturation` look better?

ghost: I think it would be better as it would fit how methods are already named in the  existing BlockSettings and ItemSettings builders e.g. new Item.Settings().durability(500).build()

YanisBft (approved):

modmuss50 (approved):

PR #527, opened 2019-03-15
Updated Villager Mapping

Gegy:


mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -33,6 +33,7 @@ CLASS aun net/minecraft/entity/player/PlayerEntity
 	FIELD g spawnPosition Leu;
 	METHOD a sendVillagerRecipes (ILbfc;IIZ)V
 		ARG 1 syncId
+		ARG 5 canLevel
```
liach: Isn't this like showing level in screen which is true for villagers but false for wandering traders?

mappings/net/minecraft/util/math/GlobalPos.mapping
```diff
@@ -0,0 +1,14 @@
+CLASS fb net/minecraft/util/math/GlobalPos
```
liach: `UniversalPos`? `Global` does not fit the minecraft theme.

mappings/net/minecraft/village/VillagerData.mapping
```diff
@@ -1,11 +1,25 @@
...
 	METHOD a withType (Lauf;)Laub;
+		ARG 1 type
 	METHOD a serialize (Lcom/mojang/datafixers/types/DynamicOps;)Ljava/lang/Object;
 	METHOD b getProfession ()Laud;
+	METHOD b getLowerLevelExperience (I)I
+		ARG 0 level
 	METHOD c getLevel ()I
+	METHOD c getUpperLevelExperience (I)I
+		ARG 0 level
+	METHOD d isLevelValid (I)Z
```
liach, line 21: I remember this one returns false for stone level. Is this `hasLevelledUp`?

liach (approved):

mappings/net/minecraft/entity/player/PlayerEntity.mapping
Gegy: Yep

mappings/net/minecraft/util/math/GlobalPos.mapping
Gegy: The Mojang name is `GlobalPos`, it seems.

mappings/net/minecraft/village/VillagerData.mapping
Gegy: Oh, potentially. Will have to check that, thanks!

mappings/net/minecraft/entity/ai/brain/EntityMemoryModule.mapping
```diff
@@ -0,0 +1,10 @@
+CLASS anw net/minecraft/entity/ai/brain/EntityMemoryModule
```
Runemoro: The vanilla name is MemoryModule**Type** (based on name in `Registry`). It doesn't actually store any data, it's used as a key to get the memory module.

Gegy:
This has pretty much been superseded by #533 -- once that is merged I'll make a PR to add the changes from here not common between the two.

liach:
hmm? why reopen?

Gegy:
Will merge into this PR

mappings/net/minecraft/village/VillagerData.mapping
Gegy: Seems the minimum level value is `1`

mappings/net/minecraft/entity/ai/brain/Brain.mapping
```diff
@@ -20,3 +25,4 @@ CLASS aiu net/minecraft/entity/ai/brain/Brain
 	METHOD b stopAllTasks (Lvc;Laif;)V
 	METHOD c getMemory (Lanw;)Ljava/util/Optional;
 	METHOD c hasActivity (Lavw;)Z
+	METHOD e clone ()Laiu;
```
liach, line 28: Are you kidding me... This is probably the worst mapping that can ever happen.

Gegy: elaborate?

liach: Looks like it overrides `Object#clone` by accident. #546

grondag: You’re probably right, but curious: if Mojang did actually override clone() would it be obfuscated? Do we have a reliable way to know if this is or isn’t clone()?

Gegy: Oops. 😄 
Apologies about that..

liach: It shouldn't be. I think now enigma correctly handles synthetic methods generated for such overrides that changes the return type in the signature (see metadata section reader classes)

PR #528, opened 2019-03-16
Increase max memory for Enigma

Runemoro:
Export source is now crashing with the default memory limit. Increasing it to 2GB fixes the problem and makes it faster than it ever was.


modmuss50:
Looks good.

PR #531, opened 2019-03-16
Map more gui

shedaniel:
Removed 1 mapping from #522, `focused` from `AbstractButtonWidget`. It just didn't make sense, I am thinking of something like `wasFocused` or `wasNarrated`.


modmuss50 (approved):
Seems easy enough. Thanks

Gegy (approved):

PR #533, opened 2019-03-16
AI mappings

Runemoro:


Gegy:
This intersects with #527

Runemoro:
> This intersects with #527

Sorry, I didn't realize you mapped that too. I'll try to remove the duplicated parts.

Gegy:
You've mapped quite a bit more than me, and I think you've got some nicer names here. What strategy should we take towards merging?

Prospector (approved):

mappings/net/minecraft/entity/Bird.mapping
```diff
@@ -0,0 +1 @@
+CLASS app net/minecraft/entity/Bird
```
Gegy, line 1: BirdEntity?

mappings/net/minecraft/entity/ai/brain/Brain.mapping
```diff
@@ -0,0 +1,22 @@
+CLASS aiu net/minecraft/entity/ai/brain/Brain
+	FIELD a memories Ljava/util/Map;
+	FIELD b sensors Ljava/util/Map;
+	FIELD c tasks Ljava/util/Map;
+	FIELD d schedule Lavy;
+	FIELD g activities Ljava/util/Set;
```
Gegy: Iirc these are all the activities that can currently be performed -- maybe this name should be more specific?

mappings/net/minecraft/entity/ai/brain/Brain.mapping
```diff
@@ -0,0 +1,22 @@
...
+	FIELD h activity Lavw;
+	METHOD a doActivity (J)V
+	METHOD a forget (Lanw;)V
+	METHOD a isMemoryInState (Lanw;Lanx;)Z
+	METHOD a setMemory (Lanw;Ljava/lang/Object;)V
```
Gegy: Maybe `putMemory`? Matches with the `Map` backend

mappings/net/minecraft/entity/ai/brain/Schedule.mapping
```diff
@@ -0,0 +1,13 @@
+CLASS avy net/minecraft/entity/ai/brain/Schedule
+	FIELD a EMPTY Lavy;
+	FIELD b SIMPLE Lavy;
+	FIELD c VILLAGER_BABY Lavy;
+	FIELD d VILLAGER_DEFAULT Lavy;
+	FIELD e scheduleRules Ljava/util/Map;
+	METHOD a getActivity (I)Lavw;
```
Gegy: Proposing `getActivityForTime`

mappings/net/minecraft/entity/ai/brain/ScheduleBuilder.mapping
```diff
@@ -0,0 +1,11 @@
+CLASS avz net/minecraft/entity/ai/brain/ScheduleBuilder
+	CLASS avz$a ActivityData
```
Gegy: iirc this is referred to as `ActivityEntry` from a string somewhere

mappings/net/minecraft/entity/ai/brain/ScheduleBuilder.mapping
```diff
@@ -0,0 +1,11 @@
...
+		FIELD b activity Lavw;
+		METHOD b getActivity ()Lavw;
+	FIELD b activities Ljava/util/List;
+	METHOD <init> (Lavy;)V
+		ARG 1 schedule
+	METHOD a build ()Lavy;
+	METHOD a addActivity (ILavw;)Lavz;
+		ARG 1 priority
```
Gegy: I may be very wrong, but I thought this is the time of day that the activity starts?

Gegy: Oh, also it may be worth calling this `withActivity` to fit builder pattern.

mappings/net/minecraft/entity/ai/brain/ScheduleRule.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS awa net/minecraft/entity/ai/brain/ScheduleRule
+	METHOD a getPriority (I)F
+		ARG 1 time
```
Gegy, line 3: Not really too important, but this is specifically time of day, maybe could be named to reflect that.

mappings/net/minecraft/entity/ai/brain/WalkTarget.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS any net/minecraft/entity/ai/brain/WalkTarget
```
Gegy, line 1: What does this class do? Seems odd for a WalkTarget to hold a LookTarget

mappings/net/minecraft/entity/ai/brain/sensor/Sensor.mapping
```diff
@@ -0,0 +1,9 @@
+CLASS aon net/minecraft/entity/ai/brain/sensor/Sensor
+	FIELD b SENSE_COOLDOWN I
```
Gegy: I don't think this is static final; also since I'm here proposing `interval` rather than cooldown

mappings/net/minecraft/entity/ai/brain/sensor/Sensor.mapping
```diff
@@ -0,0 +1,9 @@
...
+	METHOD a getRequiredMemoryModules ()Ljava/util/Set;
```
Gegy: Required Modules is correct, but I think it's more accurately described as Output Modules

mappings/net/minecraft/entity/ai/brain/task/Task.mapping
```diff
@@ -0,0 +1,33 @@
+CLASS ajf net/minecraft/entity/ai/brain/task/Task
+	CLASS ajf$a Status
+	FIELD b timeLimit J
+	FIELD c minRunTime I
+	FIELD d maxRunTime I
+	METHOD a getRequiredMemoryState ()Ljava/util/Set;
+	METHOD a isTimeLimitExceeded (J)Z
+		ARG 1 time
+	METHOD a shouldRun (Lvc;Laif;)Z
```
Gegy, line 13: I believe this is specific for checking whether the task should start running, not continue running

mappings/net/minecraft/entity/ai/brain/task/Task.mapping
```diff
@@ -0,0 +1,33 @@
...
+		ARG 3 time
+	METHOD b tryStarting (Lvc;Laif;J)Z
+		ARG 1 world
+		ARG 2 entity
+		ARG 3 time
+	METHOD c tick (Lvc;Laif;J)Z
+		ARG 1 world
+		ARG 2 entity
+		ARG 3 time
+	METHOD d keepRunning (Lvc;Laif;J)V
```
Gegy, line 27: `shouldKeepRunning`?

mappings/net/minecraft/entity/ai/brain/task/Task.mapping
```diff
@@ -0,0 +1,33 @@
+CLASS ajf net/minecraft/entity/ai/brain/task/Task
+	CLASS ajf$a Status
+	FIELD b timeLimit J
```
Gegy: `timeLimit` seems relative to start time -- iirc this is in relation to time of day

mappings/net/minecraft/entity/passive/VillagerEntity.mapping
```diff
@@ -1,6 +1,8 @@
 CLASS aua net/minecraft/entity/passive/VillagerEntity
 	FIELD bD VILLAGER_DATA Lqe;
+	METHOD a initBrainSchedule (Laiu;)V
```
Gegy: I think this is responsible for general brain initialization, not just the schedule

Gegy:
Pretty much just added a load of nitpicks. This is looks great.

mappings/net/minecraft/entity/Bird.mapping
Runemoro: We also have `CrossbowUser` (rather than `CrossbowUserEntity`) and `Npc` (rather than `NpcEntity`) interfaces. It seems like the current convention is to not add the entity suffix to interfaces (technically, they could be implemented by non-entities too).

Gegy: Ah, I assumed this was a base-class rather than an interface.

mappings/net/minecraft/entity/ai/brain/ScheduleBuilder.mapping
Runemoro: Can't find that string, but I'll rename it to `ActivityEntry` anyway.

mappings/net/minecraft/entity/ai/brain/WalkTarget.mapping
Runemoro: It's what gets stored in the `walk_target` memory module.

mappings/net/minecraft/entity/ai/brain/task/Task.mapping
Runemoro: This is called to make the task continue running, if it returned true when `shouldKeepRunning` was called.

mappings/net/minecraft/entity/ai/brain/task/Task.mapping
Runemoro: I'll name it `endTime`.

mappings/net/minecraft/entity/ai/brain/task/Task.mapping
Gegy: Ah, I didn't notice `shouldKeepRunning` existed already.

mappings/net/minecraft/entity/ai/brain/task/Task.mapping
Runemoro: Yes, it is, but the start method is called "run". Most tasks implement only the run method and not keepRunning, so it doesn't make sense to call "run" "start" since in most cases, it not only starts, but runs the whole task.

Gegy (approved):

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -45,7 +45,7 @@ CLASS ahw net/minecraft/entity/Entity
 	FIELD ax BREATH Lqe;
 	FIELD ay CUSTOM_NAME Lqe;
 	FIELD az NAME_VISIBLE Lqe;
-	FIELD b maxEntityId Ljava/util/concurrent/atomic/AtomicInteger;
+	FIELD b MAX_ENTITY_ID Ljava/util/concurrent/atomic/AtomicInteger;
```
liach, line 48: Uh no, atomic integer is mutable.

Runemoro: It's `static final`. A field like `INSTANCE` is always uppercase despite being mutable, so your argument that uppercase is immutable doesn't work.

modmuss50 (approved):

PR #534, opened 2019-03-16
Client class mappings

Runemoro:


mappings/net/minecraft/sortme/Hand.mapping
```diff
@@ -1,3 +1,3 @@
-CLASS aie net/minecraft/sortme/OptionMainHand
+CLASS aie net/minecraft/sortme/Hand
```
Prospector: What's this used for? There's already the Hand enum...maybe a different name would be best

mappings/net/minecraft/client/gui/IButton.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS cwu
+CLASS cwu net/minecraft/client/gui/IButton
```
Prospector: needs a better name

mappings/net/minecraft/client/gui/menu/RedErrorScreen.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS cxs net/minecraft/client/gui/menu/ErrorScreen
+CLASS cxs net/minecraft/client/gui/menu/RedErrorScreen
```
Prospector: why? what other error screen is there?

mappings/net/minecraft/sortme/Hand.mapping
LemmaEOF: yeah, having two classes named `Hand` is a bad idea.

mappings/net/minecraft/client/gui/IButton.mapping
LemmaEOF: We want to avoid using `I` as a prefix if we can. Currently the only I-prefixed class is `IWorld`, and we've been looking to change it too.

mappings/net/minecraft/client/gui/menu/RedErrorScreen.mapping
Runemoro: The regular one. This one is the one with the red background: https://bugs.mojang.com/secure/attachment/122747/English.png

mappings/net/minecraft/client/gui/IButton.mapping
Runemoro: It's an interface implemented by all button widgets. Any suggestions?

mappings/net/minecraft/client/gui/menu/RedErrorScreen.mapping
LemmaEOF: Ah, I'd always thought that was it showing a red-tinted world skybox with no terrain loaded. Is it ever used for something other than the "Unable to load worlds" error screen?

mappings/net/minecraft/client/gui/IButton.mapping
LemmaEOF: It could just be named `Button`.

mappings/net/minecraft/sortme/Hand.mapping
Runemoro: It's passed as an argument to ModelWithArms.setHandAngle.

Runemoro: Now that I think of it, since there's another one, I'm guessing both are called the same by Mojang and they unintentionally used the options one for that method,

mappings/net/minecraft/client/gui/menu/RedErrorScreen.mapping
Runemoro: No, but it's not hardcoded for that error so its name should stay generic in case vanilla or a mod uses it for something else in the future. It could be called SevereErrorScreen if you want a more semantic name.

LemmaEOF: I think `ServerErrorScreen` would be a lot better. Describing something by its visual features doesn't really feel right to me.

Pyrofab: `SevereErrorScreen` sounds preferable in my opinion.

Runemoro: @Boundarybreaker "severe" (as in, all your worlds are gone), not "server".

LemmaEOF: ahh, misread. Yeah that sounds great too.

mappings/net/minecraft/client/gui/widget/SliderButtonWidget.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS cxa net/minecraft/client/gui/widget/SliderButtonWidget
```
Gegy, line 1: Found this name quite confusing. What about something like `SliderControlWidget`?

mappings/net/minecraft/sortme/Hand.mapping
Gegy: The one enum represents `LEFT` or `RIGHT`, while the other represents `MAIN` or `OFFHAND`. It does not seem like a mistake to be using `LEFT`/`RIGHT` here.

LemmaEOF: `OptionMainHand` is about which hand is set to be the main hand. The current name is just fine.

Gegy: It doesn't represent how it's used, though. Like here it's used to render actions on either the left/right player arm.

LemmaEOF: hmm, that's a good point. `Hand` clashes though, which is an issue. I can think of maybe calling this one `ChiralHand`, but that's pretty obtuse.

mappings/net/minecraft/client/gui/widget/SliderButtonWidget.mapping
shedaniel: It is a "button" anyway.

mappings/net/minecraft/sortme/Hand.mapping
Prospector: How about DominantHand?

Prospector: Wait, never mind. Because it still has both dominant and non dominant hands.

Gegy: Has the same issue as the former name, imo. It describes its use as an option, but not in cases such as `ModelWithArms#setHandAngle`

Prospector: Is it used exclusively client side?

Prospector: Well even if it's not, it's used exclusively for rendering. How about VisibleHand or RenderingHand?

Gegy: I won't suggest any names because I know the MCP mapping, but probably best to consider the name just by the enum variants: current `Hand` is `MAIN`/`OFFHAND` and `OptionsMainHand` is `LEFT/RIGHT`

mappings/net/minecraft/client/gui/widget/SliderButtonWidget.mapping
Runemoro: This is named based on the unobfuscated `RealmsSliderButton`. We already have a `SliderWidget`, which is the superclass of `SliderButtonWidget`.

mappings/net/minecraft/sortme/Hand.mapping
Runemoro: Renamed it to `AbsoluteHand` (since it's not relative to whether the owner is left or right handed). It might not be the best name, but at least it's better than `OptionMainHand`.

mappings/net/minecraft/client/gui/IButton.mapping
Runemoro: After checking again, it's only implemented by buttons used on the realms screen and provides a method to get the `net.minecraft.realms` proxy class, so I named it `RealmsButton`.

Gegy (approved):

modmuss50 (approved):

PR #540, opened 2019-03-20
Dispenser Behaviors and Chunk Occlusion graph

Gegy:


YanisBft (approved):

mappings/akn.mapping
```diff
@@ -0,0 +1 @@
+CLASS akn
```
Pyrofab, line 1: This looks like a mistake

Pyrofab (approved):

mappings/akn.mapping
Gegy: Yeah, Enigma does this sometimes. Not sure why. There's a few other files like this, so shouldn't be a huge issue. Can remove if needed, though.

PR #542, opened 2019-03-20
Changed AbstractButtonWidget to use mojang names

ghost:
Changed AbstractButtonWidget to use mojang names


modmuss50:
What happens if you reset to obf, there isnt a lot of need for most of these x -> x mappings.

Only the x and y need to be mapped, along with the params

ghost:
So, resetting to obf works for all the mappings, but when x and y are mapped they still don't work in a development environment so I've removed them.

modmuss50:
yeah, x and y shouldnt be in the yarn repo, it would have been ok if they showed in dev. Im happy with this.

modmuss50 (approved):

PR #546, opened 2019-03-21
Add some villager related mappings and fix some really poor names

liach:
Poor names fixed: 
- In `Brain`: `clone` -> `copy` (`Object#clone`)
- In `FoxEntity` `MooshroomEntity`: `getType` -> `getFoxType` `getMooshroomType` (`Entity#getType`)

### DIFF
--- a/mappings/net/minecraft/village/TraderRecipe.mapping
+++ b/mappings/net/minecraft/village/TraderRecipe.mapping
@@ -6,6 +6,7 @@ CLASS bfj net/minecraft/village/TraderRecipe
 	FIELD e maxUses I
 	FIELD f rewardExp Z
 	FIELD g tax I
+	FIELD h demandBonus I
 	FIELD i priceMultiplier F
 	FIELD j rewardedExp I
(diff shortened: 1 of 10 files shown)


mappings/net/minecraft/village/TraderRecipe.mapping
```diff
@@ -5,6 +5,7 @@ CLASS bfj net/minecraft/village/TraderRecipe
 	FIELD d uses I
 	FIELD e maxUses I
 	FIELD f rewardExp Z
+	FIELD g tax I
```
Gegy, line 8: I don't understand how this field is used, but I don't think `tax` is correct here

liach: It holds a negative int as a discount, but since it is negative, I cannot call it a discount... Hence I called it a "negative" tax.

liach: On a side note, this is a flat-rate modification to first slot price after multiplier application; it is not saved to trade itself, but is calculated from a villager based on its rating with the trader and sent to the client with a packet.

PR #548, opened 2019-03-21
Shapes

Runemoro:


mappings/net/minecraft/util/shape/VoxelShape.mapping
```diff
@@ -1,7 +1,7 @@
 CLASS cri net/minecraft/util/shape/VoxelShape
-	FIELD a shape Lcqx;
+	FIELD a cellSet Lcqx;
```
Prospector: `cellSet`...why not just `cells`?

Runemoro: Renamed it to `voxels`.

Prospector: ah yes much nicer

Prospector (approved):

mappings/net/minecraft/util/shape/DoubleListPair.mapping
```diff
@@ -0,0 +1,8 @@
+CLASS cra net/minecraft/util/shape/DoubleListPair
+	CLASS net/minecraft/util/shape/DoubleListMerger$a SectionPairPredicate
```
Gegy: The outer class name is wrong here

mappings/net/minecraft/util/shape/DoubleListPair.mapping
```diff
@@ -0,0 +1,8 @@
...
+		METHOD merge test (III)Z
```
Gegy: Why not keep the Mojang name here?

mappings/net/minecraft/util/shape/DoubleListPair.mapping
Runemoro: Fixed

mappings/net/minecraft/util/shape/DoubleListPair.mapping
Runemoro: Didn't realize it was a Mojang name (using Knit so there's no highlighting). I unmapped it, though it seems weird to call the method of a predicate `merge`...

Gegy: I haven't read through the code much, but it does seem to be used specifically for merging voxel shapes.

Runemoro: It's passed as a parameter to the `forAllOverlappingSections`, which returns true if the predicate is true for all pairs of sections in both lists that overlap. It's used by both the shape comparison and shape merging code, but it still seems weird for a boolean-valued method with no side effects to be called `merge`.

Gegy: See `BitSetVoxelSet#combine`: the `SectionPairPredicate` passed to `forAllOverlappingSections` produces side effects through its implementation. Imo, `merge` does make some sense here.
Also, not necessarily a good name, but I imagine `forAllOverlappingSections` would be `merge` too

Prospector (approved):

modmuss50 (approved):

PR #551, opened 2019-03-22
Block and item stuff

YanisBft:
Almost all blockstate and shape fields are done
Renamed some block and item classes to match in-game names (IDs)


Runemoro:
Do the renamed classes match the block/item IDs too? That seems more important than matching the ingame names.

YanisBft:
Yep that's what I meant, block and item IDs, not display names.

Prospector (approved):

Gegy (approved):

PR #564, opened 2019-03-28
Some Screen and Options mappings (Primarily, ListWidget)

Prospector:


mappings/net/minecraft/client/gui/ScreenChild.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS cws net/minecraft/client/gui/InputListener
+CLASS cws net/minecraft/client/gui/ScreenChild
```
Gegy: It's referenced as `child` in a field but I doubt it's the Mojang name. I imagine this'd be something like `ScreenElement`/`ScreenComponent`

mappings/net/minecraft/client/gui/Screen.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS cyz net/minecraft/client/gui/Screen
+	FIELD minecraft client Lcus;
```
shedaniel: Isn't this an official name?

mappings/net/minecraft/client/gui/widget/ButtonWidget.mapping
```diff
@@ -1 +1,10 @@
 CLASS cvx net/minecraft/client/gui/widget/ButtonWidget
+	CLASS cvx$a PressAction
+	FIELD onPress pressAction Lcvx$a;
```
shedaniel: Isn't this an official name?

mappings/net/minecraft/client/gui/widget/ListWidget.mapping
```diff
@@ -1,11 +1,17 @@
-CLASS cwk net/minecraft/client/gui/widget/AbstractListWidget
+CLASS cwk net/minecraft/client/gui/widget/ListWidget
+	FIELD minecraft client Lcus;
+	FIELD x0 height I
+	FIELD x1 width I
+	FIELD y0 top I
+	FIELD y1 bottom I
+	FIELD yo scroll D
```
shedaniel: These should be official names, although this improves the names.
Not the case with `client` tho.

modmuss50: Yep, all these names need to go, including the client mapping.

mappings/net/minecraft/client/gui/widget/ButtonWidget.mapping
modmuss50: yes, needs to go.

modmuss50 (changes requested):
The mojang named methods and fields need to go. You can keep the method argument mappings for them however.

mappings/net/minecraft/client/gui/ScreenChild.mapping
Prospector: Screen components already exist...really it's just a clickable object. Any suggestions?

mappings/net/minecraft/client/gui/widget/ButtonWidget.mapping
Prospector: whoops yeah that field was an accident

mappings/net/minecraft/client/gui/widget/ListWidget.mapping
Prospector: I would disagree pretty strongly there. The names Mojang has given them aren't much better than if they had just obfuscated them.

mappings/net/minecraft/client/gui/ScreenChild.mapping
Gegy: Huh. The current ScreenComponent name seems wrong -- it is just a more specific implementation of `ScreenChild`.

Gegy: As it is, it doesn't hold much 'component'-like behaviour, despite implementing `ScreenChild`.

Prospector: oh. Hmm I have an idea for a more thorough PR but it will be controversial I think.

mappings/net/minecraft/client/gui/InteractionElement.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS cws net/minecraft/client/gui/ScreenChild
+CLASS cws net/minecraft/client/gui/InteractionElement
```
Gegy: `InteractableElement`?

Prospector: ok that works

Prospector: or it could just be an `Element` because there's no non-interactable elements?

Gegy: Sounds good 👍

modmuss50 (approved):
im happy with this.

Runemoro (approved):

PR #566, opened 2019-03-29
Container -> Menu, Inventory -> Container, Drop Widget Suffix, Reorganize Screens, Align more names with Mojang's

### LINKED ISSUE #351: net.minecraft.container.Container -> SharedGuiState?
Container implies inventory. It's pretty easy to mistake this class for an inventory-only thing, but:
* it carries non-inventory numeric information, such as furnace or machine progress
* it's even used to sync not-really-numeric information, such as beacon effects
* it holds the syncId that's used to manage custom-packet-delivered information
* it manages button clicks such as in EnchantingTableContainer
* Slots and buttons have x/y locations. This also makes Container function as a structural description of the gui.

So the name `Container` never really worked for me. Keeping all this in mind, SharedGuiState may not be the best name here. It's hard to smurf-name all the specific instances. To make it a better package name, perhaps net.minecraft.guistate.GuiState instead. But I think if we can come to a good consensus on the purpose of the object we can find a much better fit.

### LINKED ISSUE #386: Container -> Menu

### LINKED ISSUE #535: Drop "Widget" suffix
The `Widget` suffix (ex. `ButtonWidget`, `SliderWidget`, `TextFieldWIdget`) just makes the GUI code more verbose without adding clarity.

Also, the unobfuscated `net.minecraft.realms` classes don't use the `Widget` suffix.

Prospector:
I have a feeling this will be one of the most controversial PRs in yarn's history.

Closes #351, Closes #386, Closes #535


mappings/net/minecraft/block/InventoryProvider.mapping
```diff
@@ -1,2 +1,5 @@
 CLASS ahe net/minecraft/block/InventoryProvider
```
Gegy, line 1: Should this then be `ContainerProvider`?

mappings/net/minecraft/client/font/TextRenderer.mapping
```diff
@@ -75,7 +75,7 @@ CLASS cvk net/minecraft/client/font/TextRenderer
 	METHOD a setFonts (Ljava/util/List;)V
 		ARG 1 fonts
 	METHOD a setRightToLeft (Z)V
-	METHOD b getStringWidth (Ljava/lang/String;)I
+	METHOD b getWidth (Ljava/lang/String;)I
```
Gegy, line 78: `getStringWidth` or `getWidthOf` are more clear, imo. Any reason for changing this?

mappings/net/minecraft/client/gui/ingame/AbstractFurnaceScreen.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS czu net/minecraft/client/gui/container/AbstractFurnaceScreen
+CLASS czu net/minecraft/client/gui/ingame/AbstractFurnaceScreen
```
Gegy, line 1: Given these are all menus, should they be in a `menu` package?

mappings/net/minecraft/client/gui/menu/AnvilScreen.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS czv net/minecraft/client/gui/container/AnvilScreen
+CLASS czv net/minecraft/client/gui/menu/AnvilScreen
```
Gegy, line 1: Oh, it seems some are. Should probably make this consistent. :P

mappings/net/minecraft/client/gui/menu/BlastFurnaceScreen.mapping
```diff
@@ -0,0 +1,2 @@
+CLASS czx net/minecraft/client/gui/menu/BlastFurnaceScreen
+	FIELD l BG_TEX Lqr;
```
Gegy, line 2: Since this PR is touching all GUI files already, could be worth finding some consistency in how these textures are named. Is it known how Mojang names these?

mappings/net/minecraft/client/gui/widget/LockDifficultyButton.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS cwf net/minecraft/client/gui/widget/LockButtonWidget
+CLASS cwf net/minecraft/client/gui/widget/LockDifficultyButton
```
Gegy, line 1: It's used for just difficulty but the implementation seems to work more generally. Should the name be kept more general?

mappings/net/minecraft/client/gui/widget/RecipeBookButton.mapping
```diff
@@ -0,0 +1,45 @@
+CLASS cwd net/minecraft/client/gui/widget/RecipeBookButton
```
Gegy, line 1: This too -- is the implementation specific to the recipe book?

mappings/net/minecraft/container/Container.mapping
```diff
@@ -1,86 +1,26 @@
...
-	METHOD b unpackQuickCraftStage (I)I
-		ARG 0 clickData
-	METHOD b packClickData (II)I
-		ARG 0 buttonId
-		ARG 1 quickCraftStage
-	METHOD b calculateComparatorOutput (Lagn;)I
-		ARG 0 inventory
-	METHOD b close (Lavm;)V
-	METHOD b transferSlot (Lavm;I)Lbbp;
+	METHOD b_ onContainerCloser (Lavm;)V
```
Gegy, line 23: I imagine this should be `onContainerClosed` 😄

Gegy: Oh, though it should match with `onContainerOpen`

mappings/net/minecraft/container/DoubleContainer.mapping
```diff
@@ -0,0 +1,8 @@
+CLASS agm net/minecraft/container/DoubleContainer
```
Gegy, line 1: I'm guessing this is used for double-chests? The `Double` was quite confusing to me at first; maybe something like `JoinedContainer`?

mappings/net/minecraft/client/gui/menu/BlastFurnaceScreen.mapping
Prospector: The button texture is names something like BUTTON_LOCATION or something. I can double check

Player3324:
Menu has a completely different meaning in every UI toolkit I know, it describes only a list of selectable items.

Container is too vague/ambiguous, it isn't more descriptive than "something that may contain something else". java.util.List is a container, a bucket is a container, a network packet is a container, ...

Dropping Widget is potentially ok, but there's always the few odd ones like List that pollute the auto-complete namespace and have to be named carefully to not have another java.awt.List.

mappings/net/minecraft/client/font/TextRenderer.mapping
Prospector: yeah, some name in one of the deobf realms things. Doesn't need to change tho

modmuss50:
Closing this, some of it may have already been done. other things seem to be tracked by issues.

PR #567, opened 2019-03-29
Drop Text in TextComponent, TextFormat -> TextStyle, I18n -> Languages, translate/translation -> localize/localization

### LINKED ISSUE #174: Drop 'Text' prefix in text model
Currently the `net/minecraft/text` has all of the text model components prefixed with `TextComponent`, in contrast with Mojang's own naming scheme (`XComponent`).

Its important to note that only `TextComponent` directly represents actual text, the rest are just placeholders apart a model that is constructed into text client-side. IMO the `Text` prefix is verbose and unnecessary.

I propose that we should use Mojang's naming (where known) and the following changes:

- `net/minecraft/text/ITextComponent` -> `net/minecraft/text/Component` (this is kind of given away in some of the serialisation code).

### LINKED ISSUE #565: translate -> getLocalizedString
This seems to be what mojang calls it. Also translate doesn't make much sense because the string you pass in is unlocalized. Translation would happen between two languages, not between unlocalized and localized

image

Prospector:
AKA Controversial PR of 28-Mar-2019 number 2

Closes #174, Closes #565


mappings/net/minecraft/text/LocalizedComponent.mapping
```diff
@@ -1,9 +1,9 @@
-CLASS jw net/minecraft/text/TranslatableTextComponent
+CLASS jw net/minecraft/text/LocalizedComponent
```
kashike: Big :-1:, the real name for this is `TranslatableComponent` as seen by the `toString`.

mappings/net/minecraft/block/Block.mapping
```diff
@@ -61,7 +61,7 @@ CLASS bly net/minecraft/block/Block
 	FIELD e defaultState Lbuw;
 	FIELD f dynamicBounds Z
 	FIELD g dropTableId Lqr;
-	FIELD h translationKey Ljava/lang/String;
+	FIELD h localizationKey Ljava/lang/String;
```
kashike: :-1:, see comment regarding component class name

kashike (changes requested):

mappings/net/minecraft/text/LocalizedComponent.mapping
Prospector: Hm I see...

mappings/net/minecraft/container/ContainerType.mapping
```diff
@@ -3,24 +3,6 @@ CLASS ayr net/minecraft/container/ContainerType
 		METHOD create (ILavl;)Laxq;
 			ARG 1 syncId
 			ARG 2 playerInventory
-	FIELD c GENERIC_9X3 Layr;
```
Gegy, line 6: Are these inferred names?

mappings/net/minecraft/entity/EntityType.mapping
```diff
@@ -151,7 +151,7 @@ CLASS aid net/minecraft/entity/EntityType
 		ARG 1 world
 	METHOD d getEntityClass ()Laik;
 	METHOD e getTranslationKey ()Ljava/lang/String;
-	METHOD f getTextComponent ()Ljm;
+	METHOD f getComponent ()Ljm;
```
Gegy, line 154: Imo, getters like this should have more specific names. Something like `getName` could work

Pyrofab: I agree. `getComponent` is very confusing, especially if you start thinking about entity component systems.

liach:
As @sfPlayer1 suggested, I believe `Text` is better than `Component` as a name. https://github.com/FabricMC/yarn/issues/174#issuecomment-433780948

mappings/net/minecraft/container/ContainerType.mapping
Prospector: yup

ChloeDawn:
`I18n` should stay imo, it's a common standard and is much better shorthand than `Languages` too.

Prospector:
Languages looks prettier and is probably more in line with what mojang calls it imo (IMO I18n is probably the most hideous class name I've seen in a long time)

ChloeDawn:
I personally find `I18n.localize(key)` to be much clearer and cleaner than `Languages.localize(key)`

Prospector:
Could everyone who feels strongly about parts of this, please state what it is you feel strongly about and what you are okay with with this PR? 

Some bits of this may be scrapped, others may be reworked, others may go through. It's quite clear this PR will not be merged in its entirety so feel free to rip apart every bit of it if it so pleases you. The PR is mainly to get the ball rolling on many issues that have left undecided on.

mappings/net/minecraft/client/util/ComponentUtil.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS cvz net/minecraft/client/util/TextComponentUtil
+CLASS cvz net/minecraft/client/util/ComponentUtil
```
ChloeDawn, line 1: `Components`

ChloeDawn (changes requested):

Pyrofab:
I find `translate` -> `localize` to be a good change. 
`I18n` to `Languages` seems inadequate to me, as `Languages` implies a util class with no actual state (in my opinion at least). 

Regarding dropping `Text` in `TextComponent`, I would be for a middle ground: keep it in the base class and drop it everywhere else. That way the meaning stays clear and can be inferred for the subclasses. Or go with Player's suggestion of dropping `Component` instead. No matter the future name of `TextComponent`s, methods and fields of that type should become way more specific.

liach:
In my opinion, we should just call the class `Internationalization` instead of `I18n`

Player3324:
I stated already elsewhere that component is yet another meaningless term. If anything I'd drop component and leave Text, which is pretty much what it is. Alternatively something like message, which seems to be what Java calls these (see e.g. https://docs.oracle.com/javase/7/docs/api/java/text/MessageFormat.html )

mappings/net/minecraft/text/EntityNbtComponent.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS jr net/minecraft/text/NbtTextComponent
+CLASS jr net/minecraft/text/EntityNbtComponent
```
jamierocks, line 1: This class should simply be called `NbtComponent`, `EntityNbtComponent` is the inner class (`jr$b`).

jamierocks (commented):
I think the package should also be changed, Mojang do not call them 'text components' - rather 'chat components'. This is often seen in their update posts (such as the [latest 1.14 post](https://www.minecraft.net/en-us/article/village---pillage-out-java-)).

To be accurate to the actual mappings, I'm fairly confident that most of the chat classes (the exception being `TextFormat`) are in the `net.minecraft.network.chat` package - established by mapping order.

```
CL: iw net/minecraft/network/PacketListener
CL: ja net/minecraft/network/chat/BaseComponent
...
CL: jn net/minecraft/network/chat/TranslatableComponent
CL: jr net/minecraft/network/packet/Packet
```

`TextFormat` is a slightly different case, and should be in the root package (this is somewhat evident by its very early obf name of `c`) and called `ChatFormat` - again, established by mapping order.

```
CL: a net/minecraft/ChatFormat
CL: b net/minecraft/CrashReport
```

---

I'm well aware that Fabric doesn't always obey by Mojang's naming conventions, though being a name that is so publicly thrown around and for such a critical part of the game - I don't think we should add 'fuel to the fire' so-to-speak, lets not have another `TileEntity` :)

To be clear, the above statement was more about calling them chat components than text components - how Fabric chooses to name the packages is something that should be down to review, not dictated by Mojang as has been the standard in yarn for sometime now.

liach:
RIP

PR #569, opened 2019-03-29
19w13a mappings

Runemoro:


mappings/net/minecraft/util/math/Vec3i.mapping
```diff
@@ -3,9 +3,26 @@ CLASS fq net/minecraft/util/math/Vec3i
...
+		ARG 2 treatAsBlockPos
+	METHOD a isWithinDistance (Lfq;D)Z
+		ARG 1 vec
+		ARG 2 distance
 	METHOD d crossProduct (Lfq;)Lfq;
-	METHOD m squaredDistanceTo (Lfq;)D
+		ARG 1 vec
+	METHOD m getSquaredDistance (Lfq;)D
+		ARG 1 vec
+	METHOD n getTaxicabDistance (Lfq;)I
```
liach: Is taxicab or manhattan distance better known?

mappings/net/minecraft/util/math/Vec3i.mapping
```diff
@@ -3,9 +3,26 @@ CLASS fq net/minecraft/util/math/Vec3i
 	FIELD b y I
 	FIELD c z I
 	FIELD e ZERO Lfq;
-	METHOD a squaredDistanceToCenter (DDDZ)D
+	METHOD a getDistanceSq (DDDZ)D
```
liach: `getSquaredDistance`

Runemoro: Fixed

liach:
Oh, you need to resolve merge conflicts as well.

mappings/net/minecraft/entity/ai/brain/WalkTarget.mapping
```diff
@@ -1,9 +1,9 @@
 CLASS aot net/minecraft/entity/ai/brain/WalkTarget
 	FIELD a lookTarget Lakg;
-	FIELD b speedFactor F
+	FIELD b speed F
```
Gegy, line 3: Why this change?

mappings/net/minecraft/util/math/Vec3i.mapping
Gegy: I've only really heard Manhattan be used, imo that is more clear

mappings/net/minecraft/entity/ai/brain/WalkTarget.mapping
Runemoro: It's eventually passed to `startMovingAlong(@Nullable final Path path, final double speed)` in `EntityNavigation`, which calls it "speed", not "speedFactor".

mappings/net/minecraft/util/math/Vec3i.mapping
Runemoro: Changed it

mappings/net/minecraft/block/SmallTranslucentBlock.mapping
```diff
@@ -0,0 +1 @@
+CLASS blb net/minecraft/block/SmallTranslucentBlock
```
Virtuoel, line 1: I don't really get the meaning of the "Small" part. The methods in this class were previously part of the `GlassBlock` and `StainedGlassBlock` classes, so it would probably be more suitable as some kind of `AbstractGlassBlock`.

mappings/net/minecraft/block/BedrockBlock.mapping
```diff
@@ -0,0 +1 @@
+CLASS blu net/minecraft/block/BedrockBlock
```
Virtuoel, line 1: Despite only being used for Bedrock, it's perfectly usable as a generic spawnproof block and has nothing specific to Bedrock other than the spawnproof part. It would look rather odd if someone made use of a class called `BedrockBlock` for some breakable spawnproof block in their mod.

mappings/net/minecraft/block/SmallTranslucentBlock.mapping
Runemoro: I may have really misread this, and I can't check right now, but isn't it extended by G**r**assBlock?

Virtuoel: Nope, definitely Glass

mappings/net/minecraft/block/BedrockBlock.mapping
Runemoro: Based on the obfuscated name (in alphabetical order relative to other block classes) the Mojang name starts with a `B`, so it's probably `BedrockBlock`.

It would be a really bad idea to use that in a mod since mojang may add a method at any point.

It also doesn't really makes sense to have a "spawnproof block" class, since unlike the other block classes, "spawnproof" isn't a kind of block (glass, log, etc.).

Virtuoel: Hmm, good points. `BedrockBlock` does seem fine in that case.

mappings/net/minecraft/block/SmallTranslucentBlock.mapping
Runemoro: Hmm, then I'm blind. I'll change it when I get home in a few hours.

modmuss50:
Merging due to update, if you have any more comments please open a new issue or a new PR.

PR #570, opened 2019-03-29
A few util mappings

Runemoro:


mappings/net/minecraft/util/math/BoundingBox.mapping
```diff
@@ -26,10 +26,19 @@ CLASS crb net/minecraft/util/math/BoundingBox
 		ARG 7 maxX
 		ARG 9 maxY
 		ARG 11 maxZ
+	METHOD a fromMutableIntBoundingBox (Lchc;)Lcrb;
```
Gegy: Seems verbose, this could potentially just be `from`?

mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -119,3 +127,4 @@ CLASS zq net/minecraft/util/math/MathHelper
 	METHOD i fastInverseSqrt (D)D
 		ARG 0 value
 	METHOD j ease (D)D
+	METHOD k sgn (D)I
```
Gegy, line 130: sign?

Runemoro: I think `sgn` is a [more common name for the sign function](https://en.m.wikipedia.org/wiki/Sign_function), the same way we have `sin` and `cos` rather than `sine` and `cosine`. But that's just my opinion, I don't mind changing it if you disagree.

Gegy: Ah, thought it was a typo. Didn't realise `sgn` is commonly used

Pyrofab: `sgn` seems needlessly obfuscated to me, but if it is actually common it may not be too big of a deal. I would note however that the JDK equivalent is called `signum`.

mappings/net/minecraft/util/NumberRange.mapping
```diff
@@ -26,21 +26,31 @@ CLASS bi net/minecraft/util/NumberRange
...
 		METHOD a (Lcom/mojang/brigadier/StringReader;Ljava/lang/Integer;Ljava/lang/Integer;)Lbi$d;
 			ARG 1 min
 			ARG 2 max
+		METHOD a fromStringReader (Lcom/mojang/brigadier/StringReader;Ljava/util/function/Function;)Lbi$d;
 		METHOD a squared (Ljava/lang/Integer;)Ljava/lang/Long;
 			ARG 0 value
 		METHOD b atLeast (I)Lbi$d;
 			ARG 0 value
 		METHOD d test (I)Z
+	CLASS net/minecraft/util/NumberRange$a Factory
```
Gegy: Knit bug

modmuss50:
Merging due to update, if you have any more comments please open a new issue or a new PR.

mappings/net/minecraft/util/math/MathHelper.mapping
iTitus: `sgn` is the common mathematical short name for that function. `signum` or `sign` is the long name.

https://en.wikipedia.org/wiki/Sign_function

PR #574, opened 2019-03-30
Some player stuff

sargunv:


mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -112,6 +114,7 @@ CLASS aii net/minecraft/entity/LivingEntity
 	METHOD cx isUndead ()Z
 	METHOD cy getHealth ()F
 	METHOD cz getRecentDamageSource ()Lahi;
+	METHOD d dropAllDrops (Lahi;)V
```
liach: Is this dropping with respect to damage source? Needs a better name.

sargunv: IMO the name makes sense in the context of the other methods around it. It's near `dropInventory`, `dropEquipment`, and `dropLoot`, and it calls all three of them, so `dropAllDrops`.

liach: Then you should just name it `drop`. You need to be concise.

sargunv: sure

liach (approved):

Gegy (approved):

PR #575, opened 2019-03-30
Random mappings around projectiles etc

liach:


mappings/net/minecraft/util/SystemUtil.mapping
```diff
@@ -17,13 +17,18 @@ CLASS p net/minecraft/util/SystemUtil
 	FIELD d LOGGER Lorg/apache/logging/log4j/Logger;
 	METHOD a toMap ()Ljava/util/stream/Collector;
 	METHOD a getValueAsString (Lbvz;Ljava/lang/Object;)Ljava/lang/String;
+	METHOD a next (Ljava/lang/Iterable;Ljava/lang/Object;)Ljava/lang/Object;
```
Gegy, line 20: Given this takes an `Iterable`, I imagine it could only ever query the first element?

liach: No. This queries the element after the argument 2, or the first element if the arg 2 is null.
Same goes for `previous` (`b`) below.

Gegy (approved):

modmuss50 (approved):

PR #576, opened 2019-03-30
Chunk loader mappings + fixes

Runemoro:


mappings/net/minecraft/server/network/ServerPlayerEntity.mapping
```diff
@@ -13,7 +13,7 @@ CLASS vf net/minecraft/server/network/ServerPlayerEntity
 	FIELD cp seenCredits Z
 	FIELD cq recipeBook Lym;
 	FIELD cu enteredNetherPos Lcrg;
-	FIELD cv chunkPos Lfo;
+	FIELD cv lastCameraChunkSectionPosition Lfo;
```
Gegy: What does the last mean here? Wondering if it would potentially make more sense to elide that

Runemoro: It's the position of the last camera update. It's checked and set by `ThreadedAnvilChunkStorage.updateCameraPosition` to check if the player changed chunks aince last tick.

Gegy: Ah ok, that makes sense. 👍

Gegy (approved):

modmuss50 (approved):

PR #578, opened 2019-03-30
Add more GL and GUI mappings

thecodewarrior:
Just started poking around with fabric and I'm loving how easy it is to map things. GUIs are one of my specialties so I thought I would do some mapping for them. The non-GUI mappings are just things that are used in the GUIs that I decided to map, along with a couple GL mappings I did independently from the GUIs.

### DIFF
--- a/mappings/net/minecraft/client/gui/WrittenBookScreen.mapping
+++ b/mappings/net/minecraft/client/gui/WrittenBookScreen.mapping
@@ -21,8 +21,11 @@ CLASS czz net/minecraft/client/gui/WrittenBookScreen
 	FIELD f cachedPageIndex I
 	FIELD g lastPageButton Ldaw;
 	FIELD h nextPageButton Ldaw;
+	FIELD i playPageTurnSound Z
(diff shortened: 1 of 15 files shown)


mappings/net/minecraft/client/gui/WrittenBookScreen.mapping
```diff
@@ -21,8 +21,11 @@ CLASS czz net/minecraft/client/gui/WrittenBookScreen
 	FIELD f cachedPageIndex I
 	FIELD g lastPageButton Ldaw;
 	FIELD h nextPageButton Ldaw;
+	FIELD i playPageTurnSound Z
```
Gegy: What about just `pageTurnSound`?

thecodewarrior: Sure. That one and the couple in `ContainerScreen` were somewhat difficult to name.

Gegy (approved):

modmuss50 (approved):

PR #581, opened 2019-04-01
Make AI task lifecycle easier to understand

sargunv:
The lifecycle names as-is are hard to understand, especially `run`, `keepRunning`, `stopInternal`. I renamed the methods so anyone implementing a task will have a much more clear idea of what methods to override and what order they come in.

```
canStart
onStart
canContinue
onContinue
onStop
```

### DIFF
--- a/mappings/net/minecraft/entity/ai/brain/task/Task.mapping
+++ b/mappings/net/minecraft/entity/ai/brain/task/Task.mapping
@@ -4,6 +4,8 @@ CLASS ajj net/minecraft/entity/ai/brain/task/Task
 	FIELD b endTime J
 	FIELD c minRunTime I
 	FIELD d maxRunTime I
+	METHOD <init> (I)V
+		ARG 1 exactRunTime
 	METHOD <init> (II)V
 		ARG 1 minRunTime
 		ARG 2 maxRunTime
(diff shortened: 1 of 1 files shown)


Runemoro:
Many tasks don't implement what you call `onContinue` at all, and the whole task is executed in what you call `onStart`. Start is confusing for those since the method does the whole task rather than just starts the task.

What about:
 - `shouldRun`/`run`
 - `shouldKeepRunning`/`keepRunning`
 - `finishRunning`

sargunv:
Amended commit with the suggested names.

mappings/net/minecraft/entity/ai/brain/task/Task.mapping
```diff
@@ -33,7 +36,8 @@ CLASS ajj net/minecraft/entity/ai/brain/task/Task
 		ARG 1 world
 		ARG 2 entity
 		ARG 3 time
-	METHOD f stopInternal (Lve;Laii;J)V
+	METHOD f finishRunning (Lve;Laii;J)V
+		ARG 3 time
```
Pyrofab, line 40: `finishExecution` feels a bit more natural, but maybe that's just me.

Pyrofab (approved):

mappings/net/minecraft/entity/ai/brain/task/Task.mapping
sargunv: In that case, how's `shouldExecute`, `execute`, `shouldContinueExecution`, `continueExecution`, and `finishExecution`? I'd like to keep the lifecycle names consistent with each other.

Pyrofab: I would personally be in favour of that change, however `Goal`s use `run`, not execute. As such it may be better to stay consistent and keep `*running`.

sargunv: I'll leave it as "running" then.

modmuss50:
Updating, if anyone has any further feedback on this please open an issue or a new PR.

PR #583, opened 2019-04-03
A few random mappings

YanisBft:


mappings/net/minecraft/container/BeaconContainer.mapping
```diff
@@ -8,6 +8,10 @@ CLASS axu net/minecraft/container/BeaconContainer
 		ARG 1 syncId
 	METHOD <init> (ILagn;Laya;Layb;)V
 		ARG 1 syncId
+	METHOD c setEffects (II)V
+		ARG 1 primaryEffectId
+		ARG 2 secondaryEffectId
+	METHOD e getProperty ()I
```
Gegy: `getProperty`? What is the property?

liach: Given this is an integer, it looks like a flags field getter to me

Gegy: I imagine should at least be `getProperties` then?

YanisBft: I named it `getProperty` because of the `setProperty` method in the same class. Should I rename both so they are plural?

Gegy: Huh, I might be missing it but I don't see a `setProperty` there. Is it in a superclass?

YanisBft: yeah actually it's a method in `Container` class. Should still be plural though

Gegy (approved):

PR #586, opened 2019-04-04
Some minor advancement mappings

liach:
`(de)serialize` -> `toXxx`/`fromXxx`
`AdvancementPosition` -> `AdvancementPositioner` (calculates the position instead of holding the position)

### DIFF
--- a/mappings/net/minecraft/advancement/AdvancementPositioner.mapping
+++ b/mappings/net/minecraft/advancement/AdvancementPositioner.mapping
@@ -1,4 +1,4 @@
-CLASS ac net/minecraft/advancement/AdvancementPosition
+CLASS ac net/minecraft/advancement/AdvancementPositioner
 	FIELD a advancement Lq;
 	FIELD b parent Lac;
 	FIELD c previousSibling Lac;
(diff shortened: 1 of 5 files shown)


mappings/net/minecraft/advancement/SimpleAdvancement.mapping
```diff
@@ -1,5 +1,5 @@
 CLASS q net/minecraft/advancement/SimpleAdvancement
-	CLASS q$a Builder
+	CLASS q$a Task
```
liach, line 2: renamed this as there is `SimpleAdvancement#toTask` method mapped.

Gegy: I don't know anything about this class, so I have no inclination either way: does it make more sense to update the method to match the class, or to match the class with the method?

mappings/net/minecraft/advancement/AdvancementPositioner.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS ac net/minecraft/advancement/AdvancementPosition
+CLASS ac net/minecraft/advancement/AdvancementPositioner
```
Gegy, line 1: Maybe something along the lines of `AdvancementPositionResolver`?

liach: `Positioner` is something that positions.

mappings/net/minecraft/advancement/SimpleAdvancement.mapping
liach: The `toString` says this is "Task Advancement", and this thing, compared to the simple advancement object, is a pojo and can be more easily serialized.

Gegy (approved):

PR #587, opened 2019-04-04
A few mappings around villagers

liach:

### DIFF
--- a/mappings/net/minecraft/client/render/entity/model/VillagerResemblingModel.mapping
+++ b/mappings/net/minecraft/client/render/entity/model/VillagerResemblingModel.mapping
@@ -1,0 +1,3 @@
+CLASS dfz net/minecraft/client/render/entity/model/VillagerResemblingModel
+	FIELD a head Ldgl;
+	FIELD l nose Ldgl;
--- a/mappings/net/minecraft/entity/passive/VillagerEntity.mapping
+++ b/mappings/net/minecraft/entity/passive/VillagerEntity.mapping
@@ -15,6 +15,7 @@ CLASS ava net/minecraft/entity/passive/VillagerEntity
 	FIELD bF levellingUp Z
 	FIELD bG lastCustomer Lavn;
 	FIELD bH buddyGolemId Ljava/util/UUID;
+	FIELD bI oneMinAfterLastGolemCheckTimestamp J
 	FIELD bL foodLevel B
 	FIELD bM gossip Laoo;
 	FIELD bN gossipStartTime J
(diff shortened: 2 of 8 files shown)


mappings/net/minecraft/client/render/entity/model/VillagerResemblingModel.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS dfz net/minecraft/client/render/entity/model/VillagerResemblingModel
```
Gegy, line 1: This name is fine, just wondering if `VillagerLikeModel` might be simpler

liach: I think "resembling" is more accurate, though "like" is a correct term as well (it is used too often and slightly confusing)

mappings/net/minecraft/entity/passive/VillagerEntity.mapping
```diff
@@ -15,6 +15,7 @@ CLASS ava net/minecraft/entity/passive/VillagerEntity
 	FIELD bF levellingUp Z
 	FIELD bG lastCustomer Lavn;
 	FIELD bH buddyGolemId Ljava/util/UUID;
+	FIELD bI oneMinAfterLastGolemCheckTimestamp J
```
Gegy, line 18: I guess this is `nextGolemCheckTimestamp`?

mappings/net/minecraft/entity/passive/VillagerEntity.mapping
```diff
@@ -23,13 +24,15 @@ CLASS ava net/minecraft/entity/passive/VillagerEntity
...
 	METHOD a isLackingBuddyGolem (J)Z
+		ARG 1 time
 	METHOD a wantsGolem (Lahz;)Z
 	METHOD a initBrain (Laix;)V
 	METHOD a releaseTicketFor (Laor;)V
 	METHOD a talkWithVillager (Lava;J)V
 		ARG 1 villager
 		ARG 2 time
 	METHOD a setVillagerData (Lavb;)V
+		ARG 1 arg1
```
Gegy: arg1?

mappings/net/minecraft/entity/passive/VillagerEntity.mapping
liach: no. This is a trash field left by proguard; it is only set but never read. every time it is set, it is set to one minute after last golem check time if the field's value is less than that.

Gegy: It's state influences the return here, I don't think it's a trash field. here

Gegy: Hm, I guess this code is wrong. Might be a bug, the condition would be true every tick

modmuss50:
Merging due to update, open a new issue or PR with further comments.

PR #82, opened 2016-09-16
Rename Minecraft => MinecraftGame

copygirl:
This renames the class `Minecraft` to `MinecraftGame` along with fields and parameters from `mc` to `game`.

Also:
- `WorldClient.getMinecraft` => `getGameInstance`
- `MinecraftGame.mcResourcePack` => `gameResourcePack`

Issue #66, opened 2016-09-09
Questions on some names


liach:
Some of the names of classes might be modified. Ex. `TileEntity` -> `BlockEntity`, `Minecraft` -> `MinecraftClient`. Those classes have a more accurate name but they were not renamed because in other decompiling projects people were too lazy to change those common-used names.

copygirl:
Not quite sure what this issue is for specifically. I think we specifically try to avoid using other mapping's names, but in some cases the name stuck, causing the community to adopt it. I'd love to have a place to point out possible name changes of important classes and discuss them.

Regarding the naming of blocks / tile entities, there's also the alternative of renaming `Block` -> `Tile`, which seems to be Mojang's convention of calling blocks. (Though I myself would prefer moving to `BlockEntity`.)

And regarding `MinecraftClient`, I'm not sure if that's appropriate. Perhaps something along the lines of `Game` or `App` might me more fitting?

asiekierka:
According to MinecraftWiki at least, the TileEntities are now internally indeed called BlockEntities, which might make the TileEntity->BlockEntity rename a good idea. http://minecraft.gamepedia.com/Block_entity

MinecraftServer is already a thing, but MinecraftClient also handles the integrated server... MinecraftGame?

darkevilmac:
I think MinecraftClient makes more sense as it matches with the naming of
the MinecraftServer class.

liach:
@darkevilmac Yeah you are the one who suggested that name change in mcp but was rejected by the bad tempered guy.

darkevilmac:
Yup, almost forgot I had suggested that until now.

asiekierka:
@darkevilmac That would only make sense if Client and Server were opposites, but Client has its own Server...

darkevilmac:
@asiekierka I don't think it's unheard of to have a game client with an integrated server. If it was running the server and the client all by itself I'd understand but as it is now there's still some separation there.

As a side note, from the issue I had initially made on MCP the official Mojang name is known to be Minecraft. Considering they're such important classes it might make sense to leave them as is.

liach:
Ok. We still may actually use `BlockEntity` though.

copygirl:
The following renaming has been done:
`TileEntity` => `BlockEntity` (#76)
`Minecraft` => `MinecraftGame` (#82)

Prospector:
Ok well this has been done for 2 years. Close?

liach:
[...]

PR #596, opened 2019-04-06
client -> minecraft

shedaniel:
It is known that mojang uses `minecraft` other than `client` in classes like `Screen`

### DIFF
--- a/mappings/net/minecraft/client/Keyboard.mapping
+++ b/mappings/net/minecraft/client/Keyboard.mapping
@@ -1,11 +1,13 @@
 CLASS cuw net/minecraft/client/Keyboard
-	FIELD a client Lcux;
+	FIELD a minecraft Lcux;
 	FIELD b repeatEvents Z
 	FIELD c clipboard Lctd;
 	FIELD d debugCrashStartTime J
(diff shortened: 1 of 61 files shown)


liach:
Let's call it `client`. Newcomers may assume it is some sort of overall game instance that contains core logic like registries instead of the client when they see `minecraft`.

shedaniel:
Newcomers will get confused when they mess with Screens anyways. If the newcomer only makes content mods, it is pretty rare that they will run into this mapping change as `client` mainly exists in the client side.

Gegy:
I don't think "people will get confused anyway" is a great justification for a mappings change. Just because we can doesn't mean we should, imo.

shedaniel:
That's true, but sometimes inconsistency can be confusing too. If I want to map this better, I would map minecraft -> client.

liach:
Looking back into #66, it seems `Minecraft` as a noun is too vague about any component of the game, so we changed the `Minecraft` reference to `MinecraftGame` in #82 (Then to `MinecraftClient` in #220).

shedaniel:
I will close this as this is not really a good idea

PR #598, opened 2019-04-06
Some mappings for loot tables

NeunEinser:
This adds new mappings to loot table entries, functions and conditions and makes some of the existing ones more consistent, where inconsitencies appeared (not many).
It adds method parameters for most methods.
Enigma seems to be unable to decompile `net.minecraft.world.condition.BlockStatePropertyLootFunction.Factory.fromJson(JsonObject, JsonDeserializationContext)`, which is why I was unable to provide mappings for potential lamda declarations in that method.


mappings/net/minecraft/command/arguments/NbtPathArgumentType.mapping
```diff
@@ -3,3 +3,4 @@ CLASS cr net/minecraft/command/arguments/NbtPathArgumentType
 	FIELD b NBT_PATH_NOT_FOUND_EXCEPTION Lcom/mojang/brigadier/exceptions/DynamicCommandExceptionType;
 	FIELD c EXAMPLES Ljava/util/Collection;
 	METHOD a create ()Lcr;
+	METHOD parse parse (Lcom/mojang/brigadier/StringReader;)Ljava/lang/Object;
```
Gegy: This shouldn't need to be remapped

mappings/net/minecraft/world/loot/LootManager.mapping
```diff
@@ -8,4 +8,7 @@ CLASS cog net/minecraft/world/loot/LootManager
 	METHOD a getSupplierNames ()Ljava/util/Set;
 	METHOD a toJson (Lcoe;)Lcom/google/gson/JsonElement;
 	METHOD a check (Lcof;Lqs;Lcoe;Ljava/util/function/Function;)V
+		ARG 1 arg2
```
Gegy: arg2?

mappings/net/minecraft/world/loot/LootManager.mapping
```diff
@@ -8,4 +8,7 @@ CLASS cog net/minecraft/world/loot/LootManager
...
+		ARG 2 supplier
+		ARG 3 getSupplier
```
Gegy: `getSupplier` as an argument name seems odd. what is it for?

mappings/net/minecraft/world/loot/LootTableReporter.mapping
```diff
@@ -2,7 +2,12 @@ CLASS cof net/minecraft/world/loot/LootTableReporter
 	FIELD a messages Lcom/google/common/collect/Multimap;
 	FIELD b nameFactory Ljava/util/function/Supplier;
 	FIELD c name Ljava/lang/String;
+	METHOD <init> (Lcom/google/common/collect/Multimap;Ljava/util/function/Supplier;)V
+		ARG 1 rmessages
```
Gegy: `reportedMessages`?

mappings/net/minecraft/world/loot/LootManager.mapping
NeunEinser: It's a functional interface of type `Function<Identifier, LootSupplier>` which in the code is hooked to `this.suppliers::get`, a get method of a map.
So a function that takes an identifier and returns the corresponding lootSupplier.
Maybe `getSupplierFunction` would be more clear?

mappings/net/minecraft/world/loot/LootTableReporter.mapping
NeunEinser: No, sometimes when I press `r` to rename, Enigma puts a `r` in the rename field while doing so. That's an instance of when I missed that happening.

mappings/net/minecraft/world/loot/condition/BlockStatePropertyLootCondition.mapping
```diff
@@ -1,5 +1,38 @@
...
+		METHOD a (Lcom/google/gson/JsonObject;Lbwc;Ljava/lang/Object;)V
+			ARG 1 property
+			ARG 2 value
 	FIELD a block Lbmb;
 	FIELD b properties Ljava/util/Map;
 	FIELD c predicate Ljava/util/function/Predicate;
+	METHOD <init> (Lbmb;Ljava/util/Map;)V
+		ARG 1 block
+		ARG 2 properties
+	METHOD a create (Lbmb;)Lcqn$a;
```
NeunEinser: There are many instances of this create method, both in loot conditions and loot functions. I have used `create` because some of them already used that mapping.
However, most of these methods return a `.Builder` of the condition/function. So I think something like `createBuilder` or `getBuilder` would be more appropiate here.

mappings/net/minecraft/world/loot/LootManager.mapping
asiekierka: getterSupplier?

NeunEinser: `supplierGetter`?

mappings/net/minecraft/world/loot/function/SetTagLootFunction.mapping
```diff
@@ -1,3 +1,10 @@
 CLASS cpv net/minecraft/world/loot/function/SetTagLootFunction
```
NeunEinser: All other functions are based on the name of the function, like it is called in the vanilla game (e.g. `set_contents` becomes `SetContentsLootFunction`)
So why does `set_nbt` become `SetTagLootFunction`? Is it okay to change that?

liach: Feel free to change that.

mappings/net/minecraft/world/loot/condition/BlockStatePropertyLootCondition.mapping
liach: Just `builder`

mappings/net/minecraft/world/loot/LootManager.mapping
```diff
@@ -7,5 +7,19 @@ CLASS cog net/minecraft/world/loot/LootManager
 	FIELD f supplierNames Ljava/util/Set;
 	METHOD a getSupplierNames ()Ljava/util/Set;
 	METHOD a toJson (Lcoe;)Lcom/google/gson/JsonElement;
+		ARG 0 supplier
+	METHOD a (Lcof;Lqs;Lcoe;)V
```
liach, line 11: Hmm, you should probably map method's names when you map the parameters

NeunEinser: It's a lambda, so you can't name it, but you can map the parameters.

modmuss50 (approved):
Looks fine to me, I will leave it open for more discussion for a bit, but will end up merging before we update.

PR #599, opened 2019-04-07
Mapping for potions and trades factories

urielsalis:


mappings/net/minecraft/village/Trades.mapping
```diff
@@ -4,26 +4,49 @@ CLASS avf net/minecraft/village/Trades
 		FIELD b price I
 		FIELD c maxUses I
 		FIELD d experience I
+		METHOD <init> (Lbbm;I)V
+			ARG 2 price
+		METHOD <init> (Lbbm;III)V
+			ARG 2 price
+			ARG 3 rmaxUses
```
liach: watch the `r` next time

Prospector: the age old enigmatic R

Prospector: Shouldn't maxUses be like stock or stockSize or stockedAmount or something? Does it tell how many the villager has left to trade or how many total they can trade if the player has traded 0 so far?

liach: No. Vanilla just calls it `maxUses`, which is the same as old maxUses in villager trades. The refill logic only impacts price in some sort of non-serialized matter.

Prospector (approved):

raphydaphy (approved):

liach (approved):

PR #604, opened 2019-04-10
Map some poi and village stuff

liach:


mappings/net/minecraft/container/MerchantContainer.mapping
```diff
@@ -1,22 +1,31 @@
 CLASS ayy net/minecraft/container/MerchantContainer
-	FIELD c villager Lbgd;
-	FIELD d villagerInventory Layx;
+	FIELD c trader Lbgd;
+	FIELD d traderInventory Layx;
```
Runemoro, line 3: Why do we have `MerchantContainer` but `traderInventory`? ~~I think we should either use "merchant" or "trader". Do we know which one is the Mojang name?~~ Mojang calls them "traders", I think `MerchantContainer` should be renamed to `TraderContainer`, or even better, `TraderInventory`.

mappings/net/minecraft/container/MerchantContainer.mapping
```diff
@@ -1,22 +1,31 @@
...
 		ARG 1 syncId
+		ARG 2 playerInventory
+		ARG 3 trader
 	METHOD a setRecipes (Lbgf;)V
 	METHOD a setCanLevel (Z)V
 		ARG 1 canLevel
 	METHOD b equals (Lbbu;Lbbu;)Z
 		ARG 1 itemStack
 		ARG 2 otherItemStack
+	METHOD c pullMatchingStackFromPlayerInventory (ILbbu;)V
```
Runemoro: Why not just `takeFromPlayer`?

mappings/net/minecraft/container/TradeOutputSlot.mapping
```diff
@@ -0,0 +1,12 @@
...
+	FIELD b player Lavr;
+	FIELD g amount I
+	FIELD h trader Lbgd;
+	METHOD <init> (Lavr;Lbgd;Layx;III)V
+		ARG 1 player
+		ARG 2 trader
+		ARG 3 traderInventory
+		ARG 4 index
+		ARG 5 xPosition
+		ARG 6 yPosition
```
Runemoro: I think just `x` and `y` would be enough.

mappings/net/minecraft/entity/ai/brain/sensor/SecondaryJobSiteSensor.mapping
```diff
@@ -0,0 +1 @@
+CLASS apm net/minecraft/entity/ai/brain/sensor/SecondaryJobSiteSensor
```
Runemoro: Why was this renamed? Look at the ID in `SensorType`:
```java
SECONDARY_POIS = SensorType.register("secondary_pois", SecondaryPointsOfInterestSensor::new);
```

mappings/net/minecraft/village/PointOfInterestStorage.mapping
```diff
@@ -9,6 +9,9 @@ CLASS aqa net/minecraft/village/PointOfInterestStorage
 		FIELD d predicate Ljava/util/function/Predicate;
 		METHOD a getPredicate ()Ljava/util/function/Predicate;
 	FIELD a pointOfInterestDistanceTracker Laqa$a;
+	METHOD a initForPalette (Lbgn;Lbxd;)V
+	METHOD a mayPointOfInterestAppearIn (Lbxd;)Z
+	METHOD a createPointsOfInterest (Lbxd;Lfo;Ljava/util/function/BiConsumer;)V
```
Runemoro: I think these shouldn't have `PointOfInterest` in the names to match the other methods (`add`, `remove`, `getDistanceFromNearestOccupied`, `count`). 

Also, what does this method do differently from `add`?

liach: this one creates poises if they aren't present; then it adds poi for each matching state in the chunk

mappings/net/minecraft/container/MerchantContainer.mapping
liach: mojabg calls the gui merchant container.
btw, where does mojang call trader?

Runemoro: Weird, the ID for the trader is `wandering_trader`, but the container is `merchant` rather than `trader`...

Prospector: villagers are merchants too. The UI is a merchant UI and the entity is a trader. It makes sense to me because merchants don't necessarily have to be traders, even if that's all vanilla ones are at the moment.

mappings/net/minecraft/container/MerchantContainer.mapping
liach: It is autofill, like what recipe books do.

mappings/net/minecraft/container/MerchantContainer.mapping
liach: so `merchant` for all recipes, ui, etc.?

mappings/net/minecraft/container/MerchantContainer.mapping
Runemoro: But isn't there a shorter name? At least `Inventory` is unnecessary.

mappings/net/minecraft/entity/ai/brain/sensor/SecondaryJobSiteSensor.mapping
liach: i saw it was using secondary job site from memory module types

Runemoro (approved):

modmuss50 (approved):

PR #605, opened 2019-04-10
CheckboxWidget and other gui related mappings

NeunEinser:
A check box widget was added for the "erase cache data" option inside the optimize world screen.

### DIFF
--- a/mappings/net/minecraft/client/gui/widget/LevelListWidget.mapping
+++ b/mappings/net/minecraft/client/gui/widget/LevelListWidget.mapping
@@ -11,6 +11,11 @@ CLASS dcr net/minecraft/client/gui/widget/LevelListWidget
 			ARG 2 levelList
 			ARG 3 level
 			ARG 4 levelStorage
+		METHOD a play ()V
+		METHOD b delete ()V
+		METHOD c edit ()V
+		METHOD d recreate ()V
+		METHOD e start ()V
 		METHOD f getIconTexture ()Ldtn;
 	FIELD a LOGGER Lorg/apache/logging/log4j/Logger;
 	FIELD b DATE_FORMAT Ljava/text/DateFormat;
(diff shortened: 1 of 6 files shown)


Runemoro:
Shouldn't it be "Checkbox" rather than "CheckBox"?
 - https://en.wikipedia.org/wiki/Checkbox (uses checkbox throughout the article)
 - `textures/gui/checkbox.png` (not `check_box`)
 - https://www.google.com/search?q=%22checkbox%22 (34 million results) vs https://www.google.com/search?q=%22check+box%22 (16 million results)

NeunEinser:
Hmm, I am used to the capital B, programming languages I use always do that, e.g.

https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/CheckBox.html
https://docs.oracle.com/javase/7/docs/api/javax/swing/JCheckBox.html
https://docs.microsoft.com/de-de/dotnet/api/system.windows.forms.checkbox

But I don't really care tbh.

mappings/net/minecraft/client/gui/widget/LevelListWidget.mapping
```diff
@@ -11,6 +11,11 @@ CLASS dcr net/minecraft/client/gui/widget/LevelListWidget
 			ARG 2 levelList
 			ARG 3 level
 			ARG 4 levelStorage
+		METHOD a play ()V
```
asiekierka, line 14: Hmm - what do these methods do?

asiekierka (approved):

mappings/net/minecraft/client/gui/widget/LevelListWidget.mapping
Runemoro: I think they're called when the user clicks play, for example.

NeunEinser: Correct. It's what gets executed, when you press `play` on the level select screen.

modmuss50 (approved):

PR #607, opened 2019-04-11
Map all commands and argument types

Runemoro:


mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -107,12 +109,14 @@ CLASS avr net/minecraft/entity/player/PlayerEntity
...
 	METHOD dJ isSpawnForced ()Z
 	METHOD dK getEnchantmentTableSeed ()I
+	METHOD dL getNextLevelExperience ()I
 	METHOD dM getHungerManager ()Laxr;
 	METHOD dN canFoodHeal ()Z
 	METHOD dO canModifyWorld ()Z
 	METHOD dP getEnderChestInventory ()Laza;
 	METHOD dQ dropShoulderEntities ()V
 	METHOD dR getScoreboard ()Lcso;
+	METHOD dS getNameAndUuid ()Ljm;
```
liach, line 119: This is really a display name for an entity in the command output that allows you to copy the player name, right?

Runemoro: No, this is a "PlayerName (123e4567-e89b-12d3-a456-426655440000)" string

liach: Ah, nice

liach (approved):

mappings/net/minecraft/command/arguments/MobEffectArgumentType.mapping
```diff
@@ -2,7 +2,7 @@ CLASS cq net/minecraft/command/arguments/MobEffectArgumentType
 	FIELD a INVALID_EFFECT_EXCEPTION Lcom/mojang/brigadier/exceptions/DynamicCommandExceptionType;
 	FIELD b EXAMPLES Ljava/util/Collection;
 	METHOD a create ()Lcq;
-	METHOD a getEffectArgument (Lcom/mojang/brigadier/context/CommandContext;Ljava/lang/String;)Lahv;
+	METHOD a getMobEffect (Lcom/mojang/brigadier/context/CommandContext;Ljava/lang/String;)Lahv;
```
asiekierka, line 5: Don't we call these StatusEffects, or am I missing something?

mappings/net/minecraft/world/border/WorldBorder.mapping
```diff
@@ -29,7 +29,7 @@ CLASS bwo net/minecraft/world/border/WorldBorder
 		METHOD n recalculateBounds ()V
 	FIELD a listeners Ljava/util/List;
 	FIELD b damagePerBlock D
-	FIELD c safeZone D
+	FIELD c buffer D
```
asiekierka, line 32: Maybe we should adopt bufferSize instead.

mappings/net/minecraft/command/arguments/MobEffectArgumentType.mapping
Runemoro: Yes, I'll rename this. It was named based on the strings in that class and the class name.

NeunEinser:
- DataCommand: 2 method names beyond `executeInsert` are missing (`addModifyArguments` and `executeModify`), also 2 inner interfaces (`a` is used for modifiying, perhaps simply `Modify`, didn't look into `b`, but looks like a `ModifyBuilder`)
- ExecuteCommand: field `d` and an inner interface that looks identical to `Condition`, if you can figure these out
- LootCommand: inner interface `a`: used for adding the item, to wherever it's supposed to go. `Process` seems like a good name to me.
- RecepieCommand: Change previously mapped methods `give` and `take` to `executeGive` and `...Take`
- SpreadPlayersCommand: inner class `Pile`: `float b` method: it's an absolute value, `void a()` normalizes the vector coordinates `boolean a` method calculates and returns the height value.

Didn't look at argument types yet. I was in process of doing the same thing, which is why I had mappings for some of the missing ones. Unfortunately, I can throw most of them away now.

PR #608, opened 2019-04-11
Map a couple of random methods

SquidDev:
I realise this is incredibly disjoint - sorry. This is just me going through methods that I use and trying to clean them up a little.

### DIFF
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -154,7 +154,8 @@ CLASS bme net/minecraft/block/Block
 		ARG 3 state
 	METHOD a getTickRate (Lbhd;)I
 		ARG 1 world
-	METHOD a (Lbme;)Z
+	METHOD a isSolidSmallSquare (Lbhd;Lev;Lfa;)Z
+	METHOD a canConnect (Lbme;)Z
 		ARG 0 block
 	METHOD a getLuminance (Lbvc;)I
 		ARG 1 state
@@ -358,6 +359,7 @@ CLASS bme net/minecraft/block/Block
 		ARG 0 state
 		ARG 1 builder
 	METHOD c getRenderLayer ()Lbgm;
+	METHOD c isSolidMediumSquare (Lbgl;Lev;)Z
 	METHOD c onRainTick (Lbha;Lev;)V
 		ARG 1 world
 		ARG 2 pos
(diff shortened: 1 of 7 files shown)


mappings/net/minecraft/block/Block.mapping
```diff
@@ -389,6 +391,7 @@ CLASS bme net/minecraft/block/Block
 		ARG 1 state
 		ARG 2 view
 		ARG 3 pos
+	METHOD d isSolidFullSquare (Lbvc;Lbgl;Lev;Lfa;)Z
```
liach, line 394: What is the distinction between small, medium, and full?

SquidDev: Small is used for torches, lanterns and pressure plates - it checks against a `VoxelShape` of size 4x4. Medium is used for rails, repeaters/comparators and pressure plates - 12x12. Full is well, a full block - so 1x1.

I'm aware they're not fantastic names, but I couldn't really find a name which covered all the cases they're used.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -305,6 +305,7 @@ CLASS aid net/minecraft/entity/Entity
 	METHOD bm getMaxBreath ()I
 	METHOD bn getBreath ()I
 	METHOD bo getHeadYaw ()F
+	METHOD bp canAttack ()Z
```
liach: `isPlayerAttackable` `canPlayerAttack`

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -425,6 +426,7 @@ CLASS aid net/minecraft/entity/Entity
 	METHOD r isTeammate (Laid;)Z
 	METHOD s isPartOf (Laid;)Z
 	METHOD t isSpectator ()Z
+	METHOD t onAttacked (Laid;)Z
```
liach: Looks like `handlePlayerAttack(Entity)` to me as this is only used by players and can skip the rest of the player attack logic

mappings/net/minecraft/block/Block.mapping
liach: Consider naming them based on their usages then.

SquidDev: `isSolidForTorchOrLantern` doesn't really roll off the tongue. IMO it's better to describing what it does, and people can look at the implementation if they need to.

CreepyCre: How about something like `isHitboxPole`?

modmuss50:
Updating, any further comments should go in an issue or PR

PR #610, opened 2019-04-12
Some 1.14-pre2 and options mappings

YanisBft:
Renamed things in CloudRenderMode to be consistent with AoOption

PR #611, opened 2019-04-12
Menu mappings

NeunEinser:
~~Has overlapping on the MouseSettingsScreen with #610~~, this is more focused on menu screens, while #610 is more focused on options.


mappings/net/minecraft/client/options/GameOption.mapping
```diff
@@ -3,6 +3,7 @@ CLASS cve net/minecraft/client/options/GameOption
 	FIELD B CHAT_COLOR Lcup;
 	FIELD C CHAT_LINKS Lcup;
 	FIELD D CHAT_LINKS_PROMPT Lcup;
+	FIELD E DISCRETE_MOUSE_SCROLL Lcup;
```
YanisBft: Overlap #610, and the ingame name is "discrete scrolling"

NeunEinser: Ah, I missed that

mappings/net/minecraft/client/options/ExponentialGameOption.mapping
```diff
@@ -0,0 +1 @@
+CLASS cva net/minecraft/client/options/ExponentialGameOption
```
asiekierka: ???

NeunEinser: It's used by the new scroll sensitivity slider in the mouse settings. While all other sliders in option menus are linear, this slider uses an exponential/logarithmic scale. Roughly the first three quarters are 0 to 1, then it goes up to 10.

No idea how else to call this.

Runemoro: I think it's called a "[logarithmic scale](https://en.wikipedia.org/wiki/Logarithmic_scale)", I've never heard of "exponential scale". So I think `LogarithmicGameOption` would be better.

NeunEinser: It also uses Math.log(), so I changed it now.

PR #613, opened 2019-04-13
Command Mappings, mostly arguments and some missing methods/inner classes

NeunEinser:


mappings/net/minecraft/server/command/ExecuteCommand.mapping
```diff
@@ -1,25 +1,51 @@
...
 	METHOD a transformSourceForCondition (Lcom/mojang/brigadier/context/CommandContext;ZZ)Ljava/util/Collection;
+		ARG 0 context
 		ARG 1 positive
-		ARG 2 result
+		ARG 2 requestResultValue
 	METHOD a executeCondition (Lcom/mojang/brigadier/tree/CommandNode;Lcom/mojang/brigadier/builder/ArgumentBuilder;ZLsa$b;)Lcom/mojang/brigadier/builder/ArgumentBuilder;
-	METHOD a addBlockConditionExecutes (Lcom/mojang/brigadier/tree/CommandNode;Lcom/mojang/brigadier/builder/ArgumentBuilder;ZZ)Lcom/mojang/brigadier/builder/ArgumentBuilder;
+		ARG 0 node
+		ARG 1 builder
+	METHOD a addBlockConditionArguments (Lcom/mojang/brigadier/tree/CommandNode;Lcom/mojang/brigadier/builder/ArgumentBuilder;ZZ)Lcom/mojang/brigadier/builder/ArgumentBuilder;
```
Runemoro: Why rename it? The method adds `executes`'s to the command, not arguments. It adds the execution actions for when there is no command argumebt and just a condition argument, to tell the player whether it matches

NeunEinser: Hmm, `executes` isn't a proper noun which is why the ethod name sounded very weird to me. Gonna change it to `Execute` tho, since it is only one that gets added. Also its used for `blocks`, not `block`. `block` and `scores` uses a method called `executeCondition` for the same thing.

Runemoro: Two are being added. Also, you can name them "execute actions" or "subcommands" if you don't like "executes".

NeunEinser: No, it adds one of a set of two, depending on `positive` and forks in the beginning
```java
private static ArgumentBuilder<ServerCommandSource, ?> addBlocksConditionExecute(CommandNode<ServerCommandSource> node, ArgumentBuilder<ServerCommandSource, ?> builder, boolean positive, boolean masked) {
    return builder.fork(node, commandContext -> transformSourceForCondition(commandContext, positive, testBlockCondition(commandContext, masked).isPresent()))
    .executes(positive ?
        (commandContext -> executePositiveBlockCondition(commandContext, masked)) :
        (commandContext -> executeNegativeBlockCondition(commandContext, masked))
    );
}
```

NeunEinser: The fork is by the way, as far as I can tell, for the condition test, if there is a chained sub command. Forking a command makes it run for all sources. `transformSourceForCondition` actually returns an empty collection, if the condition fails, or a collection with the source that was there before, if it succeeds.

So the command gets forked with either an empty source or the source it already has, making sure the command will not continue running for any source at all, if the condition fails.

So the method actually takes care of both things. It stops the command if the condition fails, and it adds the execute for when there is no chained sub command.

Maybe `addBlocksConditionLogic` would be more fitting.

mappings/net/minecraft/server/command/SpreadPlayersCommand.mapping
```diff
@@ -2,16 +2,27 @@ CLASS tk net/minecraft/server/command/SpreadPlayersCommand
 	CLASS tk$a Pile
 		FIELD a x D
 		FIELD b z D
+		METHOD a normalize ()V
+		METHOD a clamp (DDDD)Z
+			ARG 1 minX
+			ARG 3 minZ
+			ARG 5 maxX
+			ARG 7 maxZ
+		METHOD a getYCoordinate (Lbgl;)I
```
Runemoro: Why not just `getY`? That's how all other getX/Y/Z methods are named.

mappings/net/minecraft/server/command/SpreadPlayersCommand.mapping
```diff
@@ -2,16 +2,27 @@ CLASS tk net/minecraft/server/command/SpreadPlayersCommand
 	CLASS tk$a Pile
 		FIELD a x D
 		FIELD b z D
+		METHOD a normalize ()V
```
Runemoro, line 5: What's weird about this method and `absolute` is that it uses the distance from (x, z) = (0, 0) rather than the area center, yet pile coordinates are absolute, not relative. This seems like a Mojang bug...

NeunEinser: Yeah, I wasn't really sure what's the purpose of these methods either.
When talking about vector arithmetic, `absolute` calculates the absolute value of the vector, tho, and this method normalizes it/ makes it a length of one, which is why I named them these.
I agree though, using the center of the spreading as origin would make more sense here.

mappings/net/minecraft/server/command/DataCommand.mapping
```diff
@@ -1,4 +1,12 @@
 CLASS ub net/minecraft/server/command/DataCommand
+	CLASS ub$a Modify
```
Runemoro: I don't think class names should be verbs... Maybe it should be `SomethingModifier` or `ModifyCallback`?

NeunEinser: Hmm, it is a functional interface that is initialized with a function for either `append`, `prepend`, `set` or `merge` and called by `executeModify` to execute the respective function.
`ModifyOperation` sounds good to me.

mappings/net/minecraft/server/command/AdvancementCommand.mapping
```diff
@@ -24,5 +24,10 @@ CLASS rl net/minecraft/server/command/AdvancementCommand
 		ARG 3 advancement
 		ARG 4 criterion
 	METHOD a register (Lcom/mojang/brigadier/CommandDispatcher;)V
-	METHOD a searchAndAdd (Lq;Ljava/util/List;)V
+		ARG 0 dispatcher
+	METHOD a addChildren (Lq;Ljava/util/List;)V
```
Gegy: The difference between the two arguments here feels unclear

mappings/net/minecraft/server/command/BossBarCommand.mapping
```diff
@@ -11,18 +11,55 @@ CLASS rp net/minecraft/server/command/BossBarCommand
...
 	FIELD k SET_VISIBILITY_UNCHANGED_VISIBLE_EXCEPTION Lcom/mojang/brigadier/exceptions/SimpleCommandExceptionType;
 	METHOD a listBossBars (Lcd;)I
+		ARG 0 source
 	METHOD a addBossBar (Lcd;Lqs;Ljm;)I
+		ARG 0 source
+		ARG 1 name
+		ARG 2 displayName
 	METHOD a getValue (Lcd;Lri;)I
+		ARG 0 source
+		ARG 1 boosBar
```
Gegy: Small typo here -- should be `bossBar` 😄

mappings/net/minecraft/server/command/RecipeCommand.mapping
```diff
@@ -1,6 +1,13 @@
 CLASS sw net/minecraft/server/command/RecipeCommand
 	FIELD a GIVE_FAILED_EXCEPTION Lcom/mojang/brigadier/exceptions/SimpleCommandExceptionType;
 	FIELD b TAKE_FAILED_EXCEPTION Lcom/mojang/brigadier/exceptions/SimpleCommandExceptionType;
-	METHOD a give (Lcd;Ljava/util/Collection;Ljava/util/Collection;)I
+	METHOD a executeGive (Lcd;Ljava/util/Collection;Ljava/util/Collection;)I
+		ARG 0 source
+		ARG 1 targets
+		ARG 2 receoies
```
Gegy: Typo here, not too sure what intended name is :P

mappings/net/minecraft/server/command/RecipeCommand.mapping
```diff
@@ -1,6 +1,13 @@
...
 	METHOD a register (Lcom/mojang/brigadier/CommandDispatcher;)V
-	METHOD b take (Lcd;Ljava/util/Collection;Ljava/util/Collection;)I
+		ARG 0 dispatcher
+	METHOD b executeTake (Lcd;Ljava/util/Collection;Ljava/util/Collection;)I
+		ARG 0 rsource
```
Gegy: Typo here too

mappings/net/minecraft/server/command/RecipeCommand.mapping
```diff
@@ -1,6 +1,13 @@
...
+		ARG 1 targets
+		ARG 2 recepies
```
Gegy: Oh, `recipies`?

mappings/net/minecraft/server/command/ServerCommandSource.mapping
```diff
@@ -5,14 +5,37 @@ CLASS cd net/minecraft/server/command/ServerCommandSource
...
+		ARG 7 nameTextComponent
+		ARG 8 server
+		ARG 9 entity
+	METHOD <init> (Lcc;Lcro;Lcrn;Lvf;ILjava/lang/String;Ljm;Lnet/minecraft/server/MinecraftServer;Laid;ZLcom/mojang/brigadier/ResultConsumer;Lck$a;)V
+		ARG 1 output
+		ARG 2 pos
+		ARG 3 rot
+		ARG 4 world
+		ARG 5 level
+		ARG 6 rsimpleName
```
Gegy: is the `r` prefix here intended?

NeunEinser: Enigma just keeps writing r in the name field when I press r to rename.

mappings/net/minecraft/server/command/RecipeCommand.mapping
NeunEinser: I managed to confuse you to the point where you don't know the spelling either anymore :D `recipes`

PR #624, opened 2019-04-15
Update method calls for player's tiny remapper update

liach:
Also updated gradlew wrapper


build.gradle
```diff
@@ -439,12 +439,12 @@ void mapJar(File output, File input, File mappings, File libraries, String from,
 	try {
 		def outputConsumer = new OutputConsumerPath(output.toPath())
 		outputConsumer.addNonClassFiles(input.toPath())
-		remapper.read(input.toPath())
+		remapper.readInputs(input.toPath())
 
 		libraries.eachFileRecurse(FileType.FILES) {file ->
-			remapper.read(file.toPath())
+			remapper.readInputs(file.toPath())
```
asiekierka: Libraries should use readClassPath, *not* readInputs - readInputs is only for remapping inputs.

asiekierka (changes requested):

build.gradle
liach: ah, that's why cfr decompiled all those fastutil things on my built jar! didn't realize that, thanks!

PR #627, opened 2019-04-16
Mojangson -> Snbt

intsuc:
SNBT is an official name of stringified NBT.


liach:
Or just `Nbt` as we don't have these named binary tags in their binary format.

asiekierka:
Source on that?

asiekierka:
Oh, right! The source code.

Issue #625, opened 2019-04-15
TextFormat has two fields named "color"


AlexIIL:
https://github.com/FabricMC/yarn/blob/8a7db997176d45ed5a72866b0a3e963443c2cfe6/mappings/net/minecraft/text/TextFormat.mapping#L5-L7

It looks like the enum field `RESET` was added, but incorrectly named as `color`. Would it make sense to make yarn detect this sort of thing automatically? (Either ensure that enum fields have their correct names and/or detecting duplicated field names - especially as this was changed in a large pr - #607).

AlexIIL:
(#626 will fix this)

AlexIIL:
(Fixed by https://github.com/FabricMC/yarn/commit/dfe0afdd2575414cf8bf849b5c81a842a05f8d4a)

PR #628, opened 2019-04-16
Remove duplicate field "TextFormat.color"

AlexIIL:
As it clashes with `TextFormat.color (Integer)`, when it should be called `TextFormat.RESET`. (#625).

(PR made manually on github - however the mapping format should be simple enough to not cause issues?)


AlexIIL:
Sorry, I should have seen #626 first :/

PR #629, opened 2019-04-16
Mapped some tasks and new classes

liach:


liach:
Closed in favor of #631

PR #646, opened 2019-04-23
Map the new 1.14 classes

Runemoro:


asiekierka:
Could also handle BlockPattern$c while at it... that's the *other* new class

modmuss50 (approved):
im happy with this, can map the other class if you want.

Runemoro:
> Could also handle BlockPattern$c while at it... that's the other new class

Mapped it too.

modmuss50 (approved):
Changes look good to me

PR #647, opened 2019-04-23
StateFactory$Builder with -> append

ChloeDawn:
The term `append` makes more sense for a mutable Builder type


liach:
How about add which is more concise

ChloeDawn:
Hmm, that works too. I was thinking more about objects such as `StringBuilder`, but I guess `add` is appropriate too as it is the term Guava uses for appending to immutable collection builders, which is what this builder effectively is

modmuss50:
add isnt a bad idea.

ChloeDawn:
Updated

liach (approved):

NeunEinser (approved):

Gegy (approved):

PR #650, opened 2019-04-24
Refactor chat components

### LINKED ISSUE #174: Drop 'Text' prefix in text model

jamierocks:
Closes #174, Succeeds #567 (partially).

This PR brings the naming of chat components in line with Mojang - in both
package and *chat* components.

### DIFF
--- a/mappings/net/minecraft/ChatFormat.mapping
+++ b/mappings/net/minecraft/ChatFormat.mapping
@@ -1,4 +1,4 @@
-CLASS c net/minecraft/text/TextFormat
+CLASS c net/minecraft/ChatFormat
 	FIELD A modifier Z
 	FIELD B code Ljava/lang/String;
 	FIELD C id I
--- a/mappings/net/minecraft/network/chat/BaseComponent.mapping
+++ b/mappings/net/minecraft/network/chat/BaseComponent.mapping
@@ -1,4 +1,4 @@
-CLASS jj net/minecraft/text/AbstractTextComponent
+CLASS jj net/minecraft/network/chat/BaseComponent
 	FIELD a siblings Ljava/util/List;
 	FIELD b style Lju;
 	METHOD equals (Ljava/lang/Object;)Z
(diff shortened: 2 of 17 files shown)


liach:
1. if we do move this to network package, what else should we move? just moving these feels too werid
2. i suggest literal component over string component

jamierocks:
> if we do move this to network package, what else should we move? just moving these feels too werid

There is plenty of packages in Yarn that don't match Mojangs - I imagine that it would be best for individual prs though. I intend to write a few more prs in the near future to work on this.

> i suggest literal component over string component

"String component" isn't a thing in this pr, that would be "text component" (herein lies how bad the previous naming was).

liach:
i actually prefer that naming style from sponge api (XxxText) as this is also promoted by player

jamierocks:
> i actually prefer that naming style from sponge api (XxxText) as this is also promoted by player

`Text` is an awful suffix, as not only does it directly conflict with Mojang's naming, there is ambiguity when talking about text components (or as you say literal components). If any suffix was to be used, it should be `Chat`.

liach:
So we will call mojang's text component literal text. straightforward and clear.
i also disagree with the network package assignment. text is mostly used to keep formatting through packets, but many mods now also use them on the client as they are more flexible and convenient than direct raw string manipulations.

modmuss50 (approved):
im not 100% sold on the network package, but the classnames seem good.

If anyone has any better ideas suggest away!

Runemoro (approved):

liach:
The only concern I have after this merge is that we should replace other non-text reference to `Component`. Other things are fine.

PR #652, opened 2019-04-26
ItemProvider -> ItemConvertible, StringRepresentable -> SnakeCaseIdentifiable

liach:
`ItemConvertible` reason:
1. This `ItemProvider` is implemented by `Item` and `Block` instead of some lambda that actually provides/supplies an item.
image
2. Another name `HasItemForm`, which is very accurate, is also considered, but because of the ideal of interface naming in adjectives, I chose `ItemConvertible`, as [`convert` as a word means "change to a different form"](https://www.dictionary.com/browse/convert).

`EnumPropertyValue` reason:
1. This is only implemented by enum classes that can be used as state property values.
image
2. This is really just used to get the enum value names in a lowercase form (or getting the proper string version of the value)
image
hence "get value name".
3. A few other usages are present, but they are only really used to be a shortcut for getting lowercase enum names for translation keys, etc.
image

### DIFF
--- a/mappings/net/minecraft/item/ItemConvertible.mapping
+++ b/mappings/net/minecraft/item/ItemConvertible.mapping
@@ -1,0 +1,2 @@
+CLASS bhg net/minecraft/item/ItemConvertible
+	METHOD g asItem ()Lbbv;
--- a/mappings/net/minecraft/item/ItemProvider.mapping
+++ b/mappings/net/minecraft/item/ItemProvider.mapping
@@ -1,2 +1,0 @@
-CLASS bhg net/minecraft/item/ItemProvider
-	METHOD g getItem ()Lbbv;
--- a/mappings/net/minecraft/state/property/EnumPropertyValue.mapping
+++ b/mappings/net/minecraft/state/property/EnumPropertyValue.mapping
@@ -1,2 +1,0 @@
-CLASS zx net/minecraft/state/property/EnumPropertyValue
-	METHOD m getValueName ()Ljava/lang/String;
--- a/mappings/net/minecraft/state/property/EnumPropertyValue.mapping
+++ b/mappings/net/minecraft/state/property/EnumPropertyValue.mapping
@@ -1,0 +1,2 @@
+CLASS zx net/minecraft/state/property/EnumPropertyValue
+	METHOD m getValueName ()Ljava/lang/String;
--- a/mappings/net/minecraft/util/SnakeCaseIdentifiable.mapping
+++ b/mappings/net/minecraft/util/SnakeCaseIdentifiable.mapping
@@ -1,0 +1,2 @@
+CLASS zx net/minecraft/util/SnakeCaseIdentifiable
+	METHOD m toSnakeCase ()Ljava/lang/String;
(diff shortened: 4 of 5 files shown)


NeunEinser (approved):

SquidDev:
I'm a little unsure about both of these:

 - IMO, `ItemProvider` is perfectly fine. It's never really used in a situation where you care about the "conversion" operation - you just want to get hold of an `Item` instance. If we were to rename it to anything, it'd just be `ItemSupplier`.

 - `StringRepresentable` is a kinda poor name, but `EnumPropertyValue` feels really ugly. Maybe something like `NamedEnum`. Eh, it doesn't feel much better.

liach:
1. this item getting interface is only implemented by block and item classes, which aren't providers or suppliers but just have an item form/state.
2. I've considered `PropertyEnum` as well, which means it's an enum used for property (values). How's that? (Might look like `EnumProperty` too much?)

ChloeDawn:
I feel the String supplying interface should be more generalized. Mojang uses it exclusively for determining the names of property values, so the most generalized form of that would be something like `Named` with `getName` as the member name.

liach:
Hmm, is there any other string supplying interfaces in vanilla? This one I see are all implemented by enums and used by enum property to retrieve a sanitized name.

mappings/net/minecraft/item/ItemConvertible.mapping
```diff
@@ -0,0 +1,2 @@
+CLASS bhg net/minecraft/item/ItemConvertible
```
Gegy, line 1: Like the method, the interface could potentially be named `AsItem` too

liach:
hmm, does this match the naming convention

mappings/net/minecraft/item/ItemConvertible.mapping
liach: hmm, does this match the naming convention? Any other class named similarly in yarn already?

ChloeDawn: I don't think `AsItem` is a very descriptive name. I think `ItemConvertible` is fine.

mappings/net/minecraft/item/ItemConvertible.mapping
```diff
@@ -0,0 +1,2 @@
...
+	METHOD g asItem ()Lbbv;
```
ChloeDawn, line 2: `getAsItem` - Following naming convention of Java interfaces that supply an explicitly typed value, such as `BooleanSupplier#getAsBoolean`, and a `get` implies side effects (for example `Block` on the first call looks up the `Item` in the registry and caches it), it is not inherently representing a constant, more a lazy object.

ChloeDawn (changes requested):

mappings/net/minecraft/item/ItemConvertible.mapping
liach: Or `toItem`?

Runemoro:
`ItemConvertible` and `StringConvertible` would be better.

ChloeDawn:
`SerializableEnum` might work for the String interface. It is explicitly used for determining a unique name for an enum constant during (de)serialization of state properties.

Runemoro:
> `SerializableEnum`

Every enum is serializable (`name` and `byName`).

liach:
Updated. Renamed `StringRepresentable` to `SnakeCaseIdentifiable` instead, which is better than `StringIdentifiable` (the only method in the interface only return snake case strings)

modmuss50:
Updating, any further discussion needs to go in a new PR or issue.

PR #657, opened 2019-04-28
Entity mappings

### LINKED ISSUE #655: PlayerEntity.experience and PlayerEntity.experienceLevel have backwards names
from the `readCustomDataFromTag`:
```java
this.experience = compoundTag_1.getInt("XpLevel");
this.experienceLevel = compoundTag_1.getInt("XpTotal");
```

"experience" should refer to the total experience points, and "experienceLevel" should refer to the number that's drawn above the XP bar

NeunEinser:
Many mappings around entities.
Resolves #655


mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -9,16 +9,19 @@ CLASS avx net/minecraft/entity/player/PlayerEntity
...
+	FIELD bF experiencePickUpDelay I
 	FIELD b SIZES Ljava/util/Map;
 	FIELD bM isInWater Z
 	FIELD bN abilities Lavu;
-	FIELD bO experience I
-	FIELD bP experienceLevel I
-	FIELD bQ experienceLevelProgress F
+	FIELD bO xpLevel I
+	FIELD bP xpTotal I
+	FIELD bQ xpProgress F
```
NeunEinser: (#655) Blayyke pointed out, it might be better to use entirely different names, so mods that used the previous names won't compile with swapped logic.
Also the corresponding nbt fields are `XpLevel`, `XpTotal` and `XpP`.

mappings/net/minecraft/entity/projectile/ExplosiveProjectileEntity.mapping
```diff
@@ -6,10 +6,21 @@ CLASS awc net/minecraft/entity/projectile/ExplosiveProjectileEntity
...
+		ARG 10 velocityY
+		ARG 12 velocityZ
+		ARG 14 world
 	METHOD <init> (Laij;Laio;DDDLbhh;)V
-		ARG 1 world
+		ARG 1 type
 		ARG 2 owner
+		ARG 3 velocityX
+		ARG 5 velocityY
+		ARG 7 velocityZ
```
NeunEinser: They are normalized in the constructor, so `velocity` might be not quite right here. However, a random spray value gets added to them before that, so a higher velocity will increase the accuracy, however the actual speed is not affected.

mappings/net/minecraft/entity/projectile/ExplosiveProjectileEntity.mapping
```diff
@@ -6,10 +6,21 @@ CLASS awc net/minecraft/entity/projectile/ExplosiveProjectileEntity
...
 	FIELD g ticks I
 	METHOD <init> (Laij;DDDDDDLbhh;)V
-		ARG 1 world
+		ARG 1 type
+		ARG 2 x
+		ARG 4 y
+		ARG 6 z
+		ARG 8 velocityX
+		ARG 10 velocityY
+		ARG 12 velocityZ
```
NeunEinser: They are just normalized, no random value gets added.

mappings/net/minecraft/entity/passive/RabbitEntity.mapping
```diff
@@ -1,20 +1,38 @@
...
+		METHOD <init> (Larh;)V
+			ARG 1 owner
+	CLASS arh$f EscapeDangerGoal
+		FIELD f rabbit Larh;
+		METHOD <init> (Larh;D)V
+			ARG 1 owner
+			ARG 2 speed
+	CLASS arh$g EatCarrotCropGoal
 		FIELD g rabbit Larh;
+		FIELD h wantsCarrots Z
```
NeunEinser, line 33: This gets assigned like this:
```java
this.wantsCarrots = this.rabbit.wantsCarrots();
this.wantsCarrots = true;
```
:thonkjang: #RabbitWantsCarrots

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -80,6 +81,7 @@ CLASS aif net/minecraft/entity/Entity
 	METHOD U getScoreboardTags ()Ljava/util/Set;
 	METHOD V kill ()V
 	METHOD W getDataTracker ()Lqk;
+	METHOD X putOnTopmostSolidBlock ()V
```
Runemoro: But this method is overriden in subclasses and does more than just move the entity to the top.

mappings/net/minecraft/entity/passive/DolphinEntity.mapping
```diff
@@ -1,13 +1,20 @@
...
+			ARG 1 owner
 	CLASS aqw$c SwimWithPlayerGoal
 		FIELD a dolphin Laqw;
 		FIELD b speed D
 		FIELD c closestPlayer Lavx;
 		METHOD <init> (Laqw;D)V
 			ARG 1 dolphin
 			ARG 2 speed
+	CLASS aqw$d PlayWithItemsGoal
+		METHOD a spitoutItem (Lbca;)V
```
Runemoro: `spitOutItem`

mappings/net/minecraft/entity/player/PlayerEntity.mapping
Runemoro: All the other experience-related names are `experience` though. They would have to all be changed for consistency.

mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -9,16 +9,19 @@ CLASS avx net/minecraft/entity/player/PlayerEntity
...
 	FIELD bR enchantmentTableSeed I
 	FIELD bT fishHook Latj;
+	FIELD bU lastLeveledUp I
```
Runemoro: What does it do? Should it be `lastLevelUpTime`/`lastLevelUpTicks`?

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -179,19 +212,31 @@ CLASS aif net/minecraft/entity/Entity
...
+		ARG 1 fluidTag
 	METHOD a isInFluid (Lzb;Z)Z
+		ARG 1 fluidTag
 		ARG 2 requireLoadedChunk
 	METHOD a toListTag ([D)Lii;
+		ARG 1 values
 	METHOD a toListTag ([F)Lii;
+		ARG 1 values
 	METHOD aA spawnSprintingParticles ()V
+	METHOD aB touchLava ()V
```
asiekierka: touchLava or onLavaTouch or?

mappings/net/minecraft/entity/mob/DrownedEntity.mapping
```diff
@@ -1,8 +1,42 @@
...
+			ARG 4 minY
+	CLASS atx$f TridentAttackGoal
+		FIELD a drowned Latx;
 	FIELD b waterNavigation Lapj;
-	FIELD bz targettingUnderwater Z
+	FIELD bz targetingUnderwater Z
 	FIELD c landNavigation Lapg;
 	METHOD eg isTargettingUnderwater ()Z
-	METHOD r setTargettingUnderwater (Z)V
+	METHOD r setTargetingUnderwater (Z)V
```
asiekierka, line 41: You missed one above.

mappings/net/minecraft/entity/passive/SnowgolemEntity.mapping
```diff
@@ -1,4 +1,5 @@
-CLASS arl net/minecraft/entity/passive/SnowmanEntity
+CLASS arl net/minecraft/entity/passive/SnowgolemEntity
```
asiekierka: SnowGolemEntity, also SNOW_GOLEM_FLAGS, if you want to go down this path

mappings/net/minecraft/entity/player/PlayerEntity.mapping
asiekierka: Yes.

asiekierka (changes requested):

mappings/net/minecraft/entity/passive/SnowgolemEntity.mapping
Runemoro: Isn't the entity ID snowman though?

NeunEinser: It's `minecraft:snow_golem`, not sure if it used to be snowman. Constants in `SoundEvent` are already named `SNOW_GOLEM_...`, too.

mappings/net/minecraft/entity/Entity.mapping
NeunEinser: It doesn't handle anything that happens when touching lava, `tick()` is doing that.
```java
public void touchLava() {
    this.touchingLava = true;
}
```

Runemoro: `setTouchingLava()`. But maybe it should be `inLava`, the same way we have `inWater`?

mappings/net/minecraft/entity/player/PlayerEntity.mapping
NeunEinser: I wanted to get more input first, before changing everything. I kinda liked the methods being `experience`.

mappings/net/minecraft/entity/player/PlayerEntity.mapping
NeunEinser: `lastPlayedLevelUpSoundTime` is more accurate actually. It's set in `addExperienceLevels` to the current entity's age and used to avoid a sound spam when players level up rapidly.

mappings/net/minecraft/entity/Entity.mapping
NeunEinser: Enigma doesn't show anything. grafik
That is the method I mapped here:
```java
@Environment(EnvType.CLIENT)
protected void putOnTopmostSolidBlock() {
	if (this.world == null) {
		return;
	}
	while (this.y > 0.0 && this.y < 256.0) {
		this.setPosition(this.x, this.y, this.z);
		if (this.world.doesNotCollide(this)) {
			break;
		}
		++this.y;
	}
	this.setVelocity(Vec3d.ZERO);
	this.pitch = 0.0f;
}
```

NeunEinser: Hmm, that doesn't seem to be used for other entities than players at all. What is it doing in `Entity`?

NeunEinser: What really confuses me now is that it's called in `onGameJoin`. When you join, your position doesn't get changed, nor does it reset the health. Or is that method only executed when the player *first* joins a server?

mappings/net/minecraft/entity/player/PlayerEntity.mapping
NeunEinser: I think I'm going to go with
`experience` -> `experienceLevel`
`experienceLevel` -> `experienceTotal`
now. It's going to break mods that use both, or only the former `experienceLevel`

mappings/net/minecraft/entity/Entity.mapping
NeunEinser: `set` implies it has a parameter.

Runemoro: No, there's lots of set methods with no parameters. For booleans, `set` and `unset` set to true and false.

NeunEinser: All boolean set methods I know (in general) use `set` with a parameter, and a command-tone for changing them. to true or false. I think that shows better what they do. I can change it, if it's wanted, though.

Runemoro: There are many `set` methods for booleans with no parameter in yarn, and `set` and `unset` are common names for setting to true and false. If you don't like set, you can use `mark`. `touchLava` makes it seem like the player was not touching lava and will be moved into the lava to touch it.

mappings/net/minecraft/entity/Entity.mapping
Runemoro: The method is only executed client-side every time a player joins or respawns. The health is probably sent in another packet after the respawn packet.

Runemoro: `respawn` (or maybe `onRespawn`?) would be a better name for it.

Runemoro (approved):

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -287,6 +343,7 @@ CLASS aif net/minecraft/entity/Entity
 	METHOD bP getEntityWorld ()Lbhh;
 	METHOD bQ getServer ()Lnet/minecraft/server/MinecraftServer;
 	METHOD bR isImmuneToExplosion ()Z
+	METHOD bS entityDataRequiresOperator ()Z
```
liach, line 346: `grantsOperatorCommandAccess`? A similar method is present in `BlockEntity` as well. How is it named?

mappings/net/minecraft/entity/ai/goal/AttackWithOwnerGoal.mapping
```diff
@@ -1,4 +1,6 @@
 CLASS aov net/minecraft/entity/ai/goal/AttackWithOwnerGoal
-	FIELD a owner Lajc;
+	FIELD a tamedEntity Lajc;
```
liach: what's wrong with owner? it indicates the goal's owner.

mappings/net/minecraft/entity/ai/goal/GoToOwnerAndPurrGoal.mapping
```diff
@@ -0,0 +1,2 @@
+CLASS ami net/minecraft/entity/ai/goal/GoToOwnerAndPurrGoal
+	FIELD g cat Laqs;
```
liach, line 2: `owner`

mappings/net/minecraft/entity/ai/goal/AttackWithOwnerGoal.mapping
Runemoro: Owner is very confusing here, because we have both the owner of the goal, and the owner of the goal's owner. We probably shouldn't name them `owner` and `ownerOwner`.

mappings/net/minecraft/entity/passive/BatEntity.mapping
```diff
@@ -1,5 +1,7 @@
 CLASS aqm net/minecraft/entity/passive/BatEntity
 	FIELD b BAT_FLAGS Lqh;
 	FIELD c CLOSE_PLAYER_PREDICATE Lapz;
+	FIELD d hangingPosition Lev;
+	METHOD dT todayIsAroundHalloween ()Z
```
liach: `isTodayAroundHalloween`

mappings/net/minecraft/entity/passive/FishEntity.mapping
```diff
@@ -1,13 +1,19 @@
 CLASS aqo net/minecraft/entity/passive/FishEntity
 	CLASS aqo$a FishMoveControl
 		FIELD i fish Laqo;
+		METHOD <init> (Laqo;)V
+			ARG 1 owner
 	CLASS aqo$b SwimToRandomPlaceGoal
+		FIELD h fish Laqo;
```
liach, line 7: `owner`

mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -9,17 +9,19 @@ CLASS avx net/minecraft/entity/player/PlayerEntity
...
-	FIELD bF experienceOrbPickupDelay I
+	FIELD bF experiencePickUpDelay I
 	FIELD b SIZES Ljava/util/Map;
 	FIELD bM isInWater Z
 	FIELD bN abilities Lavu;
-	FIELD bO experience I
-	FIELD bP experienceLevel I
-	FIELD bQ experienceLevelProgress F
+	FIELD bO experienceLevel I
+	FIELD bP experienceTotal I
```
liach: `totalExperience`

mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -34,36 +36,58 @@ CLASS avx net/minecraft/entity/player/PlayerEntity
...
+		ARG 4 experience
+		ARG 5 leveled
 	METHOD a openContainer (Lahd;)Ljava/util/OptionalInt;
 	METHOD a addCritParticles (Laif;)V
+		ARG 1 target
 	METHOD a interact (Laif;Lagz;)Laha;
 		ARG 1 entity
+		ARG 2 hand
 	METHOD a setMainHand (Lain;)V
+		ARG 1 absoluteHand
```
liach, line 50: Is this actually `HandSide` or what

mappings/net/minecraft/entity/projectile/ExplosiveProjectileEntity.mapping
liach: `directionVec`?

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
```diff
@@ -6,13 +6,19 @@ CLASS axl net/minecraft/entity/vehicle/AbstractMinecartEntity
 	METHOD <init> (Laij;Lbhh;)V
 		ARG 1 type
 	METHOD <init> (Laij;Lbhh;DDD)V
-		ARG 1 world
+		ARG 1 type
+		ARG 2 world
 		ARG 3 x
+		ARG 5 y
+		ARG 7 z
+	METHOD a (F)V
```
liach: uh, what is this?

mappings/net/minecraft/entity/ai/goal/AttackWithOwnerGoal.mapping
liach: Let's choose another universal word for the entity that goals apply to... applicant?

Runemoro: I suggest `entity`, `cat`, `zombie`, etc., see https://github.com/FabricMC/yarn/issues/661. Having a single name also causes issues with subclasses that have a more specific owner field.

liach: Ah, `actor`. How about that? We should change all `owner` to `actor`, which is sensible and not confusing any more.

mappings/net/minecraft/entity/ai/goal/GoToOwnerAndPurrGoal.mapping
liach: `actor`

mappings/net/minecraft/entity/passive/FishEntity.mapping
liach: `actor`

mappings/net/minecraft/entity/player/PlayerEntity.mapping
Runemoro: It should be `hand`, even if it's an `AbsoluteHand`.

mappings/net/minecraft/entity/Entity.mapping
NeunEinser: It's currently called afterRespawn. Is that okay?

mappings/net/minecraft/entity/ai/goal/AttackWithOwnerGoal.mapping
NeunEinser: But it will still be inconsistent because some Goals are extended by others, which then store a less general `actor` reference (e.g. super class uses `MobEntityWithAi`, sub class stores `DolphonEntity`). If the super class already uses actor, not possible anymore.
I suggest to use `actor` in every class that extends `Goal` directly and `<entity>Actor` when `actor` is not possible.
I avoided exactly that in that pr, because `catOwner` sounds like someone who owns the cat, but with actor that should not be a problem anymore.

mappings/net/minecraft/entity/ai/goal/GoToOwnerAndPurrGoal.mapping
NeunEinser: It extends MoveToTargetGoal, so `catActor`

mappings/net/minecraft/entity/passive/BatEntity.mapping
NeunEinser: No, it's not just Oct 31. It's true from Oct 10 to Sept 3.

NeunEinser: No, it's not just Sept 31. It's true from Sept 20 to Nov 3.

mappings/net/minecraft/entity/passive/FishEntity.mapping
NeunEinser: `fishActor`

mappings/net/minecraft/entity/player/PlayerEntity.mapping
NeunEinser: Hmm, no. `absoluteHand` is left or right. This changes the `mainHand` to either the left or right hand, so `MainHand` has nothing to do with `AbsoluteHand`, hence the difference is important here.

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
NeunEinser: A method that sets some tracked data, where I tried to figure out what it exactly is, and then forgor to remove the parameter name.
It has something to do with the minecart shake animation, I am pretty sure, but there are 3 different types of that, and I gave up on it.

mappings/net/minecraft/entity/passive/BatEntity.mapping
NeunEinser: Oh, I misread.

mappings/net/minecraft/entity/projectile/ExplosiveProjectileEntity.mapping
NeunEinser: I thought about that too, wasn't sure because of the accuracy change, though.

mappings/net/minecraft/entity/Entity.mapping
NeunEinser: `BlockEntity` doesn't seem to have such a method, even though it would make sense. Maybe it's somewhere in the item?
Also, `grantsOperatorCommandAccess` sounds like, when using this item, you get operator command access.
`requiresOperatorCommandAccess`, if you want that. That does not include, however, that only the usage of entity data requires that. You can set a command block minecart from your inventory without being an operator, but the `EntityTag` will not be applied

NeunEinser: This is from CommandBlockItem, without looking at much else.
```java
@Nullable
@Override
protected BlockState getBlockState(ItemPlacementContext itemPlacementContext) {
	PlayerEntity playerEntity3 = itemPlacementContext.getPlayer();
	return (playerEntity3 == null || playerEntity3.isCreativeLevelTwoOp()) ? super.getBlockState(itemPlacementContext) : null;
}
```

mappings/net/minecraft/entity/ai/goal/GoToOwnerAndPurrGoal.mapping
liach: Just `cat` as rune suggested. Disregard my `actor` suggestions.

Runemoro:
@NeunEinser Why owner -> actor? We said we'd keep them named based on the entity type.

liach:
> [@NeunEinser Why owner -> actor? We said we'd keep them named based on the entity type.](https://github.com/FabricMC/yarn/pull/657#issuecomment-488043836)

I should have clarified about our latest decision to use entity class name to prevent clashes.

NeunEinser:
@Runemoro This was the latest reply by liach in that chain:
> Ah, `actor`. How about that? We should change all `owner` to `actor`, which is sensible and not confusing any more.

Runemoro:
@NeunEinser No, the last reply was:

> We should use the entity type till mojang makes goals have a generic parameter for its owner type like brain class does.

in https://github.com/FabricMC/yarn/issues/661.

NeunEinser:
Of course, when working on this, I read stuff on my pr and don't worry so much about (usually) unrelated issues elsewhere on this repo.
It was a miscommunication. No need to discuss that any further.

It should not be too much effort to change it in classes that have an Actor suffix now.
Is that sufficient for your desires?

Runemoro:
@NeunEinser Things are still named actor...

So we now have three styles rather than two... `owner`, `actor`, and entity-type.

NeunEinser:
Intentionally did not remove actor namings from the following classes. They are generic, because Mojang likes consistency.
```
BowAttackGoal
CrossbowAttackGoal
HoldInHandsGoal
MoveToRaidCenterGoal
PatrolEntity.PatrolGoal
RaiderEntity.PickupBannerAsLeaderGoal
```

Runemoro (approved):

mappings/net/minecraft/block/InfestedBlock.mapping
```diff
@@ -1,8 +1,11 @@
...
+	FIELD b REGULAR_INFESTED_MAP Ljava/util/Map;
+	METHOD <init> (Lbml;Lbml$c;)V
+		ARG 1 regularBlock
+		ARG 2 settings
 	METHOD d getRegularBlock ()Lbml;
 	METHOD e getRegularBlock (Lbml;)Lbvj;
 		ARG 0 infestedBlockState
-	METHOD j hasRegularBlock (Lbvj;)Z
-		ARG 0 infestedBlockState
+	METHOD j isInfestableBlock (Lbvj;)Z
```
jamierocks: I'd drop the *Block.

liach: `canInfest` is shorter.

NeunEinser: `canInfest` is not a good property imo, though. I'd go for `canBeInfested`, then. `isInfestable` is good, too and shorter than `canBeInfested`

Runemoro (approved):

mappings/net/minecraft/block/InfestedBlock.mapping
liach: Can we just use `contains` given the class name `InfestableBlock` already has the `infest` context and this is a static method?

Runemoro: No, `contains` would mean that the block somehow contains another block.

modmuss50 (approved):
Looks good, I can sort the merge conflicts out if needed.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -489,7 +489,7 @@ CLASS aif net/minecraft/entity/Entity
 	METHOD j getCameraPosVec (F)Lcrv;
 		ARG 1 tickDelta
 	METHOD j allowsPermissionLevel (I)Z
-	METHOD j onBubbleColumnEjection (Z)V
+	METHOD j onBubbleColumnSurfaceCollision (Z)V
```
NeunEinser, line 492: Actually also fixes a typo in `surface`, compared to main branch.
It is the better name though, since it's also used for bubble columns that drag, so it can also suck entities in, with a higher downwards velocity than usually.
top collision might be an alternative, since the bubble column block at the surface is also the topmost of the column.

PR #659, opened 2019-04-29
Update missing Items mappings

doublej472:
This PR adds some of the missing mappings in the Items class


asiekierka (approved):

kashike (approved):

doublej472:
@asiekierka @kashike I've added a much bigger commit marking the Enigma names as unobfusicated (which was surprisingly annoying to do), please let me know what you think!

asiekierka:
@doublej472 if they're dark gray *don't touch them* unless they need corrections - they're auto-derived

doublej472:
@asiekierka alright, I've removed that part of the PR now.

It did fix the generated code from `Export Source...` in yarn, otherwise it uses the obfusicated name, maybe an Enigma bug / feature?

asiekierka:
The Enigma source code export isn't all that great - using fabric-loom (or, in other words, applying Fernflower (or Procyon, or CFR) to a tiny-remapper-remapped JAR) is by far preferable. Either way, decompiled source code should only be used as a lookup aid and not recompiled due to the chances of decompiler artifacts causing different behaviour.

NeunEinser (approved):

PR #669, opened 2019-05-05
VerticalEntityPosition -> EntityContext

ChloeDawn:
This object is used more than for just "vertical position".

- `isHolding(Item)`: Determines if the entity is holding the given item in their main hand.
- `isSneaking()`: Determines if the entity is sneaking.
- `isAbove(VoxelShape,BlockPos,boolean)`: Determines if the entity is above the given shape and position.
  - Returns the `defaultValue` if the entity is absent.

### DIFF
--- a/mappings/net/minecraft/entity/EntityContext.mapping
+++ b/mappings/net/minecraft/entity/EntityContext.mapping
@@ -1,0 +1,11 @@
+CLASS csa net/minecraft/entity/EntityContext
+	METHOD a absent ()Lcsa;
+	METHOD a of (Laif;)Lcsa;
+		ARG 0 entity
+	METHOD a isHolding (Lbbv;)Z
+		ARG 1 item
+	METHOD a isAbove (Lcsp;Lev;Z)Z
+		ARG 1 shape
+		ARG 2 pos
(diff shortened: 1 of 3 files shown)


ChloeDawn:
These were named with the mindset that a parameter would be declared as `entity`, with checks on `entity.isHolding`, `entity.isSneaking`, `entity.isAbove`. If it is decided that the type should be declared as `context` in parameters, then we will need to add `Entity` to the method names.. but I feel it is redundancy here that is easily solved by just referring to the type as `entity` in parameters (afaik there are no applications where an entity *and* a context are both provided in parameters, it is either one or the other.

Pyrofab (approved):

Gegy (approved):

NeunEinser:
Hmm, I don't like `Context` that much. It is very likely that this will get called `context` in parameters, rather than `entity` and then it is confusing, as you said.
`isEntityHolding` and such is too verbose, though.

It seems to store some current attributes of the entity, so maybe `EntityState`? that at least is more likely to get called `entityState`, which would make a little more sense imo.

modmuss50 (approved):

PR #671, opened 2019-05-05
Add *.sh to .gitignore

ChloeDawn:
I use scripts to load Enigma and setup commits etc. Would be nice if Yarn itself ignored script files in Git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ out/
 *.iws
 *.ipr
 *.jar
+
+*.sh


modmuss50:
Moving to a git include might not be a bad idea, I think we use it on all if not most of the other projects.

ChloeDawn:
Sounds good to me.

modmuss50 (changes requested):
this can be changed to a git include, either via this PR or I can do it later.

ChloeDawn:
I can make a new PR for it tomorrow

PR #674, opened 2019-05-06
Map resources (pack, reload) stuff

liach:
Mapped:
* resources (pack, reload) stuff (net.minecraft.resource package should be fully covered now)
* some tag stuff
* command suggestion window in both chat and command block ui (mojang should just unify those instead of copy pasta)

Edit: 
1. Visit https://github.com/FabricMC/yarn/pull/674#issuecomment-489503792 for a first round of questions to debate.


liach:
@Runemoro Requesting a review from you.

mappings/net/minecraft/client/gui/AbstractCommandBlockScreen.mapping
```diff
@@ -1,21 +1,61 @@
 CLASS dak net/minecraft/client/gui/AbstractCommandBlockScreen
+	CLASS dak$a SuggestionWindow
+		FIELD b area Ldkt;
+		FIELD c suggestions Lcom/mojang/brigadier/suggestion/Suggestions;
+		FIELD d typedText Ljava/lang/String;
+		FIELD e inWindowIndex I
+		FIELD f selection I
+		FIELD g mouse Lcru;
+		FIELD h chosen Z
```
Runemoro: What does this mean?

mappings/net/minecraft/client/gui/AbstractCommandBlockScreen.mapping
```diff
@@ -1,21 +1,61 @@
...
 	FIELD c doneButton Lcwo;
 	FIELD d cancelButton Lcwo;
 	FIELD e toggleTrackingOutputButton Lcwo;
 	FIELD f trackingOutput Z
+	FIELD g exceptions Ljava/util/List;
 	FIELD j parsedCommand Lcom/mojang/brigadier/ParseResults;
-	FIELD m suggestionsDisabled Z
+	FIELD k suggestionsFuture Ljava/util/concurrent/CompletableFuture;
+	FIELD l suggestionWindow Ldak$a;
+	FIELD m notSuggesting Z
```
Runemoro: This is just temporarily set by a method in the suggestion window when completing the command to avoid infinite recursion. `notSuggesting` makes it sound like the window is closed. `suggestionsDisabled` or `completingCommand` better describes this.

mappings/net/minecraft/client/gui/ingame/ChatScreen.mapping
```diff
@@ -1,31 +1,46 @@
...
+	FIELD b commandExceptions Ljava/util/List;
 	FIELD c commandExceptionsX I
 	FIELD d commandExceptionsWidth I
 	FIELD e WHITESPACE_PATTERN Ljava/util/regex/Pattern;
 	FIELD i parseResults Lcom/mojang/brigadier/ParseResults;
 	FIELD j suggestionsFuture Ljava/util/concurrent/CompletableFuture;
 	FIELD k suggestionsWindow Lcyr$a;
-	FIELD m suggestionsTemporarilyDisabled Z
-	METHOD a openSuggestionsWindow ()V
+	FIELD m notSuggesting Z
```
Runemoro: Same here

mappings/net/minecraft/client/gui/AbstractCommandBlockScreen.mapping
liach: Will do `completingCommand`

mappings/net/minecraft/client/gui/AbstractCommandBlockScreen.mapping
liach: Means if a suggestion has been chosen, as if you have typed a tab after you have input the initial characters. When this is true, typing tab will go to next suggestion instead of filling this suggestion.

mappings/net/minecraft/client/gui/widget/TextFieldWidget.mapping
```diff
@@ -55,6 +55,8 @@ CLASS cwu net/minecraft/client/gui/widget/TextFieldWidget
 		ARG 3 mouseY
 	METHOD j getMaxLength ()I
 	METHOD k hasBorder ()Z
+	METHOD k getGuiXForStringPosition (I)I
```
Runemoro: `getCharacterX(int index)`?

mappings/net/minecraft/resource/AbstractFilenameResourcePack.mapping
```diff
@@ -1,11 +1,20 @@
 CLASS we net/minecraft/resource/AbstractFilenameResourcePack
```
Runemoro: Maybe also rename this to `AbstractFileResourcePack`?

mappings/net/minecraft/client/gui/AbstractCommandBlockScreen.mapping
liach: `completingSuggestion`

mappings/net/minecraft/client/gui/AbstractCommandBlockScreen.mapping
liach: Will do `completed`

mappings/net/minecraft/resource/DefaultResourcePack.mapping
```diff
@@ -3,12 +3,15 @@ CLASS wk net/minecraft/resource/DefaultResourcePack
...
+	FIELD e typeToFileSystem Ljava/util/Map;
+	METHOD <init> ([Ljava/lang/String;)V
+		ARG 1 namespaces
 	METHOD a getIdentifiers (ILjava/lang/String;Ljava/nio/file/Path;Ljava/lang/String;Ljava/util/function/Predicate;)Ljava/util/Collection;
 		ARG 1 maxDepth
 		ARG 2 namespace
 		ARG 3 path
 		ARG 4 searchLocation
-		ARG 5 filter
+		ARG 5 fileNameFilter
```
Runemoro: For `DefaultResourcePack`, the file name is the same as the identifier, so it's really an `identifierFilter` (it's also the same argument that gets passed to `ResourceManager.findResources`, which is not file-specific), and since the method is `getIdentifier`, `filter` is enough.

mappings/net/minecraft/resource/ReloadableResourceManagerImpl.mapping
```diff
@@ -5,5 +5,12 @@ CLASS xi net/minecraft/resource/ReloadableResourceManagerImpl
 	FIELD d initialListeners Ljava/util/List;
 	FIELD e namespaces Ljava/util/Set;
 	FIELD f type Lwi;
+	FIELD g engineThread Ljava/lang/Thread;
```
Runemoro: `engineThread`? Could just `thread` be enough (as in, "the thread of the resource manager").

mappings/net/minecraft/resource/ResourceImpl.mapping
```diff
@@ -4,9 +4,14 @@ CLASS xj net/minecraft/resource/ResourceImpl
 	FIELD c packName Ljava/lang/String;
 	FIELD d id Lqs;
 	FIELD e inputStream Ljava/io/InputStream;
-	FIELD f metadataInputStream Ljava/io/InputStream;
+	FIELD f metaInputStream Ljava/io/InputStream;
```
Runemoro, line 7: This is called `metadata` everywhere else? Should the other names be changed too?

mappings/net/minecraft/resource/ReloadableResourceManagerImpl.mapping
liach: It's the thread of the engine (client, server threads). Not used by the manager.

mappings/net/minecraft/resource/ResourcePack.mapping
```diff
@@ -1,12 +1,19 @@
 CLASS wh net/minecraft/resource/ResourcePack
 	METHOD a getName ()Ljava/lang/String;
 	METHOD a getNamespaces (Lwi;)Ljava/util/Set;
+		ARG 1 type
 	METHOD a findResources (Lwi;Ljava/lang/String;ILjava/util/function/Predicate;)Ljava/util/Collection;
 		ARG 1 type
 		ARG 2 namespace
+		ARG 3 maxDepth
+		ARG 4 fileNameFilter
```
Runemoro: `ResourcePack` is not file-specific, it could be implemented using something other than files. `identifierFilter` is better (note that file name = identifier for the `DefaultResourcePack`).

mappings/net/minecraft/resource/ResourceImpl.mapping
liach: Hmm? I think we have simplified it at some resource-related places... Might be worthy for an issue for discussion.

mappings/net/minecraft/resource/ResourceReloadHandler.mapping
```diff
@@ -1,9 +1,28 @@
 CLASS xh net/minecraft/resource/ResourceReloadHandler
+	CLASS xh$a Creator
```
Runemoro: This should be `Factory`, to be consistent with all the other `Factory` classes we have.

mappings/net/minecraft/resource/ResourceReloadListener.mapping
```diff
@@ -1,9 +1,10 @@
 CLASS wz net/minecraft/resource/ResourceReloadListener
-	CLASS wz$a Helper
-		METHOD a waitForAll (Ljava/lang/Object;)Ljava/util/concurrent/CompletableFuture;
+	CLASS wz$a Anchor
```
Runemoro: Just curious, what do you mean by `Anchor`?

mappings/net/minecraft/tag/ItemTags.mapping
```diff
@@ -1,6 +1,11 @@
 CLASS yz net/minecraft/tag/ItemTags
+	CLASS yz$a AutoUpdatingTag
+		FIELD a modCount I
```
Runemoro: `cachedModificationId`

mappings/net/minecraft/tag/ItemTags.mapping
```diff
@@ -1,6 +1,11 @@
...
+		FIELD b delegate Lzb;
 	FIELD N container Lzc;
-	FIELD O containerChanges I
+	FIELD O modCount I
```
Runemoro: `modificationId`

mappings/net/minecraft/tag/ItemTags.mapping
```diff
@@ -1,6 +1,11 @@
 CLASS yz net/minecraft/tag/ItemTags
+	CLASS yz$a AutoUpdatingTag
```
Runemoro: `CachingTag`

mappings/net/minecraft/resource/ResourcePack.mapping
liach: This only filters the path of the id, so I will call it `pathFilter`

mappings/net/minecraft/tag/TagManager.mapping
```diff
@@ -1,4 +1,14 @@
 CLASS zd net/minecraft/tag/TagManager
+	CLASS zd$a PreparedBuilders
```
Runemoro: A plural class name? This contains maps between ids and builders for each builder type, so it's really just a `IdentifierToBuilderMap`.

mappings/net/minecraft/resource/ReloadableResourceManagerImpl.mapping
Runemoro: Ah. I think it's called `mainThread` in other places (chunk loading code, and also the thread names), so it should probably be called that to be consistent.

mappings/net/minecraft/resource/ResourceImpl.mapping
Runemoro: I'm talking about the other names right below this one.

mappings/net/minecraft/resource/ResourcePack.mapping
Runemoro: Oh, never mind then. `fileNameFilter` is better. You're right, it includes the extension so it's not just the ID.

mappings/net/minecraft/resource/ResourceReloadListener.mapping
liach: anchor is like a fixed point, like a spike. Idea came from mediawiki's anchor for link and railcraft's world anchor.

mappings/net/minecraft/tag/ItemTags.mapping
liach: `expectedModCount`, like java standard library?

mappings/net/minecraft/resource/ResourceReloadListener.mapping
Runemoro: Yes, I understand that. But what does it mean for resource packs?

mappings/net/minecraft/tag/ItemTags.mapping
Runemoro: No. That's very different! `expectedModCount` checks modification counts to detect and throw `ConcurrentModificationException`s. This just keeps track of the last cached modification ID to check if it needs to be updated.

mappings/net/minecraft/tag/ItemTags.mapping
liach: Auto updating is more important here as these tags aren't reloaded with tag manager but still can reflect latest tag contents.

Runemoro: Yes, so it's caching the data and only updating it from the tag manager when necessary. That's the definition of what caching is.

mappings/net/minecraft/tag/TagManager.mapping
liach: This is really an intermediary object produced by prepare stage of reload. Hence `PreparedBuilders`. Or `TagReloadTask`? Need a better name indeed.

mappings/net/minecraft/resource/ResourceReloadListener.mapping
liach: for listeners, it is an anchor point so that the completable future for apply stage can be executed after this anchor point's completable future is done.

mappings/net/minecraft/tag/TagManager.mapping
Runemoro: It's not a task. It's just a container for maps between ids and builders:
```java
    public static class a {
        final Map<Identifier, Tag.Builder<Block>> a;
        final Map<Identifier, Tag.Builder<Item>> b;
        final Map<Identifier, Tag.Builder<Fluid>> c;
        final Map<Identifier, Tag.Builder<EntityType<?>>> d;

        public a(final Map<Identifier, Tag.Builder<Block>> map1, final Map<Identifier, Tag.Builder<Item>> map2, final Map<Identifier, Tag.Builder<Fluid>> map3, final Map<Identifier, Tag.Builder<EntityType<?>>> map4) {
            this.a = map1;
            this.b = map2;
            this.c = map3;
            this.d = map4;
        }
    }
```

The task just uses it.

liach: I say task for we have `Advancement.Task` "Task Advancement" which is similar to `Tag.Builder`.

liach: Should we name this `PreparedData` instead?

Runemoro: `BuilderMap`/`BuilderMapContainer`/`BuilderContainer`?

mappings/net/minecraft/tag/ItemTags.mapping
liach: `version`?

mappings/net/minecraft/tag/TagManager.mapping
Runemoro: > Should we name this PreparedData instead?

Why PreparedData? It's a *public* class that stores all the builders for each tag.

mappings/net/minecraft/tag/ItemTags.mapping
Runemoro: Yes.

mappings/net/minecraft/resource/ResourcePack.mapping
liach: Will say `pathFilter` because in resource managers, the raw path stored in ids still include the file suffix. E.g. when game scans for resource type `advancements` with suffix `.json`, `minecraft:advancements/story/root.json` will be returned by the resource manager as the id, which the advancement loader will then truncate to `minecraft:story/root`.

mappings/net/minecraft/resource/ResourceImpl.mapping
liach: Hmm, I kept the full metadata name in interface methods, just simplified that for internal fields and methods and such for now.

mappings/net/minecraft/resource/DefaultResourcePack.mapping
liach: Discussion below.

mappings/net/minecraft/resource/ResourcePack.mapping
Runemoro: It seems like `fileNameFilter` was originally correct: `filter.test(path2.getFileName().toString()`

mappings/net/minecraft/tag/ItemTags.mapping
liach: Done. `latestVersion` and `version`.

mappings/net/minecraft/tag/TagManager.mapping
liach: > [a *public* class](https://github.com/FabricMC/yarn/pull/674#discussion_r281057773)

Nah. Mojang makes a lot of unnecessary inner classes public, such as `CommandFunctionManager$Entry` and all those caching tags you just visited.

mappings/net/minecraft/tag/ItemTags.mapping
liach: Done.

liach:
So up to this point, we have a few points to debate:

- metadata -> meta: Currently meta in fields and metadata in established usages (interfaces, etc.)
- `ResourceReloadListener.Helper` -> `Anchor`: This provides a completable future so that apply executions can be done after all prepare executions are done (apply and prepare names are from mojang logs in profiling resource reload handler) Any alternative names?
- `TagManager$a` (`zd$a`) -> `TagManager.Intermediate`: Only used in tag manager for the transition data between prepare and apply stages. Any alternative names?

liach:
@modmuss50 Requesting opinion on those 3 things above and a review.

mappings/net/minecraft/tag/TagManager.mapping
Runemoro: Still, a container class should not be named based on its only usage.

Runemoro:
> TagManager -> RegistryTagManager (only manages registry tags)

What other tags are there? All tags are registered...

liach:
@Runemoro Function tags, the only tag that is ordered (broken for a missing `.ordered(this.ordered)` builder call) and is not registry-based. 

Edit: How this `ordered` thing is broken
image

mappings/net/minecraft/resource/ResourceImpl.mapping
modmuss50: yes, I think the old name for this is fine tbh, the surrounding methods are `metadata`. Either change it everywhere or not all. Seeing your other comment using meta in fields and methods is fine tbh, and then keeping the full MetaData for class names?

mappings/net/minecraft/resource/ResourceReloadListener.mapping
```diff
@@ -1,9 +1,10 @@
 CLASS wz net/minecraft/resource/ResourceReloadListener
```
modmuss50, line 1: You should be careful not to change any names here in yarn that would make the API make no sense. This looks fine to me, just leaving this comment here to make you aware.

mappings/net/minecraft/tag/TagManager.mapping
modmuss50: the names @Runemoro suggested make sense. I dont see many people using this class, but that doesnt need to limit its name.

modmuss50 (changes requested):
A few things that @Runemoro has mostly spotted. Just double check that this doesnt clash with the names already present in fabric api: see here: 

https://github.com/FabricMC/fabric/tree/master/fabric-resource-loader/src/main/java/net/fabricmc/fabric/api/resource

mappings/net/minecraft/resource/ResourceReloadListener.mapping
liach: The name changes should be fine; I just renamed helper to Synchronizer (player suggestion)

mappings/net/minecraft/tag/TagManager.mapping
liach: Using `BuilderHolder` now.

PR #679, opened 2019-05-06
net.minecraft.util.Void -> Unit

### LINKED ISSUE #676: net.minecraft.util.Void -> Unit
It should be renamed to avoid having the same name as `java.lang.Void`. It's exactly the same as https://github.com/Mojang/DataFixerUpper/blob/master/src/main/java/com/mojang/datafixers/util/Unit.java except for not having a `toString` method.

ChloeDawn:
Resolves #676


modmuss50:
#680 also has this change. Gonna close this PR, sorry about that.

PR #681, opened 2019-05-06
Fix some strange settings names

Chocohead:
#611 added two spelling mistakes of `biome` (as far as I can tell) and named a button callback `tmp` which seems unintentional.


ChloeDawn:
It also added a parameter name to `AddServerScreen#onClose` called `unused` which should be `text` or `fieldText`

Chocohead:
Strictly speaking it is unused but might as well name it descriptively anyway

liach (approved):

ChloeDawn (approved):

modmuss50 (approved):

PR #684, opened 2019-05-07
Update LlamaEntity.mapping

tyra314:
Adds mappings for two more methods

### DIFF
--- a/mappings/net/minecraft/entity/passive/LlamaEntity.mapping
+++ b/mappings/net/minecraft/entity/passive/LlamaEntity.mapping
@@ -19,3 +19,7 @@ CLASS arv net/minecraft/entity/passive/LlamaEntity
 		ARG 0 color
 	METHOD u setVariant (I)V
 	METHOD v setStrength (I)V
+	METHOD h spitAt (Laio;)V
+		ARG 1 target
+	METHOD a breedWith (Laid;)Larv;
+		ARG 1 mate


mappings/net/minecraft/entity/passive/LlamaEntity.mapping
```diff
@@ -19,3 +19,7 @@ CLASS arv net/minecraft/entity/passive/LlamaEntity
 		ARG 0 color
 	METHOD u setVariant (I)V
 	METHOD v setStrength (I)V
+	METHOD h spitAt (Laio;)V
+		ARG 1 target
+	METHOD a breedWith (Laid;)Larv;
```
liach, line 24: Looks like `createChild` to me by param and return types.

tyra314: There is already a different function literally called `createChild`, which is used in this function. Also, there is a function `canBreedWith`.

liach: what does this do in addition to calling create child?

tyra314: createChild gives a plain new Entity instance. This function, however, does the inherit-stats-from-parents thing and gives the child the color of one of the parents. One parent is the instance, the function is called on and the other is the parameter.

modmuss50:
merging, any issues should go to a new PR or an issue.

PR #685, opened 2019-05-07
Add some obf classes and fix my snake case thing

liach:


modmuss50 (approved):
Also looks good to me

PR #686, opened 2019-05-07
Some new 1.14.1-pre1 methods and fields

NeunEinser:
This maps some of new 1.14.1-pre1 methods and fields and some other random stuff along the way. Does not clash with #685


modmuss50 (approved):
Looks good to me

PR #693, opened 2019-05-10
Pass over DataFixers

quat1024:
This PR maps a bunch of things in `net/minecraft/datafixers/fixers`, mostly focusing on missing parameter names and method names. Hopefully clearer names in this area help people understand DataFixer better.

Some highlights:

* Constructor parameters changed to `outputSchema` and `changesType`.
  * Obvious change since this is what DataFixerUpper calls them. There's just lots of these parameters.
* `EntityPaintingFix` -> `HangingEntityFix`
  * It applies to paintings *and* item frames, not just paintings.
* `ItemOminousBannerRenameFix` -> `OminousBannerItemRenameFix`
  * Suffixy.
  * Matched the other one.
* `OminousBannerBlockEntityRenameFix#convertOutdatedTranslationKey` -> `fixBannerName`
  * Felt wordy? Idk.
  * It's a fixer, so of course it's converting something outdated.
  * (see also `OminousBannerItemRenameFix`'s method that I named myself to the same thing.)
* Parameters:
  * `Dynamic<?>` -> `tag` in most cases
    * Realistically we are talking about nbt tags in 90% of cases where a dynamic appears.
  * `DynamicOps<?>` -> `ops`
    * What DataFixerUpper calls them.

Some questions:

* Parameters: What to call "Typed"s? I can't find any Typed names in DataFixerUpper.
* `LeavesFix`, `SwimStatsRenameFix`, `VillagerXpRebuildFix`
  * These don't wanna decompile for me!
* `SavedDataVillageCropFix#a, b`
  * Maybe I've been staring at Yarn too long but I can't figure these ones out.
  * Seem to be "plumbing" functions.

### DIFF
--- a/mappings/net/minecraft/datafixers/fixes/AddTrappedChestFix.mapping
+++ b/mappings/net/minecraft/datafixers/fixes/AddTrappedChestFix.mapping
@@ -1,2 +1,5 @@
 CLASS aea net/minecraft/datafixers/fixes/AddTrappedChestFix
 	FIELD a LOGGER Lorg/apache/logging/log4j/Logger;
+	METHOD <init> (Lcom/mojang/datafixers/schemas/Schema;Z)V
+		ARG 1 outputSchema
+		ARG 2 changesType
--- a/mappings/net/minecraft/datafixers/fixes/AdvancementsFix.mapping
+++ b/mappings/net/minecraft/datafixers/fixes/AdvancementsFix.mapping
@@ -1,2 +1,5 @@
 CLASS aai net/minecraft/datafixers/fixes/AdvancementsFix
 	FIELD a RENAMED_ADVANCEMENTS Ljava/util/Map;
+	METHOD <init> (Lcom/mojang/datafixers/schemas/Schema;Z)V
+		ARG 1 outputSchema
+		ARG 2 changesType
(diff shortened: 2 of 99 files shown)


Gegy (approved):

asiekierka (approved):

asiekierka:
The questions can be left for another PR.

ChloeDawn:
The package is still riddled with prefixing so the two class renames have actually made the inconsistency worse

ChloeDawn:
VBO in OptionsForceVBOFix should be Vbo

SquidDev:
I think it's worth noting that a lot of the names are just lifted from the data fixer names within the file. For instance, the VBO fix contains `OptionsForceVBOFix` in the name, and so the class is most likely named that. Likewise `OminousBannerItemRenameFix` is (probably) just called `OminousBannerRenameFix`. A lot of the the prefixed names (say, `BlockEntityJukeboxFix`) are only because that's what name is used within the class.

Of course, that doesn't mean we have to go with (assumed) Mojang names here, it was just the decision I made when originally mapping some of these classes out :).

liach:
these prefix names are straight from mojang names, so things don't need change

quat1024:
Hmm, I think even if you want to stick to the internal string names, there's definitely a lot of cases where it's appropriate to slightly change the spelling or casing of data fixer classes, since not all fixer names have a string that works as a class name anyways (like `Map id fix`)

I like `OminousBannerItemRenameFix`: the mojang string is just `OminousBannerRenameFix` and idk, part of me likes the `Item` clarifying that it's not the `OminousBannerBlockEntityRenameFix`? Agree with the `VBO` -> `Vbo` tho

NeunEinser:
If you change `ItemOminousBannerRenameFix`, you'd need to change all other class names that are prefixed like that, too though, for consistency.

PR #698, opened 2019-05-13
Hotfix for #696

ChloeDawn:
A conflict has arisen! Yarn/Tiny does not support shared implementations due to the uniqueness of names in obfuscation. I guarantee Mojang shares an implementation here (literally returns what super of `Entity#getWorld` does (`this.world`), but declared for the `Hopper#getWorld` implementation). So we need to make `Hopper#getWorld` unique. Maybe we could look into supporting shared implementations in the future?


modmuss50:
Pulled in manually with my travis PR. https://github.com/FabricMC/yarn/commit/1c92f132ee92058a36d2691850ab3590e2d16e49

PR #699, opened 2019-05-13
More accurate coral block and egg material names

Virtuoel:
Changed since only the blue coral is Tube Coral and Turtle Eggs use the same Material as Dragon Eggs.

### DIFF
--- a/mappings/net/minecraft/block/CoralFanBlock.mapping
+++ b/mappings/net/minecraft/block/CoralFanBlock.mapping
@@ -1,4 +1,4 @@
-CLASS bnl net/minecraft/block/CoralTubeFanBlock
+CLASS bnl net/minecraft/block/CoralFanBlock
 	FIELD a deadCoralBlock Lbmm;
 	METHOD <init> (Lbmm;Lbmm$c;)V
 		ARG 1 deadCoralBlock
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/block/CoralFanBlock.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS bnl net/minecraft/block/CoralTubeFanBlock
+CLASS bnl net/minecraft/block/CoralFanBlock
```
asiekierka, line 1: Weren't these based on Mojang's registry IDs? Could you double-check with Blocks?

mappings/net/minecraft/block/Material.mapping
```diff
@@ -35,7 +35,7 @@ CLASS clf net/minecraft/block/Material
 	FIELD L PISTON Lclf;
 	FIELD M UNUSED_PLANT Lclf;
 	FIELD N PUMPKIN Lclf;
-	FIELD O DRAGON_EGG Lclf;
+	FIELD O EGG Lclf;
```
asiekierka, line 38: OK

mappings/net/minecraft/block/CoralFanBlock.mapping
Prospector: I would assume they don't have individual classes for every type of coral fan? That wouldn't make sense

Gegy (approved):

Runemoro (approved):

PR #701, opened 2019-05-15
Particle pos arguments

YanisBft:
Also renamed TurtleEntityModel to be consistent with other model classes
Mapped missing block sound group

### DIFF
--- a/mappings/net/minecraft/world/LightType.mapping
+++ b/mappings/net/minecraft/world/LightType.mapping
@@ -1,1 +1,4 @@
 CLASS bhr net/minecraft/world/LightType
+	FIELD c LIGHT I
+	METHOD <init> (Ljava/lang/String;II)V
+		ARG 3 value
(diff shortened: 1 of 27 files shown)


Runemoro (approved):

mappings/net/minecraft/world/LightType.mapping
```diff
@@ -1 +1,4 @@
 CLASS bhr net/minecraft/world/LightType
+	FIELD c LIGHT I
```
Gegy: What's this constant? Maybe missing context but `LIGHT` seems pretty vague

YanisBft: Luminance better?

Gegy: How is it used?

YanisBft: It seems to be just used in LightType class, in the constructor

Gegy: It's not a `static final` constant, it should be camelCase. The value doesn't seem to be used anywhere (unless Enigma is reporting wrong) I think it's best to not map it meaninglessly.

Gegy (approved):

PR #702, opened 2019-05-16
Name some stuff in and around BoatEntity

howtonotwin:


mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
```diff
@@ -25,7 +29,14 @@ CLASS axm net/minecraft/entity/vehicle/AbstractMinecartEntity
 		ARG 7 type
 	METHOD a setCustomBlockPresent (Z)V
 	METHOD b setCustomBlock (Lbvk;)V
+	METHOD c setDamageWobbleTicks (I)V
+		ARG 1 wobbleTicks
+	METHOD d setDamageWobbleSide (I)V
+		ARG 1 wobbleSide
 	METHOD k setCustomBlockOffset (I)V
+	METHOD l getDamageWobbleStrength ()F
```
Runemoro, line 37: What's the wobble strength? Is it an angle? If yes, then should it be `getDamageWobbleAngle`?

mappings/net/minecraft/entity/vehicle/BoatEntity.mapping
```diff
@@ -7,22 +7,72 @@ CLASS axo net/minecraft/entity/vehicle/BoatEntity
...
 		ARG 1 paddle
+	METHOD a interpolatePaddlePhase (IF)F
+		ARG 1 paddle
+		ARG 2 tickDelta
 	METHOD a copyEntityData (Laii;)V
+		ARG 1 entity
 	METHOD a setBoatType (Laxo$b;)V
-	METHOD a setPaddleState (ZZ)V
+		ARG 1 type
+	METHOD a setPaddleMovings (ZZ)V
```
Runemoro, line 51: I think `setPaddlesMoving` would be better.

mappings/net/minecraft/entity/vehicle/BoatEntity.mapping
```diff
@@ -7,22 +7,72 @@ CLASS axo net/minecraft/entity/vehicle/BoatEntity
...
+		ARG 1 leftMoving
+		ARG 2 rightMoving
+	METHOD a setInputs (ZZZZ)V
```
Runemoro, line 54: `setInput` (all four booleans together form the input)

mappings/net/minecraft/entity/vehicle/BoatEntity.mapping
```diff
@@ -7,22 +7,72 @@ CLASS axo net/minecraft/entity/vehicle/BoatEntity
...
+		ARG 1 tickDelta
+	METHOD b setDamageWobbleTicks (I)V
+		ARG 1 wobbleTicks
+	METHOD c setDamageWobbleSide (I)V
+		ARG 1 side
+	METHOD d setBubbleWobbleTicks (I)V
+		ARG 1 wobbleTicks
 	METHOD f asItem ()Lbbw;
 	METHOD i getPaddleSoundEvent ()Lyj;
+	METHOD m getDamageWobbleStrength ()F
```
Runemoro, line 69: Same here

mappings/net/minecraft/entity/vehicle/BoatEntity.mapping
```diff
@@ -7,22 +7,72 @@ CLASS axo net/minecraft/entity/vehicle/BoatEntity
...
+	METHOD n getDamageWobbleTicks ()I
+	METHOD o getDamageWobbleSide ()I
 	METHOD p getBoatType ()Laxo$b;
+	METHOD q handleBubbleColumn ()V
 	METHOD s checkLocation ()Laxo$a;
 	METHOD u checKBoatInWater ()Z
 	METHOD v getUnderWaterLocation ()Laxo$a;
+	METHOD w updateVelocity ()V
+	METHOD z updatePaddles ()V
```
Runemoro, line 78: Is this called once per tick? If yes, this should be `tickPaddles`.

modmuss50:
Updating, any further feedback or changes should go into a new PR/issue.

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
howtonotwin: I think `getDamageWobbleAngle` would be wrong. That's calculated in the renderer using this and `getDamageWobbleTicks()`. Really, there should be `damageWobble`, `lastDamageWobble`, and `interpolateDamageWobble`, like there are for bubble columns, but Mojang appears to have taken a shortcut. If you want to make it clear it's an angle, it should be `getDamageWobbleMaxAngle`, but then we also need `getBubbleWobbleStrength` ~> `getBubbleWobbleMaxAngle` etc, and *then* I think we may want to add `angle` to *everything* `wobble` related.

Perhaps we can instead name this `getDamageWobbleMax` and ~~leave it up to discovery that everything "wobble" is in angles~~ name the parameter to `setDamageWobbleMax` `angle`.

Issue #683, opened 2019-05-06
Package visibility issues


JamiesWhiteShirt:
[Starting with 1.14 these were intended to be resolved](https://github.com/FabricMC/yarn/issues/615), but there are currently many issues.

A somewhat shortened report of classes that must be moved to the same packages:

```
ERROR: Must be in one package:
net/minecraft/world/loot/LootChoiceProvider
net/minecraft/world/loot/entry/AlternativeEntry
net/minecraft/world/loot/entry/AlternativeEntry$Builder
net/minecraft/world/loot/entry/CombinedEntry
net/minecraft/world/loot/entry/CombinedEntry$1
net/minecraft/world/loot/entry/CombinedEntry$Factory
net/minecraft/world/loot/entry/CombinedEntry$Serializer
net/minecraft/world/loot/entry/GroupEntry
net/minecraft/world/loot/entry/LootEntries
net/minecraft/world/loot/entry/LootEntries$Serializer
net/minecraft/world/loot/entry/LootEntry
net/minecraft/world/loot/entry/LootEntry$Builder
net/minecraft/world/loot/entry/LootEntry$Serializer
net/minecraft/world/loot/entry/SequenceEntry
ERROR: Must be in one package:
net/minecraft/client/gui/Screen
net/minecraft/client/gui/ingame/ChatScreen
net/minecraft/client/gui/ingame/ChatScreen$1
net/minecraft/client/gui/ingame/ChatScreen$SuggestionWindow
net/minecraft/client/gui/menu/CustomizeBuffetLevelScreen
net/minecraft/client/gui/menu/CustomizeBuffetLevelScreen$1
net/minecraft/client/gui/menu/CustomizeBuffetLevelScreen$BuffetBiomesListWidget
net/minecraft/client/gui/menu/CustomizeBuffetLevelScreen$BuffetBiomesListWidget$BuffetBiomeItem
net/minecraft/client/gui/menu/CustomizeFlatLevelScreen
net/minecraft/client/gui/menu/CustomizeFlatLevelScreen$1
net/minecraft/client/gui/menu/CustomizeFlatLevelScreen$SuperflatLayersListWidget
net/minecraft/client/gui/menu/CustomizeFlatLevelScreen$SuperflatLayersListWidget$SuperflatLayerItem
net/minecraft/client/gui/menu/NewLevelPresetsScreen
net/minecraft/client/gui/menu/NewLevelPresetsScreen$SuperflatPreset
[...]

Issue #649, opened 2019-04-24
Mojang package structure


Runemoro:
[Here's a list of all the class names in alphabetical order (which Proguard preserves), along with guesses for package names that work with the ordering](https://gist.github.com/Runemoro/e8c45aabc8f7048d0414e42ea338d510)

I'm not saying we should copy the entire structure in yarn (I'm not saying we shouldn't either), but it may help us figure out which classes are in the wrong package, misnamed (there's lots!), and which packages need to be merged to avoid [this](https://github.com/FabricMC/yarn/issues/615) in the future.

NeunEinser:
Few notes (omitting `net.minecraft.` in names):

- ❌  `predicate` -> `advancement.criterion`: Some predicates are at least used by loot tables as well (since 1.14). In 1.13 predicates as a subpackage of criterions would have meade sense, though.
- ❌  util.NumberRange -> advancements: What the hell is NumberRange doing in advancements? It's used in a variety of places :jangthonk:
- ✔️ `predicates.*` -> `predicates` It might make sense to not have subpackages in predicates, since it seems like what Mojang is doing and it's not that many classes anyways
- ✔️ `server.function.CommandFunction` -> `command`: It's used in `command.arguments.FunctionArgumentType` too
- ✔️ `server.command`... `.CommandOutput`, `.ServerCommandSource`, `.CommandManager`, `.CommandManager`  -> `command`: I never liked stuff that are not commands in that package. This makes sense imo, although it is debatable.
- ✔️  `util.TagHelper` -> `nbt`: It's an NBT util, so debatable. But it does make sense and we might as well adapt Mojang's structure.
- ❓  `server.dedicated.command` -> `server.command`: I think the dedicated seperation does make sense, but it might also be more convinient to have all comands in one place. Mojang probably thinks of the `dedicated` package as classes that do not belong in any other package and are only for the dedicated server. Two `commands` packages that have the same meaning at two places is probably unwanted.
  - ✔️  `server.dedicated.command` -> `server.command.dedicated`: That might be a better solution to the current structure (see also previous point), though it goes against the "everything that's dedicated only goes into `server.dedicated`" moral.
[...]

PR #703, opened 2019-05-16
fix visibility issues

NeunEinser:
Fixes most of #683 
Not fixed:
```
ERROR: Must be in one package:
net/minecraft/client/gui/MapRenderer
net/minecraft/client/gui/MapRenderer$1
net/minecraft/client/gui/MapRenderer$MapTexture
net/minecraft/client/network/ClientPlayNetworkHandler
net/minecraft/client/network/ClientPlayNetworkHandler$1
net/minecraft/client/network/PlayerListEntry
net/minecraft/client/network/PlayerListEntry$1
```
I think this is a false positive. Their obfuscated names are very far apart.

This pull request adopts the package structure for gui screens from #649.
I used `ingame` for what Rune called `inventory` in #649, because there are screens in there that are not inventory related, like command block screens, and book (edit) screen. Not sure about that one, though, since there are also screens that are in separate packages, like the PauseMenuScreen. Most ingame screens are in that package, though
Also changes some names where they were bad.


NeunEinser:
Requesting review from @JamiesWhiteShirt to confirm my branch fixes the visibility issues and does not create new ones somehow.

liach:
The so called map renderer is in fact a client map manager, hence it qualifies for the network package.

Runemoro (approved):

Gegy (approved):

JamiesWhiteShirt:
I can confirm that the partition that was not fixed is likely a false positive. There is a chance we are missing some of them, but these are definitely great to have fixed now.

liach:
It is not a false positive, given the client network handler accesses `MapRenderer$MapTexture`, which is package private. It is totally possible that mojang has two same packages in two different modules.

NeunEinser:
> It is not a false positive, given the client network handler accesses `MapRenderer$MapTexture`, which is package private.

From Discord (by `@JamiesWhiteShirt#8888`) (I didn't really check myself)
>MapRenderer has a public method that returns a package-private which I had not accounted for. this is apparently legal in java land
>and a public method with a package-private parameter
>so the descriptor of the called method doesn't matter

So `MapTexture` is the return type of a public method, which is apparently legal.

>It is totally possible that mojang has two same packages in two different modules.

While this is theoretically possible, looking at the context and other classes and packages in between, it is highly unlikely that these are in different modules.

PR #705, opened 2019-05-17
Change some BoatEntity names

howtonotwin:
Continuation of old PR #702 incorporating the given feedback.

### DIFF
--- a/mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
+++ b/mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
@@ -2,7 +2,7 @@ CLASS axm net/minecraft/entity/vehicle/AbstractMinecartEntity
 	CLASS axm$a Type
 	FIELD b WOBBLE_TICKS Lqh;
 	FIELD c WOBBLE_SIDE Lqh;
-	FIELD d WOBBLE_MAX_ANGLE Lqh;
+	FIELD d WOBBLE_MAX_ROLL Lqh;
 	FIELD e CUSTOM_BLOCK_ID Lqh;
 	FIELD f CUSTOM_BLOCK_OFFSET Lqh;
 	FIELD g CUSTOM_BLOCK_PRESENT Lqh;
@@ -14,8 +14,8 @@ CLASS axm net/minecraft/entity/vehicle/AbstractMinecartEntity
 		ARG 3 x
 		ARG 5 y
 		ARG 7 z
-	METHOD a setWobbleMaxAngle (F)V
-		ARG 1 maxAngle
+	METHOD a setWobbleMaxRoll (F)V
+		ARG 1 maxRoll
 	METHOD a onActivatorRail (IIIZ)V
 		ARG 1 x
 		ARG 2 y
@@ -35,7 +35,7 @@ CLASS axm net/minecraft/entity/vehicle/AbstractMinecartEntity
 	METHOD d setWobbleSide (I)V
 		ARG 1 wobbleSide
 	METHOD k setCustomBlockOffset (I)V
-	METHOD l getWobbleMaxAngle ()F
+	METHOD l getWobbleMaxRoll ()F
 	METHOD m getWobbleTicks ()I
 	METHOD n getWobbleSide ()I
 	METHOD o getMinecartType ()Laxm$a;
--- a/mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
+++ b/mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
@@ -1,8 +1,8 @@
 CLASS axm net/minecraft/entity/vehicle/AbstractMinecartEntity
 	CLASS axm$a Type
-	FIELD b DAMAGE_WOBBLE_TICKS Lqh;
-	FIELD c DAMAGE_WOBBLE_SIDE Lqh;
-	FIELD d DAMAGE_WOBBLE_MAX Lqh;
+	FIELD b WOBBLE_TICKS Lqh;
+	FIELD c WOBBLE_SIDE Lqh;
+	FIELD d WOBBLE_MAX_ANGLE Lqh;
 	FIELD e CUSTOM_BLOCK_ID Lqh;
 	FIELD f CUSTOM_BLOCK_OFFSET Lqh;
 	FIELD g CUSTOM_BLOCK_PRESENT Lqh;
@@ -14,7 +14,7 @@ CLASS axm net/minecraft/entity/vehicle/AbstractMinecartEntity
 		ARG 3 x
 		ARG 5 y
 		ARG 7 z
-	METHOD a setDamageWobbleMax (F)V
+	METHOD a setWobbleMaxAngle (F)V
 		ARG 1 maxAngle
 	METHOD a onActivatorRail (IIIZ)V
 		ARG 1 x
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
```diff
@@ -34,7 +35,7 @@ CLASS axm net/minecraft/entity/vehicle/AbstractMinecartEntity
 	METHOD d setDamageWobbleSide (I)V
 		ARG 1 wobbleSide
 	METHOD k setCustomBlockOffset (I)V
-	METHOD l getDamageWobbleStrength ()F
+	METHOD l getDamageWobbleMax ()F
```
Runemoro: What does this fix? I was asking if the strength is measured as an angle. If yes, then `getDamageWobbleAngle` would be more descriptive. Otherwise, `getDamageWobbleStrength` is ok.

howtonotwin: The problem is that your suggestion is (IMO) wrong and you (justifiably) do not like my initial naming. I attempted to compromise by making it clear that we're working in angles by naming the argument of `setDamageWobbleMax` `maxAngle`. If you want more than that, `wobble` can be globally replaced with `wobbleAngle` (I think it would be inconsistent to only change it for just the two methods you pointed out), but I have not done so yet because, IMO, that would be very verbose. See also my replies on the other PR.

Runemoro: > The problem is that your suggestion is (IMO) wrong

How is it wrong, is it not an angle?

> you (justifiably) do not like my initial naming

If a more descriptive name exists (`getDamageWobbleAngle`), it should be named that, rather than forcing people who want to use it to search through the code to figure out what the unit is.

> making it clear that we're working in angles by naming the argument of setDamageWobbleMax

I really don't see how adding `max` make it clear what the unit is?

>  If you want more than that, wobble can be globally replaced with wobbleAngle (I think it would be inconsistent to only change it for just the two methods you pointed out)

There are no names involving `wobble` other than those you just named...

liach: `getDamageWobbleMax` is definitely wrong; is it just `getMaxWobbleDamage`, where `damage` is the actual unit?

Runemoro: @liach Based on the current name, I'm assuming this is the angle by which the minecart wobbles when damaged.

liach: `getWobbleRoll`

asiekierka:
Half the pull request is undecided, and 1.14.2pre2 is coming up. I'll re-open if it doesn't break much.

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
howtonotwin: The units are degrees, yes. Other names with `wobble` that we may want to replace with `wobbleAngle` in them:

* `DAMAGE_WOBBLE_MAX`, `getDamageWobbleMax`, `setDamageWobbleMax` (these appear in twice, in `AbstractMinecartEntity` and `BoatEntity`)
* `BUBBLE_WOBBLE_MAX`, `getBubbleWobbleMax`, `setDamageWobbleMax`
* `lastBubbleWobble`, `bubbleWobble`, `interpolateBubbleWobble`

The damage wobble is not a roll (side-to-side), it is a pitch (front-to-back). I don't know of a name for the bubble wobble; boats wobble sort of diagonally, roll and pitch, while over bubble columns. Also, in `AbstractMinecartEntity`, this is not actually damage wobble, just wobble, as it wobbles while hitting activator rails, too. I didn't see that the first time around.

Now, (as I've said before), *any* name like `getWobbleRoll` or `getWobbleAngle` is inaccurate. This variable is only the maximum allowed wobble (as an angle). The *actual* wobble, for the minecart and for the boat's damage wobble, is only calculated in the renderer, incorporating a `sin` function and the world time. The boat's `bubbleWobble` field holds the calculated wobble, which is updated every `tick()`, and the `interpolateBubbleWobble` method is what actually gets the rendered angle. I am OK with naming `interpolateBubbleWobble` as such, because it actually tells you the wobble. I am not OK with naming `getDamageWobbleMax` `getWobble` etc., because it doesn't.

If it's going to be causing this much confusion, then I guess it's best to just be as explicit as possible. These six data tracker methods should be named `{g,s}et[Reason]WobbleMaxAngle`, the data trackers `[REASON_]WOBBLE_MAX_ANGLE`, and the three stragglers in `BoatEntity` should be `[last]BubbleWobbleAngle` and `interpolateBubbleWobbleAngle`.

asiekierka:
Nope, it fine. Carry on.

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
Runemoro: > Other names with wobble that we may want to replace with wobbleAngle in them

You named all of these. My suggestion of using "wobble angle" instead of wobble applies to all your names from your last PR (except wobbleSide).

>  I am not OK with naming getDamageWobbleMax getWobble etc., because it doesn't.

When did I ever suggest `getWobble`? I am suggesting `getDamageWobbleAngle` (or `MaxAngle` if you prefer...

> If it's going to be causing this much confusion, then I guess it's best to just be as explicit as possible. These six data tracker methods should be named {g,s}et[Reason]WobbleMaxAngle, the data trackers [REASON_]WOBBLE_MAX_ANGLE, and the three stragglers in BoatEntity should be [last]BubbleWobbleAngle and interpolateBubbleWobbleAngle.

I don't understand what you're saying. Adding the word `angle` to the name was my original suggestion, which you argue against, and you then suggest the exact same thing.

howtonotwin: For nine of the listed names, `maxAngle` is correct, but `angle` is wrong (to me, it would imply that it's the *current* angle, which it isn't). As I read it, you suggested just `angle`, which I argued against, because, again, just `angle` is not correct. Also, I thought it would be too verbose, as all 12 names listed above would have to contain `wobbleAngle` instead of `wobble`, for consistency, which is without factoring in that 9 of them also need a `max` tacked on.  I said that we *could* use those verbose names, but, when creating this PR, I chose different names that I hoped would do instead. Now, I'm realizing that this is more trouble than it's worth, and I'm just going to use the most verbose option because we've been talking past each other for a whole day now.

> When did I ever suggest `getWobble`? I am suggesting `getDamageWobbleAngle`

I was referring to the absence of the word `Max` in the names, which is what makes the difference between correct and incorrect. `getDamageWobbleAngle` is just as incorrect as `getWobble`, without the `Max` in it. The clarification `Angle` and the qualifier `Damage` (where it's correct, sorry about that) are useful, of course.

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
```diff
@@ -14,7 +14,8 @@ CLASS axm net/minecraft/entity/vehicle/AbstractMinecartEntity
 		ARG 3 x
 		ARG 5 y
 		ARG 7 z
-	METHOD a setDamageWobbleStrength (F)V
+	METHOD a setWobbleMaxAngle (F)V
```
liach: This is a roll angle. Can we just call it a `roll`, which is more straightforward?

howtonotwin: I called it `angle` for consistency with the bubble column wobble, which is not a roll, pitch, or yaw (and therefore the only name I know for it is just `angle`). Also, what do you mean, exactly, by "call it a `roll`"? Do you mean `wobbleMaxRoll`? (Also, oops. Apparently I misremembered and called this a pitch before).

liach: This wobble is applied to the left-right roll of the minecart only, and this angle is only a roll. This is not a pitch (changed when going uphill/downhill) or a yaw (determines which direction minecart goes forward)

howtonotwin: If you don't have any qualms about the name, I'll go ahead and use it.

liach (approved):

Runemoro (approved):

PR #712, opened 2019-05-25
Item refactoring

ChloeDawn:
The entire package is covered for the most part, bar one or two methods, and cases where an Enigma bug is preventing mapping. Various changes have been made to base methods to better describe their purpose.

**Class name changes:**

- `BaseMapItem`, `isMap` -> `NetworkSyncedItem`, `isNetworkSynced`: See thread https://github.com/FabricMC/yarn/pull/712#discussion_r287575071
  - It is used for items that can send an update packet to the client on each player tick
- `FoodSetting(s)` -> `FoodComponent(s)`: See thread https://github.com/FabricMC/yarn/pull/712#discussion_r287574907
- `BaseBowItem` -> `RangedWeaponItem`
  - The base class shared by bows and crossbows that provides a projectile predication API

**Noteworthy changes in Item/ItemStack:**
- What was previously named "durability" is now "max damage". It represents the maximum amount of damage that can be applied to a stack before it breaks.
  - The inherent problem with describing the maximum amount of damage allowed as "durability" is that durability can easily imply "uses remaining", whereas in this context it describes the total amount of times an item can be damaged before it breaks, regardless of what the stacks current damage is.
- The concept of "amount" is now "count" as this is the Mojang name for this attribute of stacks.
- The method for increasing the stack's count is now `increment`, and for decreasing, `decrement`.
- What was previously "stack size" is now "max count". It represents the maximum count that an item can be stacked up to.


ChloeDawn:
~~I do not believe `ProjectileShootingItem` accurately represents the class, as it provides no actual shooting logic, only an API for querying held projectiles, and testing ItemStack's for projectile viability. I am open to better name suggestions.~~

ChloeDawn:
I strongly believe that `ItemPropertyGetter` is a wildly incorrect name, but my attempts at discussing it on Discord were fruitless, and responders either were not interested in discussion or believed the class name was fine. The API is intended for computing values that are then tested against expected minimum values declared in item model overrides. It is a client-side API bar the actual class file and the method for registering an implementation that is called from item constructors. The biggest issue I have with the current API naming is that it implies they are general un-sided item properties, but they are not.

However, as there was no luck in my attempts to discuss this, I have left the class name as-is, but have updated the accessors of the API in the Item class to better represent the name.

mappings/net/minecraft/item/Item.mapping
```diff
@@ -1,149 +1,182 @@
...
-	METHOD a onTagDeserialized (Lib;)Z
+		ARG 3 pos
+		ARG 4 miner
+	METHOD a postProcessTag (Lib;)Z
 		ARG 1 tag
-	METHOD a getProperty (Lqs;)Lbca;
-	METHOD a addProperty (Lqs;Lbca;)V
+	METHOD a getPropertyGetter (Lqs;)Lbca;
 		ARG 1 id
-	METHOD a matches (Lzb;)Z
```
Juuxel, line 112: Should `Block.matches` and `Fluid.matches` be also changed?

ChloeDawn: Yes, however, it would be out of scope for this pull request; I intend to cover blocks and fluids separately, to avoid a huge mess of changes in one ticket.

mappings/net/minecraft/item/CompassItem.mapping
```diff
@@ -1,5 +1,16 @@
...
 	CLASS baq$1
+		FIELD b angle D
+		FIELD c step D
+		FIELD d lastTick J
+		METHOD a getYaw (Latf;)D
+			ARG 1 entity
+		METHOD a getAngle (Lbhi;D)D
+			ARG 1 world
+			ARG 2 entityYaw
+		METHOD a getSpawnArctangent (Lbhj;Laii;)D
```
Gegy: Recall this was brought up on Discord, but not sure where that went. `getAngleToSpawn`?

mappings/net/minecraft/block/dispenser/ItemDispenserBehavior.mapping
```diff
@@ -1,11 +1,18 @@
 CLASS fu net/minecraft/block/dispenser/ItemDispenserBehavior
-	METHOD a dispenseItem (Lbhi;Lbcb;ILfa;Lfk;)V
+	METHOD a spawnItem (Lbhi;Lbcb;ILfa;Lfk;)V
```
Pyrofab, line 2: Is this change just to be in line with `spawnParticles` ? The method does dispense the item.

mappings/net/minecraft/item/CompassItem.mapping
ChloeDawn: Resolved by 54d9bb3

mappings/net/minecraft/item/CrossbowItem.mapping
```diff
@@ -1,16 +1,83 @@
...
+		ARG 1 world
+		ARG 2 entity
 	METHOD e getPullTime (Lbcb;)I
 		ARG 0 stack
-	METHOD j getChargedProjectiles (Lbcb;)Ljava/util/List;
+	METHOD j getProjectiles (Lbcb;)Ljava/util/List;
+		ARG 0 crossbow
 	METHOD k clearProjectiles (Lbcb;)V
+		ARG 0 crossbow
+	METHOD l getVelocityMultiplier (Lbcb;)F
```
Gegy: Would these `velocityMultiplier` names work as just `speed`?

mappings/net/minecraft/item/ArmorStandItem.mapping
```diff
@@ -2,5 +2,5 @@ CLASS azy net/minecraft/item/ArmorStandItem
 	METHOD <init> (Lbbw$a;)V
 		ARG 1 settings
 	METHOD a setRotations (Latd;Ljava/util/Random;)V
-		ARG 1 armorStand
+		ARG 1 entity
```
Pyrofab: Why would you change that ? The parameter is not any entity, it's an armor stand.

mappings/net/minecraft/item/ArrowItem.mapping
```diff
@@ -1,7 +1,7 @@
 CLASS azz net/minecraft/item/ArrowItem
 	METHOD <init> (Lbbw$a;)V
 		ARG 1 settings
-	METHOD a createProjectile (Lbhi;Lbcb;Lair;)Lawc;
+	METHOD a createArrow (Lbhi;Lbcb;Lair;)Lawc;
 		ARG 1 world
 		ARG 2 stack
-		ARG 3 livingShooter
+		ARG 3 entity
```
Pyrofab: Isn't that less descriptive than previously ?

mappings/net/minecraft/item/CrossbowItem.mapping
ChloeDawn: Possibly? It determines a multiplier that is applied to an existing projectile velocity. If you think speed is appropriate for that I will change it. See `ProjectileEntity#setVelocity`.

ChloeDawn:
```java
private static float getVelocityMultiplier(ItemStack stack) {
    if (stack.getItem() == Items.CROSSBOW && hasProjectile(stack, Items.FIREWORK_ROCKET)) {
        return 1.6f;
    }
    return 3.15f;
}
```
```java
new Vec3d(x, y, z).normalize().add(this.random.nextGaussian() * 0.007499999832361937 * trajectoryMultiplier, this.random.nextGaussian() * 0.007499999832361937 * trajectoryMultiplier, this.random.nextGaussian() * 0.007499999832361937 * trajectoryMultiplier).multiply((double)velocityMultiplier);
```

mappings/net/minecraft/item/Item.mapping
```diff
@@ -1,149 +1,182 @@
...
+		ARG 4 slot
 		ARG 5 selected
-	METHOD a onItemFinishedUsing (Lbcb;Lbhi;Lair;)Lbcb;
+	METHOD a finishUsing (Lbcb;Lbhi;Lair;)Lbcb;
 		ARG 1 stack
 		ARG 2 world
-		ARG 3 livingEntity
-	METHOD a onItemStopUsing (Lbcb;Lbhi;Lair;I)V
+		ARG 3 user
+	METHOD a onStoppedUsing (Lbcb;Lbhi;Lair;I)V
```
Gegy, line 75: Should this also drop the `on` prefix?

mappings/net/minecraft/item/BlockItem.mapping
```diff
@@ -1,29 +1,46 @@
...
+		ARG 2 state
+	METHOD a writeTagToBlockEntity (Lbhi;Lavy;Lev;Lbcb;)Z
 		ARG 0 world
 		ARG 1 player
 		ARG 2 pos
 		ARG 3 stack
 	METHOD a getPlaceSound (Lbvk;)Lyj;
-	METHOD a addProperty (Lbvk;Lbwn;Ljava/lang/String;)Lbvk;
+		ARG 1 state
+	METHOD a (Lbvk;Lbwn;Ljava/lang/Comparable;)Lbvk;
```
Pyrofab, line 18: Is this normal ?

mappings/net/minecraft/item/Item.mapping
ChloeDawn: I retained `on` prefix for event-based methods as it is the current standard in Yarn afaik. Would rather that change be a global migration rather than create inconsistencies now I think.

ChloeDawn: If you are referring to the differing name pattern for `finishUsing`, it is because that is not an event listener, it is a transformer that returns a "finished" stack. See `FoxEntity#tickMovement` when the fox eats food, for example.

mappings/net/minecraft/item/CrossbowItem.mapping
Gegy: I think `speed` works nicely here.
Also, for `trajectoryMultiplier`: could maybe be something like `inaccuracy` or `deviation`? Not sure, but the current name is fine too.

mappings/net/minecraft/item/Item.mapping
Gegy: Ah, yeah, had thought the prefix was being dropped in other places. 👍

mappings/net/minecraft/item/ComplexItem.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS bar net/minecraft/item/ComplexItem
```
Pyrofab: Unsure about this name. It's an item that can send a packet every tick when it is in a player's inventory.

mappings/net/minecraft/item/CrossbowItem.mapping
ChloeDawn: `inaccuracy` was considered (I think it is also the name MCP opted for), but I don't believe it represents the parameter well. The default value for full precision is `1.0`, and if I saw an "inaccuracy" I would expect `0.0` to be full precision. It is very clearly a multiplier imo.

mappings/net/minecraft/item/DefaultFoodSettings.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS aya net/minecraft/item/FoodItemSettings
+CLASS aya net/minecraft/item/DefaultFoodSettings
```
Pyrofab: Should we name `Items` `DefaultItems` as well ?

mappings/net/minecraft/item/ArmorStandItem.mapping
ChloeDawn: Ah, this was from my initial pass at parameter mappings, cleaning up with generalized names. Do you think `stand` would suffice within this context?

mappings/net/minecraft/item/ArrowItem.mapping
ChloeDawn: Another case of my initial generalized name pass. I will rename this to `shooter`.

mappings/net/minecraft/item/ArmorStandItem.mapping
Pyrofab: sure

mappings/net/minecraft/item/ArrowItem.mapping
ChloeDawn: Resolved by 07f735c

mappings/net/minecraft/item/DefaultFoodSettings.mapping
ChloeDawn: Good point. Maybe we should re-evaluate the overall class name. Do you think `FoodMetadata` could work? Then again.. the plural `FoodMetadatas` feels a little odd.

mappings/net/minecraft/block/dispenser/ItemDispenserBehavior.mapping
ChloeDawn: The semantics of the methods in that class are actually rather interesting. I agree that this is likely a poor name choice I've made, but I don't think `dispenseItem` would be a good choice either, as it is too ambiguous alongside `dispense` (deobfuscated name) with identical signature. The difference is this method does not produce a sound or particles. Maybe it should be `dispenseSilently`?

mappings/net/minecraft/item/ArmorStandItem.mapping
ChloeDawn: Resolved by 60cd923

mappings/net/minecraft/item/ComplexItem.mapping
ChloeDawn: The reasoning for this class name is listed in the initial ticket message. I am open to suggestions for a different name, though. I also agree it is fairly ambiguous. I could not think of an appropriate name to describe the behaviour at the time though, so opted for the Mojang name used in Bedrock.

Data provided by @kashike through Discord:
```
02a08829 T ComplexItem::ComplexItem(std::string const&, int)
02a08829 T ComplexItem::ComplexItem(std::string const&, int)
02a08845 T ComplexItem::~ComplexItem()
02a08841 T ComplexItem::~ComplexItem()
02a08841 T ComplexItem::~ComplexItem()
```

mappings/net/minecraft/item/CrossbowItem.mapping
gegy1000: Though, `1.0` wouldn't be full precision based on that usage?

mappings/net/minecraft/item/DefaultFoodSettings.mapping
Pyrofab: Good question. `FoodMetadata` seems alright, alternatives are `FoodInformation` and `FoodComponent`.

mappings/net/minecraft/item/BlockItem.mapping
ChloeDawn: It is a synthetic method with mapped parameter `value`:

```java
.map(value -> state.with(property, value))
```

mappings/net/minecraft/item/HorseArmorItem.mapping
```diff
@@ -1,9 +1,9 @@
 CLASS bbv net/minecraft/item/HorseArmorItem
 	FIELD a bonus I
-	FIELD b texture Ljava/lang/String;
+	FIELD b entityTexture Ljava/lang/String;
 	METHOD <init> (ILjava/lang/String;Lbbw$a;)V
 		ARG 1 bonus
 		ARG 2 name
 		ARG 3 settings
-	METHOD d getHorseArmorTexture ()Lqs;
+	METHOD d getEntityTexture ()Lqs;
```
Pyrofab, line 8: I feel like this could be more descriptive. Maybe emphasize that it is an "in world" texture ? Or that it is for when it is worn.

mappings/net/minecraft/item/CrossbowItem.mapping
Gegy: `1.0` wouldn't be full precision, though, would it? Maybe I'm missing something, I haven't looked into it at all.
(Oops, double posted. Somehow thought GitHub hadn't sent it lol)

mappings/net/minecraft/item/Item.mapping
```diff
@@ -1,149 +1,182 @@
...
+		ARG 1 slot
+	METHOD a isIn (Lbas;)Z
+		ARG 1 group
+	METHOD a appendStacks (Lbas;Lfj;)V
+		ARG 1 group
+		ARG 2 stacks
+	METHOD a getRawId (Lbbw;)I
 		ARG 0 item
-	METHOD a onEntityDamaged (Lbcb;Lair;Lair;)Z
+	METHOD a attack (Lbcb;Lair;Lair;)Z
```
Pyrofab: I don't think you can say a sword attacks a zombie, except as a figure of speech. Would `hurt` make more sense ?

mappings/net/minecraft/item/Item.mapping
```diff
@@ -1,149 +1,182 @@
...
-		ARG 3 target
+		ARG 2 user
+		ARG 3 entity
+		ARG 4 hand
 	METHOD a canRepair (Lbcb;Lbcb;)Z
-		ARG 1 tool
+		ARG 1 stack
 		ARG 2 ingredient
-	METHOD a onEntityTick (Lbcb;Lbhi;Laii;IZ)V
+	METHOD a tick (Lbcb;Lbhi;Laii;IZ)V
```
Pyrofab: Maybe this should emphasize that it happens only in a player inventory. A mistake seems too easy.

mappings/net/minecraft/item/ItemStack.mapping
```diff
@@ -2,132 +2,186 @@ CLASS bcb net/minecraft/item/ItemStack
...
+	FIELD e cooldown I
 	FIELD f item Lbbw;
 	FIELD g tag Lib;
 	FIELD h empty Z
-	FIELD i holdingItemFrame Latf;
-	FIELD j lastCheckedCanHarvestBlock Lbvo;
-	FIELD k lastCheckedCanHarvestResult Z
-	FIELD l lastCheckedCanPlaceBlock Lbvo;
-	FIELD m lastCheckedCanPlaceResult Z
+	FIELD i frame Latf;
```
Pyrofab, line 10: This should be at least `holdingFrame`. `frame` doesn't give any more information than the field type.

mappings/net/minecraft/item/ItemStack.mapping
```diff
@@ -2,132 +2,186 @@ CLASS bcb net/minecraft/item/ItemStack
...
-	METHOD B toTextComponent ()Ljm;
-	METHOD C getUpdateCooldown ()I
-	METHOD D getAmount ()I
+	METHOD B toHoverableText ()Ljm;
+	METHOD C getCooldown ()I
+	METHOD D getCount ()I
 	METHOD E isFood ()Z
 	METHOD F createModifierFormat ()Ljava/text/DecimalFormat;
-	METHOD G updateEmptyFlag ()V
+	METHOD G updateEmptyState ()V
```
Pyrofab, line 28: What was the issue with `flag` ?

mappings/net/minecraft/item/CrossbowItem.mapping
ChloeDawn: The following is appended to the velocity for a trajectory. A multiplier of `1.0` means no change:
`
.add(this.random.nextGaussian() * 0.008 * trajectoryMultiplier, this.random.nextGaussian() * 0.008 * trajectoryMultiplier, this.random.nextGaussian() * 0.008 * trajectoryMultiplier)
`

mappings/net/minecraft/item/SpawnEggItem.mapping
```diff
@@ -15,6 +15,6 @@ CLASS bcz net/minecraft/item/SpawnEggItem
 	METHOD a isOfSameEntityType (Lib;Laim;)Z
 		ARG 1 tag
 		ARG 2 type
-	METHOD b entityTypeFromTag (Lib;)Laim;
+	METHOD b getEntityType (Lib;)Laim;
 		ARG 1 tag
-	METHOD d iterator ()Ljava/lang/Iterable;
+	METHOD d getSpawnEggs ()Ljava/lang/Iterable;
```
Pyrofab: Much needed change. Although it is a static method, so it could be made simpler. Would something like `SpawnEggItem.getAll()` work ?

mappings/net/minecraft/item/CrossbowItem.mapping
Gegy: With `1.0` would it not add velocity up to `0.008` on each axis?

ChloeDawn: Oh.. Yeah, right... I'm bad at math. So, `inaccuracy` then?

Gegy: Yeah. Could also be like `inaccuracyMultiplier` or `inaccuracyFactor`.

mappings/net/minecraft/block/dispenser/ItemDispenserBehavior.mapping
ChloeDawn: Changed to `dispenseSilently` in 176d9bc, marking as resolved

mappings/net/minecraft/item/Item.mapping
ChloeDawn: Would `inventoryTick` be explicit enough?

Pyrofab: I think so

mappings/net/minecraft/item/ItemStack.mapping
ChloeDawn: The field is named relative to the verbosity of the encapsulation, which is descriptive enough within context imo. The methods would be called as such:
```java
itemStack.isInFrame();
itemStack.getFrame();
itemStack.setFrame(...);
```
Essentially, it is the item's frame (`ItemFrameEntity`).

mappings/net/minecraft/item/DefaultFoodSettings.mapping
ChloeDawn: I like `FoodComponent`, it better describes its use in Mojang's step towards a component system for behavioural attributes.

ChloeDawn: On the other hand, they have a hardcoded getter in Item... Not sure if `item.getFoodComponent()` is very nice, and `item.getFood()` would be too vague.

Pyrofab: That's what I was thinking. The tweet introducing this change did mention ECS.

mappings/net/minecraft/item/Item.mapping
ChloeDawn: Maybe `hit`? It handles stack damaging for the implementations in `MiningToolItem`, `HoeItem`, `SwordItem` and `TridentItem`, then returns true. The default return value in `Item` is `false`.

mappings/net/minecraft/item/DefaultFoodSettings.mapping
Pyrofab: I don't have a problem with `item.getFoodComponent()` but there may more fluent names

mappings/net/minecraft/item/Item.mapping
ChloeDawn: Resolved by a57ec18

mappings/net/minecraft/item/ItemStack.mapping
ChloeDawn: It updates a cached state used internally to avoid redundant checks. If anything it should be `updateCachedEmptyState` but it feels a little verbose for an internal method.

mappings/net/minecraft/item/SpawnEggItem.mapping
ChloeDawn: Resolved by cc5abf1

mappings/net/minecraft/item/ItemStack.mapping
Pyrofab: Fair enough

mappings/net/minecraft/item/ItemStack.mapping
Pyrofab: Dunno, I still think it would not hurt to be a little more verbose here.

mappings/net/minecraft/item/Item.mapping
Pyrofab: `hit` sounds good to me.

ChloeDawn: Resolved by bf7a33a, after a discussion on Discord

mappings/net/minecraft/item/ItemStack.mapping
ChloeDawn: Resolved until further notice through discussion on Discord
> InsomniaKitten: how else would an item's frame be interpreted

> Pyrofab: Fair

mappings/net/minecraft/item/HorseArmorItem.mapping
ChloeDawn: Resolved through brief discussion in Discord. It represents the entity texture of the horse armour and is reasonably descriptive in context. The real concern here is why the hell Mojang store their entity textures in an item class, set up via the constructor... Dammit, Mojang.

mappings/net/minecraft/item/DefaultFoodSettings.mapping
ChloeDawn: Resolved by 794bc6e. Feedback is welcome on the getter name in `Item`.

mappings/net/minecraft/item/ComplexItem.mapping
ChloeDawn: Resolved by d06eb13 after discussion in Discord

mappings/net/minecraft/item/CrossbowItem.mapping
ChloeDawn: Resolved by b4a92fd after discussion in Discord

modmuss50:
Just going to pick out some names that I think might need some more feedback on from others. Im currently unsure on them. These are very commonly used so I think getting more feedback is needed.

# ItemStack

- amount -> count

- addAmount -> increment 

- subtractAmount -> decrement 

`count` is the mojang names, but that is by no means a solid reason to use it.

mappings/net/minecraft/item/FoodComponent.mapping
```diff
@@ -1,37 +1,38 @@
-CLASS axz net/minecraft/item/FoodItemSetting
+CLASS axz net/minecraft/item/FoodComponent
```
modmuss50, line 1: This suffix conflicts with the recent chat changes. I think this is ok as the chat stuff needs more work.

ChloeDawn:
`increment` and `decrement` are not Mojang names, they were names discussed on Discord. What else would you be incrementing other than the count? It doesn't really need to be explicit.

modmuss50:
ah ok, I like `increment` and `decrement`, but im still unsure on count.

It could just be me, thats why im trying to get more feedback on them.

mappings/net/minecraft/item/Item.mapping
```diff
@@ -1,149 +1,182 @@
 CLASS bbw net/minecraft/item/Item
 	CLASS bbw$a Settings
-		FIELD a fullStackSize I
-		FIELD b durability I
+		FIELD a maxCount I
+		FIELD b maxDamage I
 		FIELD c recipeRemainder Lbbw;
-		FIELD d itemGroup Lbas;
+		FIELD d group Lbas;
```
modmuss50, line 6: Is just `group` a bit vague?

ChloeDawn: It's the group (`ItemGroup`) bound to the item. The getter is also `getGroup()`, so you would call as `item.getGroup()` returning an `ItemGroup`.

modmuss50: Ah yeah, thats a fair point.

grondag:
> ah ok, I like `increment` and `decrement`, but im still unsure on count.
> 
> It could just be me, thats why im trying to get more feedback on them.

I like count (and the other two.)

mappings/net/minecraft/item/ItemPropertyGetter.mapping
```diff
@@ -1,5 +1,5 @@
 CLASS bca net/minecraft/item/ItemPropertyGetter
-	METHOD call (Lbcb;Lbhi;Lair;)F
+	METHOD call call (Lbcb;Lbhi;Lair;)F
```
liach: Did enigma glitch here?

mappings/net/minecraft/item/ItemPlacementContext.mapping
```diff
@@ -1,15 +1,16 @@
 CLASS baf net/minecraft/item/ItemPlacementContext
-	FIELD a canReplaceHitBlock Z
-	FIELD g placedPos Lev;
+	FIELD a canReplaceExisting Z
```
liach, line 2: `existingReplaceable`? Consider a name that doesn't clash with the method name if possible.

liach (approved):
lgtm. Since you have named `FoodComponent`, we should consider moving the existing `Component` classes to `Text` instead.

mappings/net/minecraft/item/ItemPropertyGetter.mapping
ChloeDawn: Seems so, yes. This is a deobfuscated method and it has somehow generated a mapping for it...

ChloeDawn: ~~Is it safe to just edit this line in the mapping file to fix this @modmuss50?~~

mappings/net/minecraft/item/ItemPlacementContext.mapping
ChloeDawn: Why does it need to differ from the method? The field is private.

liach: It is fine. Just in case there is a setter, but there is not.

mappings/net/minecraft/item/ItemPropertyGetter.mapping
ChloeDawn: Resolved by 9efcc43

Pyrofab (approved):

modmuss50 (approved):
Im happy with this, its got some changes that will effect a lot of people, but the mappings migration tool exists.

Nice work.

modmuss50:
Im going to wait for another maintainer to approve this as well, but it seems good.

Sollace:
Based on Discord discussions, I'm going to suggest: `ProjectileShootingItem` -> `RangedWeaponItem`

> **Ranged weapon**
> A ranged weapon is any weapon that can engage targets beyond hand-to-hand distance, i.e. at distances greater than the physical reach of the weapon itself. It is sometimes also called projectile weapon or missile weapon because it typically works by launching projectiles, though technically a directed-energy weapon is also a ranged weapon. In contrast, a weapon intended to be used in hand-to-hand combat is called a melee weapon. -- Wikipedia

ChloeDawn:
> Based on Discord discussions, I'm going to suggest: `ProjectileShootingItem` -> `RangedWeaponItem`
> 
> > **Ranged weapon**
> > A ranged weapon is any weapon that can engage targets beyond hand-to-hand distance, i.e. at distances greater than the physical reach of the weapon itself. It is sometimes also called projectile weapon or missile weapon because it typically works by launching projectiles, though technically a directed-energy weapon is also a ranged weapon. In contrast, a weapon intended to be used in hand-to-hand combat is called a melee weapon. -- Wikipedia

Resolved by ed069d3

liach (approved):

PR #713, opened 2019-05-26
Inventory refactor

ChloeDawn:
I'm about to go to bed, so I won't be able to respond to feedback for a while... But there's no harm in me pushing this now anyway, I guess.


mappings/net/minecraft/inventory/Inventory.mapping
```diff
@@ -1,28 +1,28 @@
...
 		ARG 1 slot
-	METHOD b isValidInvStack (ILbcb;)Z
+	METHOD b isValidStack (ILbcb;)Z
 		ARG 1 slot
 		ARG 2 stack
-	METHOD b_ onInvClose (Lavy;)V
+	METHOD b_ onInventoryClosed (Lavy;)V
 		ARG 1 player
-	METHOD c isInvEmpty ()Z
+	METHOD c hasStacks ()Z
```
Draylar: Maybe I'm going crazy, but isn't this opposite (function returns true if it's empty)? I think isEmpty may be a better choice.

ChloeDawn: My god... Don't map code at 3am, kids. I will fix this tomorrow. Fml.

mappings/net/minecraft/inventory/Inventory.mapping
```diff
@@ -1,28 +1,28 @@
 CLASS agw net/minecraft/inventory/Inventory
-	METHOD K_ getInvMaxStackAmount ()I
-	METHOD N_ getInvSize ()I
-	METHOD a getInvStack (I)Lbcb;
+	METHOD K_ getSlotCapacity ()I
+	METHOD N_ getInventorySize ()I
+	METHOD a getStack (I)Lbcb;
 		ARG 1 slot
-	METHOD a takeInvStack (II)Lbcb;
+	METHOD a removeStack (II)Lbcb;
```
kgscialdone: This could just be me, but this change *might* not be the best idea. If I remember correctly, `takeInvStack` is only called when a player manually picks up a stack from the inventory, while the existing `removeInvStack` is used more generally for several different things (hoppers, etc.). It might be better to simply call this `takeStack`, removing the redundant `Inv`, in order to keep the distinction more clear.

ChloeDawn: From what I can tell, they have effectively the same semantics, with one just handling splitting logic for removing only a certain amount.

ChloeDawn: On further inspection, the semantics differ more than I thought. In `BasicInventory` for example, the splitting method also notifies change. Everything else seems identical though. Mutation-safe item removal.

ChloeDawn: Worth noting that *only* `BasicInventory` appears to notify change there among the default implementations. EDIT: Seemingly, CraftingInventory has a different approach to change notification.

mappings/net/minecraft/inventory/Inventory.mapping
```diff
@@ -1,6 +1,6 @@
 CLASS agw net/minecraft/inventory/Inventory
 	METHOD K_ getSlotCapacity ()I
-	METHOD N_ getInventorySize ()I
+	METHOD N_ getTotalSlots ()I
```
SquidDev: I feel `getSlotCount` is more reasonable (or just keep it as is!). `getTotalSlots` really doesn't seem that clear a name.

ChloeDawn: I opted to not use `getSlotCount` because I felt it could be easily confused with `getSlotCapacity` (which returns the max count an item can be stacked to within slots).

SquidDev: Yeah, I had that same problem with it. Honestly, just leave as is then - `getInventorySize` isn't perfect, but it's clearer than `getTotalSlots`.

ChloeDawn: Resolved by 7af4c0e

mappings/net/minecraft/inventory/SidedInventory.mapping
```diff
@@ -1,11 +1,11 @@
 CLASS ahm net/minecraft/inventory/SidedInventory
-	METHOD a canInsertInvStack (ILbcb;Lfa;)Z
+	METHOD a canAdd (ILbcb;Lfa;)Z
```
modmuss50: Any reason over canAdd and not `isValidStack` (same as normal inv), or `canAddStack`?

modmuss50:
Overall I think this looks good, just need to think if an inv is on a tile could a method clash with something else.

See #330 for the first PR of theses changes

mappings/net/minecraft/inventory/SidedInventory.mapping
kgscialdone: It's worth noting that this method has *zero* impact on whether a stack is valid for a slot. It is specifically and only used by vanilla for automation purposes, i.e. hoppers - calling it `isValidStack` would lead to significant confusion, since A. that name already exists on the superclass and serves a completely different purpose and B. it doesn't actually have anything to do with whether a stack is valid or not. It's more of a "can you access the slot you want from this side", really.

ChloeDawn: Lack of `Stack` suffix was an oversight on my part, I'll resolve this.

mappings/net/minecraft/inventory/DoubleInventory.mapping
```diff
@@ -1,4 +1,8 @@
 CLASS agv net/minecraft/inventory/DoubleInventory
 	FIELD a first Lagw;
 	FIELD b second Lagw;
-	METHOD a isPart (Lagw;)Z
+	METHOD <init> (Lagw;Lagw;)V
+		ARG 1 first
+		ARG 2 second
+	METHOD a contains (Lagw;)Z
```
Runemoro: `contains` is wrong, it should stay `isPart`. An inventory can't contain another inventory since an inventory isn't an item. Someone could also think that "contains" means "contains all the items that the other inventory contains".

mappings/net/minecraft/inventory/Inventory.mapping
```diff
@@ -1,28 +1,28 @@
...
+	METHOD a canUseContainer (Lavy;)Z
 		ARG 1 player
-	METHOD a getInvAmountOf (Lbbw;)I
+	METHOD a getTotalCount (Lbbw;)I
 		ARG 1 item
-	METHOD a containsAnyInInv (Ljava/util/Set;)Z
+	METHOD a anyMatch (Ljava/util/Set;)Z
 		ARG 1 items
-	METHOD a_ onInvOpen (Lavy;)V
+	METHOD a_ onContainerOpened (Lavy;)V
```
Runemoro: It should be `onOpened`, or `onInventoryOpened`. Some inventories don't have containers, like the player inventory.

mappings/net/minecraft/inventory/Inventory.mapping
```diff
@@ -1,28 +1,28 @@
...
 		ARG 1 slot
-	METHOD b isValidInvStack (ILbcb;)Z
+	METHOD b isValidStack (ILbcb;)Z
 		ARG 1 slot
 		ARG 2 stack
-	METHOD b_ onInvClose (Lavy;)V
+	METHOD b_ onContainerClosed (Lavy;)V
 		ARG 1 player
-	METHOD c isInvEmpty ()Z
+	METHOD c hasNoStacks ()Z
```
Runemoro, line 27: Why not just `isEmpty`?

mappings/net/minecraft/inventory/DoubleInventory.mapping
ChloeDawn: How about `includes`?

mappings/net/minecraft/inventory/Inventory.mapping
ChloeDawn: I could name it `isInventoryEmpty` if preferred. But `isEmpty` is too ambiguous e.g. `blockEntity.isEmpty()`.

Runemoro: It won't be `BlockEntity.isEmpty`, but rather `ChestBlockEntity.isEmpty`, `FurnaceBlockEntity.isEmpty`, etc., which are all clear.

mappings/net/minecraft/inventory/DoubleInventory.mapping
Runemoro: "includes X" makes me think even more of "contains everything that X contains". The two chest blocks (inventories) that make up a chest (double inventory) are called "parts" (parts of the same chest). What's wrong with `isPart`?

asiekierka:
Unfinished PR and easy enough to update, I'm leaving this open

mappings/net/minecraft/inventory/SidedInventory.mapping
```diff
@@ -1,11 +1,11 @@
 CLASS ahn net/minecraft/inventory/SidedInventory
-	METHOD a canInsertInvStack (ILbcc;Lfa;)Z
+	METHOD a canAddk (ILbcc;Lfa;)Z
```
Gegy, line 2: Typo here I think

PR #714, opened 2019-05-26
DefaultedList -> NonNullList

ChloeDawn:
The list implementation does not always have a default value. It has two perceived states.
- Fixed size, returning a default value for absent elements.
- Variable size, always returning an element, with no default defined.
Effectively, this implementation is just to ensure there is never a null return value, not to ensure there is a default value. The default value just happens to exist as a way to handle absent values in fixed size delegate lists.

The constructors have been updated to be more contextual too.

I'm not sure if the semantics of resizing should be more explicit in the constructors. Essentially, all but the no-args constructor produce a fixed-size list (Backed by an instance of `Arrays$ArrayList`).

### DIFF
--- a/mappings/net/minecraft/util/DefaultedList.mapping
+++ b/mappings/net/minecraft/util/DefaultedList.mapping
@@ -1,16 +1,23 @@
 CLASS fj net/minecraft/util/DefaultedList
-	FIELD a underlying Ljava/util/List;
+	FIELD a delegate Ljava/util/List;
 	FIELD b defaultValue Ljava/lang/Object;
 	METHOD <init> (Ljava/util/List;Ljava/lang/Object;)V
-		ARG 1 underlying
-	METHOD a create ()Lfj;
-	METHOD a create (ILjava/lang/Object;)Lfj;
+		ARG 1 delegate
+		ARG 2 defaultValue
+	METHOD a of ()Lfj;
+	METHOD a withExpectedSize (ILjava/lang/Object;)Lfj;
 		ARG 0 size
 		ARG 1 defaultValue
-	METHOD a create (Ljava/lang/Object;[Ljava/lang/Object;)Lfj;
+	METHOD a of (Ljava/lang/Object;[Ljava/lang/Object;)Lfj;
 		ARG 0 defaultValue
 		ARG 1 values
 	METHOD add (ILjava/lang/Object;)V
-		ARG 1 value
+		ARG 1 index
+		ARG 2 element
+	METHOD get (I)Ljava/lang/Object;
+		ARG 1 index
+	METHOD remove (I)Ljava/lang/Object;
+		ARG 1 index
 	METHOD set (ILjava/lang/Object;)Ljava/lang/Object;
 		ARG 1 index
+		ARG 2 element
--- a/mappings/net/minecraft/util/NonNullList.mapping
+++ b/mappings/net/minecraft/util/NonNullList.mapping
@@ -1,15 +1,15 @@
 CLASS fj net/minecraft/util/NonNullList
 	FIELD a delegate Ljava/util/List;
-	FIELD b defaultValue Ljava/lang/Object;
+	FIELD b fallback Ljava/lang/Object;
 	METHOD <init> (Ljava/util/List;Ljava/lang/Object;)V
 		ARG 1 delegate
-		ARG 2 defaultValue
+		ARG 2 fallback
 	METHOD a of ()Lfj;
 	METHOD a ofSize (ILjava/lang/Object;)Lfj;
 		ARG 0 size
-		ARG 1 defaultValue
+		ARG 1 fallback
 	METHOD a copyOf (Ljava/lang/Object;[Ljava/lang/Object;)Lfj;
-		ARG 0 defaultValue
+		ARG 0 fallback
 		ARG 1 values
 	METHOD add (ILjava/lang/Object;)V
 		ARG 1 index
--- a/mappings/net/minecraft/util/NonNullList.mapping
+++ b/mappings/net/minecraft/util/NonNullList.mapping
@@ -1,16 +1,16 @@
 CLASS fj net/minecraft/util/NonNullList
 	FIELD a delegate Ljava/util/List;
-	FIELD b fallback Ljava/lang/Object;
+	FIELD b initialElement Ljava/lang/Object;
 	METHOD <init> (Ljava/util/List;Ljava/lang/Object;)V
 		ARG 1 delegate
-		ARG 2 fallback
+		ARG 2 initialElement
 	METHOD a of ()Lfj;
 	METHOD a ofSize (ILjava/lang/Object;)Lfj;
 		ARG 0 size
-		ARG 1 fallback
+		ARG 1 initialElement
 	METHOD a copyOf (Ljava/lang/Object;[Ljava/lang/Object;)Lfj;
-		ARG 0 fallback
-		ARG 1 values
+		ARG 0 initialElement
+		ARG 1 elements
 	METHOD add (ILjava/lang/Object;)V
 		ARG 1 index
 		ARG 2 element
--- a/mappings/net/minecraft/util/NonNullList.mapping
+++ b/mappings/net/minecraft/util/NonNullList.mapping
@@ -1,14 +1,14 @@
-CLASS fj net/minecraft/util/DefaultedList
+CLASS fj net/minecraft/util/NonNullList
 	FIELD a delegate Ljava/util/List;
 	FIELD b defaultValue Ljava/lang/Object;
 	METHOD <init> (Ljava/util/List;Ljava/lang/Object;)V
 		ARG 1 delegate
 		ARG 2 defaultValue
 	METHOD a of ()Lfj;
-	METHOD a withExpectedSize (ILjava/lang/Object;)Lfj;
+	METHOD a ofSize (ILjava/lang/Object;)Lfj;
 		ARG 0 size
 		ARG 1 defaultValue
-	METHOD a of (Ljava/lang/Object;[Ljava/lang/Object;)Lfj;
+	METHOD a copyOf (Ljava/lang/Object;[Ljava/lang/Object;)Lfj;
 		ARG 0 defaultValue
 		ARG 1 values
 	METHOD add (ILjava/lang/Object;)V


asiekierka (approved):

mappings/net/minecraft/util/DefaultedList.mapping
```diff
@@ -1,16 +1,23 @@
...
-	METHOD a create ()Lfj;
-	METHOD a create (ILjava/lang/Object;)Lfj;
+		ARG 1 delegate
+		ARG 2 defaultValue
+	METHOD a of ()Lfj;
+	METHOD a withExpectedSize (ILjava/lang/Object;)Lfj;
 		ARG 0 size
 		ARG 1 defaultValue
-	METHOD a create (Ljava/lang/Object;[Ljava/lang/Object;)Lfj;
+	METHOD a of (Ljava/lang/Object;[Ljava/lang/Object;)Lfj;
```
InsomniaKitten: Should this be `copyOf`?

asiekierka:
Where did the NonNullList name come from and why use it over DefaultedList (if we have DefaultedRegistry)?

LemmaEOF:
NonNullList is voldemap.

Runemoro:
> NonNullList is voldemap.

It's not a problem if @InsomniaKitten wasn't aware of that. NonNullList is an obvious name for a list whose elements are non-null.

I'd prefer NonNullObjectList, though. It's not the list that isn't null, but rather it's a list of non-null objects.

modmuss50:
I have not got the code in front of me, and have a few question.

Can the default value be null?

Can null values be inserted?

LemmaEOF:
Yes and no.

~~*The `DefaultedList.create()` method validates that something's not null, but you can call a `new DefaultedList<>()` for a null default value.~~ UPDATE: actually, it looks like you *can* create a DefaultedList with a null default value. The `public static <E> DefaultedList<E> create(E object_1, E... objects_1)` just calls `Arrays.asList(object_1)` and is annotated with `@SafeVarargs`, so it's just fine to pass a null there. It only validates on creating a set-size DefaultedList and on adding/setting something to/in the list.

Player3324:
The class is quite odd.

clear() violates the j.u.Collection contract, it only sets all existing elements to the default value. This is the only location using the default value other than the pre-sizing static factory.

remove() however modifies the list as usual.

size(), isEmpty() include cleared/presized elements, but not remove()d ones

set() at end, add() grow the list as usual and reject null

get() is annotated as Nonnull

None of the indexing methods work outside the bounds, which somewhat conflicts with the default value slot theme.

A null default value is banned in one of the factory methods. Another always uses null as the default (empty collection) and the 3rd one accepts null for both default and contents. The protected constructors also use or permit null.

So overall it may be an attempt at a nonnull list, but the implementation isn't. The @Nonnull on get() is a lie. It is never fixed size, just presized. It always uses the default value, including null.

Overall DefaultedList is not wrong, NonnullList is technically wrong, but possibly in line with the intention. The implementation is definitely incorrect. I tend towards keeping the current name, but NonnullList or NonNullList is ok as well.

Nonnull vs NonNull is another debate, the annotation uses the former.

LemmaEOF:
would `PreSetList` be any better? (capitalizing the S in Set because it's a pre-set list, not a list of presets)

liach:
can you call the ofSize ofFixedSize to emphasize the fixed size of ju arrays arraylist; also consider renaming defaultValue to fallback?

ChloeDawn:
> Where did the NonNullList name come from and why use it over DefaultedList (if we have DefaultedRegistry)?

The name was devised through a discussion on Discord.
`DefaultedList` does not represent the semantics, as the list is not guaranteed to provide default values in-place of removed elements. Default values are only substituted when the list is fixed size, else you would just be given the next available element on a query for an index (as indices would shift on element removal).

ChloeDawn:
>Can the default value be null?

Yes, it can - but only in the case where the list is resizable, in which case a default value is unnecessary. For example, if you have a list of three elements, and remove the second element, querying the second element would return the former third element, as the list size is now 2 and indices have shifted.

>Can null values be inserted?

No, never. Both `add` and `set` verify the nullity of element arguments.

ChloeDawn:
> Nonnull vs NonNull is another debate, the annotation uses the former.

All but the jsr305 annotation typically use `NonNull`.

ChloeDawn:
>Overall DefaultedList is not wrong.

It is very much wrong, as it implies removed elements will always be substituted with a default value, which is not the case. This behaviour only occurs when the list is of a fixed size, which is not the case for the no-args constructor. All but the no-args constructor form a delegate using an `Arrays$ArrayList` which is a fixed size wrapper for object arrays. The no-args implementation uses a standard `ArrayList`.

mappings/net/minecraft/util/NonNullList.mapping
```diff
@@ -0,0 +1,23 @@
+CLASS fj net/minecraft/util/NonNullList
+	FIELD a delegate Ljava/util/List;
+	FIELD b defaultValue Ljava/lang/Object;
```
liach: `fallback`. Since this is no longer `DefaultedList`, a `defaultValue` appears weird.

mappings/net/minecraft/util/NonNullList.mapping
```diff
@@ -0,0 +1,23 @@
...
+	METHOD <init> (Ljava/util/List;Ljava/lang/Object;)V
+		ARG 1 delegate
+		ARG 2 defaultValue
+	METHOD a of ()Lfj;
+	METHOD a ofSize (ILjava/lang/Object;)Lfj;
```
liach, line 8: `ofFixedSize` to emphasize the fixed size

ChloeDawn: Personally, I would reserve this semantic to documentation, as this makes the constructor seem differing to the semantics of the array copying constructor, and both return fixed-size lists.

ChloeDawn: If anything we could probably make the no-args constructor imply a resizable list, which implicitly implies other constructors are fixed-size. But I still think it is a semantic best reserved for documentation to avoid verbose naming.

ChloeDawn:
>I'd prefer NonNullObjectList, though

We don't call lists `ObjectList`'s. Just because it is describing the element intrinsics does not mean it needs to be explicit in the type name. We don't have `ObjectHashEqualitySet` either.

ChloeDawn:
> The @nonnull on get() is a lie.

The class retains a contract in its implementation that a null value will *never* be returned. I studied this closely to be sure before making this proposal.

> It is never fixed size, just presized. 

For the "pre-sized" and array copying implementations, `Arrays$ArrayList` is used, which is a fixed-size wrapper for object arrays. It can never be resized, contractually nor semantically.

Pyrofab:
> It is very much wrong, as it implies removed elements will always be substituted with a default value, which is not the case.

Well, if the underlying list has a fixed size, `remove` flat out doesn't work, and if it doesn't, it arguably doesn't need to be replaced with the default.
I am personally not fond of `NonNullList` as not only is it a MCP name, it also gives a strong guarantee against nullability that the implementation does not actually fulfil (because of the third factory method).

ChloeDawn:
>Well, if the underlying list has a fixed size, `remove` flat out doesn't work


The `remove` functionality works fine. In a fixed-size collection that wraps an object array, a removal is replacing an element with a null reference. And this null is handled appropriately with the fallback value that you define during list construction.

>it also gives a strong guarantee against nullability that the implementation does not actually fulfil (because of the third factory method)

The list is guaranteed to never return a null value, the third factory method requires a non-null fallback value for this exact reason.

mappings/net/minecraft/util/NonNullList.mapping
ChloeDawn: Resolved by 20b995c

Player3324:
Right, I missed the bit about Arrays.asList.

It is however incorrect that it can't have nulls, the 3rd factory method or a subclass accessing the constructors don't prevent construction with nulls from the start (I tested it).

asiekierka:
I like the new names. Except for NonNullList, I still think it should be DefaultedList at least until the vararg thing is taken care of.

liach (approved):

modmuss50:
I think a good idea would be to use the current classname, with the new member names.

ChloeDawn:
I think I would prefer to wait for Mojang to fix the bug with their constructor if there is so much disagreement here on naming because of it.

asiekierka:
Okay then

liach:
From another perspective, what are the usages of this list? E.g. if this is used only for inventories, we can call it an `InventoryList`

Player3324:
I only checked fields, but it appears to be used exclusively for substituting null with empty ItemStack, specifically around inventories, slots and recipes.

The divided views on Defaulted vs NonNull make sense, one more or less implies the other. IMO Defaulted is closer to the behavior, while NonNull is more about implementation details.

Factory methods named of*() may be better reserved for immutable results, in line with Java9's, which isn't the case for this class.

My preference is currently: DefaultedList, delegate, defaultValue, create(), createdFixedSize(int,Object), copyOf(Object,Object[]).

ChloeDawn:
I would not expect a defaulted list to throw an exception on element removal when it is fixed size, or to disallow null values in the setter/adder (I would expect it to substitute nulls with the default, so I can call set(1, null) on a fixed-size to reset the element without needing to know the default element). I would also not expect a defaulted list to have a Nonnull annotated getter - what if you want a null default value? It is extremely clear that the varargs constructor has a poor implementation bug, it is *not* a feature of the API.

Player3324:
My interpretation is based on what was probably intended by the authors, not the mess that is actually there. Considering the very specific uses it makes sense that it is supposedly fixed size in some cases and consistently substituting even if not fully/properly implemented as such.

The "default" pov makes a lot more sense when considering what null actually is: the standard default value. This makes DefaultedList a default value substitution mechanism. Its inability to substitute with null with null doesn't matter, that'd be a no-op and can be provided by e.g. a regular ArrayList.

To add to my previous comment, I don't think NonNullList sufficiently describes the substituting behavior, which means that the class couldn't be expected to set/reset anything to a default value.

modmuss50 (changes requested):
Needs updating to pre2 before even thinking about merging, im still not sure on this PR.

asiekierka:
https://github.com/FabricMC/yarn/pull/769

PR #716, opened 2019-05-27
Refactor properties

ChloeDawn:
- Mapped all parameters
- Renamed `create` to `of`
- Renamed `IntegerProperty` to `IntProperty`
- Cleaned up constant names in `Properties`
- Renamed `getValueAsString` to `getName`
- Renamed `getValueClass` to `getValueType`


mappings/net/minecraft/state/property/IntProperty.mapping
```diff
@@ -0,0 +1,12 @@
+CLASS bwk net/minecraft/state/property/IntProperty
```
asiekierka, line 1: It's a property of Integer objects, not so much primitive ints.

mappings/net/minecraft/state/property/Property.mapping
```diff
@@ -1,6 +1,8 @@
 CLASS bwn net/minecraft/state/property/Property
 	METHOD a getName ()Ljava/lang/String;
-	METHOD a getValueAsString (Ljava/lang/Comparable;)Ljava/lang/String;
-	METHOD b getValueClass ()Ljava/lang/Class;
+	METHOD a getName (Ljava/lang/Comparable;)Ljava/lang/String;
```
asiekierka, line 3: That method is clearly for getting the name of a value, IIRC, and should not be confused with getName(). As such, I'm saying no to this rename

mappings/net/minecraft/state/property/Properties.mapping
```diff
@@ -2,23 +2,23 @@ CLASS bwb net/minecraft/state/property/Properties
...
 	FIELD C WATERLOGGED Lbwc;
-	FIELD D AXIS_XZ Lbwi;
-	FIELD E AXIS_XYZ Lbwi;
-	FIELD F UP_BOOL Lbwc;
-	FIELD G DOWN_BOOL Lbwc;
-	FIELD H NORTH_BOOL Lbwc;
-	FIELD I EAST_BOOL Lbwc;
-	FIELD J SOUTH_BOOL Lbwc;
-	FIELD K WEST_BOOL Lbwc;
+	FIELD D HORIZONTAL_AXIS Lbwi;
```
liach, line 5: Plural of `AXIS` is `AXES`, so `HORIZONTAL_AXES`. Same below.

mappings/net/minecraft/state/property/Properties.mapping
```diff
@@ -83,3 +83,7 @@ CLASS bwb net/minecraft/state/property/Properties
 	FIELD x SHORT Lbwc;
 	FIELD y SIGNAL_FIRE Lbwc;
 	FIELD z SNOWY Lbwc;
+	METHOD a (Lbwo;)Z
```
liach, line 86: what does this boolean return value mean?

mappings/net/minecraft/state/property/Property.mapping
ChloeDawn: `property.getName(value)` is perfectly clear as to its intent imo.

liach: `getSerializedName`? Referring to that `StringRepresentable` horror.

liach: Or `nameValue`?

mappings/net/minecraft/state/property/IntProperty.mapping
ChloeDawn: The constructor takes unboxed ints and every use of the property handles values as unboxed ints. This is a cleaner name, and the verbosity of Integer over Int is redundant imo.

mappings/net/minecraft/state/property/Property.mapping
ChloeDawn: Imo this method should be the same name as the property name getter. Both return a name. So if we rename this method the other should be reconsidered too. I personally think getName is fine though, anything else appended is just redundant verbosity.

mappings/net/minecraft/state/property/Properties.mapping
ChloeDawn: It represents a key for a singular value: `state.get(AXIS)`. If you argue that this should be plural you should argue that FACING should be FACINGS too. But it is incorrect naming for such a property imo.

mappings/net/minecraft/state/property/Properties.mapping
ChloeDawn: It is a synthetic method for a predicate.

mappings/net/minecraft/state/property/Properties.mapping
liach: nice new name! old name is confusing as i thought it was a swizzle. please mark this as resolved.

liach (approved):

PR #722, opened 2019-05-31
Few gui & containers changes

shedaniel:
- Mostly mapping gui & containers parameters
- RecipeBookButtonWidget -> TexturedButtonWidget
  - Why? Because RecipeBookButtonWidget is just a class for rendering custom textured buttons


liach (approved):

Prospector:
Not sure if there's a better name or not but just FYI, the textured buttons don't have an inactive state.

modmuss50 (approved):
Im happy with this.

PR #727, opened 2019-06-03
Changes SheepWoolEntityModel to SheepNakedBodyEntityModel

matjojo:
Since that is what this class actually is.

You can see this in action by grabbing this repo: https://github.com/matjojo/bigSheep and changing some values in the body of the `BigSheepWoolEntityModel`, loading up the game, summoning a big sheep, and seeing that it's naked body is out of place (Depending on what you changed, comments are provided.)

Now I only changed this one classname since that seems to be enough to be clear about what this class is, but if there are any other classes that take their name from here those should probably also be changed. I couldn't find any.

### DIFF
--- a/mappings/net/minecraft/client/render/entity/model/SheepNakedBodyModel.mapping
+++ b/mappings/net/minecraft/client/render/entity/model/SheepNakedBodyModel.mapping
@@ -1,0 +1,1 @@
+CLASS dfy net/minecraft/client/render/entity/model/SheepNakedBodyModel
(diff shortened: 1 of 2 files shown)


ChloeDawn:
Technically all sheep are naked, they wear no clothing... This terminology is rather silly (and/or indecent) for this kind of environment, imo. A better and more accurate name would be **`ShearedSheepEntityModel`**.
###### Also, you have the naming pattern reversed, it should've been `NakedSheepBodyEntityModel`.

asiekierka:
Since it's a one-line fix, please rebase when 1.14.3-pre1 comes along

matjojo:
> This terminology is rather silly (and/or indecent) for this kind of environment, imo.

This is definitely a translation mistake then, that is very normal over here. Anyways. I made a mistake rebasing and had to make a new local repo. So the new name, now changed to: `SheepShearedEntityModel` is now here: #730 

Also insomnia, the naming pattern for these classes are mostly <entityname><featurename>'EntityModel', so I don't think that changing it to 'ShearedSheepBodyEntityModel is the best idea.
See `SkullEntityModel` and `SkullOverlayEntityModel` as well as `VillagerResemblingModel`.

PR #730, opened 2019-06-03
Changes SheepWoolEntityModel to SheepShearedEntityModel

matjojo:
Since that is what this model actually is.

Due to a mistake by me this change was first in "https://github.com/FabricMC/yarn/pull/727", sorry about that.

The considerations made there where:
>Technically all sheep are naked, they wear no clothing... This terminology is rather silly (and/or indecent) for this kind of environment, imo. A better and more accurate name would be ShearedSheepEntityModel.
> Also, you have the naming pattern reversed, it should've been NakedSheepBodyEntityModel.

>> This terminology is rather silly (and/or indecent) for this kind of environment, imo.

>This is definitely a translation mistake then, that is very normal over here. So the name, now changed to: `SheepShearedEntityModel`. 

>Also insomnia, the naming pattern for these classes are mostly \<entityname\>\<featurename\>'EntityModel', so I don't think that changing it to 'ShearedSheepBodyEntityModel is the best idea.
See `SkullEntityModel` and `SkullOverlayEntityModel` as well as `VillagerResemblingModel`.

I still feel like this is saying that this class is the complete sheep when Sheared, but it's a step forwards from saying that it's wool.

### DIFF
--- a/mappings/net/minecraft/client/render/entity/model/ShearedSheepEntityModel.mapping
+++ b/mappings/net/minecraft/client/render/entity/model/ShearedSheepEntityModel.mapping
@@ -1,0 +1,1 @@
+CLASS dfz net/minecraft/client/render/entity/model/ShearedSheepEntityModel
--- a/mappings/net/minecraft/client/render/entity/model/ShearedSheepEntityModel.mapping
+++ b/mappings/net/minecraft/client/render/entity/model/ShearedSheepEntityModel.mapping
@@ -1,1 +1,0 @@
-CLASS dfz net/minecraft/client/render/entity/model/ShearedSheepEntityModel
--- a/mappings/net/minecraft/client/render/entity/model/ShearedSheepSkinModel.mapping
+++ b/mappings/net/minecraft/client/render/entity/model/ShearedSheepSkinModel.mapping
@@ -1,1 +1,0 @@
-CLASS dfz net/minecraft/client/render/entity/model/ShearedSheepSkinModel
--- a/mappings/net/minecraft/client/render/entity/model/ShearedSheepSkinModel.mapping
+++ b/mappings/net/minecraft/client/render/entity/model/ShearedSheepSkinModel.mapping
@@ -1,0 +1,1 @@
+CLASS dfz net/minecraft/client/render/entity/model/ShearedSheepSkinModel
--- a/mappings/net/minecraft/client/render/entity/model/SheepEntityModel.mapping
+++ b/mappings/net/minecraft/client/render/entity/model/SheepEntityModel.mapping
@@ -1,1 +1,1 @@
-CLASS dfy net/minecraft/client/render/entity/model/SheepEntityModel
+CLASS dfz net/minecraft/client/render/entity/model/SheepEntityModel
--- a/mappings/net/minecraft/client/render/entity/model/SheepShearedEntityModel.mapping
+++ b/mappings/net/minecraft/client/render/entity/model/SheepShearedEntityModel.mapping
@@ -1,0 +1,1 @@
+CLASS dfz net/minecraft/client/render/entity/model/SheepShearedEntityModel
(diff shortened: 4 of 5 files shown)


matjojo:
I just took at look at that failed build, and it's a network timeout, got no idea how to run it again though

Runemoro:
`ShearedSheep`, not `SheepSheered`. We call it a "sheared sheep entity model", not a "sheep sheared entity model". The second would imply we have a single `ShearedEntity` used by all sheared entities.

matjojo:
There seems to be no general order that the current mappings take, so I agree with changing it to the order that makes more sense when reading. So `ShearedSheepEntityModel` it is.

One consideration to make is that this is purely about the 'body' of the sheep, and has nothing to do with the legs. I don't see an easy way to add that into the name though. So if no-one else has a better idea I feel like this describes the class well enough.

Runemoro:
> One consideration to make is that this is purely about the 'body' of the sheep, and has nothing to do with the legs.

What about `ShearedSheepSkinModel`?

Is there also a model for only the wool?

matjojo:
No, the wool is in the sheep model. `ShearedSheepSkinModel` is a good idea. It displays the 'sheep' and 'sheared' parts and the fact that it is not the whole sheep but only the skin. I'll push that change and think that that should be it if there are no other concerns.

Runemoro:
> No, the wool is in the sheep model

But then if the legs are in the same model as the wool and this model doesn't include the legs, how do you display a sheared sheep with legs?

matjojo:
> how do you display a sheared sheep with legs?

Yeah that is the core of why the name change was needed, this model is really only the shaven part of the sheep without legs. So the legs of the real sheep are still used when this model is 'appended' to the sheep without it's wool.

In fact this model is always rendered, even if the wool is also being rendered.

Here an image: https://i.imgur.com/fYmjhYy.png

Where is arrow is pointed at is what is currently called the `SheepWoolEntityModel` moved 10 units to the side for visibility.

You can see the difference between the sheared one and the one with wool. The wool around the body, legs and face are gone, but these have nothing to do with this model, except, and this is where it seems to get crazy. This is an image of what is currently called the `SheepWoolEntityModel`'s head made smaller.:https://i.imgur.com/okjw2Pp.png Sheared and not sheared. It seems to me like this model is really all the non sheared bits except for the legs, which seem to be done in the sheep itself. This is the constructor for this model, where you can see that it only defines a head and a body.

```
   public SheepWoolEntityModel() {
      super(12, 0.0F);
      this.head = new Cuboid(this, 0, 0);
      this.head.addBox(-3.0F, -4.0F, -6.0F, 6, 6, 8, 0.0F);
      // this is the head part that I made smaller in the second image
      this.head.setRotationPoint(0.0F, 6.0F, -8.0F);
      this.body = new Cuboid(this, 28, 8);
      this.body.addBox(-4.0F, -10.0F, -7.0F, 8, 16, 6, 0.0F);
      // this is the skin part that I moved in the first image.
      this.body.setRotationPoint(0.0F, 5.0F, 2.0F);
   }
```

Runemoro:
I just looked at the code, and `SheepWoolEntityModel` actually extends `QuadrupedEntityModel` which provides the legs. The legs "defined" in `SheepEntityModel` are actually redefined to be shorter and wider and they're not actually legs, but wool.

So the names are just inverted. `SheepWoolEntityModel` should be `SheepEntityModel`, and `SheepEntityModel` should be `SheepWoolEntityModel`.

matjojo:
Ahh, so that `QuadrupedEntityModel` is the framework, which the wool is laid over, but the wool is currently called sheep, and the sheep is currently called wool.

matjojo:
So as far as I know this can be merged now right? Or is there anything else?

Runemoro (approved):

asiekierka:
Done on -pre2

PR #732, opened 2019-06-03
getPlayerFacing -> getPlayerHorizontalFacing

modmuss50:
Undo a recent change to getPlayerHorizontalFacing as the horizontal bit is very important.

### DIFF
--- a/mappings/net/minecraft/item/ItemUsageContext.mapping
+++ b/mappings/net/minecraft/item/ItemUsageContext.mapping
@@ -15,7 +15,7 @@ CLASS bdn net/minecraft/item/ItemUsageContext
 		ARG 4 stack
 		ARG 5 hit
 	METHOD a getBlockPos ()Lev;
-	METHOD f getPlayerFacing ()Lfa;
+	METHOD f getPlayerHorizontalFacing ()Lfa;
 	METHOD g isPlayerSneaking ()Z
 	METHOD h getPlayerYaw ()F
 	METHOD i getSide ()Lfa;


Prospector (approved):

Runemoro:
I disagree. The direction of the player is the direction in which the body is pointing (while standing), not the head. A player can be facing north yet looking up, but a player can't be facing upwards. So "horizontal" is redundant.

I suggest changing "facing" to "direction" though. What's a "facing"? I'm assuming it's taken from the F3 screen, but that correctly uses facing as a verb, not a noun ("Facing north").

modmuss50:
There could well be room for more changes with this, as it just returns `player.getHorizontalFacing` when the player is not null.

ChloeDawn:
`getPlayerFacing` is fine imo. It does not conflict with any of the other names, and the intent is clear. If I want to know which way the player is facing, I would expect a horizontal direction. A player cannot *face* up/down, they only *look* up/down, which is what `ItemPlacementContext#getPlayerLookDirection` is for.

Prospector:
I also think it is fine to not specify horizontal, however I think it should say direction instead of facing, `getPlayerDirection`

ChloeDawn:
`getPlayerFacing` follows the current naming of `getHorizontal`**`Facing`**`()` in `Entity`. If you believe "direction" would better describe it, it should be a full migration. Until then, this method should refer to it as a "facing".

Prospector:
personally I think Direction implies horizontal a bit better (more associated with N/S/E/W on maps) and makes more sense than facing. Facing isn't a noun in the direction sense, it's a verb.

modmuss50:
We can leave this as-is then.

Issue #715, opened 2019-05-27
EntitySize -> EntityDimensions


Sollace:
EntityDimensions and all the fields there-in are indirectly confirmed by Mojang through string constants in their code:

```
   public String toString() {
      return "EntityDimensions w=" + this.width + ", h=" + this.height + ", fixed=" + this.constant;
   }
```

EntitySize -> EntityDimensions
EntitySize.constant -> EntityDimensions.constant

In addition to that, I would also change these two methods to be more in line with what they seem to be doing:

EntitySize.resizeable(...) -> EntityDimensions.changing(...)
EntitySize.constant(...) -> EntityDimensions.fixed(...)

ChloeDawn:
It was decided after discussion that EntitySize better represents this data class. I believe constant is fine too, representing an immutable size. Imo fixed doesn't really imply that well.

Pyrofab:
I am personally not fond of having class names differ from the string representation, especially when the name is not particularly bad. It's just confusing to have blatant conflicting information.

liach:
this mojang is indeed a bad obe. we have dimensions for worlds already! but the rename from constant to fixed i support.

liach:
on a side note, i initially mapped this class before mojang overrode the toString method. we should add to the guideline about whether we are mapping to mojang names (which are sometimes incorrect or off) or correct names.

Runemoro:
EntitySize is wrong. [The word "size" means only one value, and using it to refer to the dimensions of an object is bad English.](https://www.google.com/search?q=size+definition)

This class represents the *dimensions* of an entity (width, height), not the *size* of the entity (big, small). In fact, I think we already have another `getEntitySize` method somewhere (in the slime entity?) that returns a single number.

Prospector:
This would be a good issue to get a PR going for

liach:
`EntityDimensions` should be good as long as we call it `Dimensions` instead of `Dimension`. See the second definition at https://www.dictionary.com/browse/dimension

Runemoro:
Yes, `Dimensions`, since it's a pair of two dimensions (width and height).

Prospector:
EntityDimensions is literally what it says in the issue name and string...

Sollace:
If everyone is on the side of doing a PR I can get one going later today.

Abd yes, I mean to say `Dimensions` not `Dimension`.

Prospector:
Yup, go for it

Sollace:
[...]

PR #733, opened 2019-06-04
Add naming conventions

Runemoro:
[Click for rendered view](https://github.com/Runemoro/yarn/blob/conventions/CONVENTIONS.md)

Adds a `CONVENTIONS.md` file which lists some naming conventions that should be followed when naming things.

There is some overlap with section 4 of https://github.com/FabricMC/yarn/pull/477, but the two documents have very different purposes (this one lists conventions that should be used in yarn names, the other lists more general information about how to contribute).

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -2,23 +2,34 @@
 
 ## General
 
-Use `UpperCamelCase` for class names. Use `lowerCamelCase` for method names, variable names, and names of fields that are not static and final. Always use `UPPER_SNAKE_CASE` for names of fields that are both static and final.
+Use `UpperCamelCase` for class names. Use `lowerCamelCase` for method names, variable names, and names of fields that are not static and
+final. Always use `UPPER_SNAKE_CASE` for names of fields that are both static and final.
 
-Method names should always be verbal phrases (`tick`, `getCarversForStep`). Non-boolean field and variable names and class names should be noun phrases (`ChunkRegion`, `color`). Boolean field and variable names should always be adjectival phrases or present-tense verbal phrases (`powered`, `canOpen`), avoiding the `is` and `has` prefixes when possible (`colored`, not `isColored` or `hasColor`).
+Method names should always be verbal phrases (`tick`, `getCarversForStep`). Non-boolean field and variable names and class names should
+be noun phrases (`ChunkRegion`, `color`). Boolean field and variable names should always be adjectival phrases or present-tense verbal
+phrases (`powered`, `canOpen`), avoiding the `is` and `has` prefixes when possible (`colored`, not `isColored` or `hasColor`).
 
-To make code as easy to read as possible, keep names in the natural language order. For example, a class representing a "chest block entity" should be named `ChestBlockEntity`, not `BlockEntityChest`. Though prefix naming may be helpful for grouping classes together in an IDE's tree view, reading and writing code is done much more often than browsing files.
+To make code as easy to read as possible, keep names in the natural language order. For example, a class representing a "chest block
+entity" should be named `ChestBlockEntity`, not `BlockEntityChest`. Though prefix naming may be helpful for grouping classes together in
+an IDE's tree view, reading and writing code is done much more often than browsing files.
 
 Use American English (`color`, not `colour`) for consistency throughout Yarn and with known Mojang names.
 
-Omit words that are redundant because of method parameter or the owner class's name. For example, use `getChunk(BlockPos pos)` rather than `getChunkAtPosition(BlockPos pos)`, and `create` rather than `createAchievement` for a method is in the `Achievement` class. Don't avoid overloading methods or shadowing fields.
+Omit words that are redundant because of method parameter or the owner class's name. For example, use `getChunk(BlockPos pos)` rather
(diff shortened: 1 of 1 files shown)


CONVENTIONS.md
```diff
@@ -0,0 +1,118 @@
...
+ - "init" for "initialize"
+ - Any abbreviations used by Java or libraries ("json", "html", etc.)
+
+Treat acronyms as single words rather than capitalising every letter. This improves readability (compare `JsonObject` and
+`JSONObject`) and is consistent with Mojang naming (a known name is `NbtIo`).
+
+## Mojang names
+
+Always use names that match names in strings in the vanilla code, unless that string is outdated or inaccurate. This avoids
+confusion, especially from new modders who may not understand what an class exception message is referring to.
```
liach: Nah, mojang names are usually confusing and should be considered separately. e.g. EntitySize was added before mojang added a toString that says it is called `EntityDimension`, which is confusing since `Dimension` is already present.

Prospector:
Personally, I think it would be good to always use Mojang names, but others seem to disagree. I don't blame them, but I feel like even if they're bad, we should use the Mojang ones because they always could be deobfuscated one day.

CONVENTIONS.md
Runemoro: @liach `EntityDimensions` is a much better name than than entity size. As I [explained in a comment on that issue](https://github.com/FabricMC/yarn/issues/715#issuecomment-498521419), size and dimension don't mean the same thing: a size is a single value telling you how big something is overall, while the dimensions are how big something is in each direction.

In any case, the point of this section isn't to always use Mojang names for everything (this is why I say "unless that string is outdated or inaccurate"), but rather to use terms that are more likely to be understood by everyone. For example, if Mojang calls something an "entity goal", don't call it an "ai target", or players will get confused when they get an exception message mentioning goals.

I'll edit this section and make this more clear, and mention that names should also match the terms commonly used by Minecraft players and modders.

CONVENTIONS.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Abbreviations
+
+Avoid abbreviations unless it's a common one everyone knows and other yarn names involving the same word use its abbreviated
+form. Full names are easier to read quickly and remember ("Which words were abbreviated?") and they often don't take more
+time to type thanks to IDE autocompletion. Common abbreviations you should use are:
+
+ - "id" for "identifier"
+ - "pos" for "position"
+ - "NBT" for "named binary tag"
```
liach: `"nbt"`

Runemoro: I'm just giving the English acronym we use for "named binary tag". In code, it should be `nbt`, as described below ("Treat acronyms as single words rather than capitalising every letter.")

Prospector: I think it's a little confusing to have it capitalized here tbh. Should just show lowercase there too

CONVENTIONS.md
```diff
@@ -0,0 +1,125 @@
+# Naming conventions
+
+## General
+
+Use `UpperCamelCase` for class names. Use `lowerCamelCase` for method names, variable names, and names of fields that are not
+both static and final. Use `UPPER_SNAKE_CASE` for names of fields that are both static and final.
+
+Method names should always be verb phrases (`tick`, `getCarversForStep`), except for builder methods and "withX" methods.
```
Pyrofab: "should always" -> "should generally". Just in this document, you cite 4 exceptions (builders, "withX", "fromX", "toX"), and that's forgetting fluent naming strategies like `of()`.

CONVENTIONS.md
Runemoro: Done

CONVENTIONS.md
Runemoro: Done

CONVENTIONS.md
```diff
@@ -0,0 +1,124 @@
...
+way.
+
+There are however three exceptions to this rule:
+ - Use "world" for what Mojang calls "level" (see https://github.com/FabricMC/yarn/issues/89)
+ - Use "container" for what Mojang calls "menu" (see https://github.com/FabricMC/yarn/issues/386)
+ - Use "inventory" for what Mojang calls "container" (no issue yet, requires renaming "container" first)
+
+## Things to avoid
+
+Don't naming methods based on implementation details. Names should describe what methods do, now how they work.
```
JamiesWhiteShirt: Did you mean "Don't name methods based on..."?

Runemoro: Yes, fixed.

JamiesWhiteShirt (approved):

Player3324:
Conciseness: I'd add something about making the desired verbosity depend on how niche a feature is.  An element used all over the place benefits more from brevity, while obscure parts of the code are better described more exhaustively.

Getters: This should specifically describe what to do with boolean fields, they often use isFieldName getter names.

Abstract is - contrary to I and Enum prefixes -  commonly used in the Java standard library. IMO it's better than alternatives like "Base" or <nothing> and should also be used for classes that don't work standalone or shouldn't be used as such, even if they don't carry the abstract modifier (behavior vs actual implementation).

Prospector:
I agree with everything Player just said

Player3324 (approved):

modmuss50 (approved):
looks good. Will merge

PR #737, opened 2019-06-05
Text refactoring + related mappings

ChloeDawn:


mappings/net/minecraft/Formatting.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS c net/minecraft/ChatFormat
+CLASS c net/minecraft/Formatting
```
mattmess1221: Should this go in the `text` package as well?

Prospector:
I'm honestly still for using Mojang names and don't like changing known ones. However, if people really want this, I suppose I can tolerate it. It's just gonna be annoying if Mojang ever deobfs a field or method relating to this at some point. 

I'm also not sure about changing a bunch of methods that were previously all getTextComponent to a mix of toText, getText, getName, and more.

mattmess1221:
If you really wanted Mojang's names, it would be using the original name from when the concept was revealed in 1.7 (ChatComponent). The same goes for Identifier from 1.6.

Prospector:
The names of the component classes are literally in the strings. The current ones are 90% accurate, i believe. Think there's one that's a bit off.

Pyrofab:
> I'm also not sure about changing a bunch of methods that were previously all getTextComponent to a mix of toText, getText, getName, and more.

It's like you had a bunch of methods named `getString`. The type is not giving you any information, that's why the method name is here.

mappings/net/minecraft/Formatting.mapping
ChloeDawn: It is used by more than text, as implicit toString within concatenation treats it as a literal Unicode formatting. Maybe it should be in the `util` package though, I overlooked that.

ChloeDawn:
Worth noting that the method for exposing the Text object as a String is `asString` now. I considered `getContents` but the implications of it conflicted with designs of the subclasses of Text. I am open to suggestions for a better name. I'm personally fine with it, although without support for Javadocs in Yarn, there is a lack of distinction between `getString`, `asString`, and `toString` (Yes, the class has all three of these, yikes).

`getString` is essentially a deep asString, that returns a concatenated `asString` result of all siblings. `asString` returns the String value of the object the receiver represents, ignoring siblings. `toString` represents the object as a named data structure.

Side-note, we cannot rename `getString`, it is a method in [Brigadier's Message API](https://github.com/Mojang/brigadier/blob/559d8f39727e98d374b0726ad88aed9832beee4e/src/main/java/com/mojang/brigadier/Message.java#L7).

mappings/net/minecraft/advancement/Advancement.mapping
```diff
@@ -53,4 +53,4 @@ CLASS q net/minecraft/advancement/Advancement
 	METHOD g getRequirementCount ()I
 	METHOD h getId ()Lqs;
 	METHOD i getRequirements ()[[Ljava/lang/String;
-	METHOD j getTextComponent ()Ljm;
+	METHOD j toText ()Ljm;
```
liach: `toHoverableText`

mappings/net/minecraft/scoreboard/ScoreboardObjective.mapping
```diff
@@ -1,19 +1,24 @@
 CLASS csv net/minecraft/scoreboard/ScoreboardObjective
 	FIELD a scoreboard Lcsy;
-	FIELD b name Ljava/lang/String;
+	FIELD b id Ljava/lang/String;
 	FIELD c criterion Lctb;
-	FIELD d displayName Ljm;
+	FIELD d name Ljm;
```
liach: Retain `name` and `displayName`. The scoreboard command's usage is [`scoreboard objectives add <name> <criteria> [display name...]`](https://minecraft.gamepedia.com/Scoreboard#Command_reference)

mappings/net/minecraft/scoreboard/ScoreboardObjective.mapping
```diff
@@ -1,19 +1,24 @@
...
+	METHOD a setName (Ljm;)V
+		ARG 1 name
+	METHOD a (Lju;)V
+		ARG 1 style
+	METHOD b getId ()Ljava/lang/String;
 	METHOD c getCriterion ()Lctb;
-	METHOD d getDisplayName ()Ljm;
-	METHOD e getTextComponent ()Ljm;
+	METHOD d getName ()Ljm;
+	METHOD e getHoverableName ()Ljm;
```
liach: `toHoverableText` for consistency.

mappings/net/minecraft/text/ClickEvent.mapping
```diff
@@ -0,0 +1,23 @@
+CLASS jl net/minecraft/text/ClickEvent
+	CLASS jl$a Action
+		FIELD f BY_NAME Ljava/util/Map;
+		FIELD g userDefinable Z
```
liach, line 4: I recall this field actually means if this action can be read from a network input; e.g. opening a local file does not have `true` for this field

mappings/net/minecraft/text/HoverEvent.mapping
```diff
@@ -0,0 +1,23 @@
+CLASS jp net/minecraft/text/HoverEvent
+	CLASS jp$a Action
+		FIELD d BY_NAME Ljava/util/Map;
+		FIELD e userDefinable Z
```
liach, line 4: Same here.

mappings/net/minecraft/text/KeybindText.mapping
```diff
@@ -0,0 +1,11 @@
+CLASS jq net/minecraft/text/KeybindText
+	FIELD b i18n Ljava/util/function/Function;
+	FIELD c key Ljava/lang/String;
+	FIELD d name Ljava/util/function/Supplier;
```
liach, line 4: Why is this a supplier?

mappings/net/minecraft/text/NbtText.mapping
```diff
@@ -0,0 +1,61 @@
+CLASS jr net/minecraft/text/NbtText
+	CLASS jr$a BlockPosNbtText
```
liach: This is actually getting nbt from a block entity, so I doubt the `BlockPos` name.

mappings/net/minecraft/text/NbtText.mapping
```diff
@@ -0,0 +1,61 @@
...
+			ARG 1 nbt
+			ARG 2 interpret
+			ARG 3 rawSelector
+		METHOD b getSelector (Ljava/lang/String;)Lec;
+			ARG 0 rawSelector
+		METHOD equals (Ljava/lang/Object;)Z
+			ARG 1 o
+		METHOD k getSelector ()Ljava/lang/String;
+	FIELD b rawJson Z
+	FIELD c nbt Ljava/lang/String;
```
liach: Mojang name is a bit confusing here... This is actually the raw path.

mappings/net/minecraft/text/NbtText.mapping
```diff
@@ -0,0 +1,61 @@
...
+		ARG 3 text
+	METHOD a (Lib;)Ljava/util/stream/Stream;
+		ARG 1 nbt
+	METHOD a (Ljm;Ljm;)Ljm;
+		ARG 0 a
+		ARG 1 b
+	METHOD b getPath (Ljava/lang/String;)Lcr$h;
+		ARG 0 nbt
+	METHOD i getNbt ()Ljava/lang/String;
+	METHOD j isRawJson ()Z
```
liach: `shouldInterpret`

mappings/net/minecraft/text/NbtText.mapping
```diff
@@ -0,0 +1,61 @@
...
+	METHOD a toNbt (Lcd;)Ljava/util/stream/Stream;
+		ARG 1 source
+	METHOD a (Lcd;Laij;ILjava/lang/String;)Ljava/util/stream/Stream;
+		ARG 3 text
+	METHOD a (Lib;)Ljava/util/stream/Stream;
+		ARG 1 nbt
+	METHOD a (Ljm;Ljm;)Ljm;
+		ARG 0 a
+		ARG 1 b
+	METHOD b getPath (Ljava/lang/String;)Lcr$h;
```
liach: Is this a static method? If yes, appears like a raw path -> path parser.

mappings/net/minecraft/text/NbtText.mapping
```diff
@@ -0,0 +1,61 @@
...
+			ARG 1 o
+		METHOD k getSelector ()Ljava/lang/String;
+	FIELD b rawJson Z
+	FIELD c nbt Ljava/lang/String;
+	FIELD d path Lcr$h;
+	FIELD e LOGGER Lorg/apache/logging/log4j/Logger;
+	METHOD <init> (Ljava/lang/String;Lcr$h;Z)V
+		ARG 1 nbt
+		ARG 2 path
+		ARG 3 rawJson
```
liach: `interpret`

mappings/net/minecraft/text/Style.mapping
```diff
@@ -1,12 +1,12 @@
-CLASS ju net/minecraft/network/chat/Style
+CLASS ju net/minecraft/text/Style
 	CLASS ju$a Serializer
 		METHOD deserialize (Lcom/google/gson/JsonElement;Ljava/lang/reflect/Type;Lcom/google/gson/JsonDeserializationContext;)Ljava/lang/Object;
-			ARG 1 functionJson
-			ARG 2 unused
+			ARG 1 style
```
liach: `json`?

mappings/net/minecraft/text/Style.mapping
```diff
@@ -21,27 +21,37 @@ CLASS ju net/minecraft/network/chat/Style
...
+		ARG 1 obj
 	METHOD f isObfuscated ()Z
 	METHOD g isEmpty ()Z
 	METHOD h getClickEvent ()Ljl;
 	METHOD i getHoverEvent ()Ljp;
 	METHOD j getInsertion ()Ljava/lang/String;
-	METHOD k getFormatString ()Ljava/lang/String;
-	METHOD m clone ()Lju;
+	METHOD k asString ()Ljava/lang/String;
+	METHOD m deepCopy ()Lju;
```
liach, line 55: :heart:

mappings/net/minecraft/text/Text.mapping
```diff
@@ -0,0 +1,57 @@
+CLASS jm net/minecraft/text/Text
+	CLASS jm$a Serializer
+		FIELD a GSON Lcom/google/gson/Gson;
+		FIELD b JSONREADER_POS Ljava/lang/reflect/Field;
```
liach: `JSON_READER_POS`; same below

mappings/net/minecraft/text/TranslatableText.mapping
```diff
@@ -0,0 +1,21 @@
+CLASS jw net/minecraft/text/TranslatableText
+	FIELD b translation Ljava/util/List;
+	FIELD c ARG_FORMAT Ljava/util/regex/Pattern;
+	FIELD d EMPTY_LANGUAGE Lhw;
+	FIELD e LANGUAGE Lhw;
+	FIELD f key Ljava/lang/String;
+	FIELD g args [Ljava/lang/Object;
+	FIELD h lock Ljava/lang/Object;
+	FIELD i languageTimeLoaded J
```
liach: What is this? `lastLanguageUpdateTimestamp`?

mappings/net/minecraft/text/TranslatableText.mapping
```diff
@@ -0,0 +1,21 @@
...
+	METHOD <init> (Ljava/lang/String;[Ljava/lang/Object;)V
+		ARG 1 key
+		ARG 2 args
+	METHOD b getArgument (I)Ljm;
```
liach: `getArg` for consistency with `args`/`getArgs`.

mappings/net/minecraft/world/Difficulty.mapping
```diff
@@ -4,10 +4,11 @@ CLASS ahb net/minecraft/world/Difficulty
...
 	METHOD <init> (Ljava/lang/String;IILjava/lang/String;)V
 		ARG 3 id
+		ARG 4 translationKey
 	METHOD a getId ()I
 	METHOD a getDifficulty (I)Lahb;
 		ARG 0 difficulty
 	METHOD a getDifficulty (Ljava/lang/String;)Lahb;
 		ARG 0 translationKey
-	METHOD b toTextComponent ()Ljm;
+	METHOD b getName ()Ljm;
```
liach: Consider unifying `getTranslatbleName`, `getName`, and `toHoverableText`/`toText`

mappings/net/minecraft/text/KeybindText.mapping
ChloeDawn: It's a computer for the name. `i18n` creates a localized string supplier, the supplier localizes the keycode on each invocation.

mappings/net/minecraft/text/HoverEvent.mapping
ChloeDawn: It determines if the action can be deserialized from arbitrary JSON

mappings/net/minecraft/text/ClickEvent.mapping
ChloeDawn: It determines if the action can be deserialized from arbitrary JSON

liach (approved):
Love it!

Gegy (approved):

modmuss50 (approved):
Looks good after taking a quick, look we are updating and everyone else seems to be happy with this so ill merge it.

jamierocks:
I must protest this PR being merged, reverting a recent change with 0 reasoning and giving a 2 day window for opinions to be collected...

Text is a **bad** name for Fabric, or just Minecraft mappings in general.

1. Text, or Chat, is a bad name for describing what they are - components of a chat message. Text makes it seem very singular, and not a chain of components as they are.
2. Text is a lingering name from MCP. Fabric isn't about going with the *incorrect* names that have been propagated by our competitors (what's next, the return of `TileEntity`?).
3. It conflicts with Mojang's own notion of Text, or as it is called here `LiteralText`. This causes all sorts of *avoidable* problems: Mojang use the 'chat component' naming publically, making an appearance in plenty of changelogs; the wiki will use different names.

liach:
@jamierocks Given we have `FoodComponent`, in my opinion, the `LiteralText`'s mojang own notion thing is comparatively trivial. I'd say `JsonText` is even better than `Component` (as that's closer to the minecraft wiki)

as far as I know, mcp did not use "text" name.

PR #740, opened 2019-06-07
AbsoluteHand -> Arm

ChloeDawn:
It represents an arm, used for changing the arm angles or models, and used for determining the arm that the main hand is bound to.

### DIFF
--- a/mappings/net/minecraft/client/render/FirstPersonRenderer.mapping
+++ b/mappings/net/minecraft/client/render/FirstPersonRenderer.mapping
@@ -31,6 +31,8 @@ CLASS dkp net/minecraft/client/render/FirstPersonRenderer
 		ARG 3 item
 	METHOD a resetEquipProgress (Lahf;)V
 	METHOD a renderArm (Lait;)V
+	METHOD a (Lait;F)V
+		ARG 1 side
 	METHOD a renderItem (Laiu;Lbce;Ldlx$b;)V
 		ARG 1 holder
 		ARG 2 stack
--- a/mappings/net/minecraft/client/options/GameOptions.mapping
+++ b/mappings/net/minecraft/client/options/GameOptions.mapping
@@ -85,7 +85,7 @@ CLASS cvr net/minecraft/client/options/GameOptions
 	FIELD n hideServerAddress Z
 	FIELD o advancedItemTooltips Z
 	FIELD p pauseOnLostFocus Z
-	FIELD q mainHandSide Lait;
+	FIELD q mainArm Lait;
 	FIELD r overrideWidth I
 	FIELD s overrideHeight I
 	FIELD t heldItemTooltips Z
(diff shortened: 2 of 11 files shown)


Runemoro:
It used to be `MainHand` and I renamed it. It's used in many other places other than the main hand option, such as armor stands.

It's absolute because we already have a `Hand` enum relative to the main hand.

ChloeDawn:
That is meaningless reasoning and still doesn't explain what "absolute" means at all in this context. Anyhow, I have further reviewed the code, and realise `MainHand` is too ambiguous and implies wrong intent, as you may have felt too. A better name would be `HandSide`, which I am now putting forward as the change proposal. Related mappings have been updated for this too.

mappings/net/minecraft/client/render/FirstPersonRenderer.mapping
```diff
@@ -31,6 +31,8 @@ CLASS dkp net/minecraft/client/render/FirstPersonRenderer
 		ARG 3 item
 	METHOD a resetEquipProgress (Lahf;)V
 	METHOD a renderArm (Lait;)V
+	METHOD a (Lait;F)V
```
ChloeDawn, line 34: Just noting that I have not studied this method specifically, and have therefore not mapped it fully. I merely mapped the relating parameter.

kashike (approved):

Runemoro:
"Hand side"?? To me a `HandSide` would be an enum containing `PALM` and `BACK`...

mattmess1221:
It makes sense to me. There's the phrase "left-hand side" and "right-hand side."

Gegy (approved):

asiekierka:
I think the general agreement is on HandSide?

Player3324:
I agree with Runemoro, it's too strongly implying to be about the side of a hand, not a side the hand may be located at.

The translation string implies this is exclusively about the main hand and "HandSide" is trivially confused with "Hand". Maybe something like MainHandPosition works better and separates this niche use class further from the other "Hand" class. Edit: This suggestion is a bad idea because "AbsoluteHand" is also being used by entity rendering to reference the left/right arm/hand.

ChloeDawn:
After much further review, I have concluded that the ideal name for this class would really be `Arm`. It represents the arm that the main hand or off hand resides on, and is also used as an argument when positioning the arms of models.

`model.setArmAngle(90.0F, Arm.LEFT) // sets the angle of the left arm to 90`
`livingEntity.[getMainArm() or getMainHandArm()] // returns the arm that the main hand is on`

`// in GameOptions, the Arm that the client player's main hand is on`
`public Arm mainHand = Arm.RIGHT;`

Runemoro:
Using arm for one and hand for the other just seems like a trick to avoid naming two classes the same (the names can be inverted and they would still be correct).

I think we should use a meaningful name that describes what this class does different. Maybe this one should be Hand, and the other one should be renamed to HandType, or RelativeHand?

Pyrofab:
The other should not be touched, it is the reference Hand used throughout the code. This one can afford being more verbose.

liach:
Handedness?

ChloeDawn:
I will be reopening this as a new ticket for a fresh slate on reviews and approvals

PR #741, opened 2019-06-07
Mapped (almost) all the missing methods and fields in Entity

Sollace:
`EntityS2CPacket.a(D)J` -> `encodePacketCoordinate`
`EntityS2CPacket.a(JJJ)` -> `decodePacketCoordinate`

These are both being used when translating between actual world coordinates and the coordinate format used by the packets and by extension the entity tracker too.

`ac` -> `trackedXPosition`
`ad` -> `trackedYPosition`
`ae` -> `trackedZPosition`
All values used by the entity tracker. Their values are set (added to) in `ClientPlayNetworkHandler` and come from `packet.getDeltaXShort`, `packet.getDeltaYShort`, and `packet.getDeltaZShort` respectively.

`U` -> `firstUpdate`
Initialized to true and set to false at the bottom of `Entity.baseUpdate`. It's pretty straight forward.

`D` -> `previousHorizontalSpeed`
`E` -> `horizontalSpeed`
`F` -> `distanceWalked`
`av` -> `nextStepDistance`
`aw` -> `nextAerialStepDistance`
These seem to be used to track stepping sounds when walking on the ground. `horizontalSpeed` is moved to `previousHorizontalSpeed` on every tick, and recalculated using the already-known horizontal speed method. They are also used with `distanceWalked` when determining when to play the footstep sounds.

`i` -> `inanimate`
Stays false most of the time and is set to true by boats, end crystals, armour stands, falling block entities, tnt, and minecarts. Used in one of the predicates for `EntityView.intersectsEntities`

`R` -> `wet`
`m()V` -> `updateInWater()V`
`n()V` -> `updateWetState ()V`
`wet` is set to true by `updateWetState` if an entity is in a water/fluid (state) block, and checked against by `Entity.isInWater`

`updateWetState` is the first method called by `updateInWater` 

`Y` -> `updateNeeded`
Set to true when an entity is added to the world and used all over the place by chunks when they're ticked. Also shortcuts some checks in the riding update functions, too.

`aF` -> `teleportRequested`
The teleport functions set this to true, and it gets reset atomically by the server when checking if an entity has requested a teleport.

`T` -> `regenTimer`
Used in `ClientPlayerEntity.updateHealth`.
This is set to 10 or 20 after health is set, and is decremented every tick in `LivingEntity.baseTick`

`setPositionAndRotations ` -> `updateClientPositionAndAngles`
`a` -> `updateClientHeadRotation`
This is a client-only method, and called in response to position/rotation change packets.
Ditto for `updateClientHeadRotation` which takes the yaw and pitch and just sets head Yaw.

`getBoundingBoxMarginForTargeting` -> `getTargettingDistance`

I think that is the most notable changes I can think of right now from reading the diff.

Note to self: really need to mark these as I do them, rather than after.

### DIFF
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -606,7 +606,7 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD u copyPositionAndRotation (Lail;)V
 		ARG 1 entity
 	METHOD u_ initDataTracker ()V
-	METHOD v initNewDimension (Lail;)V
+	METHOD v copyFrom (Lail;)V
 		ARG 1 original
 	METHOD w hasPassenger (Lail;)Z
 		ARG 1 passenger
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -237,7 +237,7 @@ CLASS ail net/minecraft/entity/Entity
 		ARG 1 observerBounds
 		ARG 2 world
 		ARG 3 subject
-	METHOD a calculateTangentalMotionVector (Lcsa;Lcrv;Lbho;Lcsf;Lzx;)Lcsa;
+	METHOD a calculateTangentialMotionVector (Lcsa;Lcrv;Lbho;Lcsf;Lzx;)Lcsa;
 		ARG 0 collisionVector
 		ARG 1 observerBounds
 		ARG 2 world
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -18,7 +18,7 @@ CLASS ail net/minecraft/entity/Entity
 	FIELD Q waterHeight D
 	FIELD R inWater Z
 	FIELD S inLava Z
-	FIELD T regenTimer I
+	FIELD T timeUntilRegen I
 	FIELD U firstUpdate Z
 	FIELD V dataTracker Lql;
 	FIELD W FLAGS Lqi;
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -301,7 +301,7 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD aO getMountedHeightOffset ()D
 	METHOD aP isLiving ()Z
 	METHOD aQ removeAllPassengers ()V
-	METHOD aR getTargetingDistance ()F
+	METHOD aR getTargetingMargin ()F
 	METHOD aS getRotationVector ()Lcsa;
 	METHOD aT getRotationClient ()Lcrz;
 	METHOD aU getRotationVecClient ()Lcsa;
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -301,7 +301,7 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD aO getMountedHeightOffset ()D
 	METHOD aP isLiving ()Z
 	METHOD aQ removeAllPassengers ()V
-	METHOD aR getTargettingDistance ()F
+	METHOD aR getTargetingDistance ()F
 	METHOD aS getRotationVector ()Lcsa;
 	METHOD aT getRotationClient ()Lcrz;
 	METHOD aU getRotationVecClient ()Lcsa;
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -324,7 +324,6 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD ai getSplashSound ()Lym;
 	METHOD aj getHighSpeedSplashSound ()Lym;
 	METHOD ak checkBlockCollision ()V
-	METHOD al playsFlyingSounds ()Z
 	METHOD am isSilent ()Z
 	METHOD an hasNoGravity ()Z
 	METHOD ao canClimb ()Z
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -606,7 +606,7 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD u copyPositionAndRotation (Lail;)V
 		ARG 1 entity
 	METHOD u_ initDataTracker ()V
-	METHOD v initNewDimention (Lail;)V
+	METHOD v initNewDimension (Lail;)V
 		ARG 1 original
 	METHOD w hasPassenger (Lail;)Z
 		ARG 1 passenger
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -407,7 +407,7 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD bU getPrimaryPassenger ()Lail;
 	METHOD bV getPassengerList ()Ljava/util/List;
 	METHOD bW getPassengersDeep ()Ljava/util/Collection;
-	METHOD bX hasPlayerMount ()Z
+	METHOD bX hasRider ()Z
 	METHOD bY getRootVehicle ()Lail;
 	METHOD bZ isLogicalSideForUpdatingMovement ()Z
 	METHOD ba getItemsEquipped ()Ljava/lang/Iterable;
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -37,9 +37,9 @@ CLASS ail net/minecraft/entity/Entity
 	FIELD aJ standingEyeHeight F
 	FIELD aa chunkY I
 	FIELD ab chunkZ I
-	FIELD ac trackedXPosition J
-	FIELD ad trackedYPosition J
-	FIELD ae trackedZPosition J
+	FIELD ac trackedX J
+	FIELD ad trackedY J
+	FIELD ae trackedZ J
 	FIELD af ignoreCameraFrustum Z
 	FIELD ag velocityDirty Z
 	FIELD ah portalCooldown I
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -406,7 +406,7 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD bU getPrimaryPassenger ()Lail;
 	METHOD bV getPassengerList ()Ljava/util/List;
 	METHOD bW getPassengersDeep ()Ljava/util/Collection;
-	METHOD bX hasRider ()Z
+	METHOD bX hasPlayerRider ()Z
 	METHOD bY getRootVehicle ()Lail;
 	METHOD bZ isLogicalSideForUpdatingMovement ()Z
 	METHOD ba getItemsEquipped ()Ljava/lang/Iterable;
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -407,7 +407,7 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD bU getPrimaryPassenger ()Lail;
 	METHOD bV getPassengerList ()Ljava/util/List;
 	METHOD bW getPassengersDeep ()Ljava/util/Collection;
-	METHOD bX hasPlayerJockey ()Z
+	METHOD bX hasPlayerMount ()Z
 	METHOD bY getPrimaryVehicle ()Lail;
 	METHOD bZ isLogicalSideForUpdatingMovement ()Z
 	METHOD ba getItemsEquipped ()Ljava/lang/Iterable;
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -2,7 +2,7 @@ CLASS ail net/minecraft/entity/Entity
 	FIELD A velocityModified Z
 	FIELD B movementMultiplier Lcsa;
 	FIELD C removed Z
-	FIELD D previousHorizontalSpeed F
+	FIELD D prevHorizontalSpeed F
 	FIELD E horizontalSpeed F
 	FIELD F distanceWalked F
 	FIELD G fallDistance F
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -16,7 +16,7 @@ CLASS ail net/minecraft/entity/Entity
 	FIELD O age I
 	FIELD P insideWater Z
 	FIELD Q waterHeight D
-	FIELD R wet Z
+	FIELD R inWater Z
 	FIELD S inLava Z
 	FIELD T regenTimer I
 	FIELD U firstUpdate Z
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -139,7 +139,7 @@ CLASS ail net/minecraft/entity/Entity
 		ARG 2 pitch
 	METHOD a updateTrackedHeadRotation (FI)V
 		ARG 1 yaw
-		ARG 2 pitch
+		ARG 2 interpolationSteps
 	METHOD a updateVelocity (FLcsa;)V
 		ARG 1 speed
 		ARG 2 movementInput
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -147,18 +147,18 @@ CLASS ail net/minecraft/entity/Entity
 		ARG 1 source
 		ARG 2 amount
 	METHOD a (Laij;)Ljava/util/stream/Stream;
-		ARG 1 collissionVector
+		ARG 1 collisionVector
 	METHOD a updateKilledAdvancementCriterion (Lail;ILahu;)V
 		ARG 1 killer
 		ARG 2 score
 		ARG 3 damageSource
 	METHOD a calculateMotionVector (Lail;Lcsa;Lcrv;Lbhl;Lcsf;Lzx;)Lcsa;
 		ARG 0 subject
-		ARG 1 collissionVector
+		ARG 1 collisionVector
 		ARG 2 observerBounds
 		ARG 3 world
 		ARG 4 context
-		ARG 5 collissions
+		ARG 5 collisions
 	METHOD a startRiding (Lail;Z)Z
 		ARG 1 entity
 		ARG 2 force
@@ -238,11 +238,11 @@ CLASS ail net/minecraft/entity/Entity
 		ARG 2 world
 		ARG 3 subject
 	METHOD a calculateTangentalMotionVector (Lcsa;Lcrv;Lbho;Lcsf;Lzx;)Lcsa;
-		ARG 0 collissionVector
+		ARG 0 collisionVector
 		ARG 1 observerBounds
 		ARG 2 world
 		ARG 3 context
-		ARG 4 collissions
+		ARG 4 collisions
 	METHOD a isTeamPlayer (Lctc;)Z
 		ARG 1 team
 	METHOD a populateCrashReport (Le;)V
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -117,7 +117,7 @@ CLASS ail net/minecraft/entity/Entity
 		ARG 5 z
 		ARG 7 yaw
 		ARG 8 pitch
-	METHOD a updateClientPositionAndAngles (DDDFFIZ)V
+	METHOD a updateTrackedPositionAndAngles (DDDFFIZ)V
 		ARG 1 x
 		ARG 3 y
 		ARG 5 z
@@ -137,7 +137,7 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD a setRotation (FF)V
 		ARG 1 yaw
 		ARG 2 pitch
-	METHOD a updateClientHeadRotation (FI)V
+	METHOD a updateTrackedHeadRotation (FI)V
 		ARG 1 yaw
 		ARG 2 pitch
 	METHOD a updateVelocity (FLcsa;)V
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -408,7 +408,7 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD bV getPassengerList ()Ljava/util/List;
 	METHOD bW getPassengersDeep ()Ljava/util/Collection;
 	METHOD bX hasPlayerMount ()Z
-	METHOD bY getPrimaryVehicle ()Lail;
+	METHOD bY getRootVehicle ()Lail;
 	METHOD bZ isLogicalSideForUpdatingMovement ()Z
 	METHOD ba getItemsEquipped ()Ljava/lang/Iterable;
 	METHOD bb isOnFire ()Z
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -2,6 +2,9 @@ CLASS ail net/minecraft/entity/Entity
 	FIELD A velocityModified Z
 	FIELD B movementMultiplier Lcsa;
 	FIELD C removed Z
+	FIELD D previousHorizontalSpeed F
+	FIELD E horizontalSpeed F
+	FIELD F distanceWalked F
 	FIELD G fallDistance F
 	FIELD H prevRenderX D
 	FIELD I prevRenderY D
@@ -13,22 +16,30 @@ CLASS ail net/minecraft/entity/Entity
 	FIELD O age I
 	FIELD P insideWater Z
 	FIELD Q waterHeight D
+	FIELD R wet Z
 	FIELD S inLava Z
+	FIELD T regenTimer I
+	FIELD U firstUpdate Z
 	FIELD V dataTracker Lql;
 	FIELD W FLAGS Lqi;
 	FIELD X POSE Lqi;
+	FIELD Y updateNeeded Z
 	FIELD Z chunkX I
 	FIELD aA NAME_VISIBLE Lqi;
 	FIELD aB SILENT Lqi;
 	FIELD aC NO_GRAVITY Lqi;
 	FIELD aD invulnerable Z
 	FIELD aE scoreboardTags Ljava/util/Set;
+	FIELD aF teleportRequested Z
 	FIELD aG pistonMovementDelta [D
 	FIELD aH pistonMovementTick J
 	FIELD aI size Laim;
 	FIELD aJ standingEyeHeight F
 	FIELD aa chunkY I
 	FIELD ab chunkZ I
+	FIELD ac trackedXPosition J
+	FIELD ad trackedYPosition J
+	FIELD ae trackedZPosition J
 	FIELD af ignoreCameraFrustum Z
 	FIELD ag velocityDirty Z
 	FIELD ah portalCooldown I
@@ -36,13 +47,17 @@ CLASS ail net/minecraft/entity/Entity
 	FIELD aj portalTime I
 	FIELD ak dimension Lbyh;
 	FIELD al lastPortalPosition Lev;
+	FIELD am lastPortalDirectionVector Lcsa;
+	FIELD an lastPortalDirection Lfa;
 	FIELD ao uuid Ljava/util/UUID;
 	FIELD ap uuidString Ljava/lang/String;
 	FIELD aq glowing Z
 	FIELD ar passengerList Ljava/util/List;
 	FIELD as vehicle Lail;
 	FIELD at velocity Lcsa;
-	FIELD au boundingBox Lcrv;
+	FIELD au entityBounds Lcrv;
+	FIELD av nextStepDistance F
+	FIELD aw nextAerialStepDistance F
 	FIELD ax fireTime I
 	FIELD ay BREATH Lqi;
 	FIELD az CUSTOM_NAME Lqi;
@@ -53,6 +68,7 @@ CLASS ail net/minecraft/entity/Entity
 	FIELD f type Laip;
 	FIELD g entityId I
 	FIELD h LOGGER Lorg/apache/logging/log4j/Logger;
+	FIELD i inanimate Z
 	FIELD j ridingCooldown I
 	FIELD k teleporting Z
 	FIELD l world Lbhl;
@@ -92,21 +108,23 @@ CLASS ail net/minecraft/entity/Entity
 		ARG 1 cursorDeltaX
 		ARG 3 cursorDeltaY
 	METHOD a requestTeleport (DDD)V
-		ARG 1 x
-		ARG 3 y
-		ARG 5 z
+		ARG 1 destX
+		ARG 3 destY
+		ARG 5 destZ
 	METHOD a setPositionAnglesAndUpdate (DDDFF)V
 		ARG 1 x
 		ARG 3 y
 		ARG 5 z
 		ARG 7 yaw
 		ARG 8 pitch
-	METHOD a setPositionAndRotations (DDDFFIZ)V
+	METHOD a updateClientPositionAndAngles (DDDFFIZ)V
 		ARG 1 x
 		ARG 3 y
 		ARG 5 z
 		ARG 7 yaw
 		ARG 8 pitch
+		ARG 9 interpolationSteps
+		ARG 10 interpolate
 	METHOD a rayTrace (DFZ)Lcry;
 		ARG 1 maxDistance
 		ARG 3 tickDelta
@@ -119,18 +137,31 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD a setRotation (FF)V
 		ARG 1 yaw
 		ARG 2 pitch
+	METHOD a updateClientHeadRotation (FI)V
+		ARG 1 yaw
+		ARG 2 pitch
 	METHOD a updateVelocity (FLcsa;)V
 		ARG 1 speed
 		ARG 2 movementInput
 	METHOD a damage (Lahu;F)Z
 		ARG 1 source
 		ARG 2 amount
+	METHOD a (Laij;)Ljava/util/stream/Stream;
+		ARG 1 collissionVector
 	METHOD a updateKilledAdvancementCriterion (Lail;ILahu;)V
 		ARG 1 killer
 		ARG 2 score
 		ARG 3 damageSource
+	METHOD a calculateMotionVector (Lail;Lcsa;Lcrv;Lbhl;Lcsf;Lzx;)Lcsa;
+		ARG 0 subject
+		ARG 1 collissionVector
+		ARG 2 observerBounds
+		ARG 3 world
+		ARG 4 context
+		ARG 5 collissions
 	METHOD a startRiding (Lail;Z)Z
 		ARG 1 entity
+		ARG 2 force
 	METHOD a setEquippedStack (Laiq;Lbce;)V
 		ARG 1 slot
 		ARG 2 stack
@@ -192,6 +223,8 @@ CLASS ail net/minecraft/entity/Entity
 		ARG 2 target
 	METHOD a setBoundingBox (Lcrv;)V
 		ARG 1 boundingBox
+	METHOD a applyPistonMovement (Lcsa;)Lcsa;
+		ARG 1 offset
 	METHOD a movementInputToVelocity (Lcsa;FF)Lcsa;
 		ARG 0 movementInput
 		ARG 1 speed
@@ -199,6 +232,17 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD a clipSneakingMovement (Lcsa;Laiz;)Lcsa;
 		ARG 1 offset
 		ARG 2 type
+	METHOD a calculatePerpendicularMotionVector (Lcsa;Lcrv;Lbho;Lail;)Lcsa;
+		ARG 0 collisionVector
+		ARG 1 observerBounds
+		ARG 2 world
+		ARG 3 subject
+	METHOD a calculateTangentalMotionVector (Lcsa;Lcrv;Lbho;Lcsf;Lzx;)Lcsa;
+		ARG 0 collissionVector
+		ARG 1 observerBounds
+		ARG 2 world
+		ARG 3 context
+		ARG 4 collissions
 	METHOD a isTeamPlayer (Lctc;)Z
 		ARG 1 team
 	METHOD a populateCrashReport (Le;)V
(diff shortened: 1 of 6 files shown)


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -303,6 +346,8 @@ CLASS aij net/minecraft/entity/Entity
 		ARG 2 value
 	METHOD b isInvulnerableTo (Lahs;)Z
 		ARG 1 damageSource
+	METHOD b (Laij;)Z
+		ARG 1 e
```
modmuss50, line 362: e? does the method not have a name either?

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -269,11 +305,13 @@ CLASS aij net/minecraft/entity/Entity
 	METHOD ac setOnFireFromLava ()V
 	METHOD ad extinguish ()V
 	METHOD ae destroy ()V
+	METHOD af calculateNextStepDistance ()F
 	METHOD ag moveToBoundingBoxCenter ()V
 	METHOD ah getSwimSound ()Lyk;
 	METHOD ai getSplashSound ()Lyk;
 	METHOD aj getHighSpeedSplashSound ()Lyk;
 	METHOD ak checkBlockCollision ()V
+	METHOD al isJesus ()Z
```
modmuss50: needs changing, I have no clue what this is supposed to mean.

mappings/net/minecraft/entity/Entity.mapping
Sollace: Wait what? I think that's inside a lambda.

The `e` is a lambda-esque parameter name.

modmuss50 (changes requested):

mappings/net/minecraft/entity/Entity.mapping
modmuss50: ah, ok didnt know it was a lambada

Sollace: Maybe Enigma doesn't yet handle lambdas correctly?

ChloeDawn: Lambdas are handled correctly, it is just confusing to read in this form

mappings/net/minecraft/entity/Entity.mapping
Sollace: Is `walksOnAir` better?

That method controls whether an entity plays stepping sounds when in the air, so probably overriden by bats. I don't know what the correct term for it would be.

mappings/net/minecraft/entity/Entity.mapping
Sollace: Ah, okay. Well you might see a few other instances of that where I changed parameter names within lambdas.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -523,10 +588,13 @@ CLASS aij net/minecraft/entity/Entity
 	METHOD s isPartOf (Laij;)Z
 		ARG 1 entity
 	METHOD t isSpectator ()Z
-	METHOD t handlePlayerAttack (Laij;)Z
+	METHOD t onAttack (Laij;)Z
+		ARG 1 attacker
 	METHOD u copyPositionAndRotation (Laij;)V
 		ARG 1 entity
 	METHOD u_ initDataTracker ()V
+	METHOD v initNewDimention (Laij;)V
```
Pyrofab: Typo (dimen**s**ion)

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -483,9 +539,16 @@ CLASS aij net/minecraft/entity/Entity
...
 	METHOD j stopRiding ()V
+	METHOD j teleport (DDD)V
+		ARG 1 destX
+		ARG 3 destY
+		ARG 5 destZ
 	METHOD j getCameraPosVec (F)Lcry;
 		ARG 1 tickDelta
 	METHOD j allowsPermissionLevel (I)Z
+		ARG 1 permissionLevel
+	METHOD j getHardCollissionBox (Laij;)Lcrt;
```
Pyrofab: Typo (collision)

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -483,9 +539,16 @@ CLASS aij net/minecraft/entity/Entity
...
+		ARG 1 colliddingEntity
```
Pyrofab: Typo (colliding)

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -423,10 +475,14 @@ CLASS aij net/minecraft/entity/Entity
 		ARG 1 x
 		ARG 3 y
 		ARG 5 z
+	METHOD e calculateNextAerialStepDistance (F)F
+		ARG 1 distance
 	METHOD e setEntityId (I)V
 		ARG 1 id
 	METHOD e getEyeHeight (Lajc;)F
 		ARG 1 pose
+	METHOD e handleCollissions (Lcry;)Lcry;
```
Pyrofab: Typo (collision)

Pyrofab (changes requested):

mappings/net/minecraft/entity/Entity.mapping
ChloeDawn: > probably overridden by bats

Have you not look at implementations? *Always* look at implementations and uses when mapping a member.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -13,36 +16,48 @@ CLASS aij net/minecraft/entity/Entity
 	FIELD O age I
 	FIELD P insideWater Z
 	FIELD Q waterHeight D
+	FIELD R wet Z
```
Pyrofab: This is much more than wet. This field is true if the entity has their head in water, I would recommend `submergedInWater`.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -231,6 +264,9 @@ CLASS aij net/minecraft/entity/Entity
 	METHOD a isInFluid (Lzc;Z)Z
```
Pyrofab: While we are at it, this could be renamed to `isSubmergedIn`. It would make it more explicit that it checks for whether the entity has said fluid up to their eyes.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -345,10 +390,12 @@ CLASS aij net/minecraft/entity/Entity
 	METHOD bQ getServer ()Lnet/minecraft/server/MinecraftServer;
 	METHOD bR isImmuneToExplosion ()Z
 	METHOD bS entityDataRequiresOperator ()Z
+	METHOD bT teleportRequested ()Z
 	METHOD bU getPrimaryPassenger ()Laij;
 	METHOD bV getPassengerList ()Ljava/util/List;
 	METHOD bW getPassengersDeep ()Ljava/util/Collection;
-	METHOD bY getTopmostVehicle ()Laij;
+	METHOD bX hasPlayerJockey ()Z
+	METHOD bY getPrimaryVehicle ()Laij;
```
Pyrofab: What's the reasoning for not just calling it `getBottommostVehicle` ?

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -345,10 +390,12 @@ CLASS aij net/minecraft/entity/Entity
 	METHOD bQ getServer ()Lnet/minecraft/server/MinecraftServer;
 	METHOD bR isImmuneToExplosion ()Z
 	METHOD bS entityDataRequiresOperator ()Z
+	METHOD bT teleportRequested ()Z
 	METHOD bU getPrimaryPassenger ()Laij;
 	METHOD bV getPassengerList ()Ljava/util/List;
 	METHOD bW getPassengersDeep ()Ljava/util/Collection;
-	METHOD bY getTopmostVehicle ()Laij;
+	METHOD bX hasPlayerJockey ()Z
```
Pyrofab: IMO Jockey implies that the player is controlling this entity. In `PlayerManager#remove`, we can see the following code:
```java
if (entity_1.method_5817()) {
    LOGGER.debug("Removing player mount");
```
So why not call it `isPlayerMount` ?

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -413,6 +463,8 @@ CLASS aij net/minecraft/entity/Entity
 	METHOD d getOppositeRotationVector (FF)Lcry;
 		ARG 1 pitch
 		ARG 2 yaw
+	METHOD d calculatePoseBounds (Lajc;)Lcrt;
```
Pyrofab: I'd suggest `calculateBoundsForPose`. It would make it slightly clearer that it is calculating the bounds of *the entity* for the given pose.

mappings/net/minecraft/entity/Entity.mapping
Sollace: I look at the uses. I'm used to not being able to easily find implementations. (iirc the only thing I've used that had that feature was Android Studio)

I did later check and eventually found the correct place where it's being used is in `ParrotEntity` (overriden to return true)

mappings/net/minecraft/entity/Entity.mapping
Sollace: Why do I always do this? >.> I get confused about what does and does not have double-Ds and Ss.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -523,10 +588,13 @@ CLASS aij net/minecraft/entity/Entity
 	METHOD s isPartOf (Laij;)Z
 		ARG 1 entity
 	METHOD t isSpectator ()Z
-	METHOD t handlePlayerAttack (Laij;)Z
+	METHOD t onAttack (Laij;)Z
```
Pyrofab: This is a weird one. It is only called by players, has a special behaviour only when the attacker is a player, but still accepts an Entity. And it is used as such:
```java
   public void attack(Entity entity_1) {
      if (entity_1.canPlayerAttack()) {
         if (!entity_1.handlePlayerAttack(this)) {
```
So `handleAttack` makes a lot more sense than `onAttack`. Now, whether we keep `player` in the name is up to debate.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -305,13 +305,13 @@ CLASS aij net/minecraft/entity/Entity
...
 	METHOD ae destroy ()V
-	METHOD af calculateNextStepDistance ()F
+	METHOD af calculateStepDelta ()F
 	METHOD ag moveToBoundingBoxCenter ()V
 	METHOD ah getSwimSound ()Lyk;
 	METHOD ai getSplashSound ()Lyk;
 	METHOD aj getHighSpeedSplashSound ()Lyk;
 	METHOD ak checkBlockCollision ()V
-	METHOD al isJesus ()Z
+	METHOD al playsFlyingSounds ()Z
```
Sollace: This still doesn't seem right to me. The name sounds too long when I try to read it.

If anyone has any ideas for a better alternative, I'm all ears.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -119,18 +137,24 @@ CLASS aij net/minecraft/entity/Entity
...
+		ARG 1 yaw
+		ARG 2 pitch
 	METHOD a updateVelocity (FLcry;)V
 		ARG 1 speed
 		ARG 2 movementInput
 	METHOD a damage (Lahs;F)Z
 		ARG 1 source
 		ARG 2 amount
+	METHOD a (Laij;)Ljava/util/stream/Stream;
+		ARG 1 collissionVector
```
Pyrofab: typo (collision)

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -192,13 +216,20 @@ CLASS aij net/minecraft/entity/Entity
...
+		ARG 1 offset
 	METHOD a movementInputToVelocity (Lcry;FF)Lcry;
 		ARG 0 movementInput
 		ARG 1 speed
 		ARG 2 yaw
 	METHOD a clipSneakingMovement (Lcry;Laix;)Lcry;
 		ARG 1 offset
 		ARG 2 type
+	METHOD a calculateMotionVector (Lcry;Lcrt;Lbhm;Lcsd;Lzv;)Lcry;
+		ARG 0 collissionVector
```
Pyrofab: typo (collision)

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -192,13 +216,20 @@ CLASS aij net/minecraft/entity/Entity
...
+		ARG 2 world
+		ARG 3 context
+		ARG 4 collissions
```
Pyrofab: typo (collision)

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -252,7 +288,7 @@ CLASS aij net/minecraft/entity/Entity
 	METHOD aO getMountedHeightOffset ()D
 	METHOD aP isLiving ()Z
 	METHOD aQ removeAllPassengers ()V
-	METHOD aR getBoundingBoxMarginForTargeting ()F
+	METHOD aR getTargettingDistance ()F
```
Pyrofab: typo (targeting)

Pyrofab (changes requested):

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -523,10 +588,13 @@ CLASS aij net/minecraft/entity/Entity
 	METHOD s isPartOf (Laij;)Z
 		ARG 1 entity
 	METHOD t isSpectator ()Z
-	METHOD t handlePlayerAttack (Laij;)Z
+	METHOD t onAttack (Laij;)Z
+		ARG 1 attacker
 	METHOD u copyPositionAndRotation (Laij;)V
 		ARG 1 entity
 	METHOD u_ initDataTracker ()V
+	METHOD v initNewDimension (Laij;)V
```
Pyrofab: At first glance, with this name, I would assume that this method initializes a new dimension. In fact, it copies attributes from the original entity in the context of a dimension change. I think the name should reflect that, by containing "copy" or equivalent.

mappings/net/minecraft/entity/Entity.mapping
Prospector: isPlayerMount or isPlayerMounted?

mappings/net/minecraft/entity/Entity.mapping
Prospector: When's it used? When it's on the player's shoulder or what?

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -13,36 +16,48 @@ CLASS aij net/minecraft/entity/Entity
...
 	FIELD aD invulnerable Z
 	FIELD aE scoreboardTags Ljava/util/Set;
+	FIELD aF teleportRequested Z
 	FIELD aG pistonMovementDelta [D
 	FIELD aH pistonMovementTick J
 	FIELD aI size Laik;
 	FIELD aJ standingEyeHeight F
 	FIELD aa chunkY I
 	FIELD ab chunkZ I
+	FIELD ac trackedXPosition J
```
Gegy: Maybe these could be just `trackedX`, `trackedY` and `trackedZ`? Fits better with other coordinate fields on the class

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -92,21 +108,23 @@ CLASS aij net/minecraft/entity/Entity
...
+		ARG 3 destY
+		ARG 5 destZ
 	METHOD a setPositionAnglesAndUpdate (DDDFF)V
 		ARG 1 x
 		ARG 3 y
 		ARG 5 z
 		ARG 7 yaw
 		ARG 8 pitch
-	METHOD a setPositionAndRotations (DDDFFIZ)V
+	METHOD a updateClientPositionAndAngles (DDDFFIZ)V
```
Gegy: This could make specific reference to being related to tracking -- the current name doesn't make that immediately clear. Not sure of alternative, though. Possibly something like `updateTrackedPositionAndAngles`?

mappings/net/minecraft/entity/Entity.mapping
Sollace: Trying to avoid MCP names, and in this case the bottom-most vehicle is also the vehicle being steered, the primary means of transportation, so I felt Primary Vehicle would be more suitable.

mappings/net/minecraft/entity/Entity.mapping
Sollace: `hasPlayerMount`?

mappings/net/minecraft/entity/Entity.mapping
liach: `getRootVehicle`

mappings/net/minecraft/entity/Entity.mapping
liach: `hasPlayerMount`

mappings/net/minecraft/entity/Entity.mapping
Sollace: Sure, `tracked` instead of `client` works for me. The intention was to make it clear this method only exists on the client, but you can probably extend it to anything in the `tracked` category.

mappings/net/minecraft/entity/Entity.mapping
Sollace: I'm not a fan for using so many words in a method name like that. It gets into the territory of calling methods `calculateNewBoundsUsingTheInverseConstantLawOfThermalDynamics()`

I don't have any other objections to the name, so I'll go ahead and use that one.

mappings/net/minecraft/entity/Entity.mapping
Sollace: Props to @liach , the MVP of method names.

mappings/net/minecraft/entity/Entity.mapping
Sollace: `submergedInWater` is such a mouth-full though. I'd more likely just called it `submerged`.

mappings/net/minecraft/entity/Entity.mapping
Sollace: * ditto `isSubmerged`

mappings/net/minecraft/entity/Entity.mapping
Sollace: A previous version used `copyFrom` for this method, however I noted that it seems to specifically exclude dimension information when copying, so it doesn't appear to be a generic copy function.

mappings/net/minecraft/entity/Entity.mapping
Pyrofab: What ? No, this returns true if the entity is mounted by a player.

Sollace:
The PR's been updated to be compatible with the new branch. I found there were some new methods added in 1.14.3-pre3 handling entity collisions that I had to map in the process.

`a` -> `calculatePerpendicularMotionVector`
`a` -> `calculateTangentalMotionVector`

`calculateMotionVector` does a check against the x/yz of the collision vector coming in and will call either one or the other method. So far all the difference I can see between them is one calls `VoxelShapes.calculateMaxOffset` and the other calls `VoxelShapes.a` (which I've dubbed `calculateSoftOffset`).

Difference between `calculateMaxOffset` and `calculateSoftOffset` is the former will repeat the check multiple times whilst the latter only does it once. (Hypothesising incoming) The logic between this seems to be that if the collisions are face-on on certain plains, a lot of the logic can be short-circuited to avoid ping-ponging, and they do this to improve performance in special cases.

Aside from that, all I'm waiting on is some feedback on the "is submerged" changes and that method for copying across dimensions.

Runemoro:
> calculateTangentalMotionVector

tangential

mappings/net/minecraft/entity/Entity.mapping
Sollace: I've decided to change this to `inWater` to be in line with `inLava`.

mappings/net/minecraft/entity/Entity.mapping
Sollace: Decided to go back to the `copyFrom` idea after reviewing its usages and overrides. (Don't see any places where it gets overriden)

Sollace:
@Runemoro Correction made, thank you!

mappings/net/minecraft/entity/Entity.mapping
Sollace: It's used in `move()` on the base entity class and just as a check whether to update stepping sounds when not on the ground. (there are separate timers for that).

imo I think it's used so the Parrots can make wing flapping sounds when they fly.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -345,10 +403,12 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD bQ getServer ()Lnet/minecraft/server/MinecraftServer;
 	METHOD bR isImmuneToExplosion ()Z
 	METHOD bS entityDataRequiresOperator ()Z
+	METHOD bT teleportRequested ()Z
 	METHOD bU getPrimaryPassenger ()Lail;
 	METHOD bV getPassengerList ()Ljava/util/List;
 	METHOD bW getPassengersDeep ()Ljava/util/Collection;
-	METHOD bY getTopmostVehicle ()Lail;
+	METHOD bX hasPlayerMount ()Z
```
Pyrofab: Sorry, I was not clear enough. `hasPlayerJockey` was viable, but this implies the entity can be riding the player. I was just noting that `jockey` implied control, but this does not check that this entity can be controlled by the rider. Now that I think about it, `hasPlayerRider` seems like an adequate name.

Pyrofab (changes requested):

mappings/net/minecraft/entity/Entity.mapping
Sollace: Ah, I see what you mean now. I can change it to `hasPlayerRider` if you would prefer. That seems almost an equivalent to what's currently there.

Pyrofab: `hasPlayerMount` is wrong IMO, it implies the player is ridden by the entity, when it is the opposite. `isPlayerMount` would be fine, as would be `hasPlayerRider`

Sollace: Lé done. Ì hope `hasRider` is fine. I only saw that I left out the `Player` after I pushed.

Sollace:
Unmapped `playsFlyingSounds`. Everything should be good to merge whenever y'all are ready.

mappings/net/minecraft/entity/Entity.mapping
Pyrofab: The `player` part is important in my opinion, but at least it is not wrong anymore

Pyrofab (approved):

mappings/net/minecraft/entity/Entity.mapping
Sollace: Eeyup. Added the `player` already.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -2,6 +2,9 @@ CLASS ail net/minecraft/entity/Entity
 	FIELD A velocityModified Z
 	FIELD B movementMultiplier Lcsa;
 	FIELD C removed Z
+	FIELD D previousHorizontalSpeed F
```
Runemoro: We use `last` rather than `previous` for the value last tick (see WorldRenderer, for example).

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -13,36 +16,48 @@ CLASS ail net/minecraft/entity/Entity
 	FIELD O age I
 	FIELD P insideWater Z
 	FIELD Q waterHeight D
+	FIELD R inWater Z
 	FIELD S inLava Z
+	FIELD T regenTimer I
```
Runemoro: This isn't a timer, it's just an int, so it should be `regenTime`, or `timeUntilRegen`. What does "regen" mean though? How foes a non-living entity regen?

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -92,21 +108,23 @@ CLASS ail net/minecraft/entity/Entity
 		ARG 1 cursorDeltaX
 		ARG 3 cursorDeltaY
 	METHOD a requestTeleport (DDD)V
-		ARG 1 x
-		ARG 3 y
-		ARG 5 z
+		ARG 1 destX
+		ARG 3 destY
+		ARG 5 destZ
 	METHOD a setPositionAnglesAndUpdate (DDDFF)V
```
Runemoro, line 114: This isn't necessary. `requestTeleport(x, y, z)` means "request teleport to x, y, z". Look at our her methods such as Entity.move which have parameters also names just "x, y, z".

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -119,18 +137,31 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD a setRotation (FF)V
 		ARG 1 yaw
 		ARG 2 pitch
+	METHOD a updateTrackedHeadRotation (FI)V
+		ARG 1 yaw
+		ARG 2 pitch
```
Runemoro: This isn't just a simple setter? It also seems weird yaw is a float and pitch is an int...

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -252,7 +301,7 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD aO getMountedHeightOffset ()D
 	METHOD aP isLiving ()Z
 	METHOD aQ removeAllPassengers ()V
-	METHOD aR getBoundingBoxMarginForTargeting ()F
+	METHOD aR getTargetingDistance ()F
```
Runemoro: Are you sure? If I remember correctly, this is the amount by which the bounding box is expanded, meaning:

  1. The new name could make people think it's from the center of the entity, as most distances are
  2. It's not an (euclidean) distance at all. It represents a cuboid region, not a spherical one as your name implies.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -483,9 +551,16 @@ CLASS ail net/minecraft/entity/Entity
...
 	METHOD j stopRiding ()V
+	METHOD j teleport (DDD)V
+		ARG 1 destX
+		ARG 3 destY
+		ARG 5 destZ
 	METHOD j getCameraPosVec (F)Lcsa;
 		ARG 1 tickDelta
 	METHOD j allowsPermissionLevel (I)Z
+		ARG 1 permissionLevel
+	METHOD j getHardCollisionBox (Lail;)Lcrv;
```
Runemoro, line 562: What's the "hard collision box"? Is there also a soft one...?

mappings/net/minecraft/world/chunk/WorldChunk.mapping
```diff
@@ -59,6 +60,8 @@ CLASS bxn net/minecraft/world/chunk/WorldChunk
 		ARG 4 clearOld
 	METHOD b remove (Lail;)V
 	METHOD c setLoadedToWorld (Z)V
+	METHOD d setUnsaved (Z)V
+		ARG 1 saved
```
Runemoro: Either the method should be setSaved or the parameter should be unsaved.

Sollace: The field is set to `true` when a new Entity is added (see WorldChunk.addEntity).

`ChunkSerializer.serialize` calls the method with a value of `false`

    worldChunk14.setUnsaved(false);

`ChunkSerializer.addEntities` calls `addEntity` and then calls the method with the value `true`.

(there is an Enigma bug here where `chunk` has been mistakenly labelled `entity`. I've corrected the name below to properly account for the lambda.)

    private static void writeEntities(CompoundTag tag, WorldChunk chunk) {
        ListTag listTag3 = tag.getList("Entities", 10);
        World world4 = chunk.getWorld();
        for (int var5 = 0; var5 < listTag3.size(); ++var5) {
            CompoundTag compoundTag6 = listTag3.getCompoundTag(var5);
            EntityType.loadEntityWithPassengers(compoundTag6, world4, entity -> {
                chunk.addEntity(entity);
                return entity;
            });
            chunk.setUnsaved(true);
        }

 It is also used in `needsSaving` when checking the `lastSaveTime` hence it is `unsaved` or `dirty` as a certain voldemort might call it.

    @Override
    public boolean needsSaving() {
        return this.shouldSave || (this.unsaved && this.world.getTime() != this.lastSaveTime);
    }

mappings/net/minecraft/entity/Entity.mapping
Sollace: You're right, I completely missed that detail. The integer here is consistent (called with a value of three) with the `interpolationSteps` parameter used by the other `updateTrackedX` methods.

Sollace: This is also not a simple setter. It's called by `ClientPlayNetworkHandler` in response to the `EntitySetHeadYawS2CPacket` packet. It's also a client-only method, so it's being named consistently with the other methods in its family (`updateTrackedPositionAndAngles`)

mappings/net/minecraft/entity/Entity.mapping
Sollace: It's used for hard collisions, as in boats and minecarts. Returning a non-null value here makes the entity effectively solid when interacting with the player and other entities, as opposed to the default where entities tend to squish together.

mappings/net/minecraft/entity/Entity.mapping
Sollace: I've abstracted this a bit. It's used in targetting to expand the bounding box, but effectively serves as an additional radius around there area an entity would normally be hit.

Maybe `getTargettingMargin` would be a better descriptor?

mappings/net/minecraft/entity/Entity.mapping
Sollace: I don't see any such example method. Besides, "x, y, z" can have other interpretations. They can be an offset on each axis, a range within which to teleport, or width/height/depth. In this case it's very clearly the _destination_ you want to teleport to, so I see value in making that distinction with the names.

mappings/net/minecraft/entity/Entity.mapping
Sollace: `last` is being used here for the `lastPortal` value. I will shorten it to `prevHorizontalSpeed` though to match the other `prev` values in this class.

mappings/net/minecraft/entity/Entity.mapping
Runemoro: It's clear they're absolute positions. Every single other method that sets positions (and there are many) uses `x`, `y`, `z`. 

> I don't see any such example method. 

Open the "Entity" class, do a search for "move(".

mappings/net/minecraft/entity/Entity.mapping
Sollace: Generally, I dislike the use of `last` itself, anyway, as it doesn't properly convey the meaning that it is the _previous_ value. With `previous` you know there must be a _current_ counterpart, but last doesn't have that assumption. `last` just says it is the value as it was last known.

mappings/net/minecraft/entity/Entity.mapping
Runemoro: Yes, margin is better. Though we decided on "targeting", not "targetting", in the last PR involving this a similar method since that's the spelling Mojang uses (see the F3 screen, "Targeted block").

mappings/net/minecraft/world/chunk/WorldChunk.mapping
Runemoro: I'm just talking about the name of the parameter not matching the method name. Anyone seeing the code `setUnsaved(true)` will think it's marking the chunk as *unsaved*. But according to your parameter name, it's actually marking it as *saved*.

If the parameter is named correctly, the method should be `setSaved`. Otherwise, the parameter should be `unsaved`

mappings/net/minecraft/entity/Entity.mapping
Sollace: This is definitely a timer of some kind. It's used in `LivingEntity.baseTick()` (which I can't see).
It is being set to a value of `60` in `SkeletonHorseTrapTriggerGoal` both by `getSkeleton` and by `getHorse`. 
It's set to `10` or `20` in `ClientPlayerEntity.setHealth()`
It is decremented every tick in `ServerPlayerEntity.tick()` (if greater than 0)

        if (this.regenTimer > 0) {
            --this.regenTimer;
        }

It is also referenced in `InGameHub.renderStatusBar` and is thus very clearly closeley related to `health` and `time`.

        int integer3 = MathHelper.ceil(playerEntity2.getHealth());
        boolean boolean4 = this.C > this.ticks && (this.C - this.ticks) / 3L % 2L == 1L;
        long long5 = SystemUtil.getMeasuringTimeMs();
        if (integer3 < this.z && playerEntity2.regenTimer > 0) {
            this.B = long5;
            this.C = this.ticks + 20;
        }
        else if (integer3 > this.z && playerEntity2.regenTimer > 0) {
            this.B = long5;
            this.C = this.ticks + 10;

After some investigation of this method. `B` and `C` are both specific to those lines, however `C` is used to calculate `boolean4` whilst goes on to be used later when rendering health:

            int integer25 = 0;
            if (boolean4) {
                integer25 = 1;
            }

Changing `integer25` to 1 only affects the first `blit` call after that, shifting the texture y position down 9 pixels, which you can probably attribute to the white "flash" hearts create when the player's health is regenerating.

mappings/net/minecraft/world/chunk/WorldChunk.mapping
Sollace: Oh, derp. Thanks for that.

Sollace: In cases like this I normally just name the parameter `value`, though. I can do the `unsaved` change for now, but maybe later we need to get a standard on what to call parameters in this situation.

mappings/net/minecraft/entity/Entity.mapping
Pyrofab: The thing is, calling the int a timer is semantically wrong. A timer is an object that keeps track of a time variable. What you call `regenTimer` is not an actual timer, it is the time that is ticked down by the entity. Hence `timeUntilRegen`.

mappings/net/minecraft/entity/Entity.mapping
Sollace: Yes, sorry. I always type targetting the way I've known it to be, even when I don't mean to. Same occurs with collissions.

mappings/net/minecraft/entity/Entity.mapping
Sollace: There is only one method here I see named `move`

    public void move(MovementType type, Vec3d offset) {

mappings/net/minecraft/world/chunk/WorldChunk.mapping
Runemoro: The standard is to name setter parameters the same as the property they're setting. There are over 1000 setters in yarn all using this convention, it's the convention Mojang uses in their public libraries, and it's the convention used in the JDK.

mappings/net/minecraft/entity/Entity.mapping
Sollace: Renamed then. :/

mappings/net/minecraft/world/chunk/WorldChunk.mapping
asiekierka: It is the standard? TIL.

~~Well, it makes sense, but I've always used ``value``~~

asiekierka (approved):

mappings/net/minecraft/entity/Entity.mapping
LemmaEOF: Yeah, Mojang moved everything over to Vec3d recently, which uses `x`, `y`, and `z` as it's general-purpose. I'm okay with both `x,y,z` and `destX,destY,destZ` here.

LemmaEOF (approved):

LemmaEOF:
@modmuss50 it looks like your review needs an update.

modmuss50 (approved):
Looks fine to me.

modmuss50:
I will leave this open for a bit longer to allow others to give more feedback.

mappings/net/minecraft/util/shape/VoxelShapes.mapping
```diff
@@ -42,6 +42,7 @@ CLASS csr net/minecraft/util/shape/VoxelShapes
 	METHOD a union (Lcsu;[Lcsu;)Lcsu;
 		ARG 0 first
 		ARG 1 others
+	METHOD a calculateSoftOffset (Lfa$a;Lcrv;Lbho;DLcsf;Ljava/util/stream/Stream;)D
```
Runemoro, line 45: What does this mean...?

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -13,36 +16,48 @@ CLASS ail net/minecraft/entity/Entity
...
 	FIELD ao uuid Ljava/util/UUID;
 	FIELD ap uuidString Ljava/lang/String;
 	FIELD aq glowing Z
 	FIELD ar passengerList Ljava/util/List;
 	FIELD as vehicle Lail;
 	FIELD at velocity Lcsa;
-	FIELD au boundingBox Lcrv;
+	FIELD au entityBounds Lcrv;
+	FIELD av nextStepDistance F
+	FIELD aw aerialStepDelta F
```
Runemoro: What's this?

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -192,13 +223,26 @@ CLASS ail net/minecraft/entity/Entity
...
+		ARG 0 collisionVector
+		ARG 1 observerBounds
+		ARG 2 world
+		ARG 3 subject
+	METHOD a calculateTangentialMotionVector (Lcsa;Lcrv;Lbho;Lcsf;Lzx;)Lcsa;
+		ARG 0 collisionVector
+		ARG 1 observerBounds
+		ARG 2 world
+		ARG 3 context
+		ARG 4 collisions
```
Runemoro, line 245: These were named wrong. They have nothing to do with perpendicular and tangential motion. Both adjust a movement for collisions, but one is an optimized version that can only be used when the movement is along a single axis (walking up stairs calls the collision movement adjustment method many times with movement along only one axis):

```java
    public static Vec3d adjustMovementForCollisions(@Nullable Entity entity, Vec3d movement, Box entityBoundingBox, World world, EntityContext context, ReusableStream<VoxelShape> collisions) {
        boolean noX = movement.x == 0.0;
        boolean noY = movement.y == 0.0;
        boolean noZ = movement.z == 0.0;

        if (noX && noY || noX && noZ || noY && noZ) {
            return adjustSingleAxisMovementForCollisions(movement, entityBoundingBox, world, context, collisions);
        }

        ReusableStream<VoxelShape> blockCollisions = new ReusableStream<>(Stream.concat(collisions.stream(), world.getBlockCollisions(entity, entityBoundingBox.stretch(movement))));
        return adjustMovementForCollisions(movement, entityBoundingBox, blockCollisions);
    }
```

mappings/net/minecraft/entity/Entity.mapping
Runemoro, line 60: This isn't a delta (delta means difference), it's the value of `distanceWalked` at which the next step sound should be played:

```java
    protected float calculateStepDelta() {
        return (float) ((int) this.distanceWalked + 1);
    }
```

mappings/net/minecraft/entity/Entity.mapping
Sollace: Since this has already been merged why don't you make a new PR/issue so we can continue discussion?

Runemoro: I found these issues while working on a new PR, and the PR will include changes to these methods.

PR #744, opened 2019-06-08
Add message history related mappings.

occanowey:


mappings/net/minecraft/client/gui/hud/ChatHud.mapping
```diff
@@ -10,12 +11,15 @@ CLASS cwu net/minecraft/client/gui/hud/ChatHud
 		ARG 1 x
 		ARG 3 y
 	METHOD a draw (I)V
+	METHOD a addMessageHistory (Ljava/lang/String;)V
```
asiekierka: addMessageHistory sounds weird

occanowey: Maybe addToMessageHistory?

liach:
does this field exist in command block gui? consider mapping if it does.

occanowey:
It doesn't look like command blocks have command history.

liach (approved):

Issue #89, opened 2016-09-21
"World" vs "Dimension" vs "Level" and where to use them?


copygirl:
Currently, `World` is meant for anything regarding single instances of the `World` class, which is unlikely to change. `Level` is being adopted as the term used for a single save - confusingly also called "World" in the game (such as "Create New World"), though sometimes "Level" in the code (like "preparing level <name> ...").

Previously some of our classes were using "Dimension" to describe an instance of a world (usually which ID - `dimensionId`) or its type (Overworld / Nether / End - `DimensionType`), though I've done some changes to get rid of the mention of "dimension" completely.

I'm not sure this is the correct way to go about it. We should retain the name "Dimension" for something, though I'm not entirely sure _what_ to use it for.

exaskye:
The impression I've gotten is "Level" was the name used in earlier versions, before multiple dimensions got implemented. The main place I recall seeing it was the flashing "Saving level..." text that was removed at some point in Beta, and that was a remnant of some early Alpha version, and it appeared despite doing nothing in SMP. (IIRC, Classic also used the .mclevel extension for save games.)

The default name for a save game on a server is `world` (and it has been for a _long_ time), which seems to imply a World is a top-level container. The sub-directories are called `DIM`, implying Dimension is the next level in. Those would seem like the most "correct" names as such, and are closer to the names people are most used to.

`World` → `Dimension`, `Level` → `World`, would follow that, but that's kind of strange.

asiekierka:
net.minecraft.world.Dimension would be awkward, though, on the other hand.

Player3324:
My preference is WorldSave + World with WorldSave being the collection of Worlds.

"Dimension" afaik is just wrong from an English language POV, describing possibly some original intention instead of the actual meaning/implementation these days. The different worlds don't have a particular dimensional meaning in the mathematical sense and with mods adding more "dimensions" that's even closer to just adding more Worlds.

Level is commonly used to describe a contiguous region/map, which is the equivalent of a world in Minecraft. The proposed use is orthogonal to that.

I didn't go for just save since that's rather ambiguous, so WorldSave seems like a good option. Mojang's own choices are poor, not very feasible to copy.

copygirl:
> WorldSave

Only problem with that is it clashes with `WorldSaveHandler`.

exaskye:
> being the collection of Worlds

WorldCollection? :P

[...]

PR #745, opened 2019-06-08
EntitySize -> EntityDimensions

Sollace:
Changes for the issue at https://github.com/FabricMC/yarn/issues/715

### DIFF
--- a/mappings/net/minecraft/entity/EntityDimensions.mapping
+++ b/mappings/net/minecraft/entity/EntityDimensions.mapping
@@ -1,11 +1,11 @@
 CLASS aim net/minecraft/entity/EntityDimensions
 	FIELD a width F
 	FIELD b height F
-	FIELD c constant Z
+	FIELD c fixed Z
 	METHOD <init> (FFZ)V
 		ARG 1 width
 		ARG 2 height
-		ARG 3 constant
+		ARG 3 fixed
 	METHOD a scaled (F)Laim;
 		ARG 1 ratio
 	METHOD a scaled (FF)Laim;
@@ -14,6 +14,6 @@ CLASS aim net/minecraft/entity/EntityDimensions
 	METHOD b changing (FF)Laim;
 		ARG 0 x
 		ARG 1 z
-	METHOD c constant (FF)Laim;
+	METHOD c fixed (FF)Laim;
 		ARG 0 x
 		ARG 1 z
--- a/mappings/net/minecraft/entity/EntityDimensions.mapping
+++ b/mappings/net/minecraft/entity/EntityDimensions.mapping
@@ -11,7 +11,7 @@ CLASS aim net/minecraft/entity/EntityDimensions
 	METHOD a scaled (FF)Laim;
 		ARG 1 widthRatio
 		ARG 2 heightRatio
-	METHOD b resizeable (FF)Laim;
+	METHOD b changing (FF)Laim;
 		ARG 0 x
 		ARG 1 z
 	METHOD c constant (FF)Laim;
--- a/mappings/net/minecraft/entity/EntityDimensions.mapping
+++ b/mappings/net/minecraft/entity/EntityDimensions.mapping
@@ -12,8 +12,8 @@ CLASS aim net/minecraft/entity/EntityDimensions
 		ARG 1 widthRatio
 		ARG 2 heightRatio
 	METHOD b changing (FF)Laim;
-		ARG 0 x
-		ARG 1 z
+		ARG 0 width
+		ARG 1 height
 	METHOD c fixed (FF)Laim;
-		ARG 0 x
-		ARG 1 z
+		ARG 0 width
+		ARG 1 height
--- a/mappings/net/minecraft/entity/EntityDimensions.mapping
+++ b/mappings/net/minecraft/entity/EntityDimensions.mapping
@@ -1,4 +1,4 @@
-CLASS aim net/minecraft/entity/EntitySize
+CLASS aim net/minecraft/entity/EntityDimensions
 	FIELD a width F
 	FIELD b height F
 	FIELD c constant Z
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -140,7 +140,7 @@ CLASS ail net/minecraft/entity/Entity
 	METHOD a move (Laiz;Lcsa;)V
 		ARG 1 type
 		ARG 2 offset
-	METHOD a getSize (Laje;)Laim;
+	METHOD a getDimensions (Laje;)Laim;
 		ARG 1 pose
 	METHOD a getEyeHeight (Laje;Laim;)F
 		ARG 1 pose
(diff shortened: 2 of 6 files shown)


Prospector (approved):

liach:
please change all getter methods to getDimensions instead of getSize; same for fields/method params

liach:
also consider renaming constant to fixed

mappings/net/minecraft/entity/EntityDimensions.mapping
```diff
@@ -11,7 +11,7 @@ CLASS aim net/minecraft/entity/EntitySize
 	METHOD a scaled (FF)Laim;
 		ARG 1 widthRatio
 		ARG 2 heightRatio
-	METHOD b resizeable (FF)Laim;
+	METHOD b changing (FF)Laim;
 		ARG 0 x
 		ARG 1 z
 	METHOD c constant (FF)Laim;
```
ChloeDawn: `fixed`

ChloeDawn (changes requested):

ChloeDawn:
The parameter names for the static constructors are incorrect. They should be `width` and `height` not `x` and `z`.

mappings/net/minecraft/entity/EntityDimensions.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS aim net/minecraft/entity/EntitySize
+CLASS aim net/minecraft/entity/EntityDimensions
 	FIELD a width F
 	FIELD b height F
 	FIELD c constant Z
```
ChloeDawn: `fixed`

mappings/net/minecraft/entity/EntityDimensions.mapping
ChloeDawn: Constructor parameter and field also need to be changed to `fixed`

ChloeDawn:
**`field`**`EnderDragonPart#partSize` -> `EnderDragonPart#dimensions`
**`field`**`Entity#size` -> `Entity#dimensions`
**`field`**`EntityType#size` -> `EntityType#dimensions`
**`field`**`EntityType$Builder#size` -> `EntityType$Builder#dimensions`
**`field`**`LivingEntity.SLEEPING_SIZE` -> `LivingEntity.SLEEPING_DIMENSIONS`
**`field`**`PlayerEntity.SIZES` -> `PlayerEntity.POSE_DIMENSIONS`
**`field`**`PlayerEntity.STANDING_SIZE` -> `PlayerEntity.STANDING_DIMENSIONS`
**`method`**`Entity#getSize(EntityPose)` -> `Entity#getDimensions(EntityPose)`
**`method`**`Entity#refreshSize` -> `Entity#refreshDimensions`
**`method`**`EntityType#getDefaultSize` -> `EntityType#getDefaultDimensions`
**`param`**`Entity#getEyeHeight@size` -> `Entity#getEyeHeight@dimensions`
**`param`**`EntityType.<init>@size` -> `EntityType.<init>@dimensions`
**`param`**`LivingEntity#getActiveEyeHeight@size` -> `LivingEntity#getActiveEyeHeight@dimensions`

ChloeDawn:
Migration required for Fabric's `EntityTypeBuilder`:
**`field`**`size` -> `dimensions`
**`method`**`size` -> `dimensions`
**`param`**`size@size` -> `dimensions`

kashike:
I'm not a fan of this name change - it will cause confusion with world dimensions.

ChloeDawn:
I agree, but either way, I do not want a partial migration. All or nothing. Personally, I would call this data class `EntityShape` as it implies the width and height, and separates it away from the concept of dimensions that the game already has.

liach:
i originally opposed this change, but i find that it will be clear if we call this one dimensions (with the s suffix) instead of dimension.

Sollace:
@liach @kashike Again, I remind everyone that it is `Dimensions` _with an S_.

@liach I did consider doing name changes for all the getters as well, however I don't believe it's neccessary to have it use the same terminology as the class name. It's clear enough to see that `getSize()` returns a some form of Dimensions. It also avoids confusion with a method like `getDimension()`.

Sollace:
@liach I thought people specifically didn't want that changed to fixed? (which is why it was left out of the PR)
@InsomniaKitten  (pinged wrong person)
>I believe constant is fine too, representing an immutable size. Imo fixed doesn't really imply that well.

kashike:
@Sollace I am well aware it has an `s`, my statement is not changing.

Sollace:
I was also thinking about world dimensions, and what's stopping us from calling them `Level`s or `LevelType`s? Minecraft has used that term in the past when referring to the overworld, so I don't see why they wouldn't still have it internally.

liach:
how about EntityExtent?

liach:
`EntityMeasure` would be clear and non-confusing.

See defintion 4 at https://www.dictionary.com/browse/measures (noun section)
>the extent, dimensions, quantity, etc., of something, ascertained especially by comparison with a standard: 
>to take the measure of a thing.

This covers "dimensions" "extent" in a sense while it won't be confusing as "extent" or "dimension" (both has been used in spongeapi to refer to world stuff)

@kashike @InsomniaKitten @Prospector wanna hear your opinions.

copygirl:
> I was also thinking about world dimensions, and what's stopping us from calling them `Level`s or `LevelType`s?

The term is already standard in the entire community.
Very early on we had a discussion on this. It didn't really lead anywhere, but see #89.

----

That said, I'm split on the change.
**Pro:** `toString` clearly mentioning the actual name of the class.
**Con:** It kind of clashes with `Dimension`, for auto-complete suggestions and such.

I also don't necessarily see `Size` as a scalar value. It is often used in a 2D context to describe an element's width and height. And I don't think we should introduce another name that is different from Mojang's name.

Sollace:
> EntityMeasure would be clear and non-confusing.

@liach Even though you didn't ask for it, I don't know if I like using `EntityMeasure`. What is it a measure of? Volume? Age? Does it specify width/height in meters, or centimeters?

Measure to me is something of standard, fixed, size or dimensions that you use to _measure_ something. Like how a meter stick is a measure of length.

liach:
this way, dimensions is a plain scalar as well. i'd say stay with size; we should not be trolled by mojang for such a tostring statement after all.

NeunEinser:
> this way, dimensions is a plain scalar as well.

Not sure what you are referring to. You can measure in m² (or m³), too, which to my understanding is the case when you measure dimensions since you are measuring *all* dimensions of the object.

I think Wikitonary's definition pretty much nails it and describes perfectly what this class is. (definition of a **single** dimension, not the plural form)
>A measure of spatial extent in a particular direction, such as height, width or breadth, or depth.

Runemoro (approved):

Runemoro:
To me a size is the overall size of something. I wouldn't expect a class named `EntitySize` to give both a width and height.

@kashike Currently Entity.getSize (which returns its dimensions, not size!) conflicts with SlimeEntity.getSlimeSize (which correctly returns a single integer size!).

asiekierka:
I say EntityDimensions.

LemmaEOF (approved):

NeunEinser (approved):

Gegy (approved):

modmuss50 (approved):

PR #747, opened 2019-06-09
AbsoluteHand -> Arm

ChloeDawn:
It represents an arm, used for changing the arm angles of models, and for determining the arm that the main hand is bound to. Some example uses of this enum can be seen below:
```java
// ModelWithArms

void setArmAngle(float <unmapped>, Arm arm);
```
```java
// BipedEntityModel

protected Arm getPreferredArm(T entity) {
    Arm arm = entity.getMainArm();
    return (entity.preferredHand == Hand.MAIN_HAND) ? arm : arm.getOpposite();
}
```
```java
// HeldItemFeatureRenderer

boolean mainHand = hand == Hand.MAIN_HAND;
Arm arm = mainHand ? player.getMainArm() : player.getMainArm().getOpposite();
```
```java
// GameOptions

if ("mainHand".equals(key)) {
    this.mainArm = "left".equals(value) ? Arm.LEFT : Arm.RIGHT;
}
```

### DIFF
--- a/mappings/net/minecraft/client/options/GameOptions.mapping
+++ b/mappings/net/minecraft/client/options/GameOptions.mapping
@@ -85,7 +85,7 @@ CLASS cvr net/minecraft/client/options/GameOptions
 	FIELD n hideServerAddress Z
 	FIELD o advancedItemTooltips Z
 	FIELD p pauseOnLostFocus Z
-	FIELD q mainHand Lait;
+	FIELD q mainArm Lait;
 	FIELD r overrideWidth I
 	FIELD s overrideHeight I
 	FIELD t heldItemTooltips Z
--- a/mappings/net/minecraft/client/render/entity/model/BipedEntityModel.mapping
+++ b/mappings/net/minecraft/client/render/entity/model/BipedEntityModel.mapping
@@ -14,7 +14,9 @@ CLASS dfa net/minecraft/client/render/entity/model/BipedEntityModel
 		ARG 1 scale
 		ARG 3 textureWidth
 	METHOD a getArm (Lait;)Ldhg;
-	METHOD a getPreferedHand (Laiu;)Lait;
+		ARG 1 arm
+	METHOD a getPreferredArm (Laiu;)Lait;
+		ARG 1 entity
 	METHOD a setAttributes (Ldfa;)V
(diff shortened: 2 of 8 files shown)


asiekierka:
kashike's vote holds a lot of weight to me, so I'll side with kashike

LemmaEOF:
I think this makes a lot more sense than making it about hands.

LemmaEOF (approved):

NeunEinser (approved):

liach (approved):

modmuss50 (approved):

shedaniel (approved):

Runemoro:
By the way, the Mojang name for this is for sure not "Arm". It's probably "Hand":
```
    net/minecraft/entity/EquipmentSlot (ain)
    net/minecraft/entity/ExperienceOrbEntity (aio)
    net/minecraft/entity/mob/FlyingEntity (aip)
    net/minecraft/util/AbsoluteHand (aiq) <---- Hand
    net/minecraft/entity/LivingEntity (air)
    net/minecraft/entity/mob/MobEntity (ais)
```

The change from hand to arm makes no sense. In Minecraft, arms and hands are exactly the same thing. This class (`AbsoluteHand`) as well as the other class (`Hand`) can both be used to represent both an arm or a hand. 

Also, the strings in `AbsoluteHand` call it a "hand", not an "arm".

Pyrofab:
I think calling it just `Hand` would be a mistake, as it would conflict with the other one. And the other one should not be changed to accomodate this one, as it is used way more often. It is also effectively used as an arm (yes, both are basically the same thing, but when rendered it is more accurate to call it an arm).
Note I don't personally have anything against `AbsoluteHand` but I do not consider the change to be terrible either.

Runemoro:
> I think calling it just Hand would be a mistake, as it would conflict with the other one

I agree. But so would calling it `Arm`, because they are interchangable.

> It is also effectively used as an arm

In `options.txt`, the `mainHand` option is an `AbsoluteHand` (which you want to name arm). That would be confusing.

modmuss50:
Merging due to overwhelming support, any further comments should go as an issue or a PR. Thanks for understanding.

PR #749, opened 2019-06-12
Identifier.ofNullable -> Identifier.tryParse

ChloeDawn:
It is not the argument that is nullable, it parses the given `String` into an `Identifier`, and returns null if it is invalid. 

Example use:
```java
@Nullable Identifier id = Identifier.tryParse(this.getPossibilyInvalidPath());
```


liach:
Consider how guava's `Ints.tryParse` returns a nullable `Integer`, I would name this method `tryParse`.

ChloeDawn:
Updated accordingly.

liach (approved):

PR #752, opened 2019-06-15
Fix #750

### LINKED ISSUE #750: Make Item/Stack naming consistent in entity and player classes
Inside PlayerEntity class, methods use *Stack* or *Item* in their name rather randomly. That's just terrible for search or IDE suggestions if you don't remember exact name. Also, a few names are unnecessary complicated, overly long and in backwards order.

Majority of names use *Stack*, but I suggest change everything to *Item* as doing other way would result in worse names (`isUsingStack`, `clearActiveStack`, `getStackCooldownManager` just to name a few). Probably this is can also be done in `Inventory` classes.
```
dropStack -> dropItem
dropEquippedStack -> getEquippedItem
getMainHandStack -> getMainHandItem
getOffHandStack -> getOffHandItem
getStackInHand -> getHeldItem

```

Items in head/chest/legs/feet slots don't have to be armor
```
getArmorItems -> getWornItems
```

`getItemsHand` was probably meant to be counterpart to `getItemsEqipped` but mistyped word *held* into *hand*. Also, both don't need to be in backwards order.
```
getItemsEquipped -> getEquippedItems
getItemsHand -> getHeldItems
```

this one doesn't seem to check any "validity" other than simply presence of item in slot
```java
public boolean isEquippedStackValid(EquipmentSlot equipmentSlot_1) {
  return !this.getEquippedStack(equipmentSlot_1).isEmpty();
}
```
```
isEquippedStackValid -> hasItemInSlot or hasEquippedItem 
```

*ItemStack* is unnecessary here, just `Item` is ok
```
giveItemStack -> giveItem
```
or it can be `insertItem` as in `Inventory`, those methods do basically same thing

asiekierka:
Up for discussion regarding other places which should have Stack->Item applied

### DIFF
--- a/mappings/net/minecraft/entity/LivingEntity.mapping
+++ b/mappings/net/minecraft/entity/LivingEntity.mapping
@@ -30,7 +30,7 @@ CLASS aiu net/minecraft/entity/LivingEntity
 	FIELD bb sidewaysSpeed F
 	FIELD bc upwardSpeed F
 	FIELD bd forwardSpeed F
-	FIELD bn activeItemStack Lbce;
+	FIELD bn activeItem Lbce;
 	FIELD bo itemUseTimeLeft I
 	FIELD br brain Lajj;
 	FIELD bs SLEEPING_POSITION Lqi;
@@ -60,7 +60,7 @@ CLASS aiu net/minecraft/entity/LivingEntity
 		ARG 5 z
 		ARG 7 particleEffects
 	METHOD a swingHand (Lahf;)V
-	METHOD a setStackInHand (Lahf;Lbce;)V
+	METHOD a setHeldItem (Lahf;Lbce;)V
 		ARG 1 hand
 	METHOD a onDeath (Lahu;)V
 		ARG 1 damageSource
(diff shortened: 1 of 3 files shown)


NeunEinser:
I don't think this should be done in general. While individual names may sound better, all of these methods operate on an Item**Stack**.
For example, you are not dropping a single item in `dropStack`/`dropItem`, but an entire stack.

Also, if you want to go for that, method parameters are still inconsistent with the new namings.

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -106,17 +106,17 @@ CLASS aiu net/minecraft/entity/LivingEntity
 		ARG 1 jumping
 	METHOD b heal (F)V
 		ARG 1 amount
-	METHOD b getStackInHand (Lahf;)Lbce;
+	METHOD b getItemInHand (Lahf;)Lbce;
```
Gegy, line 109: Should this change to `getHeldItem` in line with the setter?

asiekierka: Yes, but I don't think this will be merged before the Intermediary move

Runemoro:
I agree with @NeunEinser. We shouldn't use "item" when there might be more than a stack of more than one item involved.

quat1024:
^ That, but `stack` always when referring to any `ItemStack` even if you expect it to have a count of 1. Reserve `Item` for only actual `Item`s.

modmuss50:
Not fully approved, and I feel needs further discussion. Seems easy enough to update.

Sollace:
I agree with most of the responses here. We should normalize on using "stack" and only use "item" for usages of Item.

modmuss50:
Closing this as it would need redoing for intermediary names. And it seems most people are against it.

PR #753, opened 2019-06-15
Fix #723

### LINKED ISSUE #723: LoopingStream -> MultiUseStream
The name came from https://github.com/FabricMC/yarn/pull/338#discussion_r242899941

The comment is correctly describing the functionality, the class allows using a stream multiple times. But in my opinion the current name "LoopingStream" suggests that the stream continues at the beginning once it reached the end.

A name like "MultiUseStream" might fit better. Similarly the method `getStream` should be named `newStream` to indicate that a new stream and not an existing one is returned.

The fields could use the following names:

Obfuscated | Deobfuscated
--- | ---
`a` | `collectedElements`
`b` | `delegate` / `source`

And if it is possible to name the field of the anonymous class, then it could be named `index` / `currentIndex`.

asiekierka:


mappings/net/minecraft/util/ReusableStream.mapping
```diff
@@ -0,0 +1,10 @@
+CLASS zx net/minecraft/util/ReusableStream
+	CLASS zx$1
+		FIELD b pos I
+		METHOD tryAdvance (Ljava/util/function/Consumer;)Z
+			ARG 1 consumer
+	FIELD a collectedElements Ljava/util/List;
+	FIELD b source Ljava/util/Spliterator;
+	METHOD <init> (Ljava/util/stream/Stream;)V
+		ARG 1 stream
+	METHOD a getStream ()Ljava/util/stream/Stream;
```
NeunEinser: `newStream` as mentioned in the issue.

asiekierka: Why not just .stream()?

NeunEinser: I was just going by what Marcono1234 said in the original comment. I was expecting a static method here, but looking at the code `stream` is the better name as it just creates a new stream that streams all the source stream's elements.

asiekierka: Done

NeunEinser (approved):

Gegy (approved):

PR #755, opened 2019-06-15
Mapping additions

mikroskeem:
I saw few unmapped methods and fields I use/hook into, so I mapped them.

### DIFF
--- a/mappings/net/minecraft/server/network/ServerPlayerEntity.mapping
+++ b/mappings/net/minecraft/server/network/ServerPlayerEntity.mapping
@@ -16,6 +16,7 @@ CLASS vj net/minecraft/server/network/ServerPlayerEntity
 	FIELD cv cameraPosition Lfo;
 	FIELD cw containerSyncId I
 	FIELD d interactionManager Lvk;
+	FIELD f pingMilliseconds I
 	FIELD g notInAnyWorld Z
 	METHOD A getStatHandler ()Lys;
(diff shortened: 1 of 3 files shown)


mappings/net/minecraft/server/network/ServerPlayerEntity.mapping
```diff
@@ -16,6 +16,7 @@ CLASS vj net/minecraft/server/network/ServerPlayerEntity
 	FIELD cv cameraPosition Lfo;
 	FIELD cw containerSyncId I
 	FIELD d interactionManager Lvk;
+	FIELD f ping I
```
asiekierka: Too vague.

mikroskeem: Other options I could come up with are `pingMilliseconds` or shorter `pingMs`

asiekierka: Either would be better.

Runemoro: Personally I think `ping` was enough. The same way we don't use `distanceBlocks` or `colorArgb`, we shouldn't use `pingMilliseconds`. The unit belongs in a javadoc, not the name.

modmuss50 (approved):
Seems fine to me

modmuss50:
Merging due to update, any more comments should go into an issue or another PR. Thanks :)

PR #756, opened 2019-06-18
Fix #751

### LINKED ISSUE #751: Use unique names for XxxArgumentType#create
[Brigadier ArgumentTypes](https://github.com/Mojang/brigadier/tree/master/src/main/java/com/mojang/brigadier/arguments) always use a unique name for their static constructors and unwrappers, i.e. `IntegerArgumentType.integer` instead of `IntegerArgumentType.create`. 

This is probably so you can statically import these methods from as many ArgumentTypes as you want. The code snippet on the Brigadier github readme uses a ton of static imports.

However since Yarn maps all of the Minecraft argument type constructor methods as `create`, you lose that ability due to name collisions.

So, `BlockPosArgumentType.create` → `BlockPosArgumentType.blockPos`, for all Minecraft argument types. Consistent with Mojang, allows for cleaner static-import code.

NeunEinser:
Additionally maps some ArgumentType parameters.
#751 Has 6 upvotes and no downvotes. I think the pattern to use static imports is quite nice and should be enabled in mods.
Plus it makes things consistent with e.g. `EntityArgumentType`, where multiple parameterless create methods are already named in the same fashion.

### DIFF
--- a/mappings/net/minecraft/command/arguments/BlockArgumentParser.mapping
+++ b/mappings/net/minecraft/command/arguments/BlockArgumentParser.mapping
@@ -18,6 +18,9 @@ CLASS dh net/minecraft/command/arguments/BlockArgumentParser
 	FIELD q tagId Lqt;
 	FIELD r cursorPos I
 	FIELD s suggestions Ljava/util/function/Function;
+	METHOD <init> (Lcom/mojang/brigadier/StringReader;Z)V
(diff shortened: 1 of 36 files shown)


modmuss50 (approved):

modmuss50:
Merging due to update, any more comments to a PR or issue, thanks a lot 👍

PR #695, opened 2019-05-13
Add proposed field names to the mapped named jar

liach:
Since @modmuss50 pointed out in #674 that these proposed names should not exist in the tiny gzip'd data, I decided to decompress the tiny data, put enum names in, and then finally map the jar.

Signed-off-by: liach <liach@users.noreply.github.com>

PR #757, opened 2019-06-18
Explain how to use an IDE in README.md

Lytigas:
masa explained this to me in the scicraft discord after he found out I was using enigma directly. I didn't know better because there wasn't documentation, so I figured I'd document what little I know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,12 @@ However, it is a good idea to consult name changes with other people - use pull 
 2. Run `./gradlew yarn` (Linux, macOS) or `gradlew yarn` (Windows)
 3. Profit
 
+To use Yarn in your IDE of choice, or as part of a full Fabric development environment, check out the [Fabric example mod](https://github.com/FabricMC/fabric-example-mod).
+
+1. Fork/Clone the repo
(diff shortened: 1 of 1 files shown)


modmuss50:
wat? this repo doesnt have genSources? there is no reason to use an ide on this repo?

NeunEinser (commented):
Yarn is only about providing mappings, which enigma is intended to be used for.

If you want to only look at decompiled code, yarn might simply not be the right place for that. This doku does not belong into yarn though, more in a general info page about fabric.

liach (changes requested):
Seriously, look at #695. That pr already includes how you gets a deobfuscated minecraft combined jar `mapNamedJar` which will give deobfuscated source after piping through a decompiler.

The fabric-loom usage of yarn is not related to yarn itself and should not be present!

liach:
I think I will add to my readme change in #695 that for vanilla research purposes, `mapNamedJar` is the most feasible; for modding purposes, fabric-loom will remap the jar for you.

Lytigas:
Clearly I don't have a full grasp of the way these components interact yet; sorry to bother. If there's a guide and/or documentation I missed, I'd love to see it; otherwise I hope it can be worked on soon.

PR #758, opened 2019-06-20
Mapped two unmapped classes and some villager stuff

liach:

### DIFF
--- a/mappings/net/minecraft/entity/passive/VillagerEntity.mapping
+++ b/mappings/net/minecraft/entity/passive/VillagerEntity.mapping
@@ -9,6 +9,7 @@ CLASS avo net/minecraft/entity/passive/VillagerEntity
 	FIELD bI foodLevel B
 	FIELD bJ gossip Lapb;
 	FIELD bK gossipStartTime J
+	FIELD bL lastGossipDecay J
 	FIELD bM experience I
 	FIELD bN lastRestock J
 	FIELD bO MEMORY_MODULES Lcom/google/common/collect/ImmutableList;
@@ -17,6 +18,13 @@ CLASS avo net/minecraft/entity/passive/VillagerEntity
 		ARG 1 entityType
 		ARG 2 world
 		ARG 3 type
+	METHOD a canSummonGolem (J)Z
+		ARG 1 time
+	METHOD a summonGolem (JI)V
+		ARG 1 time
+		ARG 3 requiredVillagerCount
+	METHOD a notifyDeath (Lail;)V
+		ARG 1 killer
 	METHOD a initBrain (Lajj;)V
 		ARG 1 brain
 	METHOD a releaseTicketFor (Lape;)V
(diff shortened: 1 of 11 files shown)


LemmaEOF (approved):
Looks good to me.

Runemoro (approved):

mappings/net/minecraft/inventory/Inventory.mapping
```diff
@@ -11,7 +11,7 @@ CLASS agz net/minecraft/inventory/Inventory
 		ARG 2 stack
 	METHOD a canPlayerUseInv (Lawb;)Z
 		ARG 1 player
-	METHOD a getInvAmountOf (Lbbz;)I
+	METHOD a countInInv (Lbbz;)I
```
kashike, line 14: This sounds weird.

liach: It counts the occurrence of an item type in this inventory. What do you suggest?

valoeghese: this mapping seems perfectly fine to me

mappings/net/minecraft/entity/passive/VillagerEntity.mapping
```diff
@@ -9,6 +9,7 @@ CLASS avo net/minecraft/entity/passive/VillagerEntity
 	FIELD bI foodLevel B
 	FIELD bJ gossip Lapb;
 	FIELD bK gossipStartTime J
+	FIELD bL lastGossipDecay J
```
valoeghese, line 12: what is "lastGossipDecay" supposed to do? maybe this can be clarified.

mappings/net/minecraft/inventory/Inventory.mapping
Blayyke: getCountOf would make more sense if we ever remove the Inv parts of the method names for Inventory methods, but I think this change is fine.

asiekierka: Same, it's fine to me

mappings/net/minecraft/entity/passive/VillagerEntity.mapping
```diff
@@ -26,22 +34,33 @@ CLASS avo net/minecraft/entity/passive/VillagerEntity
...
 	METHOD eq wantsToStartBreeding ()Z
 	METHOD er canBreed ()Z
 	METHOD es hasSeedToPlant ()Z
 	METHOD et getLastRestock ()J
 	METHOD ev sayNo ()V
 	METHOD ew clearCurrentBonus ()V
+	METHOD ex lacksFood ()Z
 	METHOD ey canLevelUp ()Z
 	METHOD ez levelUp ()V
+	METHOD f getReputationFor (Lawb;)I
```
valoeghese: this seems fine, but would getReputation be a more concise way of naming it?

liach: feared that it may be confusing because it passes down to another reputation getter in another class which has another param (only getting reputation sum from certain types of gossips)

mappings/net/minecraft/entity/passive/VillagerEntity.mapping
liach: This is the last timestamp of a gossip decay (as it appears as `LastGossipDecay` in nbt)
After a while (like 5 minutes?) all gossips will undergo a time-caused decay and its value will decrease (will be removed when its value is less than 2) there is also a gossip exchange mechanism as well, and there is another decrement for strength of gossip when it is shared.

modmuss50 (approved):

liach:
Anything else wrong with this?

modmuss50:
I dont think so, looks good to merge.

PR #759, opened 2019-06-20
Add a decompileCFR task

modmuss50:
Might be useful if you want to diff the source.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -486,7 +486,7 @@ task decompileCFR(dependsOn: mapNamedJar, type: JavaExec) {
 
 	classpath = configurations.cfrRuntime
 	main = 'org.benf.cfr.reader.Main'
-	args '--outputdir', 'src', '--extraclasspath', Stream.of(libraries.listFiles()).map{it.name}.collect(Collectors.joining(";")), namedJar.getAbsolutePath()
+	args '--outputdir', 'src', namedJar.getAbsolutePath()
 }
 
 publishing {
--- a/build.gradle
+++ b/build.gradle
@@ -88,6 +88,9 @@ import net.fabricmc.weave.CommandFindMappingErrors
 import net.fabricmc.weave.CommandTinyify
 import org.apache.commons.io.FileUtils
 
+import java.util.stream.Stream
+import java.util.stream.Collectors
+
 import java.util.zip.GZIPOutputStream
 
 boolean validateChecksum(File file, String checksum) {
@@ -483,7 +486,7 @@ task decompileCFR(dependsOn: mapNamedJar, type: JavaExec) {
 
 	classpath = configurations.cfrRuntime
 	main = 'org.benf.cfr.reader.Main'
-	args '--outputdir', 'src', namedJar.getAbsolutePath()
+	args '--outputdir', 'src', '--extraclasspath', Stream.of(libraries.listFiles()).map{it.name}.collect(Collectors.joining(";")), namedJar.getAbsolutePath()
 }
 
 publishing {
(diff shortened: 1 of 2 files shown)


liach (approved):
Mind adding this to readme as well.

build.gradle
```diff
@@ -473,6 +476,15 @@ task mapNamedJar(dependsOn: [buildTinyWithEnum, mapIntermediaryJar]) {
...
 }
 
+task decompileCFR(dependsOn: mapNamedJar, type: JavaExec) {
+	doFirst {
+		file('src').deleteDir()
+	}
+
+	classpath = configurations.cfrRuntime
+	main = 'org.benf.cfr.reader.Main'
+	args '--outputdir', 'src', namedJar.getAbsolutePath()
```
liach, line 489: Given we have already get the libraries when we've map the jar, we can just supply the libraries to the decompiler. The flag is `--extraclasspath`, and it seems to take a parameter like java's classpath (a string)

liach (changes requested):

build.gradle
modmuss50: It seems to decompile just fine without it, ill look into setting that up but I dont think its needed.

modmuss50: Done.

liach (approved):

modmuss50:
humm, that last change was bad as it assumes the libs have been downloaded before the task is evaluated, and in the cast of Travis they hadnt

liach:
I think you should have written `{it -> it.name}` instead of just `{it.name}` possibly

ChloeDawn:
`{it.name}` is perfectly valid. Although I would personally add whitespace for readability: `{ it.name }`

modmuss50:
Yeah, the issue is the library’s haven’t been downloaded when the args are processed. I’m going to diff the source with and without the libs to see if it’s even needed.

build.gradle
```diff
@@ -473,6 +479,15 @@ task mapNamedJar(dependsOn: [buildTinyWithEnum, mapIntermediaryJar]) {
...
 }
 
+task decompileCFR(dependsOn: mapNamedJar, type: JavaExec) {
+	doFirst {
+		file('src').deleteDir()
+	}
+
+	classpath = configurations.cfrRuntime
+	main = 'org.benf.cfr.reader.Main'
+	args '--outputdir', 'src', '--extraclasspath', Stream.of(libraries.listFiles()).map{it.name}.collect(Collectors.joining(";")), namedJar.getAbsolutePath()
```
ChloeDawn:
```groovy
Stream.of(libraries.listFiles()).map{it.name}.collect(Collectors.joining(";"))
```
could just be
```groovy
libraries.listFiles().collect { it.name }.join ';'
```

ChloeDawn (changes requested):

build.gradle
modmuss50: yeah, thats better but it doesnt solve the first issue of the libs not being there. ive removed adding the libs as the code looks fine without it (I havent been able to diff it just yet)

liach:
mind updating the base branch the pull request is targetting?

Runemoro (approved):

Runemoro:
> Undo adding libs as I dont think its required

It is required. Without libs, the decompiler will:

 1. Be forced to add casts to every library method call (`LibraryClass.doSomething((Something) something)`) since it doesn't know that the call isn't ambiguous.
 2. Be unable to correctly infer generics or remove unchecked casts since it needs the function signature which is in the library class.

liach:
now as we always download libs after the intermediary migration, consider adding libs back.

Issue #761, opened 2019-06-25
MinecraftServer.WORLD_INFO is the prefilled world info for beta


Blayyke:
Could probably be named as DEMO_WORLD_INFO or something. 
```java
   static {
      WORLD_INFO = (new LevelInfo((long)"North Carolina".hashCode(), GameMode.SURVIVAL, true, false, LevelGeneratorType.DEFAULT)).setBonusChest();
   }
```

Issue #760, opened 2019-06-25
MinecraftServer#isServerThreadAlive is the opposite of what it does


Blayyke:
This method is as below. It is never overridden, 
```java
   @Environment(EnvType.CLIENT)
   public boolean isServerThreadAlive() {
      return !this.serverThread.isAlive();
   }
```

and it is only called in MinecraftClient
```java
            while(!integratedServer_1.isServerThreadAlive()) {
               this.render(false);
            }
``` 
here.

Issue #704, opened 2019-05-16
PauseScreen -> GameMenuScreen


Prospector:
In-game, it's called a game menu. In multiplayer, the game is not paused when this screen is opened, and in addition to that you can pause the game in singleplayer without the game menu screen open.

PR #762, opened 2019-06-25
Fix three easy issues

quat1024:
Three issues that had unanimous thumb-ups and seem like nobrainers:

#761  `WORLD_INFO` -> `DEMO_LEVEL_INFO`

#760 `MinecraftServer#isServerThreadAlive` -> ~~`isServerThreadDead`~~ `isStopping`

#704 `PauseScreen` -> `GameMenuScreen`

### DIFF
--- a/mappings/net/minecraft/server/MinecraftServer.mapping
+++ b/mappings/net/minecraft/server/MinecraftServer.mapping
@@ -235,7 +235,7 @@ CLASS net/minecraft/server/MinecraftServer
 	METHOD t exit ()V
 	METHOD u isNetherAllowed ()Z
 	METHOD v start ()V
-	METHOD w isServerThreadDead ()Z
+	METHOD w isStopping ()Z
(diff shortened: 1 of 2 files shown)


liach (approved):

mappings/net/minecraft/server/MinecraftServer.mapping
```diff
@@ -235,7 +235,7 @@ CLASS net/minecraft/server/MinecraftServer
 	METHOD t exit ()V
 	METHOD u isNetherAllowed ()Z
 	METHOD v start ()V
-	METHOD w isServerThreadAlive ()Z
+	METHOD w isServerThreadDead ()Z
```
ChloeDawn:
```suggestion
	METHOD w isStopping ()Z
```
Used by the client to determine whether the integrated server is shutting down.

ChloeDawn (changes requested):

PR #763, opened 2019-06-25
Replace Weave commands with Enigma commands

Runemoro:
This PR replaces the findMappingErrors and tinyfy commands from Weave with the checkMappings and convertMappings commands from Enigma, fully replacing Weave with Enigma (see https://github.com/FabricMC/weave/issues/2).

I checked that the output produced by `gradlew build` is exactly the same as before.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
 		}
 	}
 	dependencies {
-		classpath "net.fabricmc:weave:0.4.0.19"
+		classpath "cuchaz:enigma:0.14.0.122"
 		classpath "net.fabricmc:stitch:0.2.0.52"
 		classpath "commons-io:commons-io:2.6"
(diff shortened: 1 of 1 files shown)


modmuss50 (approved):
Looks fine to me

build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
 		}
 	}
 	dependencies {
-		classpath "net.fabricmc:weave:0.4.0.19"
+		classpath "cuchaz:enigma:0.14.0.122"
```
liach, line 10: Mind using a `+` to keep up with future updates?

liach (approved):

build.gradle
Prospector: shouldn't really imo. That way if an enigma update breaks people can still use yarn as if nothing happened

liach: eh fine

PR #767, opened 2019-06-29
Update stitch and buildscript to improve name guessing

liach:
See FabricMC/stitch#12

Dunno why the build does not appear on [maven](https://maven.fabricmc.net/net/fabricmc/stitch/), but it appears on [jenkins](https://jenkins.modmuss50.me/job/FabricMC/job/stitch/job/master/61/).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ buildscript {
 	}
 	dependencies {
 		classpath "cuchaz:enigma:0.14.0.128"
-		classpath "net.fabricmc:stitch:0.2.1.60"
+		classpath "net.fabricmc:stitch:0.2.1.61"
 		classpath "commons-io:commons-io:2.6"
 		classpath "com.google.guava:guava:28.0-jre"
 		classpath 'de.undercouch:gradle-download-task:3.4.3'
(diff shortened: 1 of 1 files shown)


asiekierka:
It's CloudFlare caching, if you're curious.

build.gradle
```diff
@@ -515,8 +516,7 @@ task buildTinyWithEnum(dependsOn: "mergeTiny", type: FileOutput) {
 		String[] argsPropose = [
 			mergedFile.getAbsolutePath(), // must use official jar
 			noEnum.getAbsolutePath(),
-			namedWithEnum.getAbsolutePath(),
-			"--writeAll"
```
asiekierka, line 519: Why'd you remove this?

asiekierka:
Also, the improved name guessing is already in - build 61 adds your *other* PR, which performs the act of changing the semantics of a command.

build.gradle
liach: The arg was added incorrectly before (before this took 3 args, but since we don't go thorough command line main it did not report the wrong number of args (4 instead of 3)). Now the 4th arg is the input jar mapping.

liach:
yes, this pr is for the changing sematics.

build.gradle
asiekierka: Ah, I see. So it didn't do anything... I guess I forgot to merge a feature, hah

PR #772, opened 2019-06-30
Mapped NativeImage

Sollace:
Oooooooh boy. This is a tricky one. Let's see if I can describe the reasoning behind all of them, starting with the easiest;

`method_4304` -> `copyRect`

This is being used to convert old player skins to the new format (64x32 -> 64x64) at runtime in the game (I bet a lot of people didn't know this, but the game actually does that).

Parameters were determined by comparing the current code to pre-1.13, and the documentation for AWT Graphics. In the end I was able to determine the parameters to be:
(srcX, srcY, dstX, dstY, width, height, flipX, flipY)

The last two were a bit hard to discern, but they do affect how the x/y values are calculated during copying as such:

                int offsetX= flipX ? (width - 1 - x) : x;
                int offsetY= flipY ? (height - 1 - y) : y;

Which makes me confident that these are for flipping the image whilst copying. X is also always flipped by vanilla when copying the components for player skins.

`fillRGBA` -> `fillRect`
Changed for consistency with `copyRect`. Also last parameter is the packed colour int, if that wasn't already clear.

`resizeSubRectTo`
The first two parameters (`x`, `y`) were obvious. For the second two (`width`, and `height`) I looked at `STBImageResize.nstbir_resize_uint8` to see what its parameters were.

The last parameter was also `channels`. The number of channels, not channel bytes, thus `NativeImage.Format.bytesPerPixel` -> `NativeImage.Format.totalChannels` (which coincides nicely with the values defined by the enums)

Speaking of enums, 
`class_1013` -> `NativeImage.GLFormat`
`class_1013.field_5015` -> `NativeImage.GLFormat.glConstant`
`class_1013.getGlConstant` -> `NativeImage.GLFormat.getGlConstant`

The values stored by these enums are GLConstants (specifically image format numbers). _More_ specifically, they are the same image format numbers used in the first parameter of their corresponding `NativeImage.Format` enum constant.

`method_4306` -> `loadFromMemory`
This method is used by the `ScreenShotUtils.method_1663` and is only called when FBOs are _off_. If FBOs are on, it goes into `loadFromTextureImage` instead, which sedgeways into the next point...

`loadFromTextureImage`
Has parameters `level`, and `removeAlpha` alpha.

The first was determined by looking at `GL11.glGetTexImage` (called by GLStateManager.getTexImage) in this method.

Second one comes from looking at the code in the method:

      if (removeAlpha && this.format.hasAlphaChannel()) {
            for (int int_2 = 0; int_2 < this.getHeight(); ++int_2) {
                for (int int_3 = 0; int_3 < this.getWidth(); ++int_3) {
                    this.setPixelRGBA(int_3, int_2, this.getPixelRGBA(int_3, int_2) | 255 << this.format.getAlphaChannelOffset());
                }
            }
        }

What it's doing here is iterating through the image and setting every pixel to the same colour, with alpha removed. From this I was able to work backwards and make the following assertions:

`method_4329` -> `hasAlphaChannel`
`method_4332` -> `getAlphaChannelOffset`
`field_5006` -> `alphaChannelOffset`
`method_4330` `getUniversalOpacityOffset` (method will either return the luminosity or alpha offset. Usually Luminosity if supported.)
(not counting things that were mapped, just not mapped everywhere)

R/G/B parameters and offsets are based on the actual values visible in the enum.

First three booleans are true, and have non-255 values for their corresponding ints for the RGB/RGBA enums, whilst both Luminosity enums have them to false and either all 255 or all 0. I went with that ordering because it's the norm, and I see no reason Mojang would do anything other.

`NativeImage.Format.writeable` come from this check in `writeToFile`:

        if (!this.format.isWriteable()) {
            throw new UnsupportedOperationException("Don't know how to write format " + this.format);
        }

`NativeImage.WriteCallback.invoke`
Parameters are `context`, `data`, and `size` (as stated by `STBIWriteCallback.invoke`).

In `makeGlyphBitmapSubpixel` the parameter names can be determined, again, by this string:

throw new IllegalArgumentException(String.format("Out of bounds: start: (%s, %s) (size: %sx%s); size: %sx%s", startX, startY, width, height, this.getWidth(), this.getHeight()))

So we have `startX` and `glyth` -> `startY`. (Sorry, whoever named that glyth)

Aaaand that's about it! I think that should cover everything. There is still `method_4319` which I don't know about, and some boolean parameters that I can't yet see what they're intended to be, but otherwise everything else should be set.


mappings/net/minecraft/client/texture/NativeImage.mapping
```diff
@@ -1,15 +1,57 @@
...
+			ARG 7 blue
+			ARG 8 luminance
+			ARG 9 alpha
+			ARG 10 redOffset
+			ARG 11 greenOffset
+			ARG 12 blueOffset
+			ARG 13 luminanceOffset
+			ARG 14 alphaOffset
+			ARG 15 writeable
+		METHOD method_4329 hasAlphaChannel ()Z
```
ChloeDawn, line 30: Does it only have an alpha channel if it has translucent pixels? If so, maybe `hasTranslucency`?

ChloeDawn (changes requested):

ChloeDawn:
`fromByteBuffer` and `fromInputStream` methods could just be `read`

mappings/net/minecraft/client/texture/NativeImage.mapping
Sollace: alpha = translucency

Alpha is the opacity channel in almost every format I know. The only one that _doesn't_ have alpha, is gifs. Those use an image-wide mapping to assign a specific colour to transparent pixels and don't support translucency at all.

Sollace:
> 
> 
> `fromByteBuffer` and `fromInputStream` methods could just be `read`

So like this?

fromBase64 -> read
fromByteBuffer -> read
fromInputStream -> read

ChloeDawn:
Yes

mappings/net/minecraft/client/texture/NativeImage.mapping
```diff
@@ -27,16 +72,44 @@ CLASS net/minecraft/class_1011 net/minecraft/client/texture/NativeImage
 		ARG 2 width
 		ARG 3 height
 		ARG 5 pointer
```
ChloeDawn: This is a pointer address, not a pointer

Sollace: So... `memoryAddress`? Either way, I don't see why we can't call it `pointer`. They seem like the same thing.

liach (approved):
Not a pro on client stuff, but the naming style looks great

mappings/net/minecraft/client/texture/NativeImage.mapping
```diff
@@ -1,15 +1,57 @@
 CLASS net/minecraft/class_1011 net/minecraft/client/texture/NativeImage
 	CLASS class_1012 Format
 		FIELD field_4993 pixelDataFormat I
-		FIELD field_4994 bytesPerPixel I
+		FIELD field_4994 totalChannels I
```
Runemoro: Are you sure?

1. If I remember correctly, it's passed to a GL function whose parameter is called "bytes per pixel" (which is not necessarily the number of channels, a channel [is not necessarily one byte](https://en.wikipedia.org/wiki/Color_depth#Deep_color_(30/36/48-bit))).
2. What does "total" mean here?
3. It should be `channelCount` (or `totalChannelCount`), it's just the channel count, not the channels themselves.

mappings/net/minecraft/client/texture/NativeImage.mapping
```diff
@@ -1,15 +1,57 @@
...
+		FIELD field_4996 writeable Z
+		FIELD field_4999 alpha Z
+		FIELD field_5000 luminance Z
```
Runemoro: These are booleans. I think `hasAlpha` and `hasLuminance` (or `hasAlphaChannel` and `hasLuminanceChannel`) would be less confusing.

mappings/net/minecraft/client/texture/NativeImage.mapping
```diff
@@ -1,15 +1,57 @@
...
+			ARG 8 luminance
+			ARG 9 alpha
+			ARG 10 redOffset
+			ARG 11 greenOffset
+			ARG 12 blueOffset
+			ARG 13 luminanceOffset
+			ARG 14 alphaOffset
+			ARG 15 writeable
+		METHOD method_4329 hasAlphaChannel ()Z
+		METHOD method_4330 getUniversalOpacityOffset ()I
```
Runemoro: What do you mean by "universal"?

mappings/net/minecraft/client/texture/NativeImage.mapping
Runemoro: We need to agree on a convention for this. Throughout yarn, LWJGL addresses are called `address`, `pointer`, and `id`.

mappings/net/minecraft/client/texture/NativeImage.mapping
```diff
@@ -27,20 +72,48 @@ CLASS net/minecraft/class_1011 net/minecraft/client/texture/NativeImage
...
+		ARG 3 height
+	METHOD <init> (Lnet/minecraft/class_1011$class_1012;IIZJ)V
+		ARG 2 width
+		ARG 3 height
+		ARG 5 glPointer
 	METHOD method_15877 writeFile (Ljava/lang/String;)V
-	METHOD method_15990 fromBase64 (Ljava/lang/String;)Lnet/minecraft/class_1011;
+		ARG 1 fileName
+	METHOD method_15990 read (Ljava/lang/String;)Lnet/minecraft/class_1011;
 		ARG 0 string
```
Runemoro: In `writeFile`, the string is a file name, and in `read`, the string is the image encoded in base64? This is confusing, the parameter name should mention it's a base 64 string.

mappings/net/minecraft/client/texture/NativeImage.mapping
```diff
@@ -27,20 +72,48 @@ CLASS net/minecraft/class_1011 net/minecraft/client/texture/NativeImage
...
+		ARG 2 xOffset
+		ARG 3 yOffset
 	METHOD method_4302 untrack ()V
-	METHOD method_4303 fromByteBuffer (Lnet/minecraft/class_1011$class_1012;Ljava/nio/ByteBuffer;)Lnet/minecraft/class_1011;
+	METHOD method_4303 read (Lnet/minecraft/class_1011$class_1012;Ljava/nio/ByteBuffer;)Lnet/minecraft/class_1011;
+	METHOD method_4304 copyRect (IIIIIIZZ)V
+		ARG 1 srcX
+		ARG 2 srcY
+		ARG 3 dstX
+		ARG 4 dstY
```
Runemoro: I suggest "rectangle" (or "copy", since the parameters make it clear it's a rectangle), "source" and "destination" (or "target"). Mre abbrs makes rding the code take more effrt.

mappings/net/minecraft/client/texture/NativeImage.mapping
```diff
@@ -27,20 +72,48 @@ CLASS net/minecraft/class_1011 net/minecraft/client/texture/NativeImage
 		ARG 2 width
 		ARG 3 height
 		ARG 5 pointer
+	METHOD <init> (Lnet/minecraft/class_1011$class_1012;IIZ)V
+		ARG 2 width
+		ARG 3 height
+	METHOD <init> (Lnet/minecraft/class_1011$class_1012;IIZJ)V
+		ARG 2 width
+		ARG 3 height
+		ARG 5 glPointer
```
Runemoro: This should match the parameter name in the constructor above (`pointer`)

mappings/net/minecraft/client/texture/NativeImage.mapping
```diff
@@ -82,12 +157,18 @@ CLASS net/minecraft/class_1011 net/minecraft/client/texture/NativeImage
 		ARG 9 clamp
 	METHOD method_4322 makePixelArray ()[I
 	METHOD method_4323 getHeight ()I
-	METHOD method_4324 fromByteBuffer (Ljava/nio/ByteBuffer;)Lnet/minecraft/class_1011;
+	METHOD method_4324 read (Ljava/nio/ByteBuffer;)Lnet/minecraft/class_1011;
 	METHOD method_4325 writeFile (Ljava/io/File;)V
-	METHOD method_4326 fillRGBA (IIIII)V
+	METHOD method_4326 fillRect (IIIII)V
```
Runemoro, line 164: Isn't just `fill` enough?

mappings/net/minecraft/client/texture/NativeImage.mapping
Sollace: Funny thing. I was just reading the CONVENTIONS.md last night, and it specifically said _not_ to do what you just suggested.

Runemoro: > has prefixes **when possible** (colored, not isColored or hasColor).

The word "color" has an adjective form (colored), so we can avoid "has" here. This is not the case for "luminance". For "alpha", "translucent" could work instead of "hasAlpha".

LemmaEOF: `hasAlpha` and `hasLuminance` are more descriptive without being verbose, because there's no direct adjective form of `alpha` and `luminance` like there is `colored`.

mappings/net/minecraft/client/texture/NativeImage.mapping
Sollace: 1. It's passed to a method where the parameter is simply named "channels" and has a description saying it is the number of channels to be used in that method.
2. Total is the total number of channels the format supports, of which RGB has 3, RGBA has 4, etc.
3. I can make it `channelCount`. `totalChannelCount` is too much.

mappings/net/minecraft/client/texture/NativeImage.mapping
Sollace: I don't see anything wrong with the current name. Also what is `target`? Is that the source or the destination?

mappings/net/minecraft/client/texture/NativeImage.mapping
Runemoro: > the total number of channels the format supports

Still not sure what "total" means here. The total of what? Is there also a non-total channel count...?

> I can make it channelCount

Yes, that's better.

mappings/net/minecraft/client/texture/NativeImage.mapping
Sollace: Agreed. We also have `handle` that's used by the window, which is technically exactly the same.

mappings/net/minecraft/client/texture/NativeImage.mapping
Sollace: Changed to `dataUri` so it's more clear.

Sollace:
`setTextureFilter`'s parameters are `blur` and `mipmap`:

    private static void setTextureFilter(boolean blur, boolean mipmap) {
        if (blur) {
            GlStateManager.texParameter(GL11.GL_TEXTURE_2D, GL11.GL_TEXTURE_MIN_FILTER, mipmap ? GL11.GL_LINEAR_MIPMAP_LINEAR : GL11.GL_LINEAR);
            GlStateManager.texParameter(GL11.GL_TEXTURE_2D, GL11.GL_TEXTURE_MAG_FILTER, GL11.GL_LINEAR);
        }
        else {
            GlStateManager.texParameter(GL11.GL_TEXTURE_2D, GL11.GL_TEXTURE_MIN_FILTER, mipmap ? GL11.GL_NEAREST_MIPMAP_LINEAR : GL11.GL_NEAREST);
            GlStateManager.texParameter(GL11.GL_TEXTURE_2D, GL11.GL_TEXTURE_MAG_FILTER, GL11.GL_NEAREST);
        }
    }

mappings/net/minecraft/client/texture/NativeImage.mapping
Sollace: I find `fillRect` a little more informative than just `fill`.

Sollace:
I made a derp. Pushed new, proper, parameter names for `copyRect` which should hopefully be more correct.

It seems my initial interpretation was correct. :/

    public void copyRect(
            int x, int y,
            int translateX, int translateY,
            int width, int height,
            boolean flipX, boolean flipY) {

        for(int offsetY = 0; offsetY < height; offsetY++) {
           for(int offsetX = 0; offsetX < width; offsetX++) {

              int pixelX = flipX ? width - 1 - offsetX : offsetX;
              int pixelY = flipY ? height - 1 - offsetY : offsetY;

              int color = getPixelRGBA(x + offsetX, y + offsetY);

              setPixelRGBA(
                      x + translateX + pixelX,
                      y + translateY + pixelY,
                      color
              );
           }
        }
     }

Runemoro (approved):

PR #775, opened 2019-07-01
Smol potion stuff

quat1024:
at the request of blayyke:

* `ThrownPotionEntity#method_7498` -> `applySplashPotion`
  * called when a splash potion lands
  * applies the splash potion effect
* `ThrownPotionEntity#method_7497` -> `applyLingeringPotion`
  * called when a lingering potion lands
  * creates an area effect cloud entity

one more:

* `LivingEntity#method_6086` -> `acceptsThrownPotions`
  * called in `applySplashPotion` and `AreaEffectCloudEntity#tick`
  * if it returns `false`, the potion effect is not applied. in vanilla armor stands have this `false`

### DIFF
--- a/mappings/net/minecraft/entity/LivingEntity.mapping
+++ b/mappings/net/minecraft/entity/LivingEntity.mapping
@@ -200,7 +200,7 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_6085 setLivingFlag (IZ)V
 		ARG 1 mask
 		ARG 2 value
-	METHOD method_6086 canApplyPotion ()Z
+	METHOD method_6086 acceptsThrownPotions ()Z
 	METHOD method_6087 pushAway (Lnet/minecraft/class_1297;)V
 	METHOD method_6088 getActiveStatusEffects ()Ljava/util/Map;
 	METHOD method_6089 containsOnlyAmbientEffects (Ljava/util/Collection;)Z
--- a/mappings/net/minecraft/entity/LivingEntity.mapping
+++ b/mappings/net/minecraft/entity/LivingEntity.mapping
@@ -200,6 +200,7 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_6085 setLivingFlag (IZ)V
 		ARG 1 mask
 		ARG 2 value
+	METHOD method_6086 canApplyPotion ()Z
 	METHOD method_6087 pushAway (Lnet/minecraft/class_1297;)V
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -200,6 +200,7 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_6085 setLivingFlag (IZ)V
 		ARG 1 mask
 		ARG 2 value
+	METHOD method_6086 canApplyPotion ()Z
```
Sollace: I'd like to use the word `immune` somewhere in there. Anyone got a suggestion?

Pyrofab: `isPotionImmune` ?
Edit: nevermind, this is the opposite of what it should be

quat1024: It's reversed, entities that return false are immune

quat1024:
Changed to `acceptsThrownPotions` through epic web editor shenanigans on my phone

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -200,6 +200,7 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_6085 setLivingFlag (IZ)V
 		ARG 1 mask
 		ARG 2 value
+	METHOD method_6086 acceptsThrownPotions ()Z
```
Runemoro: "accepts" is a weird term for this. Why not `affectedBySplashPotions`?

asiekierka: I like that suggestion., though I would debate between that and isAffectedBySplashPotions

Sollace: `doSplashPotionsHaveAnyEffectsWhenAppliedToThisEntity()`?

Pyrofab: True, `affectedBySplashPotions` would be a field name. It could be argued that `isAffectedBySplashPotions` is a tad too verbose, as Sollace noted, but at least it's clear.

quat1024: I'm a bit cautious to refer to them as "splash potions" since it also affects the behavior of lingering potions, but I suppose you could consider them a type of splash potion.

quat1024:
While `isAffectedBySplashPotions` is a bit wordy i think it's the best we've got lol

Runemoro (approved):

modmuss50 (approved):

PR #776, opened 2019-07-01
Mapped Quarternion

Sollace:
A bit torn between using mojang names or not here. The constructors are used together with Vec4f (x,y,z) but the toString in this class clearly indicates that they're (i, j, k)

    @Override
    public String toString() {
        return new StringBuilder()
             .append("Quaternion[" + this.magnitude()).append(" + ")
             .append(this.i()).append("i + ")
             .append(this.j()).append("j + ")
             .append(this.k()).append("k]")
             .toString();
    }

### DIFF
--- a/mappings/net/minecraft/client/util/math/Quaternion.mapping
+++ b/mappings/net/minecraft/client/util/math/Quaternion.mapping
@@ -4,7 +4,7 @@ CLASS net/minecraft/class_1158 net/minecraft/client/util/math/Quaternion
 		ARG 1 x
 		ARG 2 y
 		ARG 3 z
-		ARG 4 magnitude
+		ARG 4 w
 	METHOD <init> (FFFZ)V
 		ARG 1 x
 		ARG 2 y
@@ -22,10 +22,10 @@ CLASS net/minecraft/class_1158 net/minecraft/client/util/math/Quaternion
 		ARG 0 value
 	METHOD method_16003 fCos (F)F
 		ARG 0 value
-	METHOD method_4921 i ()F
-	METHOD method_4922 j ()F
-	METHOD method_4923 k ()F
-	METHOD method_4924 magnitude ()F
+	METHOD method_4921 x ()F
+	METHOD method_4922 y ()F
+	METHOD method_4923 z ()F
+	METHOD method_4924 w ()F
 	METHOD method_4925 copyFrom (Lnet/minecraft/class_1158;)V
 		ARG 1 other
 	METHOD method_4926 reverse ()V
--- a/mappings/net/minecraft/client/util/math/Quaternion.mapping
+++ b/mappings/net/minecraft/client/util/math/Quaternion.mapping
@@ -1,4 +1,31 @@
 CLASS net/minecraft/class_1158 net/minecraft/client/util/math/Quaternion
 	FIELD field_5656 components [F
+	METHOD <init> (FFFF)V
+		ARG 1 x
+		ARG 2 y
+		ARG 3 z
+		ARG 4 magnitude
+	METHOD <init> (FFFZ)V
+		ARG 1 x
+		ARG 2 y
+		ARG 3 z
+		ARG 4 w
+	METHOD <init> (Lnet/minecraft/class_1158;)V
+		ARG 1 other
+	METHOD <init> (Lnet/minecraft/class_1160;FZ)V
+		ARG 1 direction
+		ARG 2 magnitude
+		ARG 3 normalize
 	METHOD equals (Ljava/lang/Object;)Z
 		ARG 1 o
+	METHOD method_16002 fSin (F)F
+		ARG 0 value
+	METHOD method_16003 fCos (F)F
+		ARG 0 value
+	METHOD method_4921 i ()F
+	METHOD method_4922 j ()F
+	METHOD method_4923 k ()F
+	METHOD method_4924 length ()F
+	METHOD method_4925 copyFrom (Lnet/minecraft/class_1158;)V
+		ARG 1 other
+	METHOD method_4926 reverse ()V
--- a/mappings/net/minecraft/client/util/math/Quaternion.mapping
+++ b/mappings/net/minecraft/client/util/math/Quaternion.mapping
@@ -25,7 +25,7 @@ CLASS net/minecraft/class_1158 net/minecraft/client/util/math/Quaternion
 	METHOD method_4921 i ()F
 	METHOD method_4922 j ()F
 	METHOD method_4923 k ()F
-	METHOD method_4924 length ()F
+	METHOD method_4924 magnitude ()F
 	METHOD method_4925 copyFrom (Lnet/minecraft/class_1158;)V
 		ARG 1 other
 	METHOD method_4926 reverse ()V
--- a/mappings/net/minecraft/util/math/Quaternion.mapping
+++ b/mappings/net/minecraft/util/math/Quaternion.mapping
@@ -13,7 +13,7 @@ CLASS net/minecraft/class_1158 net/minecraft/util/math/Quaternion
 	METHOD <init> (Lnet/minecraft/class_1158;)V
 		ARG 1 other
 	METHOD <init> (Lnet/minecraft/class_1160;FZ)V
-		ARG 1 direction
+		ARG 1 axis
 		ARG 2 angle
 		ARG 3 isDegrees
 	METHOD equals (Ljava/lang/Object;)Z
--- a/mappings/net/minecraft/util/math/Quaternion.mapping
+++ b/mappings/net/minecraft/util/math/Quaternion.mapping
@@ -22,10 +22,10 @@ CLASS net/minecraft/class_1158 net/minecraft/util/math/Quaternion
 		ARG 0 value
 	METHOD method_16003 cos (F)F
 		ARG 0 value
-	METHOD method_4921 x ()F
-	METHOD method_4922 y ()F
-	METHOD method_4923 z ()F
-	METHOD method_4924 w ()F
+	METHOD method_4921 getX ()F
+	METHOD method_4922 getY ()F
+	METHOD method_4923 getZ ()F
+	METHOD method_4924 getW ()F
 	METHOD method_4925 copyFrom (Lnet/minecraft/class_1158;)V
 		ARG 1 other
 	METHOD method_4926 reverse ()V
--- a/mappings/net/minecraft/util/math/Quaternion.mapping
+++ b/mappings/net/minecraft/util/math/Quaternion.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_1158 net/minecraft/client/util/math/Quaternion
+CLASS net/minecraft/class_1158 net/minecraft/util/math/Quaternion
 	FIELD field_5656 components [F
 	METHOD <init> (FFFF)V
 		ARG 1 x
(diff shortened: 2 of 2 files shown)


Pyrofab:
https://en.wikipedia.org/wiki/Quaternion
The standard quaternion units are `i`, `j`, and `k`. The fields in the quaternion class are the coefficients for those units. The `toString` is a mathematical representation of the quaternion as `"Quaternion[a + bi + cj + dk]"`. As such, does it really make sense to name the fields and methods (i, j, k) ? I think (a, b, c, d) or (x, y, z, w) would be a better match.

Sollace:
> 
> 
> https://en.wikipedia.org/wiki/Quaternion
> The standard quaternion units are `i`, `j`, and `k`. The fields in the quaternion class are the coefficients for those units. The `toString` is a mathematical representation of the quaternion as `"Quaternion[a + bi + cj + dk]"`. As such, does it really make sense to name the fields and methods (i, j, k) ? I think (a, b, c, d) or (x, y, z, w) would be a better match.

Wait wat.

But you _just_ said that they must be `i`, `j`, and `k`?

Pyrofab:
`i`, `j`, and `k` are mathematical constants. The Quaternion class stores coefficients for those constants. So if the numbers are (1, 2, 3, and 4), the Quaternion will be `1 + 2i + 3j + 4k`.

Sollace:
Oooh. I get it.

Yeah, I can change the names to use the same as Vec4f then.

mappings/net/minecraft/client/util/math/Quaternion.mapping
```diff
@@ -1,4 +1,31 @@
...
 	METHOD equals (Ljava/lang/Object;)Z
 		ARG 1 o
+	METHOD method_16002 fSin (F)F
+		ARG 0 value
+	METHOD method_16003 fCos (F)F
+		ARG 0 value
+	METHOD method_4921 x ()F
+	METHOD method_4922 y ()F
+	METHOD method_4923 z ()F
+	METHOD method_4924 w ()F
```
liach: given they are getters, can we name them `getX()` `getY()` `getZ()` `getW()` instead?

liach (changes requested):

mappings/net/minecraft/client/util/math/Quaternion.mapping
```diff
@@ -1,4 +1,31 @@
...
+		ARG 1 direction
+		ARG 2 magnitude
+		ARG 3 normalize
 	METHOD equals (Ljava/lang/Object;)Z
 		ARG 1 o
+	METHOD method_16002 fSin (F)F
+		ARG 0 value
+	METHOD method_16003 fCos (F)F
+		ARG 0 value
+	METHOD method_4921 x ()F
```
ChloeDawn:
```suggestion
	METHOD method_4921 getX ()F
```

mappings/net/minecraft/client/util/math/Quaternion.mapping
```diff
@@ -1,4 +1,31 @@
...
+	METHOD method_4922 y ()F
```
ChloeDawn:
```suggestion
	METHOD method_4922 getY ()F
```

mappings/net/minecraft/client/util/math/Quaternion.mapping
```diff
@@ -1,4 +1,31 @@
...
+	METHOD method_4923 z ()F
```
ChloeDawn:
```suggestion
	METHOD method_4923 getZ ()F
```

mappings/net/minecraft/client/util/math/Quaternion.mapping
ChloeDawn:
```suggestion
	METHOD method_4924 getW ()F
```

ChloeDawn (changes requested):

ChloeDawn:
@Sollace This class should also be moved to `net.minecraft.util.math`. It is not a client-only class.

Pyrofab:
Quoting JamiesWhiteShirt on the discord,
>  by the same logic that i,j,k are the units of the quaternion, x,y,z(,w) can also be units of vectors

Essentially, calling them (r, i, j, k) is still a possibility, contrary to what I first thought.

Runemoro:
> by the same logic that i,j,k are the units of the quaternion, x,y,z(,w) can also be units of vectors

No... `x`, `y`, `z` are names for the components of the vectors and the unit vectors are usually called `i`, `j`, `k`. Similarly, `x`, `y`, `z`, `w` are names for the components of the quaternion and the unit quaternions are called `1`, `i`, `j`, `k`.

mappings/net/minecraft/client/util/math/Quaternion.mapping
Sollace: We can. Should we change the method on the vector classes too?

Sollace:
> 
> 
> @Sollace This class should also be moved to `net.minecraft.util.math`. It is not a client-only class.

How do you know that?
There's nothing I can see in Enigma that hints I'm able to move classes to other packages.

mappings/net/minecraft/client/util/math/Quaternion.mapping
Sollace: I don't want to commit these until I get an answer about the `x()`, `y()`, `z()` methods in all the vector classes.

Until then, I'd prefer to stick to the current convention.

ChloeDawn:
The action of renaming a class allows you to define the package it is located in; go to rename the class and remove `client` from the package structure.

mappings/net/minecraft/util/math/Quaternion.mapping
```diff
@@ -0,0 +1,31 @@
...
+		ARG 1 x
+		ARG 2 y
+		ARG 3 z
+		ARG 4 w
+	METHOD <init> (Lnet/minecraft/class_1158;)V
+		ARG 1 other
+	METHOD <init> (Lnet/minecraft/class_1160;FZ)V
+		ARG 1 direction
+		ARG 2 magnitude
+		ARG 3 normalize
```
Runemoro: Something's weird about this, are you sure about `magnitude`? If the second argument is really the magnitude, then normalize doesn't make sense, since you could just pass a magnitude of `1` to the constructor to make it normal.

mappings/net/minecraft/util/math/Quaternion.mapping
```diff
@@ -0,0 +1,31 @@
...
+	METHOD equals (Ljava/lang/Object;)Z
+		ARG 1 o
+	METHOD method_16002 fSin (F)F
+		ARG 0 value
+	METHOD method_16003 fCos (F)F
+		ARG 0 value
```
Runemoro, line 24: What does `f` mean in `fSin` and `fCos`?

mappings/net/minecraft/util/math/Quaternion.mapping
```diff
@@ -0,0 +1,31 @@
...
+	METHOD method_4921 x ()F
+	METHOD method_4922 y ()F
+	METHOD method_4923 z ()F
+	METHOD method_4924 w ()F
```
Runemoro: These should be `getX`, etc. to respect the getter conventions (look at other vector classes, they use `getX`, etc.).

mappings/net/minecraft/util/math/Quaternion.mapping
Runemoro: If it means that the methods work on floats, then just `sin` and `cos` is enough, the type is already in the method signature, there's no need to specify it in the name too. `f` is confusing (it could also mean "fast", for example).

Sollace: It's to be in line with the GlStateManager methods, which have `scalef` for example.

mappings/net/minecraft/util/math/Quaternion.mapping
Sollace: When I looked the other vector classes weren't using `getX()`. It was just `x()`.

mappings/net/minecraft/util/math/Quaternion.mapping
Runemoro: What? This is not a `GlStateManager` method. The methods in GlStateManager are named to match the OpenGL method names, which are not overloaded because C doesn't support overloading. And anyway, even if you were naming this as a GlStateManager method, it should be `sinf`, not `fSin`.

mappings/net/minecraft/util/math/Quaternion.mapping
Sollace: net/minecraft/client/util/math/Vector3f

    public float x() {
        return this.components[0];
    }
    
    public float y() {
        return this.components[1];
    }
    
    public float z() {
        return this.components[2];
    }

Runemoro: Not sure which vector class you looked at, but all the methods I've seen that return coordinates are named `getX`, `getY`, and `getZ`. If there's some other getter named `x` rather than `getX`, that one is wrong.

mattmess1221: Not using the `getX` naming will also make it hard for kotlin developers

Runemoro: https://github.com/FabricMC/yarn/issues/789

mappings/net/minecraft/util/math/Quaternion.mapping
Sollace: Well `normalize` alters the magnitude, and magnitude is used to alter the x/y/z components.

        if (normalize) {
            magnitude *= 0.017453292F;
        }
        float sineMag = sin(magnitude / 2F);
        this.components = new float[4];

        this.components[0] = direction.x() * sinMag;
        this.components[1] = direction.y() * sinMag;
        this.components[2] = direction.z() * sinMag;
        this.components[3] = cos(magnitude / 2F);

Maybe someone else can tell me more definitively what's going on here.

Runemoro: "Normalize" means "make the magnitude 1". `0.017453292f` is `pi / 180`, so it seems like the `magnitude` is named wrongly, it's actually an angle. The third parameter is true if the second parameter is given in degrees rather than radians.

Sollace: I was just thinking that, but couldn't figure out what `0.017453292f` was as a factor of Pi.

mappings/net/minecraft/util/math/Quaternion.mapping
Sollace: Coolio

mappings/net/minecraft/util/math/Quaternion.mapping
Runemoro: Here's a description of what the constructor does: https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles#Definition

The first parameter is the direction of the axis of rotation, the second is the angle by which to rotate around that axis, and the third is whether that angle is given in degrees rather than radians.

I suggest the names `axis`, `angle`, `degrees`.

mappings/net/minecraft/util/math/Quaternion.mapping
Sollace: @killjoy1221 How does the naming make things harder for Kotlin devs? It's just a method name, brah.

mappings/net/minecraft/util/math/Quaternion.mapping
Sollace: Done.

Runemoro (approved):

liach (approved):

asiekierka:
wait no wrong branch

Sollace:
_gasp_

PR #777, opened 2019-07-01
UntrackMemoryUtil -> Untracker

Sollace:
Turns out there wasn't really that much to change here.

This is based on the names Chloe gave in the discord chat last night. I just felt `Allocator` would be a bit more descriptive than `Untracker`.

### DIFF
--- a/mappings/net/minecraft/client/util/Allocator.mapping
+++ b/mappings/net/minecraft/client/util/Allocator.mapping
@@ -2,4 +2,4 @@ CLASS net/minecraft/class_301 net/minecraft/client/util/Allocator
 	FIELD field_1643 ALLOCATOR_UNTRACK Ljava/lang/invoke/MethodHandle;
 	METHOD method_1406 untrack (Lorg/lwjgl/system/Pointer;)V
 	METHOD method_1407 untrack (J)V
-		ARG 0 pointer
+		ARG 0 address
--- a/mappings/net/minecraft/client/util/Allocator.mapping
+++ b/mappings/net/minecraft/client/util/Allocator.mapping
@@ -1,0 +1,5 @@
+CLASS net/minecraft/class_301 net/minecraft/client/util/Allocator
+	FIELD field_1643 ALLOCATOR_UNTRACK Ljava/lang/invoke/MethodHandle;
+	METHOD method_1406 untrack (Lorg/lwjgl/system/Pointer;)V
+	METHOD method_1407 untrack (J)V
+		ARG 0 pointer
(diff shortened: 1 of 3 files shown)


Pyrofab:
This class only deallocates right ? Why name it Allocator ?

mattmess1221:
Maybe `GLPointerUtils`?

Sollace:
> 
> 
> This class only deallocates right ? Why name it Allocator ?
Well, you see, I didn't like the sound of `Untracker.untrack(...)`.

`Allocator.untrack(...)` sounds more natural, even though it only hosts the de-allocating part of allocator.

mappings/net/minecraft/client/util/Allocator.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS net/minecraft/class_301 net/minecraft/client/util/Allocator
+	FIELD field_1643 ALLOCATOR_UNTRACK Ljava/lang/invoke/MethodHandle;
+	METHOD method_1406 untrack (Lorg/lwjgl/system/Pointer;)V
```
ChloeDawn: Missing parameter name should be `pointer`. Note that it may display as `pointer` in Enigma but this is just an inferred name, and the Loom toolchain will infer it differently, producing `pointer_1` if unmapped here.

mappings/net/minecraft/client/util/Allocator.mapping
```diff
@@ -0,0 +1,5 @@
...
+	METHOD method_1407 untrack (J)V
+		ARG 0 pointer
```
ChloeDawn:
```suggestion
		ARG 0 address
```

ChloeDawn (changes requested):

ChloeDawn:
`Allocator` implies the opposite of this class' primary (and only) purpose. If anything it should be `Deallocator`, but personally, I still prefer `Untracker`.

Runemoro:
> Well, you see, I didn't like the sound of `Untracker.untrack(...)`.

I don't see any problem with `Untracker.untrack(...)`. You can name it `Untracker` or `MemoryUtil`, but `Allocator` is wrong since that's the opposite of what this class does.

mappings/net/minecraft/client/util/Allocator.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS net/minecraft/class_301 net/minecraft/client/util/Allocator
+	FIELD field_1643 ALLOCATOR_UNTRACK Ljava/lang/invoke/MethodHandle;
```
Runemoro: Why rename this? I think `_METHOD_HANDLE` (or `_HANDLE`) should stay part of the name.

mappings/net/minecraft/client/util/Allocator.mapping
Sollace: Ah. Thank you! I'll have to keep that in mind. I've been not mapping parameters in some cases because the value Enigma shows is already correct.

Runemoro (approved):

PR #778, opened 2019-07-01
Entity control mappings

Sollace:
## MoveControl
method_6243(f, f) -> strafeTo(forward, sideways)

This method is called in `BowAttackGoal.tick` with either positive or negative 0.5 for each parameter. It's not clear from this call site what the parameter are, though.

However, when you look at the two fields it sets (`field_6368`, and `field_6371`) respectively, you'll see that they're being used in `MoveControl.tick` in the following way:

            this.entity.setForwardSpeed(this.field_6368);
            this.entity.setSidewaysSpeed(this.field_6371);

That gives us the mappings:
`field_6368` -> `strafeForwardAmount`
`field_6371` -> `strafeSidewaysAmount`
And from those we get the parameter names for `strafeTo`.

## LookControl

`method_19615` and `method_20248` are just overloads for the existing, already mapped `lookAt` method, so I just filled in the missing pieces.

## BodyControl

`method_6224` -> `tick`
This is called from `MobEntity.method_6031` and seems to be the main method.

`field_6355` -> `ticksUpdated`
`field_6354` -> `lastHeadYaw`

`method_20243` -> `rotateLook`
`method_20246` -> `isIndependent`
This method checks if the mob is being ridden, and only returns true if it's being ridden by _another mob_. If it returns true, this class checks if the entity's head has turned, will reset the tick counter, then call `rotateLook` to update a field on the entity (unmapped) which I believe relates to the entity's head/look direction.

`method_20245` -> `rotateBody`
If the entity's look direction has _not_ changed, it increments the tick counter, and calls `rotateBody` after 10 ticks to turn in the direction the entity is facing.

`method_20247` -> `isMoving`
This method checks the entity's current and last x/z coordinates and returns true if the square of the distance between the two is larger than _this_ very tiny number: 2.500000277905201E-7.

`method_20244` -> `rotateHead`
If it returns true, most of the above processing is skipped, and all it does is call `rotateHead` to update the entity's `headYaw`.

### DIFF
--- a/mappings/net/minecraft/entity/ai/control/BodyControl.mapping
+++ b/mappings/net/minecraft/entity/ai/control/BodyControl.mapping
@@ -1,6 +1,6 @@
 CLASS net/minecraft/class_1330 net/minecraft/entity/ai/control/BodyControl
 	FIELD field_6354 lastHeadYaw F
-	FIELD field_6355 ticksUpdated I
+	FIELD field_6355 activeTicks I
 	FIELD field_6356 entity Lnet/minecraft/class_1308;
 	METHOD method_20243 rotateLook ()V
 	METHOD method_20244 rotateHead ()V
--- a/mappings/net/minecraft/entity/ai/control/BodyControl.mapping
+++ b/mappings/net/minecraft/entity/ai/control/BodyControl.mapping
@@ -1,2 +1,8 @@
 CLASS net/minecraft/class_1330 net/minecraft/entity/ai/control/BodyControl
+	FIELD field_6354 lastHeadYaw F
+	FIELD field_6355 ticksUpdated I
 	FIELD field_6356 entity Lnet/minecraft/class_1308;
+	METHOD method_20244 rotateHead ()V
+	METHOD method_20245 rotateBody ()V
+	METHOD method_20246 isMovingFreely ()Z
+	METHOD method_20247 isSprinting ()Z
--- a/mappings/net/minecraft/entity/ai/control/BodyControl.mapping
+++ b/mappings/net/minecraft/entity/ai/control/BodyControl.mapping
@@ -2,6 +2,7 @@ CLASS net/minecraft/class_1330 net/minecraft/entity/ai/control/BodyControl
 	FIELD field_6354 lastHeadYaw F
 	FIELD field_6355 ticksUpdated I
 	FIELD field_6356 entity Lnet/minecraft/class_1308;
+	METHOD method_20243 rotateLook ()V
 	METHOD method_20244 rotateHead ()V
 	METHOD method_20245 rotateBody ()V
 	METHOD method_20246 isMovingFreely ()Z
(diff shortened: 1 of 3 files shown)


mappings/net/minecraft/entity/ai/control/BodyControl.mapping
```diff
@@ -1,2 +1,10 @@
 CLASS net/minecraft/class_1330 net/minecraft/entity/ai/control/BodyControl
+	FIELD field_6354 lastHeadYaw F
+	FIELD field_6355 ticksUpdated I
```
liach: activeTicks
```suggestion
	FIELD field_6355 activeTicks I
```

mappings/net/minecraft/entity/ai/control/MoveControl.mapping
```diff
@@ -1,10 +1,12 @@
 CLASS net/minecraft/class_1335 net/minecraft/entity/ai/control/MoveControl
 	CLASS class_1336 State
 	FIELD field_6367 targetZ D
+	FIELD field_6368 strafeForwardAmount F
```
liach: `forwardStrafe`
```suggestion
	FIELD field_6368 forwardStrafe F
```

mappings/net/minecraft/entity/ai/control/MoveControl.mapping
```diff
@@ -1,10 +1,12 @@
...
 	FIELD field_6369 targetY D
 	FIELD field_6370 targetX D
 	FIELD field_6371 entity Lnet/minecraft/class_1308;
 	FIELD field_6372 speed D
+	FIELD field_6373 strafeSidewaysAmount F
```
liach:
```suggestion
	FIELD field_6373 sidewaysStrafe F
```

liach (changes requested):

liach (approved):
Unfortunately gradle wrapper distribution site glitched when applying the suggestions. Look great!

mappings/net/minecraft/entity/ai/control/MoveControl.mapping
```diff
@@ -1,10 +1,12 @@
 CLASS net/minecraft/class_1335 net/minecraft/entity/ai/control/MoveControl
 	CLASS class_1336 State
 	FIELD field_6367 targetZ D
+	FIELD field_6368 forwardStrafe F
```
Runemoro: This doesn't work, ["strafe" means to "move sideways"](https://en.wiktionary.org/wiki/strafe#Verb). Maybe `forwardMovement`?

mappings/net/minecraft/entity/ai/control/MoveControl.mapping
```diff
@@ -1,10 +1,12 @@
...
 	FIELD field_6369 targetY D
 	FIELD field_6370 targetX D
 	FIELD field_6371 entity Lnet/minecraft/class_1308;
 	FIELD field_6372 speed D
+	FIELD field_6373 sidewaysStrafe F
```
Runemoro: This is redundant since "strafe" implies "sideways". Maybe either `strafe` or `sidewaysMovement`? I prefer `sidewaysMovement` to match `forwardMovement`.

Runemoro (approved):

mappings/net/minecraft/entity/ai/control/BodyControl.mapping
liach: Slap... I don't know if this is problems with updates or myself, but this name is no longer right in 1.17! Will submit a fix. :zany_face:

PR #784, opened 2019-07-04
BlockPosLookTarget -> BlockPosWrapper

Sollace:
`LookTarget` -> `FocusPoint`

`getBlockPos` -> `getBlockPos`
(gets the position of this focus point in block coordinates)
`getPos` -> `getLookVector`
(gets the position vector, typically has to be the middle of the block where this focus point is centred)
`isSeenBy` -> `hasBeenVisited`

It's an interface that represents some point that the entity's AI can focus onto (whether a block or entity).
It provides a method to get the block position, and position vector of where the AI needs to direct the entity, and a method to determine whether the entity has reached that point or not.

`WalkTarget` -> `Destination`

Pretty self-explanatory. A destination operates on a `FocusPoint` and represents the destination an entity's AI must walk towards.

`BlockPosLookTarget` -> `BlockPosWrapper` (Mojang name)

Implementation of a `FocusPoint` based on a fixed block position and is consistent with `EntityPosWrapper`.

<s>Method names are up to debate. I know the normal is to say `getBlockPos` however I felt it meaningful to be explicit that this is the _position of the focus point_, and not just a generic getter for a `BlockPos`.

An alternative approach would be to call them `getEndPoint` and `getEndPointVector` but those are not quite as clear.</s>


Runemoro:
While `BlockPosWrapper` and `EntityWrapper` are Mojang names, the names we currently have are much better. The name "wrapper" describes what it is without describing what it's for.

What the classes really are are implementations of `LookTarget` for entities and block positions. LookTarget provides:
 1. A method that returns the target position towards which the villager should look
 2. A condition for whether the Entity with that target can complete the target (for blocks, always true, and for entities, if the entity has seen and looked back at the villager)

The changes `LookTarget` -> `FocusPoint` and `WalkTarget` -> `Destination` are pointless. The new names mean basically the same thing, but are more confusing and don't match the registry names of the memory modules (`look_target` and `walk_target`).

modmuss50:
I agree with @Runemoro some of the changes seem to have been done for the sake of it.

The only change that I think is required around this area is:

`EntityPosWrapper` -> `EntityLookTarget`

Sollace:
I made some changes based on feedback on discord. As @Pyrofab#2068 pointed out, `Destination` could use some improvement.

I thought of maybe:
1. `DestinationPoint` 
2. `PointOfInterest`
3. `Checkpoint`
4. Just keep `WalkTarget`

liach:
I agree with runemoro. `BlockPosWrapper` seems as if the class with this name is a subclass and an implementation in the `LookTarget` class, but it is not.

Issue #785, opened 2019-07-05
MutableIntBoundingBox -> MutableIntBox


Blayyke:
With the recent BoundingBox -> Box change, this was forgotten. Come to think of it, the Int in this is redundant as there is no other type that has non-int fields.

Juuxel:
Changed to `BlockBox` in #801.

PR #801, opened 2019-07-25
MutableIntBoundingBox -> BlockBox

Sollace:
Last PR for today!

I renamed `MutableIntBoundingBox` to `MutableBox` as per #785. I also standardised the naming between the two clases.

`x0, y0, z0, x1, y1, z1` -> `minX, minY, minZ, maxX, maxY, maxZ`

I also kept the names `getXSize`, `getYSize`, and `getZSize` for now, though I want to know what people would think of me changing them to `getWidth`, `getHeight`, and `getDepth` instead.

### DIFF
--- a/mappings/net/minecraft/util/math/BlockBox.mapping
+++ b/mappings/net/minecraft/util/math/BlockBox.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_3341 net/minecraft/util/math/MutableBox
+CLASS net/minecraft/class_3341 net/minecraft/util/math/BlockBox
 	FIELD field_14376 maxZ I
 	FIELD field_14377 maxY I
 	FIELD field_14378 maxX I
--- a/mappings/net/minecraft/util/math/BlockBox.mapping
+++ b/mappings/net/minecraft/util/math/BlockBox.mapping
@@ -39,12 +39,12 @@ CLASS net/minecraft/class_3341 net/minecraft/util/math/BlockBox
 	METHOD method_14664 getBlockCountZ ()I
 	METHOD method_14665 empty ()Lnet/minecraft/class_3341;
 	METHOD method_14666 create (IIIIII)Lnet/minecraft/class_3341;
-		ARG 0 minX
-		ARG 1 minY
-		ARG 2 minZ
-		ARG 3 maxX
-		ARG 4 maxY
-		ARG 5 maxZ
+		ARG 0 x1
+		ARG 1 y1
+		ARG 2 z1
+		ARG 3 x2
+		ARG 4 y2
+		ARG 5 z2
 	METHOD method_14667 rotated (IIIIIIIIILnet/minecraft/class_2350;)Lnet/minecraft/class_3341;
 		ARG 0 x
 		ARG 1 y
--- a/mappings/net/minecraft/util/math/BlockBox.mapping
+++ b/mappings/net/minecraft/util/math/BlockBox.mapping
@@ -56,8 +56,8 @@ CLASS net/minecraft/class_3341 net/minecraft/util/math/BlockBox
 		ARG 7 sizeY
 		ARG 8 sizeZ
 		ARG 9 facing
-	METHOD method_14668 setFrom (Lnet/minecraft/class_3341;)V
-		ARG 1 source
+	METHOD method_14668 encompass (Lnet/minecraft/class_3341;)V
+		ARG 1 region
 	METHOD method_14669 intersectsXZ (IIII)Z
 		ARG 1 minX
 		ARG 2 minZ
--- a/mappings/net/minecraft/util/math/BlockBox.mapping
+++ b/mappings/net/minecraft/util/math/BlockBox.mapping
@@ -28,15 +28,15 @@ CLASS net/minecraft/class_3341 net/minecraft/util/math/BlockBox
 		ARG 1 other
 	METHOD method_14658 toNbt ()Lnet/minecraft/class_2495;
 	METHOD method_14659 getDimensions ()Lnet/minecraft/class_2382;
-	METHOD method_14660 getXLength ()I
+	METHOD method_14660 getBlockCountX ()I
 	METHOD method_14661 offset (III)V
 		ARG 1 dx
 		ARG 2 dy
 		ARG 3 dz
 	METHOD method_14662 contains (Lnet/minecraft/class_2382;)Z
 		ARG 1 vec
-	METHOD method_14663 getYLength ()I
-	METHOD method_14664 getZLength ()I
+	METHOD method_14663 getBlockCountY ()I
+	METHOD method_14664 getBlockCountZ ()I
 	METHOD method_14665 empty ()Lnet/minecraft/class_3341;
 	METHOD method_14666 create (IIIIII)Lnet/minecraft/class_3341;
 		ARG 0 minX
--- a/mappings/net/minecraft/util/math/Box.mapping
+++ b/mappings/net/minecraft/util/math/Box.mapping
@@ -6,11 +6,17 @@ CLASS net/minecraft/class_238 net/minecraft/util/math/Box
 	FIELD field_1324 maxZ D
 	FIELD field_1325 maxY D
 	METHOD <init> (DDDDDD)V
-		ARG 1 x1
-		ARG 3 y1
-		ARG 5 z1
+		ARG 1 minX
+		ARG 3 minY
+		ARG 5 minZ
+		ARG 7 maxX
+		ARG 9 maxY
+		ARG 11 maxZ
+	METHOD <init> (Lnet/minecraft/class_2338;)V
+		ARG 1 pos
 	METHOD <init> (Lnet/minecraft/class_2338;Lnet/minecraft/class_2338;)V
 		ARG 1 min
+		ARG 2 max
 	METHOD <init> (Lnet/minecraft/class_243;Lnet/minecraft/class_243;)V
 		ARG 1 min
 		ARG 2 max
@@ -32,6 +38,14 @@ CLASS net/minecraft/class_238 net/minecraft/util/math/Box
 	METHOD method_1005 getCenter ()Lnet/minecraft/class_243;
 	METHOD method_1006 contains (Lnet/minecraft/class_243;)Z
 		ARG 1 vec
+	METHOD method_1007 traceCollissionSide (Lnet/minecraft/class_238;Lnet/minecraft/class_243;[DLnet/minecraft/class_2350;DDD)Lnet/minecraft/class_2350;
+		ARG 0 box
+		ARG 1 approachVector
+		ARG 2 traceDistanceResult
+		ARG 3 approachDirection
+		ARG 4 xDelta
+		ARG 6 yDelta
+		ARG 8 zDelta
 	METHOD method_1008 contains (DDD)Z
 		ARG 1 x
 		ARG 3 y
@@ -58,7 +72,9 @@ CLASS net/minecraft/class_238 net/minecraft/util/math/Box
 	METHOD method_17940 getYSize ()D
 	METHOD method_17941 getZSize ()D
 	METHOD method_18804 stretch (Lnet/minecraft/class_243;)Lnet/minecraft/class_238;
+		ARG 1 scale
 	METHOD method_19316 from (Lnet/minecraft/class_3341;)Lnet/minecraft/class_238;
+		ARG 0 mutable
 	METHOD method_989 offset (DDD)Lnet/minecraft/class_238;
 		ARG 1 x
 		ARG 3 y
@@ -66,8 +82,13 @@ CLASS net/minecraft/class_238 net/minecraft/util/math/Box
 	METHOD method_990 getMax (Lnet/minecraft/class_2350$class_2351;)D
 		ARG 1 axis
 	METHOD method_991 union (Lnet/minecraft/class_238;)Lnet/minecraft/class_238;
+		ARG 1 box
 	METHOD method_992 rayTrace (Lnet/minecraft/class_243;Lnet/minecraft/class_243;)Ljava/util/Optional;
+		ARG 1 min
+		ARG 2 max
 	METHOD method_993 intersects (Lnet/minecraft/class_243;Lnet/minecraft/class_243;)Z
+		ARG 1 from
+		ARG 2 to
 	METHOD method_994 intersects (Lnet/minecraft/class_238;)Z
 		ARG 1 box
 	METHOD method_995 averageDimension ()D
@@ -75,6 +96,20 @@ CLASS net/minecraft/class_238 net/minecraft/util/math/Box
 		ARG 1 blockPos
 	METHOD method_997 offset (Lnet/minecraft/class_243;)Lnet/minecraft/class_238;
 		ARG 1 vec3d
-	METHOD method_998 ([DLnet/minecraft/class_2350;DDDDDDDDLnet/minecraft/class_2350;DDD)Lnet/minecraft/class_2350;
-		ARG 18 maxZ
+	METHOD method_998 traceCollissionSide ([DLnet/minecraft/class_2350;DDDDDDDDLnet/minecraft/class_2350;DDD)Lnet/minecraft/class_2350;
+		ARG 0 traceDistanceResult
+		ARG 1 approachDirection
+		ARG 2 xDelta
+		ARG 4 yDelta
+		ARG 6 zDelta
+		ARG 8 begin
+		ARG 10 minX
+		ARG 12 maxX
+		ARG 14 minZ
+		ARG 16 maxZ
+		ARG 18 resultDirection
+		ARG 19 startX
+		ARG 21 startY
+		ARG 23 startZ
 	METHOD method_999 intersection (Lnet/minecraft/class_238;)Lnet/minecraft/class_238;
+		ARG 1 box
--- a/mappings/net/minecraft/util/math/Box.mapping
+++ b/mappings/net/minecraft/util/math/Box.mapping
@@ -6,20 +6,20 @@ CLASS net/minecraft/class_238 net/minecraft/util/math/Box
 	FIELD field_1324 maxZ D
 	FIELD field_1325 maxY D
 	METHOD <init> (DDDDDD)V
-		ARG 1 minX
-		ARG 3 minY
-		ARG 5 minZ
-		ARG 7 maxX
-		ARG 9 maxY
-		ARG 11 maxZ
+		ARG 1 x1
+		ARG 3 y1
+		ARG 5 z1
+		ARG 7 x2
+		ARG 9 y2
+		ARG 11 z2
 	METHOD <init> (Lnet/minecraft/class_2338;)V
 		ARG 1 pos
 	METHOD <init> (Lnet/minecraft/class_2338;Lnet/minecraft/class_2338;)V
-		ARG 1 min
-		ARG 2 max
+		ARG 1 pos1
+		ARG 2 pos2
 	METHOD <init> (Lnet/minecraft/class_243;Lnet/minecraft/class_243;)V
-		ARG 1 min
-		ARG 2 max
+		ARG 1 pos1
+		ARG 2 pos2
 	METHOD equals (Ljava/lang/Object;)Z
 		ARG 1 o
 	METHOD method_1001 getMin (Lnet/minecraft/class_2350$class_2351;)D
--- a/mappings/net/minecraft/util/math/Box.mapping
+++ b/mappings/net/minecraft/util/math/Box.mapping
@@ -38,7 +38,7 @@ CLASS net/minecraft/class_238 net/minecraft/util/math/Box
 	METHOD method_1005 getCenter ()Lnet/minecraft/class_243;
 	METHOD method_1006 contains (Lnet/minecraft/class_243;)Z
 		ARG 1 vec
-	METHOD method_1007 traceCollissionSide (Lnet/minecraft/class_238;Lnet/minecraft/class_243;[DLnet/minecraft/class_2350;DDD)Lnet/minecraft/class_2350;
+	METHOD method_1007 traceCollisionSide (Lnet/minecraft/class_238;Lnet/minecraft/class_243;[DLnet/minecraft/class_2350;DDD)Lnet/minecraft/class_2350;
 		ARG 0 box
 		ARG 1 approachVector
 		ARG 2 traceDistanceResult
@@ -96,7 +96,7 @@ CLASS net/minecraft/class_238 net/minecraft/util/math/Box
 		ARG 1 blockPos
 	METHOD method_997 offset (Lnet/minecraft/class_243;)Lnet/minecraft/class_238;
 		ARG 1 vec3d
-	METHOD method_998 traceCollissionSide ([DLnet/minecraft/class_2350;DDDDDDDDLnet/minecraft/class_2350;DDD)Lnet/minecraft/class_2350;
+	METHOD method_998 traceCollisionSide ([DLnet/minecraft/class_2350;DDDDDDDDLnet/minecraft/class_2350;DDD)Lnet/minecraft/class_2350;
 		ARG 0 traceDistanceResult
 		ARG 1 approachDirection
 		ARG 2 xDelta
--- a/mappings/net/minecraft/util/math/Box.mapping
+++ b/mappings/net/minecraft/util/math/Box.mapping
@@ -40,7 +40,7 @@ CLASS net/minecraft/class_238 net/minecraft/util/math/Box
 		ARG 1 vec
 	METHOD method_1007 traceCollisionSide (Lnet/minecraft/class_238;Lnet/minecraft/class_243;[DLnet/minecraft/class_2350;DDD)Lnet/minecraft/class_2350;
 		ARG 0 box
-		ARG 1 approachVector
+		ARG 1 intersectingVector
 		ARG 2 traceDistanceResult
 		ARG 3 approachDirection
 		ARG 4 xDelta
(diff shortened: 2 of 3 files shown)


mappings/net/minecraft/util/math/Box.mapping
```diff
@@ -32,6 +38,14 @@ CLASS net/minecraft/class_238 net/minecraft/util/math/Box
 	METHOD method_1005 getCenter ()Lnet/minecraft/class_243;
 	METHOD method_1006 contains (Lnet/minecraft/class_243;)Z
 		ARG 1 vec
+	METHOD method_1007 traceCollissionSide (Lnet/minecraft/class_238;Lnet/minecraft/class_243;[DLnet/minecraft/class_2350;DDD)Lnet/minecraft/class_2350;
```
Pyrofab: `traceCollisionSide`
Also, what does this method do exactly ? I can't really tell just by looking at the name.

mappings/net/minecraft/util/math/Box.mapping
```diff
@@ -32,6 +38,14 @@ CLASS net/minecraft/class_238 net/minecraft/util/math/Box
...
+		ARG 0 box
+		ARG 1 approachVector
```
Pyrofab: There may be a more appropriate term than "approach". Does `intersectingVector` make sense in this context ?

mappings/net/minecraft/util/math/Box.mapping
```diff
@@ -58,23 +72,44 @@ CLASS net/minecraft/class_238 net/minecraft/util/math/Box
...
 	METHOD method_994 intersects (Lnet/minecraft/class_238;)Z
 		ARG 1 box
 	METHOD method_995 averageDimension ()D
 	METHOD method_996 offset (Lnet/minecraft/class_2338;)Lnet/minecraft/class_238;
 		ARG 1 blockPos
 	METHOD method_997 offset (Lnet/minecraft/class_243;)Lnet/minecraft/class_238;
 		ARG 1 vec3d
-	METHOD method_998 ([DLnet/minecraft/class_2350;DDDDDDDDLnet/minecraft/class_2350;DDD)Lnet/minecraft/class_2350;
-		ARG 18 maxZ
+	METHOD method_998 traceCollissionSide ([DLnet/minecraft/class_2350;DDDDDDDDLnet/minecraft/class_2350;DDD)Lnet/minecraft/class_2350;
```
Pyrofab: traceCollisionSide

mappings/net/minecraft/util/math/MutableBox.mapping
```diff
@@ -28,23 +28,23 @@ CLASS net/minecraft/class_3341 net/minecraft/util/math/MutableIntBoundingBox
...
-		ARG 3 x2
-		ARG 4 y2
-		ARG 5 z2
+		ARG 0 minX
+		ARG 1 minY
+		ARG 2 minZ
+		ARG 3 maxX
+		ARG 4 maxY
+		ARG 5 maxZ
 	METHOD method_14667 createRotated (IIIIIIIIILnet/minecraft/class_2350;)Lnet/minecraft/class_3341;
```
Pyrofab: We could consider renaming this to `rotated` maybe ?

Pyrofab:
I think `getWidth`, `getHeight` and `getDepth` sound much better than `getXYZSize`.

mappings/net/minecraft/util/math/MutableBox.mapping
Sollace: This returns a new instance with the rotation, rather than this one. I don't feel simply calling it `rotated` would adequately portray that.

mappings/net/minecraft/util/math/Box.mapping
Sollace: `intersectionVector` works better, yes.

mappings/net/minecraft/util/math/MutableBox.mapping
Pyrofab: `rotate` would be the mutating method IMO

mappings/net/minecraft/util/math/Box.mapping
Sollace: Best I can say is it it's used for rayTracing to find either the side of the box the ray enters, or that the ray exits.

It's called once for each side, every time passing in a shared double[] array and the direction produced by the last call of the method.

    @Nullable
    public static BlockHitResult rayTrace(Iterable<Box> boxes, Vec3d from, Vec3d to, BlockPos blockPos) {
        double[] doubles_1 = { 1.0 };
        Direction direction6 = null;
        double double7 = to.x - from.x;
        double double9 = to.y - from.y;
        double double11 = to.z - from.z;
        for (final Box class_238_1 : boxes) {
            direction6 = traceCollissionSide(class_238_1.offset(blockPos), from, doubles_1, direction6, double7, double9, double11);
        }
        if (direction6 == null) {
            return null;
        }
        double double13 = doubles_1[0];
        return new BlockHitResult(from.add(double13 * double7, double13 * double9, double13 * double11), direction6, blockPos, false);
    }
    
    @Nullable
    private static Direction traceCollissionSide(Box box, Vec3d approachVector, double[] traceDistanceResult, @Nullable Direction approachDirection, double xDelta, double yDelta, double zDelta) {
        if (xDelta > 1.0E-7) {
            approachDirection = traceCollissionSide(traceDistanceResult, approachDirection, xDelta, yDelta, zDelta, box.minX, box.minY, box.maxY, box.minZ, box.maxZ, Direction.WEST, approachVector.x, approachVector.y, approachVector.z);
        }
        else if (xDelta < -1.0E-7) {
            approachDirection = traceCollissionSide(traceDistanceResult, approachDirection, xDelta, yDelta, zDelta, box.maxX, box.minY, box.maxY, box.minZ, box.maxZ, Direction.EAST, approachVector.x, approachVector.y, approachVector.z);
        }
        if (yDelta > 1.0E-7) {
            approachDirection = traceCollissionSide(traceDistanceResult, approachDirection, yDelta, zDelta, xDelta, box.minY, box.minZ, box.maxZ, box.minX, box.maxX, Direction.DOWN, approachVector.y, approachVector.z, approachVector.x);
        }
        else if (yDelta < -1.0E-7) {
            approachDirection = traceCollissionSide(traceDistanceResult, approachDirection, yDelta, zDelta, xDelta, box.maxY, box.minZ, box.maxZ, box.minX, box.maxX, Direction.UP, approachVector.y, approachVector.z, approachVector.x);
        }
        if (zDelta > 1.0E-7) {
            approachDirection = traceCollissionSide(traceDistanceResult, approachDirection, zDelta, xDelta, yDelta, box.minZ, box.minX, box.maxX, box.minY, box.maxY, Direction.NORTH, approachVector.z, approachVector.x, approachVector.y);
        }
        else if (zDelta < -1.0E-7) {
            approachDirection = traceCollissionSide(traceDistanceResult, approachDirection, zDelta, xDelta, yDelta, box.maxZ, box.minX, box.maxX, box.minY, box.maxY, Direction.SOUTH, approachVector.z, approachVector.x, approachVector.y);
        }
        return approachDirection;
    }
    
    @Nullable
    private static Direction traceCollissionSide(double[] traceDistanceResult, @Nullable Direction approachDirection, double xDelta, double yDelta, double zDelta, double begin, double minX, double maxX, double minZ, double maxZ, Direction resultDirection, double startX, double startY, double startZ) {
        double double26 = (begin - startX) / xDelta;
        double double28 = startY + double26 * yDelta;
        double double30 = startZ + double26 * zDelta;
        if (0.0 < double26 && double26 < traceDistanceResult[0] && minX - 1.0E-7 < double28 && double28 < maxX + 1.0E-7 && minZ - 1.0E-7 < double30 && double30 < maxZ + 1.0E-7) {
            traceDistanceResult[0] = double26;
            return resultDirection;
        }
        return approachDirection;
    }

Sollace:
@Pyrofab If I can get more people to agree, I'll put it in.

Another related thing I meant to ask is:
What becomes width, and what becomes depth?

I know `y = height`, but `must x = width` and `z =depth` or must it be `x = depth` and `z = width`?

mappings/net/minecraft/util/math/MutableBox.mapping
Sollace: Again, this method does not mutate the box. Returns a new box.

JamiesWhiteShirt: Which `rotated` adequately portrays in my opinion.

Pyrofab: `rotate` != `rotated`
Rotate implies mutation to a rotated state, rotated implies the creation of a rotated copy

Pyrofab:
Seeing how width and depth are not immediately clear, I have reconsidered ~~and I would suggest `getXYZLength` instead~~. I think those names are fine.

Pyrofab (approved):

mappings/net/minecraft/util/math/MutableBox.mapping
```diff
@@ -28,27 +28,33 @@ CLASS net/minecraft/class_3341 net/minecraft/util/math/MutableIntBoundingBox
 		ARG 1 other
 	METHOD method_14658 toNbt ()Lnet/minecraft/class_2495;
 	METHOD method_14659 getSize ()Lnet/minecraft/class_2382;
```
Pyrofab: New suggestion: `getSize` -> `getDimensions`. The latter is technically more correct and disambiguates it from the `getXYZSize` methods.

Sollace: So `getXDimention`?

I haven't looked at this for a few days, but iirc `getSize` here returns a scalar value, so perhaps `getVolume` would be better?

Pyrofab: Nop, keep `getXSize`, but rename the base `getSize` to `getDimensions`. That is because sizes are single values, whereas dimensions are vectors.

Runemoro:
I don't see why width, height, and depth would be x, y, z and not z, y, x... 

I suggest using `getSizeX` instead of `getXSize` to respect [the convention](https://github.com/FabricMC/yarn/blob/1.14.4/CONVENTIONS.md#coordinates) (it's the size's x component, not the "x size").

Sollace:
Whilst looking at MurableBox, I notice that the constructor not specifying a y range results in a box defaulted to 0-256 (world height).

This class is also really only used with structures and commands, so it makes me wonder...

This could either just be a `SelectionBox` or a `BlockSelectionArea`, or some mix of the two - something that represents an area in the world rather than entity hitboxes* like the oher one is used for.

* inb4 `Box` -> `Hitbox`

Pyrofab:
`Box` is used for a lot more than hitboxes, and MutableBox uses the [1, 512] y range by default. However, the fact that the latter uses whole numbers and that it is only used for blocks could be taken into account for naming it. I am thinking `BlockBox` may be a fairly straightforward name, in the same vein as `BlockPos`.

mappings/net/minecraft/util/math/Box.mapping
```diff
@@ -54,27 +68,48 @@ CLASS net/minecraft/class_238 net/minecraft/util/math/Box
 	METHOD method_1013 isValid ()Z
 	METHOD method_1014 expand (D)Lnet/minecraft/class_238;
 		ARG 1 value
-	METHOD method_17939 getXSize ()D
-	METHOD method_17940 getYSize ()D
-	METHOD method_17941 getZSize ()D
+	METHOD method_17939 getXLength ()D
+	METHOD method_17940 getYLength ()D
+	METHOD method_17941 getZLength ()D
```
Runemoro, line 73: Why change this? Every other size getter uses `getSizeX`/`Y`/`Z`. Using `length` in this one would be weird.

mappings/net/minecraft/util/math/Box.mapping
```diff
@@ -54,27 +68,48 @@ CLASS net/minecraft/class_238 net/minecraft/util/math/Box
...
 	METHOD method_992 rayTrace (Lnet/minecraft/class_243;Lnet/minecraft/class_243;)Ljava/util/Optional;
+		ARG 1 min
+		ARG 2 max
 	METHOD method_993 intersects (Lnet/minecraft/class_243;Lnet/minecraft/class_243;)Z
+		ARG 1 from
+		ARG 2 to
 	METHOD method_994 intersects (Lnet/minecraft/class_238;)Z
 		ARG 1 box
-	METHOD method_995 averageDimension ()D
+	METHOD method_995 getAverageSideLength ()D
```
Runemoro, line 94: Why change this? "Dimension" means "side length" but is more concise.

> [a measurable extent of some kind, such as **length, breadth, depth, or height**](https://www.google.com/search?q=dimension)

mappings/net/minecraft/util/math/BlockBox.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_3341 net/minecraft/util/math/MutableIntBoundingBox
+CLASS net/minecraft/class_3341 net/minecraft/util/math/BlockBox
```
Runemoro, line 1: We already have a class named `Box` which it's just an immutable version of this one (the same way we have `BlockPos` and `MutableBlockPos`). Either this class should be `MutableBox` and `Box` should stay `Box`, or this class should be `MutableBlockBox` and `Box` should be named `BlockBox`.

Pyrofab: The idea behind calling this one `BlockBox` is that it's the equivalent of the `Box` class but for blocks - it uses ints and is used only in block contexts. The main difference between the other `Box` and this one is not mutability, it's the type of data it contains.

mappings/net/minecraft/util/math/Box.mapping
Pyrofab: I think "side length" is clearer, but that's really a matter of taste.

mappings/net/minecraft/util/math/Box.mapping
Pyrofab: This change was originally a suggestion of mine, but I agree that we should probably stick to established conventions.

mappings/net/minecraft/util/math/BlockBox.mapping
Runemoro: > The main difference between the other Box and this one is not mutability, it's the type of data it contains.

They contain exactly the same data (ints minX, minY, minZ, maxX, maxY, maxZ) and have the same methods.

Pyrofab: Not quite, the original `Box` uses doubles, and is used in a lot of contexts. This one uses ints and is limited to block contexts.

Runemoro: > the original Box uses doubles,

Sorry, I thought it was also using ints, I should have looked at the code.

mappings/net/minecraft/util/math/BlockBox.mapping
```diff
@@ -27,28 +27,34 @@ CLASS net/minecraft/class_3341 net/minecraft/util/math/MutableIntBoundingBox
...
+	METHOD method_14661 offset (III)V
 		ARG 1 dx
 		ARG 2 dy
 		ARG 3 dz
 	METHOD method_14662 contains (Lnet/minecraft/class_2382;)Z
 		ARG 1 vec
-	METHOD method_14663 getBlockCountY ()I
-	METHOD method_14664 getBlockCountZ ()I
+	METHOD method_14663 getYLength ()I
+	METHOD method_14664 getZLength ()I
```
Runemoro: I think it should stay block count to avoid confusion between "length" and "dimensions". The `getDimensions` method returns max - min:

    public Vec3i getSize() {
        return new Vec3i(this.maxX - this.minX, this.maxY - this.minY, this.maxZ - this.minZ);
    }

while the `getBlockCount` methods return max - min + 1 (the number of blocks, so a box from (0, 0, 0) to (0, 0, 0) will have a block count of 1 since it contains the block at (0, 0, 0):

    public int getBlockCountX() {
        return this.maxX - this.minX + 1;
    }
    
    public int getBlockCountY() {
        return this.maxY - this.minY + 1;
    }
    
    public int getBlockCountZ() {
        return this.maxZ - this.minZ + 1;
    }

mappings/net/minecraft/util/math/Box.mapping
Sollace: I'm with Pyro on this one. Not only that, but `dimension` is very similar to `world dimension`, `entity dimensions`, and just the term itself implies this method returns a vector, or therwise multiple scalars to describe the "dimensions" of the box. (see: `BlockBox.getDimensions`)

`Length` on the other hand, is always a scalar.

mappings/net/minecraft/util/math/BlockBox.mapping
Sollace: So... "conversation resolved"?

mappings/net/minecraft/util/math/BlockBox.mapping
Sollace: Yeah, I think so. It was originally changed to `getX/Y/ZSize` to conform with the method in `Box` but now that the class has been renamed to `BlockBox` there is no longer any need to maintain parity.

Sollace: The names are _horribly_ unwieldy and clunky to read, though, so I welcome any, literally _any_ other suggestions.

Pyrofab: I reckon this can always be changed later if someone has a better idea.

mappings/net/minecraft/util/math/Box.mapping
Sollace: `getSizeX` IMO, is counter to the CONVENTIONS.md

> To make code as easy to read as possible, keep names in the natural language order. For example, a class representing a chest block entity should be named `ChestBlockEntity` rather than `BlockEntityChest`.

Runemoro: What? `getXSize` breaks the convention. There's no such thing as an "x size" (unless you're talking about the size of the "x" character in a font). There is though a "size's x", the x component of the size.

Also, the conventions mention this explicitly: https://github.com/FabricMC/yarn/blob/1.14.4/CONVENTIONS.md#coordinates

In any case, even if you disagree, it should still be `sizeX`/`Y`/`Z` and you should open an issue about changing *all* `somethingX`/`Y`/`Z`/ variables to `x`/`y`/`zSomething`, not just change this one and be inconsistent ([within the same class!](https://github.com/FabricMC/yarn/pull/801/files#diff-281a29125b212111a67a0e63eb6dfaffR31)). That would make people have to remember which methods have the coordinate in front and which have it at the end.

Sollace: Get X Size = get the size in the X dimension
Get X Length = get the length of the X component

Compare that to a real conversaion:

"What is your cup('s) size?"
"My cup('s) size is DD"

As opposed to:

"What is your size('s) cup?"
"My size has a cup?"

Do you even english?

Sollace: You can also consider them a mutation of `x`, `y`, and `z`.

Where each component is a scalar _range_. If the mins and maxes of each dimension where refactored into a dedicated `Range` object the equivalent of these methods would all be:
`getX().length()`, `getY().length()`, `getZ().length()`.

Or.

`getX().getSize()`, `getY().getSize()`, `getZ().getSize()`.

Runemoro: I have never heard anyone refer to the size in the x dimension as the "x size". 

> What is your size('s) cup?

A size doesn't have a cup. But a cup has a size, and that size has three components.

A size doesn't have a box. But a box has a size, and that size has three components.

> Do you even english?

Do you even logic?

Also, ironically, you forgot the capital letter in English exactly in the sentence you insult my English.

> Where each component is a scalar range. If the mins and maxes of each dimension where refactored into a dedicated Range object the equivalent of these methods would all be:
getX().length(), getY().length(), getZ().length().

I would never guess that a method named `getX` returns a range, I'd instead assume it returns the position on the x axis. If I saw `getX().length()` I'd be confused about how a position can possibly have a length, just as I'd be confused when I see `getXLength`.

If you really want to use the (weird) range naming, then it would have to be `getXRangeLength` (get the x [coordinate's] range length) to be clear.

Also, by the same argument as yours, I can argue that `getSizeX` is correct, because it represents `getSize().getX()`. It's obviously more common to see methods that return sizes than methods that return ranges, so this would be the better option even if the other one was correct (not saying it is).

I suggest that this PR doesn't get merged as it is until a PR changing `CONVENTIONS.MD` is merged, since it's currently breaking the conventions.

Sollace: I'll agree to that until someone else can weigh in on this.

As it is, I'm not against renaming the methods to something else. Just as it stands, though the pre-X version are _obviously_ better in my opnion, neither options `getSizeX`, `getLengthX`, `getXSize` or `getXLength` are very elegant names due to the mixing of "dimension-axis + arbitrary-modifier".

valoeghese:
*cue all structure mods causing syntax errors in the dev environment*

Runemoro (approved):

Runemoro:
Is anything stopping this PR from being merged? It's been open for almost a month and has two approvals since more than a week.

liach (approved):

mappings/net/minecraft/util/math/MutableBox.mapping
```diff
@@ -28,23 +28,23 @@ CLASS net/minecraft/class_3341 net/minecraft/util/math/MutableIntBoundingBox
...
-		ARG 2 z1
-		ARG 3 x2
-		ARG 4 y2
-		ARG 5 z2
+		ARG 0 minX
+		ARG 1 minY
+		ARG 2 minZ
+		ARG 3 maxX
+		ARG 4 maxY
+		ARG 5 maxZ
```
ghost: Why was this change to min/max when coords can be given in either position? e.g. `create(1, 1, 1, 2, 2, 2)` and `create(2, 2, 2, 1, 1, 1)` results in same box.
```java
public static MutableIntBoundingBox create(int int_1, int int_2, int int_3, int int_4, int int_5, int int_6) {
    return new MutableIntBoundingBox(Math.min(int_1, int_4), Math.min(int_2, int_5), Math.min(int_3, int_6), Math.max(int_1, int_4), Math.max(int_2, int_5), Math.max(int_3, int_6));
}
```

mappings/net/minecraft/util/math/Box.mapping
```diff
@@ -6,11 +6,17 @@ CLASS net/minecraft/class_238 net/minecraft/util/math/Box
...
 	METHOD <init> (DDDDDD)V
-		ARG 1 x1
-		ARG 3 y1
-		ARG 5 z1
+		ARG 1 minX
+		ARG 3 minY
+		ARG 5 minZ
+		ARG 7 maxX
+		ARG 9 maxY
+		ARG 11 maxZ
```
ghost: Same as my other comment, positions can be given in either order so why use min/max.

ghost: Other occurances exist in BlockPos, vec3d constructors aswell.

mappings/net/minecraft/util/math/MutableBox.mapping
Sollace: Because when I was working on this the general consensus was against using x/y/z1 and x/y/z2. These names are also not very friendly, and actually rather confusing as I've found in the past.

mappings/net/minecraft/util/math/Box.mapping
Pyrofab: According to the code you posted above, the `create` methods put the values in the right order for the constructor. Meaning the constructor does care about the order, and the argument names are correct.

mappings/net/minecraft/util/math/MutableBox.mapping
Runemoro: @Sollace I agree with @NinjaPhenix, these should be changed back to `x1`, `x2`. The new names are much more confusing, they're making me think I have to make sure they are min and max, when they can in fact be max and min. Please don't add extra restrictions through parameter names.

> and actually rather confusing as I've found in the past.

What's confusing about being able to make a box between two points, with no restrictions at all...?


@liach @Pyrofab (mentioning you in case like me you weren't aware of this when you reviewed) What are your thoughts?

Runemoro (changes requested):
(see https://github.com/FabricMC/yarn/pull/801#discussion_r317400638)

mappings/net/minecraft/util/math/Box.mapping
liach: Given mojang does not require max/min, consider names like `xa` `xb` etc.

liach (changes requested):

mappings/net/minecraft/util/math/MutableBox.mapping
Pyrofab: Agreed, at first I thought the `create` method just passed everything to the constructor without changing the argument order.

Sollace:
Alright, changed back to pos1/pos2 then (bleh)

Only outlier is the constructor for BlockBox

```
    public BlockBox(int minX, int minY, int minZ, int maxX, int maxY, int maxZ) {
        this.minX = minX;
        this.minY = minY;
        this.minZ = minZ;
        this.maxX = maxX;
        this.maxY = maxY;
        this.maxZ = maxZ;
    }
```
(this is why I changed them, so it's standardized between all the methods)

Runemoro (approved):

Sollace:
@Runemoro Might have to re-re-approve this. I don't know.

Runemoro (approved):

ghost (approved):

modmuss50 (approved):

modmuss50:
Merged in my IDE, thanks.

PR #802, opened 2019-07-26
LivingEntity mappings

Sollace:
Using Stitch (and the extracted sources from the gradle task) I finally managed to open LivingEntity so I could map the fields.

Notable changes:
`getHealthMaximum` -> `getMaximumHealth`
Pretty obvious.

Instances of PotionEffect become StatusEffect (we call them Status Effects almost everywhere else)
    `addPotionEffect` -> `addStatusEffect` ()
    `clearPotionEffects` -> `clearStatusEffects`
    `spawnPotionParticles` -> `tickStatusEffects` (this does a lot more than spawn particles. It will also remove expired ones

`method_18396` -> `getArmorVisibility`
Called only when the entity is invisible. The value is just the factor (0-1) of what percentage the body is covered by armour. A higher value makes you more visible to enemies.

`method_16212` -> `getBlockStateUnderFeet` (suggestions welcome)
This gets the blockstate the entity is standing in. 

`field_6239` -> `roll`
Used in the renderer when swimming or flying with an elytra.

`field_6267` -> `diagonalSpeed`
Used by the Ender Dragon.

`field_6255` -> `lookDirection` (result of turnHead) 
`turnHead` takes an angle as the second parameter, and will either return the same angle or `angle * -1`.

`field_6228` -> `jumpingCooldown`
A cooldown timer used to keep you from ~~breaking your legs~~ bunnyhopping too quickly.

`method_6024` -> `getLeaningPitch`
`field_6243` -> `leaningPitch`
`field_6264` -> `lastLeaningPitch`
This is the angles used to make the player lean forwards as it transitions between standing and crawlling/swiming.

### DIFF
--- a/mappings/net/minecraft/entity/LivingEntity.mapping
+++ b/mappings/net/minecraft/entity/LivingEntity.mapping
@@ -39,6 +39,7 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	FIELD field_6250 forwardSpeed F
 	FIELD field_6251 handSwingProgress F
 	FIELD field_6252 isHandSwinging Z
+	FIELD field_6253 lastDamageTaken F
 	FIELD field_6254 maxHurtTime I
 	FIELD field_6255 lookDirection F
 	FIELD field_6256 damageTracker Lnet/minecraft/class_1283;
--- a/mappings/net/minecraft/entity/LivingEntity.mapping
+++ b/mappings/net/minecraft/entity/LivingEntity.mapping
@@ -182,7 +182,7 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_6048 getItemUseTime ()I
 	METHOD method_6049 isPotionEffective (Lnet/minecraft/class_1293;)Z
 		ARG 1 effect
-	METHOD method_6050 spawnPotionParticles ()V
+	METHOD method_6050 tickStatusEffects ()V
 	METHOD method_6051 getRand ()Ljava/util/Random;
 	METHOD method_6052 getAttacking ()Lnet/minecraft/class_1309;
 	METHOD method_6053 initAi ()V
--- a/mappings/net/minecraft/entity/LivingEntity.mapping
+++ b/mappings/net/minecraft/entity/LivingEntity.mapping
@@ -173,6 +173,7 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_6037 spawnItemParticles (Lnet/minecraft/class_1799;I)V
 		ARG 1 item
 		ARG 2 count
+	METHOD method_6038 onDismounted (Lnet/minecraft/class_1297;)V
 	METHOD method_6039 isBlocking ()Z
 	METHOD method_6040 consumeItem ()V
 	METHOD method_6041 getFallSound (I)Lnet/minecraft/class_3414;
--- a/mappings/net/minecraft/entity/LivingEntity.mapping
+++ b/mappings/net/minecraft/entity/LivingEntity.mapping
@@ -85,7 +85,7 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_16079 getLootContextBuilder (ZLnet/minecraft/class_1282;)Lnet/minecraft/class_47$class_48;
 		ARG 1 killedByPlayer
 	METHOD method_16080 drop (Lnet/minecraft/class_1282;)V
-	METHOD method_16212 getBlockStateUnderFeet ()Lnet/minecraft/class_2680;
+	METHOD method_16212 getBlockState ()Lnet/minecraft/class_2680;
 	METHOD method_16826 setDespawnCounter (I)V
 		ARG 1 despawnCounter
 	METHOD method_17825 getScaleFactor ()F
--- a/mappings/net/minecraft/entity/LivingEntity.mapping
+++ b/mappings/net/minecraft/entity/LivingEntity.mapping
@@ -220,7 +220,7 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_6046 getGroup ()Lnet/minecraft/class_1310;
 	METHOD method_6047 getMainHandStack ()Lnet/minecraft/class_1799;
 	METHOD method_6048 getItemUseTime ()I
-	METHOD method_6049 isPotionEffective (Lnet/minecraft/class_1293;)Z
+	METHOD method_6049 canHaveStatusEffect (Lnet/minecraft/class_1293;)Z
 		ARG 1 effect
 	METHOD method_6050 tickStatusEffects ()V
 	METHOD method_6051 getRand ()Ljava/util/Random;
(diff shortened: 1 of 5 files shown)


Sollace:
Also this doesn't contain any parameter name changes, since I can't see those in IDEA.

modmuss50 (approved):
Looks good to me.

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -2,48 +2,79 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
...
 	FIELD field_6259 prevHeadYaw F
-	FIELD field_6260 attributeContainer Lnet/minecraft/class_1325;
+	FIELD field_6260 attributes Lnet/minecraft/class_1325;
+	FIELD field_6261 pushCooldown I
+	FIELD field_6262 randomSmallSeed F
+	FIELD field_6263 serverZ D
+	FIELD field_6264 lastLeaningPitch F
+	FIELD field_6265 headTrackingIncrements I
 	FIELD field_6266 preferredHand Lnet/minecraft/class_1268;
+	FIELD field_6267 diagonalSpeed F
```
Runemoro: This is `turningSpeed`, not `diagonalSpeed`. In `EnderDragonEntity`:
```java
this.yaw += this.field_6267 * 0.1f;
```

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -54,6 +85,7 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_16079 getLootContextBuilder (ZLnet/minecraft/class_1282;)Lnet/minecraft/class_47$class_48;
 		ARG 1 killedByPlayer
 	METHOD method_16080 drop (Lnet/minecraft/class_1282;)V
+	METHOD method_16212 getBlockStateUnderFeet ()Lnet/minecraft/class_2680;
```
Runemoro: "Feet" doesn't make sense for all entities, since fish don't have feet, for example. Also, this returns the block *at* the feet of a player, not under them.

It should just be `getBlockState`.

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -169,23 +212,27 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_6059 hasStatusEffect (Lnet/minecraft/class_1291;)Z
 	METHOD method_6060 knockback (Lnet/minecraft/class_1309;)V
 		ARG 1 target
-	METHOD method_6062 cannotMove ()Z
-	METHOD method_6063 getHealthMaximum ()F
+	METHOD method_6061 canPierceDefense (Lnet/minecraft/class_1282;)Z
```
Runemoro: This is wrong, it actually returns whether an attack hits the shield:

```java
    private boolean hitsShield(DamageSource source) {
        Entity attacker = source.getSource();
        boolean pierces = false;

        if (attacker instanceof ProjectileEntity) {
            ProjectileEntity projectile = (ProjectileEntity) attacker;
            if (projectile.getPierceLevel() > 0) {
                pierces = true;
            }
        }

        if (!source.bypassesArmor() && isBlocking() && !pierces) { // using shield and damage doesn't bypass armor
            Vec3d pos = source.getPosition();
            if (pos != null) {
                Vec3d rotation = getRotationVec(1.0F);
                Vec3d diff = pos.reverseSubtract(new Vec3d(x, y, z)).normalize();
                diff = new Vec3d(diff.x, 0.0D, diff.z);
                return diff.dotProduct(rotation) < 0.0D; // shield was hit
            }
        }

        return false;
    }
```

```java
            if (amount > 0.0F && hitsShield(source)) {
                damageShield(amount);
                damageAbsorbedByShield = amount;
                amount = 0.0F;
                if (!source.isProjectile()) {
                    Entity attacker = source.getSource();
                    if (attacker instanceof LivingEntity) {
                        takeShieldHit((LivingEntity) attacker);
                    }
                }

                shieldWasHit = true;
            }
```

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -209,9 +256,11 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 		ARG 1 source
 	METHOD method_6091 travel (Lnet/minecraft/class_243;)V
 		ARG 1 movementInput
-	METHOD method_6092 addPotionEffect (Lnet/minecraft/class_1293;)Z
+	METHOD method_6092 addStatusEffect (Lnet/minecraft/class_1293;)Z
 		ARG 1 effect
+	METHOD method_6093 knockDownwards ()V
 	METHOD method_6094 canBreatheInWater ()Z
+	METHOD method_6095 absorbDamage (Lnet/minecraft/class_1282;)Z
```
Runemoro: This is wrong, it should be `tryUsingTotem`:

```java
    private boolean tryUsingTotem(DamageSource source) {
        if (source.isOutOfWorld()) {
            return false;
        }

        ItemStack totem = null;

        for (Hand hand : Hand.values()) {
            ItemStack stackInHand = getStackInHand(hand);
            if (stackInHand.getItem() == Items.TOTEM_OF_UNDYING) {
                totem = stackInHand.copy();
                stackInHand.decrement(1);
                break;
            }
        }

        if (totem != null) {
            if (this instanceof ServerPlayerEntity) {
                ServerPlayerEntity item = (ServerPlayerEntity) this;
                item.incrementStat(Stats.USED.getOrCreateStat(Items.TOTEM_OF_UNDYING));
                Criterions.USED_TOTEM.handle(item, totem);
            }

            setHealth(1.0F);
            clearPotionEffects();
            addPotionEffect(new StatusEffectInstance(StatusEffects.REGENERATION, 900, 1));
            addPotionEffect(new StatusEffectInstance(StatusEffects.ABSORPTION, 100, 1));
            world.sendEntityStatus(this, (byte) 35);
        }

        return totem != null;
    }
```

mappings/net/minecraft/entity/LivingEntity.mapping
Runemoro: @Sollace You renamed this `tryUseShield`, but it doesn't actually use the shield, it just returns true if the shield can block the attack (if the player is blocking, and the line between the the entity and attacker intersects the shield). The second piece of code above shows the *usage* of the `hitsShield` method.

Sollace: @Runemoro `tryUseShield(source)` seems to work fine in that context, though.

```
        if (amount > 0 && tryUseShield(source)) {
            blockedByShield = true;

            damageShield(amount);
            amountAbsorbedByShield = amount;
            amount = 0;

            if (!source.isProjectile()) {
               Entity attacker = source.getSource();
               if (attacker instanceof LivingEntity) {
                  takeShieldHit((LivingEntity)attacker);
               }
            }
         }
```

If the amount is >0 (not fully absorbed) it tries to use the shield and then does the things that only occur if the shield succeeded in blocking the attack.

Runemoro: > it tries to use the shield

No. It checks if the shield can block the attack. Trying to use the shield would mean that it's possible for a call to `tryUseShield` to actually use the shield and damage it, subtract the absorbed damage, play shield hit sound, etc. Compare this method with `tryUseTotem`, which will actually destroy the totem, start the animation, and heal the player if it the player is holding a totem.

Checking if you can use something is in no way the same as actually using (or attempting to use) something.

Sollace: Alright, changed.

modmuss50 (approved):

Runemoro (approved):

PR #803, opened 2019-07-26
BlockPos mappings

Sollace:
I swear this is the last one! I noticed BlockPos was missing almost all of its parameter name mappings, and what we had were half wrong.

`method_10075` -> `crossProductPos`
This is just a `crossProduct`, however I couldn't name it that because Vec3i already defines a method with that name (same thing, but returning a Vec3i)

### DIFF
--- a/mappings/net/minecraft/util/math/BlockPos.mapping
+++ b/mappings/net/minecraft/util/math/BlockPos.mapping
@@ -1,6 +1,7 @@
 CLASS net/minecraft/class_2338 net/minecraft/util/math/BlockPos
 	CLASS 1
 		FIELD field_17676 connector Lnet/minecraft/class_3980;
+		FIELD field_18231 position Lnet/minecraft/class_2338$class_2339;
 		METHOD tryAdvance (Ljava/util/function/Consumer;)Z
 			ARG 1 consumer
 	CLASS 2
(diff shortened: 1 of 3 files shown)


mappings/net/minecraft/util/math/BlockPos.mapping
```diff
@@ -72,27 +116,44 @@ CLASS net/minecraft/class_2338 net/minecraft/util/math/BlockPos
 		ARG 2 y
 		ARG 3 z
 	METHOD method_10070 rotate (Lnet/minecraft/class_2470;)Lnet/minecraft/class_2338;
+		ARG 1 rotation
 	METHOD method_10071 unpackLongY (J)I
+		ARG 0 y
 	METHOD method_10072 south ()Lnet/minecraft/class_2338;
 	METHOD method_10074 down ()Lnet/minecraft/class_2338;
+	METHOD method_10075 crossProductPos (Lnet/minecraft/class_2382;)Lnet/minecraft/class_2338;
```
kvverti: Methods are allowed to overload by return type. Since `BlockPos` extends `Vec3i`, this doesn't pose a problem at source level.

Pyrofab: The argument is still `Vec3i`, which means they differ only through their return type. This is only possible in bytecode, not at source level, so they must have different names in the source.

kvverti: `BlockPos` extends `Vec3i`, so at source level this is a method with a covariant return type. There is a similar synthetic method in `BlockPos` which does override the superclass method. Enigma doesn't allow covariant returns for whatever reason, though, so if we want it to have the same name (which I believe it should), it will have to be done manually.

Pyrofab: Oh, there is a bridge, should have said so from the start. Yes, it should definitely be called `crossProduct` then.

modmuss50 (approved):

Pyrofab (approved):

PR #820, opened 2019-08-18
FishingBobberEntity Mappings

juliand665:
I hope I followed the conventions correctly!

### DIFF
--- a/mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
+++ b/mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
@@ -11,7 +11,7 @@ CLASS net/minecraft/class_1536 net/minecraft/entity/projectile/FishingBobberEnti
 	FIELD field_7173 hookCountdown I
 	FIELD field_7174 waitCountdown I
 	FIELD field_7175 state Lnet/minecraft/class_1536$class_1537;
-	FIELD field_7176 hasHitNonEntity Z
+	FIELD field_7176 stuckOnBlock Z
 	FIELD field_7177 owner Lnet/minecraft/class_1657;
 	METHOD <init> (Lnet/minecraft/class_1657;Lnet/minecraft/class_1937;II)V
 		ARG 1 thrower
--- a/mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
+++ b/mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
@@ -1,10 +1,17 @@
 CLASS net/minecraft/class_1536 net/minecraft/entity/projectile/FishingBobberEntity
 	CLASS class_1537 State
 	FIELD field_7165 hookedEntity Lnet/minecraft/class_1297;
-	FIELD field_7168 luckOfTheSeaLevel I
+	FIELD field_7166 selfHitTimer I
+	FIELD field_7167 removalTimer I
+	FIELD field_7168 lureLevel I
+	FIELD field_7169 fishAngleDegrees F
 	FIELD field_7170 HOOK_ENTITY_ID Lnet/minecraft/class_2940;
-	FIELD field_7171 lureLevel I
+	FIELD field_7171 luckOfTheSeaLevel I
+	FIELD field_7172 fishTravelCountdown I
+	FIELD field_7173 hookCountdown I
+	FIELD field_7174 waitCountdown I
 	FIELD field_7175 state Lnet/minecraft/class_1536$class_1537;
+	FIELD field_7176 hasHitNonEntity Z
 	FIELD field_7177 owner Lnet/minecraft/class_1657;
 	METHOD <init> (Lnet/minecraft/class_1657;Lnet/minecraft/class_1937;II)V
 		ARG 1 thrower
@@ -23,4 +30,12 @@ CLASS net/minecraft/class_1536 net/minecraft/entity/projectile/FishingBobberEnti
 		ARG 3 lureLevel
 		ARG 4 luckOfTheSeaLevel
 	METHOD method_6947 getOwner ()Lnet/minecraft/class_1657;
+	METHOD method_6949 tickFishingLogic (Lnet/minecraft/class_2338;)V
+		ARG 1 pos
 	METHOD method_6951 updateHookedEntityId ()V
+	METHOD method_6952 smoothenMovement ()V
+	METHOD method_6954 pullHookedEntity ()V
+	METHOD method_6957 use (Lnet/minecraft/class_1799;)I
+		ARG 1 usedItem
+	METHOD method_6958 checkForCollision ()V
+	METHOD method_6959 checkRemoval ()Z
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
```diff
@@ -1,10 +1,17 @@
...
+	FIELD field_7168 lureLevel I
+	FIELD field_7169 fishAngleDegrees F
 	FIELD field_7170 HOOK_ENTITY_ID Lnet/minecraft/class_2940;
-	FIELD field_7171 lureLevel I
+	FIELD field_7171 luckOfTheSeaLevel I
+	FIELD field_7172 fishTravelCountdown I
+	FIELD field_7173 hookCountdown I
+	FIELD field_7174 waitCountdown I
 	FIELD field_7175 state Lnet/minecraft/class_1536$class_1537;
+	FIELD field_7176 hasHitNonEntity Z
```
Pyrofab: This name seems a bit unwieldy, and more appropriate for a getter than a boolean field.

mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
```diff
@@ -23,4 +30,12 @@ CLASS net/minecraft/class_1536 net/minecraft/entity/projectile/FishingBobberEnti
 		ARG 3 lureLevel
 		ARG 4 luckOfTheSeaLevel
 	METHOD method_6947 getOwner ()Lnet/minecraft/class_1657;
+	METHOD method_6949 tickFishingLogic (Lnet/minecraft/class_2338;)V
```
Pyrofab, line 33: I believe we tend to use `update` rather than `tick`, so that would be `updateFishingLogic`.

mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
liach: missedEntity?

mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
liach: we prefer tick over update.

mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
```diff
@@ -23,4 +30,12 @@ CLASS net/minecraft/class_1536 net/minecraft/entity/projectile/FishingBobberEnti
...
+		ARG 1 pos
 	METHOD method_6951 updateHookedEntityId ()V
+	METHOD method_6952 smoothenMovement ()V
+	METHOD method_6954 pullHookedEntity ()V
+	METHOD method_6957 use (Lnet/minecraft/class_1799;)I
+		ARG 1 usedItem
+	METHOD method_6958 checkForCollision ()V
+	METHOD method_6959 checkRemoval ()Z
```
liach: shouldBeRemoved?

mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
Pyrofab: https://github.com/FabricMC/yarn/blob/1.14.4/CONVENTIONS.md#ticks-and-updates
Looks like it. Although a lot of stuff in entities doesn't obey this rule.

juliand665: I went for `tick` because it's called from the `tick` method (every tick) whenever the bobber is in water (so effectively a conditional tick method); does that change your assessment?

mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
juliand665: To me, `shouldBeRemoved` sounds like a pure function that just returns whether or not the bobber should be removed, however, calling this method when the conditions are fulfilled actually goes ahead and removes the bobber, too. The boolean returned is in fact whether or not it was removed, though.
image
I've attached a screenshot of the code with deobfuscated locals from when I was looking at it.

mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
juliand665: Ah, yeah. On closer look, `HitResult` can only be `MISS`, `ENTITY`, or `BLOCK`, so `hasHitBlock` would work. As for the wording, is `hitBlock`, `onBlock`, or `stuckOnBlock` an improvement?

mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
Pyrofab: It seems that `tick` is indeed the better choice according to conventions.

mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
Pyrofab: I think `stuckOnBlock` is the clearest one.

Pyrofab (approved):

mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
liach: `tryRemove`/`tryInvalidate`

mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
```diff
@@ -1,10 +1,17 @@
 CLASS net/minecraft/class_1536 net/minecraft/entity/projectile/FishingBobberEntity
 	CLASS class_1537 State
 	FIELD field_7165 hookedEntity Lnet/minecraft/class_1297;
-	FIELD field_7168 luckOfTheSeaLevel I
+	FIELD field_7166 selfHitTimer I
+	FIELD field_7167 removalTimer I
+	FIELD field_7168 lureLevel I
+	FIELD field_7169 fishAngleDegrees F
```
liach: `degrees` may be unnecessary as most minecraft angles are in degrees

juliand665: fair point.

mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
juliand665: hm, i'm not happy with those since neither really expresses that it should only be removed if it's invalid. since it's a method name, maybe a longer name would be fine; how about `removeIfInvalid`?

mappings/net/minecraft/entity/projectile/FishingBobberEntity.mapping
```diff
@@ -1,10 +1,17 @@
 CLASS net/minecraft/class_1536 net/minecraft/entity/projectile/FishingBobberEntity
 	CLASS class_1537 State
 	FIELD field_7165 hookedEntity Lnet/minecraft/class_1297;
-	FIELD field_7168 luckOfTheSeaLevel I
+	FIELD field_7166 selfHitTimer I
```
Sollace, line 4: Is `self` really necessary here? Could we simply this to `hitTimer` instead?

juliand665: While `selfHitTimer > 0`, the bobber won't attach to the entity holding its fishing rod. I feel like the name `selfHitTimer` conveys that information pretty well.

liach (approved):
Should we merge this into 1.14.4? Might cherry-pick into 1.15 if we do.

Juuxel (approved):

PR #822, opened 2019-08-18
A few mappings and improvements

Runemoro:

PR #826, opened 2019-08-19
BlockHitResult.field_17591 -> isEnteringBlock

Sollace:
This field is set to true in only one place (`VoxelShape.rayTrace`) where it does the below check and returns a hit with an opposite facing:

```
@Nullable
public BlockHitResult rayTrace(Vec3d start, Vec3d end, BlockPos pos) {
		if (this.isEmpty()) {
				return null;
		}

		Vec3d ray = end.subtract(start);
		if (ray.lengthSquared() < 1.0E-7) {
				return null;
		}

		Vec3d projection = start.add(ray.multiply(0.001));
		if (this.contains(projection.x - pos.getX(), projection.y - pos.getY(), projection.z - pos.getZ())) {
				return new BlockHitResult(projection, Direction.getFacing(ray.x, ray.y, ray.z).getOpposite(), pos, true);
		}
		return Box.rayTrace(this.getBoundingBoxes(), start, end, pos);
}
```

The method starts by calculating the `ray` which is the approach vector between the start and end point, then scaling that down to project the start a small amount forward.

It then checks if that new position is inside the current `VoxelShape`. and if it is will return a hit at the projected position (`pos + projection`), and inverts the direction to be facing away from the block (towards the origin of the ray).

The value is then only used in `ScaffoldingItem` where it reverses the direction again so players can still place scaffolds at the block they're standing inside of (since Scaffolding will just stack higher when you do so).

```
 if (context.isEnteringBlock()) {
                Direction direction7 = context.isHitInsideBlock() ? context.getSide().getOpposite() : context.getSide();
            }
```

tl;dr
`isEnteringBlock() == false` implies `getSide()` is a direction leading in the same direction as the ray that caused the hit. (Ray is _leaving_ the block)
`isEnteringBlock() == true` implies `getSide()` is a direction _opposite_ to the ray that caused the hit. (Ray is _entering_ the block)

### DIFF
--- a/mappings/net/minecraft/item/ItemUsageContext.mapping
+++ b/mappings/net/minecraft/item/ItemUsageContext.mapping
@@ -15,6 +15,7 @@ CLASS net/minecraft/class_1838 net/minecraft/item/ItemUsageContext
 		ARG 4 stack
 		ARG 5 hit
 	METHOD method_17698 getHitPos ()Lnet/minecraft/class_243;
+	METHOD method_17699 isHitInteringBlock ()Z
 	METHOD method_20287 getHand ()Lnet/minecraft/class_1268;
 	METHOD method_8036 getPlayer ()Lnet/minecraft/class_1657;
 	METHOD method_8037 getBlockPos ()Lnet/minecraft/class_2338;
--- a/mappings/net/minecraft/item/ItemUsageContext.mapping
+++ b/mappings/net/minecraft/item/ItemUsageContext.mapping
@@ -15,7 +15,7 @@ CLASS net/minecraft/class_1838 net/minecraft/item/ItemUsageContext
 		ARG 4 stack
 		ARG 5 hit
 	METHOD method_17698 getHitPos ()Lnet/minecraft/class_243;
-	METHOD method_17699 isHitInteringBlock ()Z
+	METHOD method_17699 isHitEnteringBlock ()Z
 	METHOD method_20287 getHand ()Lnet/minecraft/class_1268;
 	METHOD method_8036 getPlayer ()Lnet/minecraft/class_1657;
 	METHOD method_8037 getBlockPos ()Lnet/minecraft/class_2338;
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/item/ItemUsageContext.mapping
```diff
@@ -15,6 +15,7 @@ CLASS net/minecraft/class_1838 net/minecraft/item/ItemUsageContext
 		ARG 4 stack
 		ARG 5 hit
 	METHOD method_17698 getHitPos ()Lnet/minecraft/class_243;
+	METHOD method_17699 isHitInteringBlock ()Z
```
liach: What is `intering`

Sollace: Whoops! Another of my typos slipped through, typing `int` when I actually mean something else...

Damn Primitives.

Runemoro:
This was already named `isInBlock` in my last PR (#822). It's just a boolean that's true if the hit was made on the inside of a block rather than the outside. 

`isEnteringBlock` sounds more like something that is true if you place a block in the empty space you're standing on (but that's not what it does, it would be false in this case).

Sollace:
@Runemoro I disagree. `isEnteringBlock` is specifically relating to the raytrace, not to block placements.

I guess another way to think of it is that it's true if the ray originated from a point inside the block it registered the hit against.

True - ray originated inside the block and is proceeding inward (direction is the opposite)
```
        \
       |  \      |
    <==|    \ray |
       |         |
        ----------
```
vs

False - ray originated outside the block and is proceeding outward (direction is unchanged)
```
          |    \   |
          |      \ |==>
          |        |\
          ----------
```

I did switch to calling it `isInBlock` as well as `isLeavingBlock` but upon investigating it closer it didn't think either of those were entirely accurate, though I can be wrong and I'm open to hear why `isInBlock` could be better.

Runemoro:
Raytraces never enter a block, they stop at block boundaries by definition.

Either they're fully inside a block or fully outside. And anyway, the hit result refers to the final position that was hit, not the whole ray.

Sollace:
Can anyone confirm what the overlaps are between this and #822 is? I can't find it, and if he's already mapped all the fields named by this PR, I can just close it.

liach:
Yeah, I've suggested the name `hitsInsideBlock` for #822, and this pr is pretty much... redundant?

Sollace:
These are all the changes this PR introduces. If you can confirm you have names for all of the, I'll close this one:

```
ItemUsageContext
	METHOD method_17699 isHitEnteringBlock ()Z
BlockHitResult
	FIELD field_17591 entering Z
	METHOD method_17781 isEnteringBlock ()Z
```

liach:
Yes, rune did name all the fields and methods.

PR #839, opened 2019-08-26
Shape mappings

Sollace:
This maps all the remaining methods in `net/minecraft/util/shape`.

Also renamed `DoubleListPair` -> `PairList`. It's a list of pairs (of doubles) not a double pair of lists.

You can see this when looking at `DoubleListPair.SectionPairPredicate` (`PairList.Consumer`) which accepts a value from the first element of the pair (`x`), the second element of the pair (`y`), and the position in the list (`index`).

```
    public interface PairList.Consumer {
        boolean merge(int x, int y, int index);
    }
```

From SimplePairList:
```
    @Override
    public boolean forEachPair(PairList.Consumer consumer) {
        for (int index = 0; index< this.valueIndices.size() - 1; index++) {
            if (!consumer.merge(minValues.getInt(index), maxValues.getInt(index), index)) {
                return false;
            }
        }
        return true;
    }
```

### DIFF
--- a/mappings/net/minecraft/block/BlockState.mapping
+++ b/mappings/net/minecraft/block/BlockState.mapping
@@ -5,9 +5,11 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 		FIELD field_16557 fullOpaque Z
 		FIELD field_16558 opaque Z
 		FIELD field_16559 DIRECTIONS [Lnet/minecraft/class_2350;
-		FIELD field_16560 shapes [Lnet/minecraft/class_265;
+		FIELD field_16560 extrudedFaces [Lnet/minecraft/class_265;
+		FIELD field_17651 exceedsCube Z
 		FIELD field_19360 collisionShape Lnet/minecraft/class_265;
 		FIELD field_19429 solidFullSquare [Z
+		FIELD field_20337 isFullCube Z
 		METHOD <init> (Lnet/minecraft/class_2680;)V
 			ARG 1 state
 	FIELD field_12290 shapeCache Lnet/minecraft/class_2680$class_3752;
@@ -60,6 +62,7 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 		ARG 3 pos
 	METHOD method_11590 initShapeCache ()V
 	METHOD method_11591 canPlaceAt (Lnet/minecraft/class_4538;Lnet/minecraft/class_2338;)Z
+		ARG 1 view
 		ARG 2 pos
 	METHOD method_11592 isSideInvisible (Lnet/minecraft/class_2680;Lnet/minecraft/class_2350;)Z
 		ARG 1 neighbor
@@ -96,6 +99,7 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 	METHOD method_11606 getOutlineShape (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;Lnet/minecraft/class_3726;)Lnet/minecraft/class_265;
 		ARG 1 view
 		ARG 2 pos
+		ARG 3 context
 	METHOD method_11607 getRayTraceShape (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
 		ARG 1 view
 		ARG 2 pos
@@ -178,7 +182,7 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 		ARG 1 view
 		ARG 2 pos
 		ARG 3 ePos
-	METHOD method_16384 getCullingShape (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;)Lnet/minecraft/class_265;
+	METHOD method_16384 getCullingFace (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;)Lnet/minecraft/class_265;
 		ARG 1 view
 		ARG 2 pos
 		ARG 3 facing
@@ -192,14 +196,18 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 	METHOD method_17770 getOutlineShape (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
 		ARG 1 view
 		ARG 2 pos
+	METHOD method_17900 exceedsCube ()Z
 	METHOD method_19287 onProjectileHit (Lnet/minecraft/class_1937;Lnet/minecraft/class_2680;Lnet/minecraft/class_3965;Lnet/minecraft/class_1297;)V
 		ARG 1 world
 		ARG 2 state
 		ARG 3 hitResult
 		ARG 4 projectile
 	METHOD method_20827 isSideSolidFullSquare (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;)Z
+		ARG 1 view
+		ARG 2 pos
+		ARG 3 direction
 	METHOD method_21743 isFullCube (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Z
-		ARG 1 world
+		ARG 1 view
 		ARG 2 pos
 	METHOD method_22360 canBucketPlace (Lnet/minecraft/class_3611;)Z
 		ARG 1 fluid
--- a/mappings/net/minecraft/block/FenceBlock.mapping
+++ b/mappings/net/minecraft/block/FenceBlock.mapping
@@ -4,4 +4,5 @@ CLASS net/minecraft/class_2354 net/minecraft/block/FenceBlock
 		ARG 1 settings
 	METHOD method_10184 canConnect (Lnet/minecraft/class_2680;ZLnet/minecraft/class_2350;)Z
 		ARG 1 state
+		ARG 2 neighborIsFullSquare
 		ARG 3 dir
--- a/mappings/net/minecraft/client/render/block/FluidRenderer.mapping
+++ b/mappings/net/minecraft/client/render/block/FluidRenderer.mapping
@@ -17,10 +17,10 @@ CLASS net/minecraft/class_775 net/minecraft/client/render/block/FluidRenderer
 		ARG 1 world
 		ARG 2 pos
 	METHOD method_3344 isSideCovered (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;F)Z
-		ARG 0 world
+		ARG 0 view
 		ARG 1 pos
-		ARG 2 side
-		ARG 3 height
+		ARG 2 direction
+		ARG 3 maxDeviation
 	METHOD method_3345 onResourceReload ()V
 	METHOD method_3346 getNorthWestCornerFluidHeight (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;Lnet/minecraft/class_3611;)F
 		ARG 1 world
--- a/mappings/net/minecraft/util/math/Direction.mapping
+++ b/mappings/net/minecraft/util/math/Direction.mapping
@@ -16,6 +16,9 @@ CLASS net/minecraft/class_2350 net/minecraft/util/math/Direction
 			ARG 3 y
 			ARG 5 z
 		METHOD method_10173 choose (III)I
+			ARG 1 x
+			ARG 2 y
+			ARG 3 z
 		METHOD method_10174 getName ()Ljava/lang/String;
 		METHOD method_10177 fromName (Ljava/lang/String;)Lnet/minecraft/class_2350$class_2351;
 			ARG 0 name
@@ -27,11 +30,19 @@ CLASS net/minecraft/class_2350 net/minecraft/util/math/Direction
 	CLASS class_2352 AxisDirection
 		FIELD field_11057 desc Ljava/lang/String;
 		FIELD field_11059 offset I
+		METHOD <init> (Ljava/lang/String;IILjava/lang/String;)V
+			ARG 3 offset
+			ARG 4 description
 		METHOD method_10181 offset ()I
 	CLASS class_2353 Type
 		FIELD field_11061 facingArray [Lnet/minecraft/class_2350;
 		FIELD field_11065 axisArray [Lnet/minecraft/class_2350$class_2351;
+		METHOD <init> (Ljava/lang/String;I[Lnet/minecraft/class_2350;[Lnet/minecraft/class_2350$class_2351;)V
+			ARG 3 directions
 		METHOD method_10183 random (Ljava/util/Random;)Lnet/minecraft/class_2350;
+			ARG 1 random
+		METHOD test (Ljava/lang/Object;)Z
+			ARG 1 direction
 	FIELD field_11030 idHorizontal I
 	FIELD field_11031 idOpposite I
 	FIELD field_11032 id I
--- a/mappings/net/minecraft/util/shape/VoxelSet.mapping
+++ b/mappings/net/minecraft/util/shape/VoxelSet.mapping
@@ -1,6 +1,18 @@
 CLASS net/minecraft/class_251 net/minecraft/util/shape/VoxelSet
-	CLASS class_252 DirectionalPointConsumer
-	CLASS class_253 VolumeConsumer
+	CLASS class_252 PositionConsumer
+		METHOD consume (Lnet/minecraft/class_2350;III)V
+			ARG 1 direction
+			ARG 2 x
+			ARG 3 y
+			ARG 4 z
+	CLASS class_253 PositionBiConsumer
+		METHOD consume (IIIIII)V
+			ARG 1 x1
+			ARG 2 y1
+			ARG 3 z1
+			ARG 4 x2
+			ARG 5 y2
+			ARG 6 z2
 	FIELD field_1372 zSize I
 	FIELD field_1373 ySize I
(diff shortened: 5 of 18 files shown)


mappings/net/minecraft/util/shape/VoxelSet.mapping
```diff
@@ -1,4 +1,18 @@
 CLASS net/minecraft/class_251 net/minecraft/util/shape/VoxelSet
+	CLASS class_252 PositionConsumer
+		METHOD consume (Lnet/minecraft/class_2350;III)V
+			ARG 1 direction
+			ARG 2 x
+			ARG 3 y
+			ARG 4 z
+	CLASS class_253 EdgeConsumer
```
Runemoro: Are you sure this is never used to iterate through boxes too? Because if I remember correctly, I think I saw this used foe both edge and box iteeation. In that case, `PositionPairConsumer` might be better.

Sollace: Maybe. I didn't specifically look for that. It seemed to be mostly used in methods that deal with edges.

Sollace: Seems you're correct! I've changed it to `BiPositionConsumer` (following the naming pattern of things like `BiConsumer` and `BiFunction`)

Runemoro: Hmm, that's OK, but I think `PositionBiConsumer` would be a bit better (it's a BiConsumer for Positions, not a Comsumer for BiPositions).

Sollace: Hm... Thing is it takes 6 parameters that make up two sets of coordinates. It doesn't take two coordinate parameters, which is where I see the distinction.

Runemoro: > takes 6 parameters that make up two sets of coordinates

Yeah, so it takes two positions, each in three parameters.

> It doesn't take two coordinate parameters

It doesn't, but I'm suggesting `PositionBiConsumer`, not `CoordinateBiConsumer`, How it passes the two components is just an implementation detail, but it's equivalent to  `BiConsumer<Vec3i>`.

> which is where I see the distinction.

I don't see how the name you're suggesting makes it clear that the positions are passed as six parameters rather than two.

At best, the names are equivalent, but then we should match the fastutil names (`IntBiConsumer`, `FloatBiConsumer`, etc.).

Sollace: > It doesn't, but I'm suggesting PositionBiConsumer, not CoordinateBiConsumer,

I'll rephrase then. It doesn't take two _position_ parameters.

It's not `accept(Position one, Position two)` which would be a `BiConsumer<T>`

It's a `accept(one, two, three, four, five, size)` which would be technically be a `HexaConsumer<T>`

Why does this matter, anyway?

Runemoro: Yes, so either `CoordinateHexaConsumer` or `PositionBiConsumer`. A `BiConsumer` is a well known name for something that consumes two things, so using `PositionBiConsumer` will make it clear it's a thing that consumes two positions. Look at fastutil's consumer names. Those are `SomethingBiConsumer`, not `BiSomethingConsumer`.

mappings/net/minecraft/block/Block.mapping
```diff
@@ -324,7 +324,7 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		ARG 1 state
 		ARG 2 mirror
 	METHOD method_9570 hasBlockEntity ()Z
-	METHOD method_9571 (Lnet/minecraft/class_2680;Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
+	METHOD method_9571 getShape (Lnet/minecraft/class_2680;Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
```
Juuxel: How does this differ from the other shape getters?

Sollace: From the commit message:
>  getShape. This delegates to the block's outline shape and is used to determine whether it is a full cube, and to get the culled shape.

It's a generic "getShape" method whereas the others are more specialised forms of the same method. i.e. `getOutlineShape()` and `getHitboxShape`. Presumably they did those so they can have blocks without the latter two but still have a shape to use where it's needed.

Runemoro: This needs a better name. Look at the blocks that override it. Why does skull return empty and fence return `VoxelShapes.union(Block.createCuboidShape(7.0, 2.0, 0.0, 9.0, 13.0, 2.0), Block.createCuboidShape(7.0, 2.0, 14.0, 9.0, 13.0, 16.0));`? It's not just a "fallback shape".

Sollace: What would you call it instead then? I'll have to look more closely at the other methods to figure that out.

Sollace: The methods we have are:

```
    @Deprecated
    public VoxelShape getOutlineShape(BlockState state, BlockView view, BlockPos pos, EntityContext ePos) {
        return VoxelShapes.fullCube();
    }
    
    @Deprecated
    public VoxelShape getCollisionShape(BlockState state, BlockView view, BlockPos pos, EntityContext ePos) {
        return this.collidable ? state.getOutlineShape(view, pos) : VoxelShapes.empty();
    }
    
    @Deprecated
    public VoxelShape getShape(BlockState state, BlockView view, BlockPos pos) {
        return state.getOutlineShape(view, pos);
    }
    
    @Deprecated
    public VoxelShape getRayTraceShape(BlockState state, BlockView view, BlockPos pos) {
        return VoxelShapes.empty();
    }
```

The `getShape` method is overriden buy `FenceBlock`, `FenceGateBlock`, `LecternBlock`, and `SkullBlock`.

```
    // FenceBlock
    @Override
    public VoxelShape getShape(BlockState state, BlockView view, BlockPos pos) {
        return this.SHAPES[this.getShapeIndex(state)];
    }
    // FenceGateBlock
    @Override
    public VoxelShape getShape(BlockState state, BlockView view, BlockPos pos) {
        if (state.<Boolean>get((Property<Boolean>)FenceGateBlock.IN_WALL)) {
            return (state.<Direction>get((Property<Direction>)FenceGateBlock.FACING).getAxis() == Direction.Axis.X) ? FenceGateBlock.field_11027 : FenceGateBlock.field_11020;
        }
        return (state.<Direction>get((Property<Direction>)FenceGateBlock.FACING).getAxis() == Direction.Axis.X) ? FenceGateBlock.field_11023 : FenceGateBlock.field_11018;
    }
    // LecternBlock
    @Override
    public VoxelShape getShape(BlockState state, BlockView view, BlockPos pos) {
        return LecternBlock.BASE_SHAPE;
    }
    // SkullBlock
    @Override
    public VoxelShape getShape(BlockState state, BlockView view, BlockPos pos) {
        return VoxelShapes.empty();
    }
```

Sollace: @Runemoro @Juuxel After looking at how this method is being used, I decided to change the name to `getCullShape` (an overload of a similar method only present on `BlockState`)

This is based on what I found in `BlockState.ShapeCache` which is pre-computing the culled faces for each direction:
```
                this.culledFaces = new VoxelShape[BlockState.ShapeCache.DIRECTIONS.length];
                VoxelShape shape = block.getCullShape(state, EmptyBlockView.INSTANCE, BlockPos.ORIGIN);
                for (final Direction i : BlockState.ShapeCache.DIRECTIONS) {
                    this.culledFaces[i.ordinal()] = VoxelShapes.cull(shape, i);
                }
```

And also how these values, and the mapped method is being used here:

```
    public VoxelShape getCullShape(BlockView view, BlockPos pos, Direction facing) {
        if (this.shapeCache != null && this.shapeCache.culledFaces != null) {
            return this.shapeCache.culledFaces[facing.ordinal()];
        }
        return VoxelShapes.cull(this.getCullShape(view, pos), facing);
    }
```

Runemoro: Interesting... Are you sure`VoxelShapes.cull` is named right though? To cull means to remove hidden faces, so if that method actually culls the shape, `BlockState.getCullShape` should actually be `getCulledShape`...

As it is the code doesn't make sense. You're taking a cull shape, culling it, and obtaining another cull shape!?

Sollace: > You're taking a cull shape, culling it, and obtaining another cull shape!?

Lol. You're taking a cull shape, culling it, and obtaining a _culled_ shaped.

Anyone is welcome to check me on that. I'm reasonably certain it has _something_ to do with culling. The method either returns `fullcube()`, `empty()`, or a `SlicedVoxelShape` cut along the direction's axis.

It seems like the slice is always going to have the face closest to the "observer" for lack of a better word.

```
public static VoxelShape cull(VoxelShape shape, Direction direction) {
	if (shape == fullCube()) {
		return fullCube();
	}
	Direction.Axis axis = direction.getAxis();
	boolean hasVolume = false;
	int sliceWidth = 0;
	
	if (direction.getDirection() == Direction.AxisDirection.POSITIVE) {
		hasVolume = DoubleMath.fuzzyEquals(shape.getMaximum(axis), 1, 1.0E-7);
		sliceWidth = shape.voxels.getSize(axis) - 1;
	} else {
		hasVolume = DoubleMath.fuzzyEquals(shape.getMinimum(axis), 0, 1.0E-7);
		sliceWidth = 0;
	}
	if (!hasVolume) {
		return empty();
	}
	return new SlicedVoxelShape(shape, axis, sliceWidth);
}
```

Runemoro: > Lol. You're taking a cull shape, culling it, and obtaining a culled shaped.

Then why is the blockstate method *cull* and not *culled*?


But that doesn't matter anymore because `VoxelShapes.cull` is named wrong. That's not culling. The code you're showing takes a single face of the shape (its cross-section at the furthest point possible) and extrudes it to make it a prism with width 1. The method should be named something like `extrudeFace` or `prismFromFace`. Its purpose is to convert the general cull shape into a side-specific cull shape.

mappings/net/minecraft/block/BlockRenderLayer.mapping
```diff
@@ -1,5 +1,4 @@
 CLASS net/minecraft/class_1921 net/minecraft/block/BlockRenderLayer
-	FIELD field_9175 CUTOUT_MIPPED Lnet/minecraft/class_1921;
```
modmuss50: What was the reasoning for removing this mapping?

modmuss50 (approved):

mappings/net/minecraft/block/BlockRenderLayer.mapping
Sollace: This is an enum value. It was not obfuscated, and in fact the mapped name is identical to the auto-generated one we get from the code.

modmuss50:
Another review is needed please

mappings/net/minecraft/block/FenceBlock.mapping
```diff
@@ -8,4 +8,5 @@ CLASS net/minecraft/class_2354 net/minecraft/block/FenceBlock
 		ARG 3 pos
 	METHOD method_10184 canConnect (Lnet/minecraft/class_2680;ZLnet/minecraft/class_2350;)Z
 		ARG 1 state
+		ARG 2 neighbourIsFullSquare
```
Juuxel: `neighborIsFullSquare`?

mappings/net/minecraft/util/math/Direction.mapping
```diff
@@ -27,11 +30,19 @@ CLASS net/minecraft/class_2350 net/minecraft/util/math/Direction
...
+			ARG 3 offset
+			ARG 4 description
 		METHOD method_10181 offset ()I
 	CLASS class_2353 Type
 		FIELD field_11061 facingArray [Lnet/minecraft/class_2350;
 		FIELD field_11065 axisArray [Lnet/minecraft/class_2350$class_2351;
+		METHOD <init> (Ljava/lang/String;I[Lnet/minecraft/class_2350;[Lnet/minecraft/class_2350$class_2351;)V
+			ARG 3 directions
 		METHOD method_10183 random (Ljava/util/Random;)Lnet/minecraft/class_2350;
+			ARG 1 rand
```
Juuxel: `random`

Juuxel (changes requested):

mappings/net/minecraft/block/FenceBlock.mapping
Sollace: Muuurika

Juuxel (approved):

liach (approved):

liach:
@Sollace mind checking if these changes can be ported to 1.15 easily? The name refactors should be fine as these contents aren't really hot stuff used by mods.

Runemoro (approved):

Sollace:
@liach If it can, do you want me to port to 1.15 and change this PR to target that branch?

I have done nothing in 1.15 yet so I don't know what it's going to look like.

Sollace:
@liach I'm going to squash this before I rebase it. Same with the other PRs. Everyone onboard with that?

Sollace:
@liach @Runemoro @liach Rebased and base branch updated to target 19w44a.

mappings/net/minecraft/block/BlockState.mapping
```diff
@@ -178,7 +182,7 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 		ARG 1 view
 		ARG 2 pos
 		ARG 3 ePos
-	METHOD method_16384 getCullingShape (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;)Lnet/minecraft/class_265;
+	METHOD method_16384 getCullingFace (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;)Lnet/minecraft/class_265;
```
liach, line 185: Nice change, reflecting the fact that this takes a direction argument

liach (approved):

Runemoro:
@Sollace could you update to the latest snapshot and fix conflicts?

Sollace:
@Runemoro Done

mappings/net/minecraft/client/render/block/FluidRenderer.mapping
```diff
@@ -17,10 +17,10 @@ CLASS net/minecraft/class_775 net/minecraft/client/render/block/FluidRenderer
 		ARG 1 world
 		ARG 2 pos
 	METHOD method_3344 isSideCovered (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;F)Z
-		ARG 0 world
+		ARG 0 view
```
liach, line 20: Player suggests to keep these parameters world even though they may be of a XxxView type.

mappings/net/minecraft/block/BlockState.mapping
```diff
@@ -192,14 +196,18 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
...
 		ARG 2 state
 		ARG 3 hitResult
 		ARG 4 projectile
 	METHOD method_20827 isSideSolidFullSquare (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;)Z
+		ARG 1 view
+		ARG 2 pos
+		ARG 3 direction
 	METHOD method_21743 isFullCube (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Z
-		ARG 1 world
+		ARG 1 view
```
liach, line 210: player suggests `world`.

mappings/net/minecraft/block/BlockState.mapping
```diff
@@ -192,14 +196,18 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
...
 		ARG 1 view
 		ARG 2 pos
+	METHOD method_17900 exceedsCube ()Z
 	METHOD method_19287 onProjectileHit (Lnet/minecraft/class_1937;Lnet/minecraft/class_2680;Lnet/minecraft/class_3965;Lnet/minecraft/class_1297;)V
 		ARG 1 world
 		ARG 2 state
 		ARG 3 hitResult
 		ARG 4 projectile
 	METHOD method_20827 isSideSolidFullSquare (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;)Z
+		ARG 1 view
```
liach, line 206: player suggests `world`.

Player3324 (approved):
Thanks. It looks fine, further discussion/work around using world vs view for the world-like arg may be in order.

mappings/net/minecraft/block/BlockState.mapping
```diff
@@ -60,6 +62,7 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 		ARG 3 pos
 	METHOD method_11590 initShapeCache ()V
 	METHOD method_11591 canPlaceAt (Lnet/minecraft/class_4538;Lnet/minecraft/class_2338;)Z
+		ARG 1 view
```
liach, line 65: player suggests `world`.

mappings/net/minecraft/client/render/block/FluidRenderer.mapping
Sollace: Yeaaaah. I can accept that. I'll do it as another PR.

Sollace: @liach #949 is the number.

Issue #832, opened 2019-08-24
Bee class issues.


ghost:
The additions of bees to 19w34a has left some classes with now inaccurate names. Im unsure as to what they should now be named so I'm using this to document their function and as for a place for discussion.

Firstly, `class_1432` which is currently named `Bird` but now also used by BeeEntity and last I checked bees are not birds. Only used in `LivingEntity` to update Bees and Parrots `limbDistance` and `limbAngle` (Their limb refers to their wings in this case) It should be noted that Ghasts and more so Bats don't use this interface despite having similar features.
For code see near towards the bottom of LivingEntity's travel method (or search for Bird)

Secondly, `class_1331` which is currently named `ParrotMoveControl` but now also used by Bee.
This class's tick method has code similar to the base implementation in `MoveControl` however gives entities and upward motion aswell as no gravity when flying towards a target. When not moving towards a target bee specifically (through a boolean argument in the class constructor) have gravity reapplied to them.

ghost:
Also had some discussion on discord [here](https://discordapp.com/channels/507304429255393322/608088354042544139/614592911555952681).

Sollace:
One suggestion:

`ParrotMoveControl` -> `FlightMoveControl`

I'm not sure about `Bird`. My previous suggestion was something like `FlyingEntity` but I see it's really just an empty interface applied to birds and bees. Can anyone find some over cases where this interface is being used? Likely it's name isn't `Bird` but more closely to some trait Mojang wanted to differentiate parrots from other entities.

Prospector:
I'd go with "Flutters" for the interface

Pyrofab:
Flutters is a bit of a weird name for an interface in my opinion. Maybe `Fluttering` may be better ?

Sollace:
`CanFlutter`

This is why I prefer the I prefix for interfaces, because then the name can
be `ICanFlutter`

ghost:
In Bedrock `ParrotMoveControl` seems to be named `FlyMoveControl` but I like Sollace's suggestion of `FlightMoveControl`. Also `Flutters` seem like an appropiate name for the interface

YanisBft:
I'm okay with `FlightMoveControl` and `FlyingEntity`, but I think `WingMoveControl` / `WingedEntity` would be more accurate, as ghasts are flying but don't have wings

Sollace:
[...]

PR #840, opened 2019-08-26
Renames some classes now shared by parrots and bees

ghost:
In 19w34a Bees were added so some old names that are now used in bees no longer make sense.
I created an issue here: https://github.com/FabricMC/yarn/issues/832 to get some feedback for new names which I've incorperated into this pull request. The following classes have been renamed:
`Bird` -> `Flutters` (edit) -> `Flutterer`
`ParrotMoveControl` -> `FlightMoveControl`

### DIFF
--- a/mappings/net/minecraft/entity/Bird.mapping
+++ b/mappings/net/minecraft/entity/Bird.mapping
@@ -1,1 +1,0 @@
-CLASS net/minecraft/class_1432 net/minecraft/entity/Bird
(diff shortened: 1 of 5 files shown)


Runemoro:
Don't bats flutter but not implement it?

ghost:
Bats don't implement it, but It's quite difficult to name with other entities have the same visual function.
But also can bats actually flutter? Or is it flapping.

Pyrofab (approved):

modmuss50 (approved):

PR #846, opened 2019-09-01
Container -> Menu

### LINKED ISSUE #386: Container -> Menu

Juuxel:
Replaces `Container` with `Menu`, as seen in various registry and logging strings. Closes #386 and #889.


mappings/net/minecraft/client/gui/screen/ingame/AbstractMenuScreen.mapping
```diff
@@ -1,9 +1,9 @@
-CLASS net/minecraft/class_465 net/minecraft/client/gui/screen/ingame/AbstractContainerScreen
+CLASS net/minecraft/class_465 net/minecraft/client/gui/screen/ingame/AbstractMenuScreen
```
Prospector: Let's drop the Abstract here

Prospector (approved):

mappings/net/minecraft/client/gui/screen/ingame/AbstractMenuScreen.mapping
Juuxel: Done.

modmuss50:
Why `Menu` ? A few days ago we were talking about names such as `ScreenController`. `ScreenSate` `ScreenHandler`.

A menu to me seems something is closer to a UI and not something that partly handles the syncing of it?

Juuxel:
`Menu` is the name used by Mojang in the menu/container/controller registry. I think it's important to use Mojang names for clarity; in addition to the registry name and strings in methods, there are unobfuscated methods like `createMenu` that don't really make sense if the class is something else than `Menu`.

`Menu` is also a much better name than `Container`, as well as simple and short. Names like `ScreenController` are long and don't work that well in some interface names, like `NameableScreenControllerProvider`.

> A menu to me seems something is closer to a UI

Menus also handle some parts of the UI, namely slots.

Pyrofab:
I would take #351 into consideration before going with the Mojang name. I think `Menu` is an unintuitive name for a class that acts as intermediate between an Inventory (or another backend) and a Screen.
As was noted in that issue, the class is rather reminiscent of a Controller, using the MVC pattern as a reference. However, as you noted, `ScreenController` becomes quite long in compound names. My suggestion would be to use `ScreenController` for the base class, but only using `Controller` in other class names. That way, the purpose can always be deducted easily from the context without relying on painfully long names.
A meaningless Mojang name is certainly a step up from a meaningless stolen name, but I think we can do more helpful than that.

liach:
To me, it feels like the receiver/controller chips in a remote. Maybe we can consider the name from this viewpoint.

Prospector:
I think the arguments for Menu are pretty solid, not to mention the unobfuscated method `createMenu`... it's just blatantly confusing to name it anything else. We have MULTIPLE places in the code that reference it as a menu. Let's just make it simple and adopt the name. I think anything else would be rather silly.

stefan134:
Yes, I agree with @Prospector. On top of having references in unobfuscated names, there are also strings (exception messages, and even the ID of the registry!) using the word "menu". 

Using a different name would be the same as using two different names. A mix of a good name and a bad name is worse than a bad name.

Why aren't we also arguing about "entity". The word entity is so vague that anything could be an entity, yet we still use it despite having better alternatives.

Also, what about "block entity"? Block entities are in no way related to entities. Wouldn't "block data" be a better description of what it does?

Runemoro (approved):

liach (approved):

liach:
This pr has no problem as I see. Should we go to #386 for whether to rename or create a new ticket for a new vote?

Runemoro:
Looking at votes and comments on this PR, #386, #566, here's who's for and against `Menu`:

For `Menu` (14):
 - gegy1000
 - Prospector
 - Juuxel
 - Boundarybreaker
 - Runemoro
 - ChloeDawn
 - wlhlm
 - falkreon
 - natanfudge
 - Valoeghese
 - UnRealDinnerbone
 - kashike
 - geniiii
 - liach

Against `Menu` (12):
 - Chocohead
 - Hypatech
 - Pyrofab
 - VADemon
 - modmuss50
 - sfPlayer1
 - Siphalor
 - Sollace
 - TheGlitch76
 - AlexIIL
 - NinjaPhenix
 - LamdAurora

See also [this spreadsheet](https://docs.google.com/spreadsheets/d/1NJTXpKBGeSAhw9fBd6dGwCNwOasZ-82vC4q1eASvskg/edit).

Player3324:
https://en.wikipedia.org/wiki/Menu_(computing)

There is nothing menu-like about this class, even less so than in the Gui class hierarchy.

liach:
Please change base branch to 19w42a.

liach:
Also consider renaming all gui packets to menu packets.

liach:
Please change base branch to 19w44a.

Sollace:
> 
> 
> To me, it feels like the receiver/controller chips in a remote. Maybe we can consider the name from this viewpoint.

`RemoteInventory`, `RemoteScreen`?

Sollace:
`SpookyInteractionsAtADistance`

Sollace:
@Juuxel This history is getting pretty messy. Would you consider rebasing?

Juuxel:
I squashed the commits, should be cleaner now with all the merges gone.

Juuxel:
*Please switch the base branch to 1.15-pre4 and resolve the conflicts.*

liach:
Wait, juuz, are you talking to yourself

Juuxel:
Updated this to 1.15-pre6.

Sollace (commented):
Best commit.

Juuxel:
Oh, 542bf10? VSCode didn't pick them up as files with conflicts so I didn't notice them when I merged.

Juuxel:
Updated to 1.15-pre7.

liach:
will we keep this open? this pr is overwhelmingly downvoted

Juuxel:
There are also [Runemoro's statistics](https://github.com/FabricMC/yarn/pull/846#issuecomment-540032393), which are a bit more comprehensive than PR votes.

modmuss50:
I think the issue still stands, im not sure Menu is the correct solution to it though

Juuxel:
Well, I'd be happy to see someone else's `Container -> X` PR as well, but so far this has been the only one (in addition to #566).

Prospector:
Can we just get Menu in because it's the obvious choice so we can fix the horrible Inventory (which should be Container) mess because we can all unanimously agree Container is incorrect. And then we can open further discussions about it we want to rename Menu.

modmuss50:
Im not sure it is the `obvious choice`, 3/4 of people have thumbs down it on this issue.

It doesnt make a lot of sense, as a menu is somehting that a user interfacts with. `GuiState` seemed to be a popular option on the discord.

Prospector:
I say `obvious choice` because it is literally the name vanilla calls it so it is literally a correct name. IMO if anyone is unhappy with this and has a better name, they should open their own PR. As long as we change this name from Container I will be happy for now.

Juuxel:
Should I do something about the conflicting `ContainerProvider`s (inherited as two different `MenuProvider`s) in this PR? One is a more general superinterface of `NameableMenuProvider` (for blocks and BEs), the other a general interface for `MenuScreen`-like screens.

Edit: I renamed the block/BE one to `MenuFactory`.

liach:
I hereby support the name `Menu`. It indeed offers a list of item (see [definition of menu on dictionary.com](https://www.dictionary.com/browse/menu) `any list or set of items, activities, etc., from which to choose: `) as its main functionality.

Yet you should add javadocs to clarify the source of name as the "list of items" rather than the definition as the electronic/digital menu.

mappings/net/minecraft/menu/Menu.mapping
```diff
@@ -1,4 +1,17 @@
 CLASS net/minecraft/class_1703 net/minecraft/menu/Menu
+	COMMENT A menu is an object used for synchronizing GUIs between the server and the client.
```
liach: Not really synchronizing GUIs but more exactly, synchronizing screens that have item lists (menus).
I'd say:
`A menu is an item list, synchronized between the server and the client to allow updating the client to modify it.`

mappings/net/minecraft/menu/Menu.mapping
```diff
@@ -1,4 +1,17 @@
...
+	COMMENT
+	COMMENT <p>Menus are usually used for syncing the GUIs of container blocks,
+	COMMENT and they have a list of {@linkplain Slot slots} that store items.
+	COMMENT The slots are usually controlled by the server.
+	COMMENT
+	COMMENT <p>Menus also contain a list of {@linkplain net.minecraft.menu.property.Property properties}
+	COMMENT that are used for syncing integers (e.g. progress bars) from the server to the client.
+	COMMENT
+	COMMENT <p>The {@link #syncId} is shared between the two sides. It is used to verify that a player
+	COMMENT has a menu open so that they can move items, for example.
```
liach, line 12: The sync id is just an identifier for the menu so as to avoid desyncs, isn't it

mappings/net/minecraft/menu/Menu.mapping
```diff
@@ -1,4 +1,17 @@
...
+	COMMENT
+	COMMENT <p>On the client, menus are visualized by a {@link net.minecraft.client.gui.screen.ingame.MenuScreen}.
```
liach, line 14: maybe mention it has a reference to a client menu?

mappings/net/minecraft/menu/GenericMenu.mapping
```diff
@@ -1,8 +1,20 @@
 CLASS net/minecraft/class_1707 net/minecraft/menu/GenericMenu
+	COMMENT A generic 9xN menu with a customizable amount of rows.
```
liach: `A generic menu with 9 columns and a customizable amount of rows.`

mappings/net/minecraft/menu/GenericMenu.mapping
```diff
@@ -1,8 +1,20 @@
...
+	COMMENT
+	COMMENT <p>The {@code createGeneric9xN} methods should usually be used only on the client,
+	COMMENT as they have an empty inventory by default.
```
liach, line 5: Maybe mention that they can be updated by s2c packets and have c2s packets to response?

mappings/net/minecraft/menu/Menu.mapping
Juuxel: IIRC it's also used to check if a player has a menu open when there's a slot update packet etc.

mappings/net/minecraft/menu/Menu.mapping
Juuxel: Sure.

mappings/net/minecraft/menu/GenericMenu.mapping
Juuxel: I think that that would work better in the doc on `Menu`.

mappings/net/minecraft/network/listener/ServerPlayPacketListener.mapping
```diff
@@ -17,7 +17,7 @@ CLASS net/minecraft/class_2792 net/minecraft/network/listener/ServerPlayPacketLi
 		ARG 1 packet
 	METHOD method_12053 onBookUpdate (Lnet/minecraft/class_2820;)V
 		ARG 1 packet
-	METHOD method_12054 onGuiClose (Lnet/minecraft/class_2815;)V
+	METHOD method_12054 onCloseMenu (Lnet/minecraft/class_2815;)V
```
liach:
```suggestion
	METHOD method_12054 onMenuClose (Lnet/minecraft/class_2815;)V
```
imo the packet doesnt' need renaming but this method name is more congruent with other methods.

mappings/net/minecraft/menu/Menu.mapping
```diff
@@ -1,4 +1,19 @@
-CLASS net/minecraft/class_1703 net/minecraft/container/Container
+CLASS net/minecraft/class_1703 net/minecraft/menu/Menu
+	COMMENT A menu is a list of items and properties, synchronized between the server and the client.
```
liach, line 2: To address @sfPlayer1's concern at https://github.com/FabricMC/yarn/pull/846#issuecomment-540034825, we should point out that this is not a menu in computing sense at all.

mappings/net/minecraft/menu/GenericMenu.mapping
Juuxel: I think the comment I have on `Menu` already describes the packet behavior (well, maybe not c2s responses but that's an impl detail).

liach: worth nothing that server-only fab mods can send a simple generic menu to client and the client will have no trouble interacting with the menu

Juuxel:
This should be up to date for 20w10a now.

mappings/net/minecraft/entity/vehicle/StorageMinecartEntity.mapping
```diff
@@ -8,11 +8,7 @@ CLASS net/minecraft/class_1693 net/minecraft/entity/vehicle/StorageMinecartEntit
 		ARG 4 y
 		ARG 6 z
 		ARG 8 world
-	METHOD <init> (Lnet/minecraft/class_1299;Lnet/minecraft/class_1937;)V
-		ARG 1 type
-	METHOD createMenu (ILnet/minecraft/class_1661;Lnet/minecraft/class_1657;)Lnet/minecraft/class_1703;
-		ARG 1 syncId
-	METHOD method_17357 createMenu (ILnet/minecraft/class_1661;)Lnet/minecraft/class_1703;
+	METHOD method_17357 getMenu (ILnet/minecraft/class_1661;)Lnet/minecraft/class_1703;
```
liach: so `getMenu` or `createMenu`?

Juuxel: It should be `createMenu`, I think. I'm pretty sure there was a `getContainer` there when I was merging.

Earthcomputer:
I suppose this can be closed now in favour of #1106?

liach:
rip our good ol' pr

PR #849, opened 2019-09-05
Add some player interaction mappings

Earthcomputer:
This PR makes significant improvements to the mappings of `ClientPlayerInteractionManager` and `ServerPlayerInteractionManager`, as well as small related improvements to other classes.

Another thing to consider, but I refrained from, is renaming the `BlockPlayerActionS2CPacket` class to something like `PlayerActionResponseS2CPacket`, as it is currently misleading in that it might suggest that the player action was blocked, whereas it is actually sent whether the action was approved or not.

### DIFF
--- a/mappings/net/minecraft/item/ItemUsageContext.mapping
+++ b/mappings/net/minecraft/item/ItemUsageContext.mapping
@@ -23,4 +23,4 @@ CLASS net/minecraft/class_1838 net/minecraft/item/ItemUsageContext
 	METHOD method_8042 getPlayerFacing ()Lnet/minecraft/class_2350;
 	METHOD method_8044 getPlayerYaw ()F
 	METHOD method_8045 getWorld ()Lnet/minecraft/class_1937;
-	METHOD method_8046 isPlayerSneaking ()Z
+	METHOD method_8046 shouldCancelInteraction ()Z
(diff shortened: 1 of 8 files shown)


Pyrofab:
I think the packet should be renamed for clarity and consistency.

mappings/net/minecraft/item/ItemUsageContext.mapping
```diff
@@ -23,4 +23,4 @@ CLASS net/minecraft/class_1838 net/minecraft/item/ItemUsageContext
 	METHOD method_8042 getPlayerFacing ()Lnet/minecraft/class_2350;
 	METHOD method_8044 getPlayerYaw ()F
 	METHOD method_8045 getWorld ()Lnet/minecraft/class_1937;
-	METHOD method_8046 isPlayerSneaking ()Z
+	METHOD method_8046 shouldCancelInteraction ()Z
```
liach, line 26: :thinking: how the heck did this got wrong

liach (approved):

mappings/net/minecraft/item/ItemUsageContext.mapping
Earthcomputer: It is only true when the player is sneaking, however I decided to rename it based on its usage rather than its implementation. Its implementation even calls `player.shouldCancelInteration`, rather than `player.isSneaking`.

liach: Is there any minecraft wiki articles or such that states how sneaking is intended to cancel interactions

Earthcomputer: I don't know, but it does.

modmuss50 (approved):

modmuss50:
Thanks for this, merged in my IDE.

PR #854, opened 2019-09-11
Some stuff

Shnupbups:
Back to mapping random things.

PR #855, opened 2019-09-11
ThrowablePotionItem

dblsaiko:


dblsaiko:
Already fixed in #854.

PR #862, opened 2019-09-15
19w36a: The art of nether update

liach:
First of all, I am no pro at biomes, so I hope people like @valoeghese or @Runemoro can review this a bit.

Background: https://minecraft.gamepedia.com/Java_Edition_19w36a#General_2

Description:
The new biome system still has object-wrapped array system for biome storage. However, we observe two different method for getting the actual biome (one blockpos, one 3 int). After investigation, the 3-int version is a direct poll to the biome storage (so the 3 ints are not block positions but biome positions that are only exact to every 4 blocks). Hence, I named it `accessBiome` to distinguish from the `getBiome(BlockPos)` that users are supposed to use.

To bridge the gap between raw biome storage and block pos in the world, mojang has some sort of biome accessor class 4543 (I named it `BiomeAccessor`) that gets the biome from a block pos with a subclass interface 4544, a long seed, and another interface 4545. 

Changed a few field names in biome sources (some static final fields have incorrect casing, etc).

Also mapped a random getCullShape method (It refers to getOutlineShape by default but seems to be used for culling), a seed mixer (which just updates the seed for biome access strategy or biome sources), two obvious block state shape related method, and a few other utilities. 

4544 exposes the method to obtain a biome from the raw biome(x/y/z) storage and is implemented by 4538 (a new world interface), the new biome array, and biome sources. I believe it would qualify for a biome storage and hence named it `Storage`.

For 4545, I named it a `BiomeAccessStrategy`. It can be defined by dimension types. It accepts the world seed, block xyz, and 4544 (the storage) and returns a specific biome. The default impl (4546) has some fuzzy logic that maybe prevents the biome edges look too blocky (as vanilla biomes are stored in 4*4*4 units now). There is also an alternative impl (4547) that ignores the y position and delegates to the default impl, which is currently used by the overworld dimension. I named 4546 `DefaultBiomeAccessStrategy` and 4547 `HorizontalBiomeAccessStrategy` as 4547 only cares about the horizontal plane.

I am eager to hear from you investigators and modders for suggestions, issues, and feedbacks!

### DIFF
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -107,6 +107,7 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		ARG 3 side
 	METHOD method_22358 (Lnet/minecraft/class_2680;Lnet/minecraft/class_3611;)Z
 		ARG 1 state
+	METHOD method_22359 isFullyLit (Lnet/minecraft/class_2680;)Z
 	METHOD method_9496 randomDisplayTick (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Ljava/util/Random;)V
 		ARG 1 state
 		ARG 2 world
@@ -322,7 +323,7 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		ARG 1 state
 		ARG 2 mirror
 	METHOD method_9570 hasBlockEntity ()Z
-	METHOD method_9571 (Lnet/minecraft/class_2680;Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
+	METHOD method_9571 getCullShape (Lnet/minecraft/class_2680;Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
 		ARG 1 state
 		ARG 2 view
 		ARG 3 pos
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -323,7 +323,7 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		ARG 1 state
 		ARG 2 mirror
 	METHOD method_9570 hasBlockEntity ()Z
-	METHOD method_9571 getCullShape (Lnet/minecraft/class_2680;Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
+	METHOD method_9571 getCullingShape (Lnet/minecraft/class_2680;Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
 		ARG 1 state
 		ARG 2 view
 		ARG 3 pos
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -107,7 +107,7 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		ARG 3 side
 	METHOD method_22358 (Lnet/minecraft/class_2680;Lnet/minecraft/class_3611;)Z
 		ARG 1 state
-	METHOD method_22359 isFullyLit (Lnet/minecraft/class_2680;)Z
+	METHOD method_22359 rendersAtFullBrightness (Lnet/minecraft/class_2680;)Z
 	METHOD method_9496 randomDisplayTick (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Ljava/util/Random;)V
 		ARG 1 state
 		ARG 2 world
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -105,9 +105,10 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		ARG 1 world
 		ARG 2 pos
 		ARG 3 side
-	METHOD method_22358 (Lnet/minecraft/class_2680;Lnet/minecraft/class_3611;)Z
+	METHOD method_22358 canBucketPlace (Lnet/minecraft/class_2680;Lnet/minecraft/class_3611;)Z
 		ARG 1 state
-	METHOD method_22359 rendersAtFullBrightness (Lnet/minecraft/class_2680;)Z
+		ARG 2 fluid
+	METHOD method_22359 hasEmissiveLighting (Lnet/minecraft/class_2680;)Z
 	METHOD method_9496 randomDisplayTick (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Ljava/util/Random;)V
 		ARG 1 state
 		ARG 2 world
--- a/mappings/net/minecraft/block/BlockState.mapping
+++ b/mappings/net/minecraft/block/BlockState.mapping
@@ -180,7 +180,7 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 		ARG 1 view
 		ARG 2 pos
 		ARG 3 ePos
-	METHOD method_16384 getCullShape (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;)Lnet/minecraft/class_265;
+	METHOD method_16384 getCullingShape (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;)Lnet/minecraft/class_265;
 		ARG 1 view
 		ARG 2 pos
 		ARG 3 facing
--- a/mappings/net/minecraft/block/BlockState.mapping
+++ b/mappings/net/minecraft/block/BlockState.mapping
@@ -6,8 +6,10 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 		FIELD field_16558 opaque Z
 		FIELD field_16559 DIRECTIONS [Lnet/minecraft/class_2350;
 		FIELD field_16560 shapes [Lnet/minecraft/class_265;
+		FIELD field_17651 oversized Z
 		FIELD field_19360 collisionShape Lnet/minecraft/class_265;
 		FIELD field_19429 solidFullSquare [Z
+		FIELD field_20337 collisionShapeFullCube Z
 		METHOD <init> (Lnet/minecraft/class_2680;)V
 			ARG 1 state
 	FIELD field_12290 shapeCache Lnet/minecraft/class_2680$class_3752;
@@ -115,7 +117,7 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 		ARG 2 pos
 		ARG 3 entity
 	METHOD method_11614 getBlock ()Lnet/minecraft/class_2248;
-	METHOD method_11615 (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
+	METHOD method_11615 getCullShape (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
 		ARG 1 view
 		ARG 2 pos
 	METHOD method_11616 hasRandomTicks ()Z
@@ -192,9 +194,12 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 	METHOD method_17770 getOutlineShape (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
 		ARG 1 view
 		ARG 2 pos
+	METHOD method_17900 isShapeOversized ()Z
 	METHOD method_19287 onProjectileHit (Lnet/minecraft/class_1937;Lnet/minecraft/class_2680;Lnet/minecraft/class_3965;Lnet/minecraft/class_1297;)V
 		ARG 1 world
 		ARG 2 state
 		ARG 3 hitResult
 		ARG 4 projectile
 	METHOD method_20827 isSideSolidFullSquare (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;)Z
+	METHOD method_21743 isCollisionShapeFullCube (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Z
+	METHOD method_22361 isFullyLit ()Z
--- a/mappings/net/minecraft/block/BlockState.mapping
+++ b/mappings/net/minecraft/block/BlockState.mapping
@@ -202,4 +202,4 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 		ARG 4 projectile
 	METHOD method_20827 isSideSolidFullSquare (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;)Z
 	METHOD method_21743 isCollisionShapeFullCube (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Z
-	METHOD method_22361 isFullyLit ()Z
+	METHOD method_22361 rendersAtFullBrightness ()Z
--- a/mappings/net/minecraft/block/BlockState.mapping
+++ b/mappings/net/minecraft/block/BlockState.mapping
@@ -6,10 +6,8 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 		FIELD field_16558 opaque Z
 		FIELD field_16559 DIRECTIONS [Lnet/minecraft/class_2350;
 		FIELD field_16560 shapes [Lnet/minecraft/class_265;
-		FIELD field_17651 oversized Z
 		FIELD field_19360 collisionShape Lnet/minecraft/class_265;
 		FIELD field_19429 solidFullSquare [Z
-		FIELD field_20337 collisionShapeFullCube Z
 		METHOD <init> (Lnet/minecraft/class_2680;)V
 			ARG 1 state
 	FIELD field_12290 shapeCache Lnet/minecraft/class_2680$class_3752;
@@ -117,7 +115,7 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 		ARG 2 pos
 		ARG 3 entity
 	METHOD method_11614 getBlock ()Lnet/minecraft/class_2248;
-	METHOD method_11615 getCullShape (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
+	METHOD method_11615 getCullingShape (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
 		ARG 1 view
 		ARG 2 pos
 	METHOD method_11616 hasRandomTicks ()Z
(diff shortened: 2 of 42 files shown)


mappings/net/minecraft/world/VoxelView.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_4538
+CLASS net/minecraft/class_4538 net/minecraft/world/VoxelView
```
valoeghese: We don't usually refer to voxels in yarn do we? It makes sense, but sounds a bit off.

mappings/net/minecraft/world/biome/BiomeArray.mapping
```diff
@@ -0,0 +1,22 @@
+CLASS net/minecraft/class_4548 net/minecraft/world/biome/BiomeArray
```
valoeghese, line 1: I assume this is a 3d array of biomes in a chunk?

mappings/net/minecraft/world/biome/source/BiomeAccessStrategy.mapping
```diff
@@ -0,0 +1,6 @@
+CLASS net/minecraft/class_4545 net/minecraft/world/biome/source/BiomeAccessStrategy
```
valoeghese: What does this class do in code exactly?

mappings/net/minecraft/world/biome/source/BiomeAccessor.mapping
```diff
@@ -0,0 +1,16 @@
...
+		METHOD method_16359 accessBiome (III)Lnet/minecraft/class_1959;
+			ARG 1 biomeX
+			ARG 2 biomeY
+			ARG 3 biomeZ
+	FIELD field_20640 storage Lnet/minecraft/class_4543$class_4544;
+	FIELD field_20641 seed J
+	FIELD field_20642 strategy Lnet/minecraft/class_4545;
+	METHOD <init> (Lnet/minecraft/class_4543$class_4544;JLnet/minecraft/class_4545;)V
+		ARG 2 seed
+		ARG 4 strategy
```
valoeghese: What is this "strategy" thing?

mappings/net/minecraft/world/biome/BiomeArray.mapping
liach: yes. This is bound to each chunk. it reflects the 3 biome xyz inputs (not block pos) into a position in the 1d array.

mappings/net/minecraft/world/biome/source/BiomeAccessStrategy.mapping
liach: > For 4545, I named it a `BiomeAccessStrategy`. It can be defined by dimension types. It accepts the world seed, block xyz, and 4544 (the storage) and returns a specific biome. The default impl (4546) has some fuzzy logic that maybe prevents the biome edges look too blocky (as vanilla biomes are stored in 444 units now). There is also an alternative impl (4547) that ignores the y position and delegates to the default impl, which is currently used by the overworld dimension. I named 4546 `DefaultBiomeAccessStrategy` and 4547 `HorizontalBiomeAccessStrategy` as 4547 only cares about the horizontal plane.

It accepts the world seed, block xyz, and 4544 (the storage) and returns a specific biome.

ghost (approved):

valoeghese:
@liach ~~what~~ why is a strategy

valoeghese:
i.e. why did you name it "strategy" exactly?

liach:
The strategy accepts the world seed, block xyz, the biome array accessor (4544) and return a biome for the block xyz.

mappings/net/minecraft/world/biome/source/BiomeAccessor.mapping
valoeghese: ok

valoeghese (approved):
Seems good

mappings/net/minecraft/world/biome/source/BiomeAccessor.mapping
```diff
@@ -0,0 +1,16 @@
+CLASS net/minecraft/class_4543 net/minecraft/world/biome/source/BiomeAccessor
+	CLASS class_4544 Storage
+		METHOD method_16359 accessBiome (III)Lnet/minecraft/class_1959;
```
Gegy: Given this function is on the storage, I think it could just be `getBiome`?

mappings/net/minecraft/world/biome/source/DefaultBiomeAccessStrategy.mapping
```diff
@@ -0,0 +1,13 @@
+CLASS net/minecraft/class_4546 net/minecraft/world/biome/source/DefaultBiomeAccessStrategy
```
Gegy: This strategy seems to use the same scale logic as the former voronoi scale biome layer -- maybe it should be named more specifically around that?

mappings/net/minecraft/world/VoxelView.mapping
```diff
@@ -35,6 +38,10 @@ CLASS net/minecraft/class_4538
 		ARG 1 pos
 	METHOD method_22351 isWater (Lnet/minecraft/class_2338;)Z
 		ARG 1 pos
+	METHOD method_22387 accessGeneratorBiome (III)Lnet/minecraft/class_1959;
```
Gegy: hm, not sure about the `access` prefix here. maybe something like `getStoredBiome`?

mappings/net/minecraft/world/biome/source/BiomeAccessStrategy.mapping
Gegy: The class could potentially just be called `BiomeAccess`

mappings/net/minecraft/world/biome/source/BiomeAccessor.mapping
```diff
@@ -0,0 +1,16 @@
+CLASS net/minecraft/class_4543 net/minecraft/world/biome/source/BiomeAccessor
```
Gegy: Thoughts on `BiomeReader` as an alternative? (random thought, not sure about it)

liach: This is like reading from the biome array, not really from the world

mappings/net/minecraft/world/biome/source/BiomeAccessor.mapping
liach: No, otherwise people will confuse it as getting biome for block xyz because this method is in world as well (present together with the actual one that gets the biome for blockpos)

mappings/net/minecraft/world/biome/source/DefaultBiomeAccessStrategy.mapping
liach: `VoronoiBiomeScaleLogic` or what? I need some clue

mappings/net/minecraft/world/biome/source/BiomeAccessor.mapping
Gegy: Ah, I see. `getStoredBiome`?

mappings/net/minecraft/world/VoxelView.mapping
liach: this is like if the world doesn't have a chunk loaded, it will "access" (call that method in BiomeAccessor.Storage) the biome stored in generator

mappings/net/minecraft/world/biome/source/DefaultBiomeAccessStrategy.mapping
Gegy: didn't mention because not sure of a good name. maybe just like `VoronoiBiomeAccess`?

liach: Hmm, I think Voroni algorithm is still in `ScaleLayer` class. This is only applying random fuzz and isn't exclusive to world gen (when entities/game logic gets biome, the call goes through this pipeline as well)

mappings/net/minecraft/world/biome/source/BiomeAccessor.mapping
liach: updated

mappings/net/minecraft/world/biome/source/DefaultBiomeAccessStrategy.mapping
valoeghese: 1. The normal ScaleLayers don’t use voronoi tesselation.
2. Why did you spell “voronoi” as “voroni” lol

valoeghese: And voronoi tesselation scaling was applied for entities/game logic getting biomes in 1.14

liach: done.

mappings/net/minecraft/world/VoxelView.mapping
liach: What name should we use instead?

mappings/net/minecraft/block/Block.mapping
```diff
@@ -318,7 +319,7 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		ARG 1 state
 		ARG 2 mirror
 	METHOD method_9570 hasBlockEntity ()Z
-	METHOD method_9571 (Lnet/minecraft/class_2680;Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
+	METHOD method_9571 getCullShape (Lnet/minecraft/class_2680;Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
```
Juuxel: This is also mapped in #839 as `getCullingShape`.

mappings/net/minecraft/block/BlockState.mapping
```diff
@@ -115,7 +117,7 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 		ARG 2 pos
 		ARG 3 entity
 	METHOD method_11614 getBlock ()Lnet/minecraft/class_2248;
-	METHOD method_11615 (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
+	METHOD method_11615 getCullShape (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
```
Juuxel: This one as well.

mappings/net/minecraft/block/Block.mapping
liach: so is cull shape or culling shape better

liach: should no longer conflict with @Sollace 's pr.

liach:
I have updated the pr based on the comments and hope to get another round of review before Wednesday the snapshot day.

ghost (approved):

mappings/net/minecraft/world/biome/source/BiomeAccessStrategy.mapping
modmuss50: Yeah, I would drop the `Strategy`

mappings/net/minecraft/world/biome/source/HorizontalVoronoiBiomeAccessStrategy.mapping
```diff
@@ -0,0 +1 @@
+CLASS net/minecraft/class_4547 net/minecraft/world/biome/source/HorizontalVoronoiBiomeAccessStrategy
```
modmuss50: Again here drop the `Strategy`

mappings/net/minecraft/world/biome/source/VoronoiBiomeAccessStrategy.mapping
```diff
@@ -0,0 +1,13 @@
+CLASS net/minecraft/class_4546 net/minecraft/world/biome/source/VoronoiBiomeAccessStrategy
```
modmuss50: And here, will let you find anymore.

modmuss50 (changes requested):
Looks close, I would stay drop the `Strategy`. I can take a look with the code later if you disagree with me.

mappings/net/minecraft/world/biome/source/BiomeAccessStrategy.mapping
liach: There is already an accessor. Imo the two class names would sound too similar

mappings/net/minecraft/world/VoxelView.mapping
Gegy: I think should rather leave this unmapped for later.

mappings/net/minecraft/world/biome/source/BiomeAccessStrategy.mapping
liach: There is already an accessor. Imo the two class names would sound too similar

liach:
@modmuss50 Updated.

liach:
Given nether update is unleashed and this is the prereq of nether update, we need another round of review!

mappings/net/minecraft/block/Block.mapping
```diff
@@ -104,6 +104,7 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		ARG 3 side
 	METHOD method_22358 (Lnet/minecraft/class_2680;Lnet/minecraft/class_3611;)Z
 		ARG 1 state
+	METHOD method_22359 rendersAtFullBrightness (Lnet/minecraft/class_2680;)Z
```
Runemoro: Just curious, what is this used for?

mappings/net/minecraft/block/BlockState.mapping
```diff
@@ -6,8 +6,10 @@ CLASS net/minecraft/class_2680 net/minecraft/block/BlockState
 		FIELD field_16558 opaque Z
 		FIELD field_16559 DIRECTIONS [Lnet/minecraft/class_2350;
 		FIELD field_16560 shapes [Lnet/minecraft/class_265;
+		FIELD field_17651 oversized Z
```
Runemoro: What does this mean? `exceedsFullCube`?

mappings/net/minecraft/world/ChunkRegion.mapping
```diff
@@ -13,6 +13,7 @@ CLASS net/minecraft/class_3233 net/minecraft/world/ChunkRegion
 	FIELD field_14097 levelProperties Lnet/minecraft/class_31;
 	FIELD field_14098 chunks Ljava/util/List;
 	FIELD field_14099 blockTickScheduler Lnet/minecraft/class_1951;
+	FIELD field_20668 biomeAccessor Lnet/minecraft/class_4543;
```
Runemoro: `biomeAccess`? Or is this something else?

mappings/net/minecraft/world/biome/layer/CachingLayerContext.mapping
```diff
@@ -8,3 +8,6 @@ CLASS net/minecraft/class_3631 net/minecraft/world/biome/layer/CachingLayerConte
 		ARG 1 cacheCapacity
 		ARG 2 seed
 		ARG 4 salt
+	METHOD method_22417 mixSeed (JJ)J
```
Runemoro: `addSalt`?

Runemoro (approved):

mappings/net/minecraft/block/Block.mapping
liach: So magma blocks renders as if it is at lighting level of 15 (full skylight+blocklight) despite the surrounding light level. This is only true for magma blocks. 

This is like the block itself renders at lighting level 15 but has no effect on neighbor blocks.

mappings/net/minecraft/world/VoxelView.mapping
liach: Renamed to `WorldView` as this is like a reader interface (no modification hooks, really)

mappings/net/minecraft/world/biome/source/BiomeAccessStrategy.mapping
liach: Yeah, now there is `BiomeAccess` and `BiomeAccessType` instead

mappings/net/minecraft/block/Block.mapping
liach: Do you have a better suggestion for the name though, rune?

mappings/net/minecraft/block/BlockState.mapping
liach: yes

mappings/net/minecraft/world/ChunkRegion.mapping
liach: Nice spotting, fixing.

mappings/net/minecraft/block/Block.mapping
Runemoro: "emissive" is a common name for materials that aren't affected by light (https://www.google.com/search?q=emissive+material) though that might make someone think the block emits light, while it doesn't?

Maybe something like `isEmissiveMaterial` or `usesEmissiveRendering`?

Gegy (approved):

mappings/net/minecraft/block/BlockState.mapping
Runemoro: @liach If that's what it means, I think naming it `exceedsFullCube` would make it clearer. I think we also have a siminarly named method in the VoxelShapes class.

Runemoro: @liach If that's what it means, I think naming it `exceedsFullCube` would make it clearer. I think we also have a siminarly named method in the VoxelShapes class.

mappings/net/minecraft/world/biome/source/HorizontalVoronoiBiomeAccessStrategy.mapping
liach: Already fixed

mappings/net/minecraft/world/biome/source/VoronoiBiomeAccessStrategy.mapping
liach: Fixed long ago

mappings/net/minecraft/block/BlockState.mapping
Runemoro: @liach You still haven't renamed this to `exceedsFullCube`. Did you just forget, or do you think `oversized` is better?

```suggestion
		FIELD field_17651 exceedsFullCube Z
```

liach: I should revert it and leave it to #839 instead.

Runemoro: That PR is very outdated (1.14.4), it will have many conflicts anyway. It would be nice to just have the name as soon as possible.

mappings/net/minecraft/block/Block.mapping
```diff
@@ -105,8 +105,10 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		ARG 1 world
 		ARG 2 pos
 		ARG 3 side
-	METHOD method_22358 (Lnet/minecraft/class_2680;Lnet/minecraft/class_3611;)Z
+	METHOD method_22358 canBucketPlace (Lnet/minecraft/class_2680;Lnet/minecraft/class_3611;)Z
```
Juuxel, line 108: Is this related to `FluidFillable.canFillWithFluid`?

liach: No, this is only called by BucketItem when the bucket tries to place a fluid on the block. This is overridden by end portal and end portal frames to prevent bucket fluid placement.

Runemoro: Shouldn't this be `canPlaceBucket`? Or even better, `canEmptyBucket`

liach: Note the block state and fluid parameters, so I called it `canBucketPlace(Fluid)`

Juuxel (approved):

mappings/net/minecraft/block/Block.mapping
liach: Called it `hasEmissiveLighting`.

mappings/net/minecraft/block/BlockState.mapping
liach: Left this so that future digs in shape can better cover this.

PR #865, opened 2019-09-18
19w38b - loot conditions from data packs, advancements go smart, data…

liach:
… storage, etc

Also renamed property containers and state factory to state and state managers. when you have something that has configurable properties, you can reliably call it a state; and the state factory is not a factory; it just create all state instances and poll the one you need on demand, which is more like a manager.

Both of these names come from when asie was working on yarn on his own almost one year ago. During this period, no one has considered a better name for these classes and I figured that such shorter names may reduce confusion (especially when you have a StateFactory.Factory class, which is cursed)

The new "predicates" data type is named loot condition manager, as that is actually loot conditions instead of predicates shared by advancements and loot table.

Also consider burning the two classes in `net/minecraft/predicate/block` in a fire as they are in no way resembling the json predicate system in vanilla; they are used for block pattern matching (the block state predicate in that package) and world generation (the block one)


LemmaEOF (approved):
I think this is good to merge.

Gegy (approved):

mappings/net/minecraft/state/property/Property.mapping
```diff
@@ -1,8 +1,8 @@
 CLASS net/minecraft/class_2769 net/minecraft/state/property/Property
 	METHOD method_11898 getValues ()Ljava/util/Collection;
 	METHOD method_11899 getName ()Ljava/lang/String;
-	METHOD method_11900 getValue (Ljava/lang/String;)Ljava/util/Optional;
+	METHOD method_11900 parse (Ljava/lang/String;)Ljava/util/Optional;
 		ARG 1 name
-	METHOD method_11901 getName (Ljava/lang/Comparable;)Ljava/lang/String;
+	METHOD method_11901 name (Ljava/lang/Comparable;)Ljava/lang/String;
```
Juuxel, line 6: Why `name`? I think `getName` sounds better (for example `Properties.FACING.getName(Direction.UP)`).

liach: It is naming the value, hence just the verb `name`. It would be confusing with the method that gets the name for property itself, e.g. `Properties.FACING.getName()` may return `"facing"` while `Properties.FACING.getName(Direction.UP)` may return `"up"`

Juuxel: Oh, I didn't see that there was another `getName()`.

liach: Yeah, that is what we have now. 
image

mappings/net/minecraft/command/StorageDataObject.mapping
```diff
@@ -0,0 +1,9 @@
+CLASS net/minecraft/class_4580 net/minecraft/command/StorageDataObject
+	FIELD field_20855 TYPE_FACTORY Ljava/util/function/Function;
+	FIELD field_20856 SUGGESTOR Lcom/mojang/brigadier/suggestion/SuggestionProvider;
```
Juuxel: `SUGGESTION_PROVIDER` to match the class name?

mappings/net/minecraft/predicate/PlayerPredicate.mapping
```diff
@@ -0,0 +1,45 @@
...
+	CLASS class_4557 Builder
+		FIELD field_20730 experienceLevel Lnet/minecraft/class_2096$class_2100;
+		FIELD field_20731 gamemode Lnet/minecraft/class_1934;
+		FIELD field_20732 stats Ljava/util/Map;
+		FIELD field_20733 recipes Lit/unimi/dsi/fastutil/objects/Object2BooleanMap;
+		FIELD field_20734 advancements Ljava/util/Map;
+		METHOD method_22507 build ()Lnet/minecraft/class_4553;
+	FIELD field_20722 ANY Lnet/minecraft/class_4553;
+	FIELD field_20723 experienceLevel Lnet/minecraft/class_2096$class_2100;
+	FIELD field_20724 gamemode Lnet/minecraft/class_1934;
```
Juuxel, line 25: `gameMode` (this capitalization is used in the class name and fields in other classes)

mappings/net/minecraft/state/property/Property.mapping
Juuxel: In that case something like `getValueName` would sound better. It's not naming the value, it's getting the value's existing name.

Also applies to method_16551.

liach: Nah, if you have two properties of the same type, they can actually have different names for values as the name-value mapping is a part of the property.

mappings/net/minecraft/predicate/PlayerPredicate.mapping
liach: https://minecraft.gamepedia.com/Java_Edition_19w38a#General
It's called `gamemode` in JSON key, so `gamemode` is fine for field name

Gegy: Imo should have a standard still -- maybe can rename other fields to `gamemode` like this, or alternatively change this name?

mappings/net/minecraft/command/StorageDataObject.mapping
liach: Done

liach:
Asking for some approval and reviews before Wednesday the snapshot day.

modmuss50:
Conflict (im on mobile so cannot merge it) after that it looks good to merge.

PR #870, opened 2019-09-21
Map bindTexture() to public method

grondag:
In the latest snapshot the old bindTexture method was made private and is now wrapped by a public method that ensures thread safety.  Mods calling `bindTexture()` will fail due to an access violate.  This PR maps `bindTexture` to the new publicMethod and renames the private method.


liach (approved):
The new name for the thread-unsafe method sounds a bit weird, but should be fine.

ghost:
Surely this would require modders to rebuild their mods, however I do agree that the thread safe method should be called `bindTexture`

ghost (approved):

grondag:
> Surely this would require modders to rebuild their mods, however I do agree that the thread safe method should be called `bindTexture`

Mods will have to be rebuilt in any case because the old method is now private.  Doing it this way means they shouldn't have to change the code.

Gegy (approved):

PR #871, opened 2019-09-22
Randoms and datafixers

Juuxel:
- Moved the new datafixers that were mapped in #866 to the correct package. (`n.m.datafixer` -> `n.m.datafixers`)
  - There was already a `LeavesFix` in the correct package, so I renamed the new `LeavesFix` to `BiomeFormatFix`. I think the `"Leaves fix"` in the class is just a mistake from the actual `LeavesFix`. Mojang seems to have used it as a template (the `makeRule` methods have a similar structure).
- Added `on` to `Block.onRandomDisplayTick` and `BlockState.onScheduledTick`.
- Renamed `rnd` and `rand` in various places to `random` for consistency.

### DIFF
--- a/mappings/net/minecraft/block/BambooBlock.mapping
+++ b/mappings/net/minecraft/block/BambooBlock.mapping
@@ -11,7 +11,7 @@ CLASS net/minecraft/class_2211 net/minecraft/block/BambooBlock
 		ARG 1 state
 		ARG 2 world
 		ARG 3 pos
-		ARG 4 rand
+		ARG 4 random
 		ARG 5 height
 	METHOD method_9386 countBambooBelow (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)I
 		ARG 1 world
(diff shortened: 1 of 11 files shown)


liach:
I think `randomDisplayTick` is fine as `tick` is already an action in this phrase.

Moreover, I am confused on `datafixers` package name. All other packages we use are singular names `block` instead of `blocks`, `entity` instead of `entities`, but `datafixers` is a plural term. I suggest we adapt a universal pattern on this.

Juuxel:
I think we should have consistency between the block tick methods, and the others use `on` prefixes.

I think `net.minecraft.datafixers` is fine. It matches Mojang's own names in DFU, where the package is `com.mojang.datafixers`.

liach:
As far as I know, most mojang packages as seen in brigadier, dfu, etc. are plural. We should stick to singular names for packages (as that's recommended)

YanisBft (approved):

Juuxel:
Now that I think of it, dropping the `on` prefix from the other block tick methods could also work. (`scheduledTick()`, `randomDisplayTick()`, `randomTick()`)

liach:
Sure, made a vote at #872

Juuxel:
I'll make another PR for the block method renames and let this one be just small tweaks.

liach (approved):

modmuss50 (approved):

liach:
Note: This pr can close #872.

Juuxel:
> Note: This pr can close #872.

No? This PR didn't change any block tick method names (notice d13fdcc; making another PR for that right now).

PR #882, opened 2019-09-28
Entity properties

Juuxel:
- Changed `Entity.BREATH` and related mappings to `AIR`
  - `air` is used by Mojang in entity NBT and scoreboards
- Mapped some `ServerPlayerEntity` fields and tick/update methods
  - `playerTick()` comes from the crash handler in the method (`Ticking player`)
- Some world border mappings
- Moved `WaterCreatureEntity` to the mob package (it extends `MobEntityWithAi`)

### DIFF
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -78,7 +78,7 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	FIELD field_6029 scoreboardTags Ljava/util/Set;
 	FIELD field_6030 EMPTY_STACK_LIST Ljava/util/List;
 	FIELD field_6031 yaw F
-	FIELD field_6032 BREATH Lnet/minecraft/class_2940;
+	FIELD field_6032 AIR Lnet/minecraft/class_2940;
 	FIELD field_6033 inanimate Z
 	FIELD field_6034 vehicle Lnet/minecraft/class_1297;
 	FIELD field_6035 z D
@@ -260,7 +260,7 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	METHOD method_5666 attemptSprintingParticles ()V
 	METHOD method_5667 getUuid ()Ljava/util/UUID;
 	METHOD method_5668 getRootVehicle ()Lnet/minecraft/class_1297;
-	METHOD method_5669 getBreath ()I
+	METHOD method_5669 getAir ()I
 	METHOD method_5670 baseTick ()V
 	METHOD method_5671 getCommandSource ()Lnet/minecraft/class_2168;
 	METHOD method_5672 getHighSpeedSplashSound ()Lnet/minecraft/class_3414;
@@ -399,7 +399,7 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	METHOD method_5747 handleFallDamage (FF)V
 		ARG 1 fallDistance
 		ARG 2 damageMultiplier
-	METHOD method_5748 getMaxBreath ()I
+	METHOD method_5748 getMaxAir ()I
 	METHOD method_5749 readCustomDataFromTag (Lnet/minecraft/class_2487;)V
 		ARG 1 tag
 	METHOD method_5750 setVelocityClient (DDD)V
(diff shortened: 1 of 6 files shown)


mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -248,8 +248,8 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 		ARG 1 source
 	METHOD method_6062 isImmobile ()Z
 	METHOD method_6063 getMaximumHealth ()F
-	METHOD method_6064 getNextBreathInAir (I)I
-		ARG 1 breath
+	METHOD method_6064 updateAirOnLand (I)I
```
Pyrofab: This method has no side effect, I do not think `update` is a good fit. It simply computes the amount of air using the given current value.

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -355,8 +355,8 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_6128 isFallFlying ()Z
 	METHOD method_6129 onStatusEffectRemoved (Lnet/minecraft/class_1293;)V
 		ARG 1 effect
-	METHOD method_6130 getNextBreathInWater (I)I
-		ARG 1 breath
+	METHOD method_6130 updateAirUnderwater (I)I
```
Pyrofab: Same here, `update` is misleading.

mappings/net/minecraft/entity/WaterCreatureEntity.mapping
```diff
@@ -1,5 +1,5 @@
 CLASS net/minecraft/class_1480 net/minecraft/entity/WaterCreatureEntity
 	METHOD <init> (Lnet/minecraft/class_1299;Lnet/minecraft/class_1937;)V
 		ARG 1 type
-	METHOD method_6673 tickBreath (I)V
-		ARG 1 breath
+	METHOD method_6673 tickAir (I)V
```
Pyrofab: I think this name can be improved. This method is specifically for fish and creatures that cannot breathe outside water, and is made no-op in dolphins.

mappings/net/minecraft/server/network/ServerPlayerEntity.mapping
```diff
@@ -1,12 +1,21 @@
 CLASS net/minecraft/class_3222 net/minecraft/server/network/ServerPlayerEntity
+	FIELD field_13963 healthScore F
+	FIELD field_13965 levelScore I
```
Pyrofab: Maybe `last<X>Score` ? Considering it is used specifically to check whether the scoreboard needs updating.

Juuxel: Sure, that'll work.

mappings/net/minecraft/entity/WaterCreatureEntity.mapping
Juuxel: I renamed it to `tickWaterBreathingAir`, but I'm not sure if that's the best mapping for the method.

modmuss50 (approved):

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -78,7 +78,7 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	FIELD field_6029 scoreboardTags Ljava/util/Set;
 	FIELD field_6030 EMPTY_STACK_LIST Ljava/util/List;
 	FIELD field_6031 yaw F
-	FIELD field_6032 BREATH Lnet/minecraft/class_2940;
```
liach, line 81: Given this applies to water mobs, breath may actually be more accurate

liach (changes requested):
After looking at water creature, I oppose your `breath` -> `air` name change as `breath` turns out to be a more accurate description.

mappings/net/minecraft/entity/Entity.mapping
Juuxel: Breath isn't a good name for a thing that's serialized to `Air`.

liach: Mojang cannot change their serialized names easily due to backward compatibility, so they might have left it as-is in nbt but changed it in their actual code.

Juuxel: > Mojang cannot change their serialized names easily due to backward compatibility

They can, though? They can make a datafixer that targets entities.

liach:
Just need to decide if we want to stick to `BREATH` or `AIR`. The rest is very good.

Runemoro (approved):

Runemoro:
I agree with breath -> air. It's the word I commonly use when playing Minecraft, and it makes more sense (a breath is an action, air is a quantity).

liach (approved):
After some discussion, let's just stay with "AIR" as "suffocate" means the stuck-in-block damage as well.

PR #885, opened 2019-09-28
isOpaque backing field/block settings

Fuyukai:
isOpaque just returns field_20720 so it seems rather obvious. BlockState method is called "nonOpaque" instead of "transparent" because stuff like beds use it.

### DIFF
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -22,11 +22,13 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		FIELD field_10668 material Lnet/minecraft/class_3614;
 		FIELD field_10669 hardness F
 		FIELD field_10670 dynamicBounds Z
+		FIELD field_20721 opaque Z
 		METHOD <init> (Lnet/minecraft/class_3614;Lnet/minecraft/class_3620;)V
 			ARG 2 materialColor
 		METHOD method_16228 dropsLike (Lnet/minecraft/class_2248;)Lnet/minecraft/class_2248$class_2251;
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/block/Block.mapping
```diff
@@ -22,11 +22,13 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		FIELD field_10668 material Lnet/minecraft/class_3614;
 		FIELD field_10669 hardness F
 		FIELD field_10670 dynamicBounds Z
+		FIELD field_20721 opaque Z
 		METHOD <init> (Lnet/minecraft/class_3614;Lnet/minecraft/class_3620;)V
 			ARG 2 materialColor
 		METHOD method_16228 dropsLike (Lnet/minecraft/class_2248;)Lnet/minecraft/class_2248$class_2251;
 			ARG 1 source
 		METHOD method_16229 dropsNothing ()Lnet/minecraft/class_2248$class_2251;
+		METHOD method_22488 nonOpaque ()Lnet/minecraft/class_2248$class_2251;
```
liach, line 31: Just curious, can you describe the usages of this method?

Juuxel (approved):

mappings/net/minecraft/block/Block.mapping
liach: Ah, this is used in blocks for initializing blocks. great

liach (approved):

liach:
Imo the mappings are right and the pr is ready.

YanisBft (approved):

modmuss50 (approved):

PR #886, opened 2019-09-29
Drop 'on' prefixes from block tick methods, Block.activate -> onUse

### LINKED ISSUE #872: Block tick method names
Here is a vote on block's tick methods' names on whether to have an `on` prefix.

`onScheduledTick()`, `onRandomDisplayTick()`, `onRandomTick()` :+1:
`scheduledTick()`, `randomDisplayTick()`, `randomTick()` :-1:

The `on` prefix exists for most methods, but `randomDisplayTick` exists previously as well in `Block` mappings. See #871.

<details>
<summary>
My personal opinion
</summary>

I personally support no `on` prefix, which is the :-1: option. When we sum up, count extra 1 to the :-1:
</details>

Juuxel:
Closes #872.

- Dropped `on` prefixes from block tick methods, see #872.
- Renamed `activate` to `use` to be consistent with the item method and the keybind name.

### DIFF
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -207,7 +207,7 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		ARG 0 stateId
 	METHOD method_9533 shouldDropItemsOnExplosion (Lnet/minecraft/class_1927;)Z
 		ARG 1 explosion
-	METHOD method_9534 use (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_1657;Lnet/minecraft/class_1268;Lnet/minecraft/class_3965;)Z
+	METHOD method_9534 onUse (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_1657;Lnet/minecraft/class_1268;Lnet/minecraft/class_3965;)Z
 		ARG 1 state
 		ARG 2 world
 		ARG 3 pos
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -126,7 +126,7 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		ARG 2 entity
 	METHOD method_9503 getBlockFromItem (Lnet/minecraft/class_1792;)Lnet/minecraft/class_2248;
 		ARG 0 item
-	METHOD method_9504 onRainTick (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;)V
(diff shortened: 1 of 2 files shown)


liach (approved):

YanisBft (approved):

natanfudge:
`onBlockRemoved`
`onBlockDestroyedByExplosion`
`onProjectileHit`
`onEntityLand`
`onEntityCollision`
`onLandedUpon`
`onStacksDropped`
`onPlaced`
`onBreak`
`onBroken`
`onSteppedOn`
`onBlockAction`
`onBlockBreakStart`
`onBlockAdded`
`use`???
It should be `onUse`.

mappings/net/minecraft/block/Block.mapping
```diff
@@ -207,7 +207,7 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		ARG 0 stateId
 	METHOD method_9533 shouldDropItemsOnExplosion (Lnet/minecraft/class_1927;)Z
 		ARG 1 explosion
-	METHOD method_9534 activate (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_1657;Lnet/minecraft/class_1268;Lnet/minecraft/class_3965;)Z
+	METHOD method_9534 use (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_1657;Lnet/minecraft/class_1268;Lnet/minecraft/class_3965;)Z
```
modmuss50: Whats the reasoning behind this? Its not wrong, but activate makes a bit more sense as use is a bit more of a generic term imo?

liach: it is when player press use key on the block

Juuxel: `use` is more consistent with the keybinding and the item methods. You're also not "activating" a chest, you're *using* it. Activating to me sounds more like activating a piston or a dispenser.

Juuxel:
>  It should be `onUse`.

I can change it to `onUse` as well.

liach:
Compared to `onUse`, I would prefer `used` (like `keyPressed`). Thoughts?

Juuxel:
Names like `used` or `keyPressed` aren't really used in `Block`, though. I don't think it would be good to change all block event method names to that format either.

natanfudge:
`onThingDone` is an event
`doThing` is a function you call
`thingDone` is a boolean indicating if thing had been done or not.
Likewise
`onUsed` would be an event
`use` would be a function you are allowed to call
`used` is a boolean indicating if the block was used before.
I therefore suggest `onUsed` or `onUse`.

liach:
I suggest `onUse` because `use` itself is a noun already.

Juuxel:
Okay, it's now `onUse`.

modmuss50 (approved):

PR #891, opened 2019-10-02
Mappings for PersistentStateManager

jmendonza:
Mappings for PersistentStateManager

### DIFF
--- a/mappings/net/minecraft/world/PersistentStateManager.mapping
+++ b/mappings/net/minecraft/world/PersistentStateManager.mapping
@@ -8,10 +8,20 @@ CLASS net/minecraft/class_26 net/minecraft/world/PersistentStateManager
 		ARG 2 dataFixer
 	METHOD method_120 get (Ljava/util/function/Supplier;Ljava/lang/String;)Lnet/minecraft/class_18;
 		ARG 1 factory
-		ARG 2 id
+		ARG 2 fileName
 	METHOD method_123 set (Lnet/minecraft/class_18;)V
+		ARG 1 persistentState
 	METHOD method_125 save ()V
+	METHOD method_17921 isCompressedNbt (Ljava/io/PushbackInputStream;)Z
+		ARG 1 pushbackInputStream
 	METHOD method_17922 getFile (Ljava/lang/String;)Ljava/io/File;
 		ARG 1 id
+	METHOD method_17923 read (Ljava/lang/String;I)Lnet/minecraft/class_2487;
+		ARG 1 fileName
+		ARG 2 worldVersion
 	METHOD method_17924 getOrCreate (Ljava/util/function/Supplier;Ljava/lang/String;)Lnet/minecraft/class_18;
 		ARG 1 factory
+		ARG 2 fileName
+	METHOD method_20786 putIfAbsent (Ljava/util/function/Supplier;Ljava/lang/String;)Lnet/minecraft/class_18;
+		ARG 1 factory
+		ARG 2 fileName
--- a/mappings/net/minecraft/world/PersistentStateManager.mapping
+++ b/mappings/net/minecraft/world/PersistentStateManager.mapping
@@ -6,7 +6,7 @@ CLASS net/minecraft/class_26 net/minecraft/world/PersistentStateManager
 	METHOD <init> (Ljava/io/File;Lcom/mojang/datafixers/DataFixer;)V
 		ARG 1 directory
 		ARG 2 dataFixer
-	METHOD method_120 get (Ljava/util/function/Supplier;Ljava/lang/String;)Lnet/minecraft/class_18;
+	METHOD method_120 load (Ljava/util/function/Supplier;Ljava/lang/String;)Lnet/minecraft/class_18;
 		ARG 1 factory
 		ARG 2 fileName
 	METHOD method_123 set (Lnet/minecraft/class_18;)V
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/world/PersistentStateManager.mapping
```diff
@@ -8,10 +8,20 @@ CLASS net/minecraft/class_26 net/minecraft/world/PersistentStateManager
 		ARG 2 dataFixer
 	METHOD method_120 get (Ljava/util/function/Supplier;Ljava/lang/String;)Lnet/minecraft/class_18;
 		ARG 1 factory
-		ARG 2 id
+		ARG 2 fileName
 	METHOD method_123 set (Lnet/minecraft/class_18;)V
+		ARG 1 persistentState
 	METHOD method_125 save ()V
+	METHOD method_17921 isCompressedNbt (Ljava/io/PushbackInputStream;)Z
+		ARG 1 pushbackInputStream
```
Pyrofab, line 16: This is not very descriptive. Maybe `stateData` ?

mappings/net/minecraft/world/PersistentStateManager.mapping
```diff
@@ -8,10 +8,20 @@ CLASS net/minecraft/class_26 net/minecraft/world/PersistentStateManager
 		ARG 2 dataFixer
 	METHOD method_120 get (Ljava/util/function/Supplier;Ljava/lang/String;)Lnet/minecraft/class_18;
```
Pyrofab: This method is specifically about the loading process, and it logs `"Error loading saved data"`. I would call it `load` or `loadSavedState`.

mappings/net/minecraft/world/PersistentStateManager.mapping
```diff
@@ -8,10 +8,20 @@ CLASS net/minecraft/class_26 net/minecraft/world/PersistentStateManager
...
 	METHOD method_17922 getFile (Ljava/lang/String;)Ljava/io/File;
-		ARG 1 id
+		ARG 1 fileName
+	METHOD method_17923 read (Ljava/lang/String;I)Lnet/minecraft/class_2487;
+		ARG 1 fileName
+		ARG 2 worldVersion
 	METHOD method_17924 getOrCreate (Ljava/util/function/Supplier;Ljava/lang/String;)Lnet/minecraft/class_18;
 		ARG 1 factory
+		ARG 2 fileName
+	METHOD method_20786 putIfAbsent (Ljava/util/function/Supplier;Ljava/lang/String;)Lnet/minecraft/class_18;
```
Pyrofab: This method attempts to retrieve the existing state from memory, and attempts to load it from disk if it fails. I would suggest `get`, given it is effectively the first part of the `getOrCreate` method.

jmendonza: ~~there is a method already named get,~~ 
and yes it does what you said + it places the state retrieved from the disk on the loadedStates map, if it can't get it from the disk, it places null instead

mappings/net/minecraft/world/PersistentStateManager.mapping
jmendonza: what do you mean?

mappings/net/minecraft/world/PersistentStateManager.mapping
Pyrofab: > there is a method already named get

Hence renaming the aforementioned `get` method. It works exactly like the Map contract: `getOrCreate` creates an object from scratch if `get` cannot obtain an object through normal means. The fact that the object is cached is an implementation detail IMO.

mappings/net/minecraft/world/PersistentStateManager.mapping
Pyrofab: The argument name, `pushbackInputStream`, is just the name of the class so it does not add any information.

mappings/net/minecraft/world/PersistentStateManager.mapping
jmendonza: https://github.com/FabricMC/yarn/pull/891/commits/e84a8b130c3161022e8941e12ce77883b50cc3bc

mappings/net/minecraft/world/PersistentStateManager.mapping
jmendonza: https://github.com/FabricMC/yarn/pull/891/commits/e84a8b130c3161022e8941e12ce77883b50cc3bc

mappings/net/minecraft/world/PersistentStateManager.mapping
```diff
@@ -6,12 +6,22 @@ CLASS net/minecraft/class_26 net/minecraft/world/PersistentStateManager
 	METHOD <init> (Ljava/io/File;Lcom/mojang/datafixers/DataFixer;)V
 		ARG 1 directory
 		ARG 2 dataFixer
-	METHOD method_120 get (Ljava/util/function/Supplier;Ljava/lang/String;)Lnet/minecraft/class_18;
+	METHOD method_120 load (Ljava/util/function/Supplier;Ljava/lang/String;)Lnet/minecraft/class_18;
 		ARG 1 factory
-		ARG 2 id
+		ARG 2 fileName
```
Runemoro, line 11: Does this include the file extension? If not, then I'd leave it "id", or "name".

mappings/net/minecraft/world/PersistentStateManager.mapping
Runemoro: I'd just call this `in` (or `stream`, or `inoutStream`). There's no need for the parameter name to specify which type of stream it is...

liach: This push back input stream, as far as I know, is used to detect if the input dat file is in gzip compressed nbt format or plain nbt format, and the marker get reset to the file beginning later

Runemoro:
Since we're not merging PRs to 1.14.4 anymore, and `PersistentStateManager` has already been fully mapped in the latest snapshot, this PR can be closed.

PR #893, opened 2019-10-03
isMinceraft

shedaniel:
From #mod-dev in fabricord

Thanks ramidzkh
https://discordapp.com/channels/507304429255393322/507982478276034570/629231365732433930

### DIFF
--- a/mappings/net/minecraft/client/gui/screen/TitleScreen.mapping
+++ b/mappings/net/minecraft/client/gui/screen/TitleScreen.mapping
@@ -2,6 +2,7 @@ CLASS net/minecraft/class_442 net/minecraft/client/gui/screen/TitleScreen
 	FIELD field_17772 backgroundFadeStart J
 	FIELD field_17774 PANORAMA_CUBE_MAP Lnet/minecraft/class_751;
 	FIELD field_17775 PANORAMA_OVERLAY Lnet/minecraft/class_2960;
+	FIELD field_17776 isMinceraft Z
 	FIELD field_18222 doBackgroundFade Z
 	FIELD field_19102 ACCESSIBILITY_ICON_TEXTURE Lnet/minecraft/class_2960;
 	FIELD field_2583 MINECRAFT_TITLE_TEXTURE Lnet/minecraft/class_2960;


natanfudge:
isMinceraft

natanfudge (approved):

ramidzkh (approved):
`isMinceraft` is probably the most appropriate name.

modmuss50 (approved):

Runemoro:
There's a spelling mistake in the name... Also, what does this do...?

Pyrofab:
Can you really not guess ?

modmuss50:
https://minecraft.gamepedia.com/Menu_screen#Minceraft

liach:
This should have been `minceraft` given this is a field instead of a boolean getter method

natanfudge:
Is that the convention? I would use isX for fields still.

liach:
In general, use adjective (phrases) for boolean fields, noun (phrases) for other fields, and verb (phrases) for methods.

```java
public void setFat(boolean fat) {
  this.fat = fat;
}
```
looks way better than 
```java
public void setIsFat(boolean isFat) {
  this.isFat = isFat;
}
```

natanfudge:
I would call it `setFat` even if the field is named `isFat`. The setter method should really only be in relation to the getter method because the field is an implementation detail.

liach:
`this.isFat = false` just doesn't make sense
and it doesn't make sense either when you have field name same as the method name
Why bother having `is` prefix when you can get rid of it?

natanfudge:
> Why bother having is prefix when you can get rid of it?

`isFat` is distinguishable as a boolean, while `fat` could easily be an int (20 mg of fat).

liach:
For example, when something like spring/gradle generates code, they do such:
E.g. you have
```
boolean isFat();
```
in the code gradle buildscript use
and you can call `sth.fat` to effectively call `isFat()` method in your gradle buildscript

This is the convention. This is effective. 

Your argument may apply for `fat` as it has a noun meaning, but not so much for `beautiful` which is exclusively an adjective.

shedaniel:
After a quick look at CONVENTIONS.md (Or whatever it is called I am on mobile) I guess I am happy to change it, but naming it `minceraft` only also doesn’t make sense, as it is still not an adjective (phrase).

liach:
`easterEggActive`?

shedaniel:
I wouldn’t know what that is until someone tell me though

natanfudge:
> For example, when something like spring/gradle generates code, they do such:
E.g. you have
boolean isFat();
in the code gradle buildscript use
and you can call sth.fat to effectively call isFat() method in your gradle buildscript


Funny you should mention that, Kotlin does it my way.
https://kotlinlang.org/docs/reference/java-interop.html#getters-and-setters

As Daniel has said mincecraft is a noun.

liach:
`minceraftShown`

PR #906, opened 2019-10-10
StatusEffect mapping and Cuboid Method rename

biom4st3r:
Mappings for statusEffects
Changed Cuboid.copyRotation -> Cuboid.copyRotationAndPoint
copyRotation doesn't just set yaw, pitch, and roll; it also copies over the rotation origin point

### DIFF
--- a/mappings/net/minecraft/client/model/ModelPart.mapping
+++ b/mappings/net/minecraft/client/model/ModelPart.mapping
@@ -77,7 +77,7 @@ CLASS net/minecraft/class_630 net/minecraft/client/model/ModelPart
 		ARG 1 model
 		ARG 2 textureOffsetU
 		ARG 3 textureOffsetV
-	METHOD method_17138 copyRotationAndPoint (Lnet/minecraft/class_630;)V
+	METHOD method_17138 copyPositionAndRotation (Lnet/minecraft/class_630;)V
 	METHOD method_22698 render (Lnet/minecraft/class_4587;Lnet/minecraft/class_4588;FIILnet/minecraft/class_1058;)V
 		ARG 1 matrix
 		ARG 2 vertexConsumer
--- a/mappings/net/minecraft/client/model/ModelPart.mapping
+++ b/mappings/net/minecraft/client/model/ModelPart.mapping
@@ -77,7 +77,7 @@ CLASS net/minecraft/class_630 net/minecraft/client/model/ModelPart
 		ARG 1 model
 		ARG 2 textureOffsetU
 		ARG 3 textureOffsetV
-	METHOD method_17138 copyRotation (Lnet/minecraft/class_630;)V
+	METHOD method_17138 copyRotationAndPoint (Lnet/minecraft/class_630;)V
 	METHOD method_22698 render (Lnet/minecraft/class_4587;Lnet/minecraft/class_4588;FIILnet/minecraft/class_1058;)V
 		ARG 1 matrix
 		ARG 2 vertexConsumer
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/client/model/ModelPart.mapping
```diff
@@ -77,7 +77,7 @@ CLASS net/minecraft/class_630 net/minecraft/client/model/ModelPart
 		ARG 1 model
 		ARG 2 textureOffsetU
 		ARG 3 textureOffsetV
-	METHOD method_17138 copyRotation (Lnet/minecraft/class_630;)V
+	METHOD method_17138 copyRotationAndPoint (Lnet/minecraft/class_630;)V
```
Runemoro: The rotation point is also the position of the model part.

Runemoro:
```suggestion
	METHOD method_17138 copyPositionAndRotation (Lnet/minecraft/class_630;)V
```

mappings/net/minecraft/entity/effect/StatusEffect.mapping
```diff
@@ -10,6 +10,8 @@ CLASS net/minecraft/class_1291 net/minecraft/entity/effect/StatusEffect
 		ARG 0 type
 	METHOD method_5555 onApplied (Lnet/minecraft/class_1309;Lnet/minecraft/class_1325;I)V
 	METHOD method_5556 getColor ()I
+	METHOD method_5559 getOrCreateTranslationKey ()Ljava/lang/String;
```
liach: A translation key isn't "created". Even though it is lazily computed, we can still just call it 
```suggestion
	METHOD method_5559 getTranslationKey ()Ljava/lang/String;
```

mappings/net/minecraft/entity/effect/StatusEffect.mapping
```diff
@@ -10,6 +10,8 @@ CLASS net/minecraft/class_1291 net/minecraft/entity/effect/StatusEffect
...
+	METHOD method_5560 getName ()Lnet/minecraft/class_2561;
```
Pyrofab, line 14: What's the name here, the localized translation key ?

liach: Should be a translateable text of the localization key.

liach (approved):

mappings/net/minecraft/entity/effect/StatusEffect.mapping
```diff
@@ -10,6 +10,8 @@ CLASS net/minecraft/class_1291 net/minecraft/entity/effect/StatusEffect
 		ARG 0 type
 	METHOD method_5555 onApplied (Lnet/minecraft/class_1309;Lnet/minecraft/class_1325;I)V
 	METHOD method_5556 getColor ()I
+	METHOD method_5559 getTranslationKey ()Ljava/lang/String;
```
liach: wait my bad... this is not a translation key

liach:
```suggestion
	METHOD method_5559 loadTranslationKey ()Ljava/lang/String;
```
Dunno why but mojang just decided to create this method and have another public method that actually offers the translation key which calls this method..

biom4st3r: Typically don't we use getXThing for that type of method?

liach: There is already a `getTranslationKey` method with same signature (method_5567)
and that 5567 just calls this method

liach (approved):
Sorry for troubling you!

biom4st3r:
oops....how do i undo those. I didn't realize pushing to my fork would edit this.

liach:
git reset --hard e8dd2d

biom4st3r:
Sorry about that. I haven't used git for anything other than personal projects

liach:
Please change base branch to 19w42a.

Runemoro (approved):

PR #912, opened 2019-10-12
Fix dust particle effect scale misnamed as alpha

JamiesWhiteShirt:
As @ChloeDawn so helpfully pointed out, this field does not contain the alpha, but rather the scale.


ChloeDawn (approved):

ChloeDawn:
The message is dripping with sarcasm, and I can confirm that I am deserving of it, but also that this change is correct.

liach (approved):

SPGoding (approved):

PR #914, opened 2019-10-16
Produce TinyV2 mappings

natanfudge:
This PR requires the following PRs in order:
- FabricMC/tiny-mappings-parser#3
- FabricMC/Enigma#164
- FabricMC/stitch#15
- FabricMC/intermediary#10


build.gradle
```diff
@@ -23,7 +23,7 @@ plugins {
 	id 'maven-publish'
 }
 
-def minecraft_version = "19w41a"
+def minecraft_version = "1.14.4"
```
liach:
```suggestion
def minecraft_version = "19w41a"
```

liach:
@natanfudge consider updating the base branch of the pr to 19w42a.

natanfudge:
I'll change it to latest when it's approved.

build.gradle
```diff
@@ -23,7 +22,7 @@ plugins {
 	id 'maven-publish'
 }
 
-def minecraft_version = "19w41a"
+def minecraft_version = "19w44a"
```
modmuss50: Not sure why this is showing as a change?

build.gradle
```diff
@@ -46,19 +45,21 @@ configurations {
...
 	}
+	intermediary
 }
 
 dependencies {
 	enigmaRuntime "net.fabricmc:stitch:0.2.1.61"
-	enigmaRuntime "cuchaz:enigma:0.14.2.134:all"
+	enigmaRuntime "cuchaz:enigma:0.14.2.134"
+	intermediary "net.fabricmc:intermediary:$minecraft_version"
+	intermediary "net.fabricmc:intermediary:$minecraft_version:v2"
```
modmuss50: I have a feeling this will break in a future gradle update as it is similar to the issue with the lwjgl natives. For now leave it be if you dont have a better idea.

natanfudge: I actually think this might not be needed anymore? let me check

natanfudge: We can keep it the  way it was before where it downloaded from Github, is that better?

build.gradle
```diff
@@ -178,25 +187,39 @@ task downloadMcJars(dependsOn: downloadWantedVersionManifest) {
...
 }
 
-task downloadIntermediary(type: Download){
+task extractIntermediary(type: FileOutput) {
 	group = buildMappingGroup
-	def url = "https://github.com/FabricMC/intermediary/raw/master/mappings/${minecraft_version}.tiny"
-	src UrlEscapers.urlFragmentEscaper().escape(url)
-	dest new File(cacheFilesMinecraft, "${minecraft_version}-intermediary.tiny")
+
+	def v1Input = configurations.intermediary.files[0]
```
modmuss50: This seems a bit fragile? Would it be a good idea to have a v1 and v2 confuration?

natanfudge: The v2 intermediary is actually not needed at all because I moved it to loom

natanfudge: look at the new commit

modmuss50:
Looks mostly ok, few things that ive commented on.

Would it be a good idea to include the auto generated names in the v2 output? as its something I woul like to do to at some point and should be quite easy to do.

natanfudge:
So you want it to produce intermediary - yarn with auto generated names?

modmuss50:
Yeah, I think so as its a bit annoying how it was done in v1. What do you (and anyone else) think about this idea?

natanfudge:
Ideally everything that can be done outside of loom should be done outside to reduce build times.

natanfudge:
Currently (in v2) we merge the mappings in loom in hopes to use newer yarn versions with older intermediary.

build.gradle
liach: @natanfudge Rebase you branch against 19w44a.

build.gradle
liach: If you do get only 1 file, use guava Iterable getOnlyElement which throws when it detects multiple possible entries.

natanfudge:
Merge https://github.com/FabricMC/stitch/pull/20 first.

build.gradle
```diff
@@ -49,7 +50,7 @@ configurations {
 }
 
 dependencies {
-	enigmaRuntime "net.fabricmc:stitch:0.2.1.61"
+	enigmaRuntime "net.fabricmc:stitch:0.3.0.65"
```
liach: Why not 0.3.1 for both stitch?

liach (approved):

build.gradle
natanfudge: 0.3.1 doesn't exist, I need FabricMC/stitch#20 to get merged and then I will change both.

build.gradle
```diff
@@ -71,30 +72,37 @@ def serverJar = new File(cacheFilesMinecraft, "${minecraft_version}-server.jar")
...
 import cuchaz.enigma.command.ConvertMappingsCommand
-import com.google.common.hash.Hashing
-import com.google.common.io.Files
-import com.google.common.net.UrlEscapers
 import groovy.io.FileType
 import groovy.json.JsonSlurper
 import net.fabricmc.stitch.commands.CommandMergeTiny
+
+//import com.google.common.io.Files
+
```
Player3324: delete these please

Player3324 (approved):
Looks not obviously wrong to me

liach:
so our param naming will get in 1.15 snapshots; i don't think we will bring this down to 114.4 for now.

natanfudge:
We'll do it right after we know it works. I already have a 1.14.4 branch with v2 branch for my mods :).

build.gradle
```diff
@@ -475,6 +475,7 @@ clean.doFirst {
 
 tasks.build.dependsOn "compressTiny"
 tasks.build.dependsOn "tinyJar"
+tasks.build.dependsOn "v2UnmergedYarnJar"
```
liach: Can just declare `tasks.build.dependsOn "compressTiny", "tinyJar", "v2UnmergedYarnJar"`

liach (approved):

modmuss50 (approved):
Im happy with the changes, tested the basic things that I need to and it seems to work fine.

Great job 👍

PR #921, opened 2019-10-24
Move loot package out of world

YanisBft:
All the datapack stuff (`advancement`, `recipe`, `tag`, `predicate` packages) is out of the `world` package, except `loot`
This PR fixes that


Juuxel (approved):

liach (approved):

modmuss50 (changes requested):
This is not how you rename mappings. Each mapping file also has the yarn name on the first line.

Please please use engima or at least be very careful when touching the mappings outside of engima.

liach:
hmm, we should have add enigma directory mapping verification for the gradle check task

modmuss50:
Yes, this should have failed to build.

YanisBft:
I was too lazy to move the files one by one, so I forgot half of the things >< Sorry!

liach (approved):

modmuss50 (approved):

Issue #929, opened 2019-10-28
BufferBuilder -> ByteBufferVertexBuffer


Runemoro:
While Mojang calls (or called) it a `BufferBuilder` (see https://github.com/FabricMC/yarn/issues/313, which renamed it from `VertexBuffer` to `BufferBuilder`), the name `VertexBuffer` is much better.

 1. `BufferBuilder` is ambiguous. What sort of buffer is it building?
 2. `BufferBuilder` makes it seem like it's meant to build a buffer and then convert it to a regular buffer afterwards (like `StringBuilder`), but in fact it has all the regular methods of Java buffers (set position, get data, get/restore state), and it's used to pass around the vertex data, not just to build the buffer. Even Mojang calls it a "buffer" in log messages: `"Allocated only {}/{} buffers"`
 3. Calling it a "buffer" would shorten names like `BlockLayeredBufferBuilderStorage`, while making them less misleading.

liach:
I support Runemoro. This buffer builder is just an intermediate buffer that makes sure the vertex buffers are properly formatted and tweaks certain input values based on its settings. However, given there may be a major snapshot this Wednesday, I suggest we revisit this topic after Wednesday snapshot is out.

Player3324:
BufferBuilder may not be sufficiently descriptive since there are indeed multiple kinds of buffers, particularly in OpenGL, like vertex buffers, pixel store buffers, element index buffers and more.

The term VertexBuffer is much worse though, it can (and should) be easily understood as the Java side representation of the OpenGL VBO, holding the buffer object id etc. The "Object" suffix doesn't matter, it doesn't change the meaning and the GL methods operating on VBOs don't use it.

Buffer builder is an utility that assembles vertex data using something close to a standard builder pattern. It buffers the data internally like StringBuilder, but the main property+purpose is the building part - not the buffering. The buffering aspect is just a practical implementation requirement/choice. The class wouldn't change much if the output was pass-through instead of buffered.

BufferBuilder can and is being used with glDrawArrays without ever touching any OpenGL buffer object.

Runemoro:
After talking with Player on Discord, I think we should `ByteBufferVertexBuffer` would be a better name for this class.

 1. It matches the log messages and debug renderer strings, which call them "buffers"
 2. It still works with the building part. A `ByteBuffer` builds byte data byte-by-byte, a `StringBuffer` builds strings, and a `PacketBuffer` builds packets, etc.
[...]

PR #930, opened 2019-10-29
Refactor rendering

### LINKED ISSUE #662: ChunkBatcher named wrong
Based on the thread names, the `ChunkBatcher` are the *threads* created by the class currently named `ChunkBatcher`:

```java
private static final ThreadFactory THREAD_FACTORY = new ThreadFactoryBuilder().setNameFormat("Chunk Batcher %d").setDaemon(true).setUncaughtExceptionHandler(new UncaughtExceptionLogger(LOGGER)).build();
```

### LINKED ISSUE #774: SystemUtil -> ?
`SystemUtil` contains a bunch of static methods. Some of them relate to a "system":

* fine-grained timers
* current operating system
* JVM flags

But many others are completely unrelated:

* patching miscellaneous holes in the Java standard library (`ifPresentOrElse`, `thenCombine`, `completeExceptionally`, `consume`, `get`, `toMap`, ...)
* a bit about the server worker
* reading and writing UUIDs from `Dynamic`s
* an identity hash strategy
* something about translation keys? lol

It appears to be less of a system-related class and more just a dumping ground for random functions that are too small to break into their own named helper class but too broadly applicable to keep near one call site.

Should this class be renamed?

Some suggestions being thrown around on Discord:

* `Misc`
* `Util`
* `Useful`
* `Varia`
* `Helper`
* `Etc`
* ~~Funny comedy suggesions:~~
  * ~~`Junk`~~
  * ~~`Stuff`~~
  * ~~`MiscellaneousGameSystemUtilities`~~
  * ~~`BucketOShi`~~

### LINKED ISSUE #926: Improve concurrency class naming
We currently have a `Mailbox` class and a `MailboxProcessor` class, but a log message in `MailboxProcessor` suggests that in fact `MailboxProcessor` is the `Mailbox`:

```java
    private void schedule() {
        if (hasMessages() && lock()) {
            try {
                executor.execute(this);
            } catch (RejectedExecutionException e1) {
                try {
                    executor.execute(this);
                } catch (RejectedExecutionException e2) {
                    LOGGER.error("Cound not schedule mailbox", e2);
                }
            }
        }
    }
```

I suggest renaming the class currently named `Mailbox` to something like `MessageQueue`, since it's really just a queue which receives messages (for example an action + a priority) and gives an action (for example the highest priority one) when polled. Then, we can rename `MailboxProcessor` (which uses a `MessageQueue` to store its messages, and provides a few extra methods for executing tasks sync/async, locking and unlocking, etc.) to `Mailbox`.

Also, as suggested by the log message in `Actor`, a better name for `createAndSendFutureActor` would be `ask`:

```java
    default <Source> CompletableFuture<Source> ask(Function<? super Actor<Source>, ? extends Msg> messageProvider) {
        CompletableFuture<Source> future = new CompletableFuture<>();
        Actor<Source> actor = create("ask future procesor handle", future::complete); // actor that acts on Source messages to complete the future
        Msg message = messageProvider.apply(actor);
        send(message);
        return future;
    }
```

It creates a future, creates an actor that receives source messages and completes the future, applies a function to convert the actor to a message, and sends that message. So it's really a helper method for sending "questions" to the `Actor` and returning a future for the reply.

The same log message would suggest `Actor` is actually named `ProcessorHandle`, but I think `Actor` is more descriptive (it accepts messages and acts on them), and [already used](https://en.wikipedia.org/wiki/Actor_model) (in fact, [this page](https://berb.github.io/diploma-thesis/original/054_actors.html) matches almost perfectly how the concurrency system works).

### LINKED ISSUE #927: ChunkRenderer -> RenderedChunk
That name would make it clearer that it's storing the actual rendered chunk (the buffers with the vertex data for each layer) rather than just a class that's called to render. It also works better with the `rebuild` method name (it's rebuilding the rendered chunk, not the chunk renderer).

This would also free up the `ChunkRenderer` name for the class currently named `ChunkBatcher` (see https://github.com/FabricMC/yarn/issues/662).

Runemoro:
- `ChunkRenderer` -> `RenderedChunk` (fixes https://github.com/FabricMC/yarn/issues/927)
 - `Mailbox` -> `MessageQueue`, `MailboxProcessor` -> `Mailbox` (fixes https://github.com/FabricMC/yarn/issues/926)
 - `SystemUtil` -> `Util` (fixes https://github.com/FabricMC/yarn/issues/774)
 - `ChunkBatcher` -> `ChunkRenderer` (fixes https://github.com/FabricMC/yarn/issues/662)


liach:
Uh, this seems to conflict with #917 on rendering stuff

Runemoro:
The only conflict is `LayeredBufferStorage` (I renamed it, and you added methods to it). I'll fix the conflict after your PR is merged.

liach:
Given #929, we might need to hold off on this.

Runemoro:
I'll be reverting the BufferBuilder rename later today so that the other changes can be merged, but leave the issue open for discussion.

liach:
Please update to 19w44a.

Runemoro:
Updated to 19w44a, but don't review yet, I'll revert #929 first.

mappings/net/minecraft/util/MessageQueue.mapping
```diff
@@ -0,0 +1,19 @@
+CLASS net/minecraft/class_3847 net/minecraft/util/MessageQueue
+	CLASS class_3848 PrioritizedQueueMailbox
```
liach: This class probably needs a rename as well.

mappings/net/minecraft/util/Util.mapping
```diff
@@ -41,9 +41,11 @@ CLASS net/minecraft/class_156 net/minecraft/util/SystemUtil
 	METHOD method_648 getMeasuringTimeNano ()J
 	METHOD method_650 getValueAsString (Lnet/minecraft/class_2769;Ljava/lang/Object;)Ljava/lang/String;
 	METHOD method_651 getJVMFlags ()Ljava/util/stream/Stream;
-	METHOD method_652 thenCombine (Ljava/util/List;)Ljava/util/concurrent/CompletableFuture;
+	METHOD method_652 combine (Ljava/util/List;)Ljava/util/concurrent/CompletableFuture;
 		ARG 0 futures
-	METHOD method_654 consume (Ljava/lang/Object;Ljava/util/function/Consumer;)Ljava/lang/Object;
+	METHOD method_654 initialize (Ljava/lang/Object;Ljava/util/function/Consumer;)Ljava/lang/Object;
```
liach: If you rename this one to initialize, I think the supplier one below can be renamed as well.

Runemoro: It would be weird calling the other one `initialize` since it just creates something using a supplier, while this one initializes a given object by applying a consumer to it.

I do agree that they should have the same name since they're used for the same purpose, so I'll name them both `create`.

liach: Both function as `setup` or so. sure

mappings/net/minecraft/client/render/LayeredVertexOutput.mapping
```diff
@@ -0,0 +1,23 @@
...
+		FIELD field_20955 startedBuilders Ljava/util/Set;
+		METHOD <init> (Lnet/minecraft/class_287;Ljava/util/Map;)V
+			ARG 1 fallbackBuilder
+			ARG 2 layerBuilders
+		METHOD method_22993 draw ()V
+		METHOD method_22994 draw (Lnet/minecraft/class_1921;)V
+			ARG 1 layer
+		METHOD method_22995 getConsumer (Lnet/minecraft/class_1921;)Lnet/minecraft/class_287;
+			ARG 1 layer
+	METHOD getBuffer getVertexConsumer (Lnet/minecraft/class_1921;)Lnet/minecraft/class_4588;
```
liach: Uh no

mappings/net/minecraft/client/render/LayeredVertexOutput.mapping
```diff
@@ -0,0 +1,23 @@
...
+		ARG 1 layer
+	METHOD getVertexConsumer get (Lnet/minecraft/class_1921;)Lnet/minecraft/class_4588;
```
liach: No either

mappings/net/minecraft/client/render/block/BlockRenderManager.mapping
```diff
@@ -1,13 +1,31 @@
...
+		ARG 5 vertexConsumer
 	METHOD method_3349 getModel (Lnet/minecraft/class_2680;)Lnet/minecraft/class_1087;
 		ARG 1 state
 	METHOD method_3350 getModelRenderer ()Lnet/minecraft/class_778;
 	METHOD method_3351 getModels ()Lnet/minecraft/class_773;
-	METHOD method_3352 tesselateFluid (Lnet/minecraft/class_2338;Lnet/minecraft/class_1920;Lnet/minecraft/class_4588;Lnet/minecraft/class_3610;)Z
-	METHOD method_3353 renderOnEntity (Lnet/minecraft/class_2680;Lnet/minecraft/class_4587;Lnet/minecraft/class_4597;II)V
-	METHOD method_3355 tesselateBlock (Lnet/minecraft/class_2680;Lnet/minecraft/class_2338;Lnet/minecraft/class_1920;Lnet/minecraft/class_4587;Lnet/minecraft/class_4588;ZLjava/util/Random;)Z
+	METHOD method_3352 renderFluid (Lnet/minecraft/class_2338;Lnet/minecraft/class_1920;Lnet/minecraft/class_4588;Lnet/minecraft/class_3610;)Z
+	METHOD method_3353 renderBlock (Lnet/minecraft/class_2680;Lnet/minecraft/class_4587;Lnet/minecraft/class_4597;II)V
```
liach: This one is different in that it uses different render layers than those of regular blocks. Beware!

mappings/net/minecraft/client/render/chunk/ChunkRenderer.mapping
```diff
@@ -44,14 +44,16 @@ CLASS net/minecraft/class_846 net/minecraft/client/render/chunk/ChunkRenderer
 		FIELD field_4459 data Ljava/util/concurrent/atomic/AtomicReference;
 		FIELD field_4461 sortTask Lnet/minecraft/class_846$class_851$class_4579;
 		FIELD field_4462 buffers Ljava/util/Map;
-		FIELD field_4463 rebuildOnClientThread Z
-		FIELD field_4464 rebuildScheduled Z
+		FIELD field_4463 needsImportantRebuild Z
```
liach: What does "important" mean?

Runemoro: The rebuild task has higher priority, and it will rebuild on the client thread if the option to rebuild near on client thread is enabled. It's used for blocks placed by a player, for example (it's important for those to show up immediately).

mappings/net/minecraft/client/render/LayeredVertexOutput.mapping
Runemoro: Oops, renamed it at its usage and didn't realize it was a Mojang name. I'll fix this.

liach: Well, if mojang did not use this as a method reference at where they render the breaking progress, we could have named it with much more ease instead

mappings/net/minecraft/util/MessageQueue.mapping
```diff
@@ -0,0 +1,22 @@
+CLASS net/minecraft/class_3847 net/minecraft/util/MessageQueue
+	CLASS class_3848 PrioritizedMessageQueue
+		FIELD field_17044 queues Ljava/util/List;
+		METHOD <init> (I)V
+			ARG 1 priorityCount
+	CLASS class_3849 SimpleMessageQueue
```
liach:
```suggestion
	CLASS class_3849 Simple
```

mappings/net/minecraft/util/MessageQueue.mapping
```diff
@@ -0,0 +1,22 @@
+CLASS net/minecraft/class_3847 net/minecraft/util/MessageQueue
+	CLASS class_3848 PrioritizedMessageQueue
```
liach:
```suggestion
	CLASS class_3848 Prioritized
```

liach (approved):

liach:
On a side note, there is a few new additions on rendering details as well.

Player3324:
The Mailbox related changes are ok.

Util is a very generic/common name, maybe it can be aligned more closely with mc.

For ChunkRenderer and ChunkBatcher I'd prefer a stricter interpretation of the term 'rendering' along the lines of drawing, rasterization or getting something to the display/frame buffer. This excludes the cases where the output is some buffer that doesn't get drawn (almost) immediately and discarded, but reused for many draws. Updating or baking chunk(section)s is one way to express this process in the client/render engine context.

mappings/net/minecraft/client/render/BuiltChunkStorage.mapping
```diff
@@ -0,0 +1,29 @@
+CLASS net/minecraft/class_769 net/minecraft/client/render/BuiltChunkStorage
+	FIELD field_4146 worldRenderer Lnet/minecraft/class_761;
+	FIELD field_4147 sizeZ I
```
Player3324, line 3: I think xSize xOffset etc fit our "natural order" naming style better, sizeX etc is prefix style

mappings/net/minecraft/util/thread/TaskExecutor.mapping
```diff
@@ -1,20 +1,21 @@
...
 	METHOD method_16902 create (Ljava/util/concurrent/Executor;Ljava/lang/String;)Lnet/minecraft/class_3846;
 		ARG 0 executor
 		ARG 1 name
-	METHOD method_16903 lock ()Z
-	METHOD method_16904 unlock ()V
+	METHOD method_16903 unpause ()Z
+	METHOD method_16904 pause ()V
 	METHOD method_16905 hasMessages ()Z
-	METHOD method_16906 isLocked ()Z
+	METHOD method_16906 isEnabled ()Z
```
Player3324: are you sure? isPaused maybe?

mappings/net/minecraft/util/thread/ThreadExecutor.mapping
```diff
@@ -1,30 +1,34 @@
...
+		ARG 1 task
 	METHOD method_20813 waitForTasks ()V
-	METHOD method_21684 getTaskQueueSize ()I
+	METHOD method_21684 getTaskCount ()I
 	METHOD method_3777 getThread ()Ljava/lang/Thread;
-	METHOD method_5382 createFuture (Ljava/lang/Runnable;)Ljava/util/concurrent/CompletableFuture;
+	METHOD method_5382 submitAsync (Ljava/lang/Runnable;)Ljava/util/concurrent/CompletableFuture;
 		ARG 1 runnable
-	METHOD method_5383 executeQueuedTasks ()V
+	METHOD method_5383 executeTasks ()V
```
Player3324: maybe runTasks? execute(Runnable) is already being used above. The same applies to executeTask()

Player3324 (commented):
Some mentions of mailbox and message seem to have slipped through, there may also be mentions of "render" in a "building/baking" context.

Overall a very nice change!

mappings/net/minecraft/client/render/BuiltChunkStorage.mapping
Runemoro: Right now we're naming coordinates the other way (see [conventions](https://github.com/FabricMC/yarn/blob/19w44a/CONVENTIONS.md#coordinates), and also almost all current coordinate fields are named this way). The idea is that the field is the "size['s] x [component]".

The other order could work too, but I think it would be better to make a separate issue to change the conventions as well as rename fields currently named this way.

Player3324: Right, I didn't even realize we added this to the conventions and I agree this should be fixed separately if desired.

For further discussion in this area it should be noted that the order isn't fixed with a natural scheme, it depends on subject ordering. There would be e.g. xSize and minX. The other form sizeX refers to a size called "x" which makes little sense and xMin implies that the x axis itself had a minimum, which it doesn't.

mappings/net/minecraft/client/render/BufferBuilderStorage.mapping
```diff
@@ -0,0 +1,10 @@
+CLASS net/minecraft/class_4599 net/minecraft/client/render/BufferBuilderStorage
+	FIELD field_20956 blockBuilders Lnet/minecraft/class_750;
+	FIELD field_20957 otherBuilders Ljava/util/SortedMap;
+	FIELD field_20958 generalDrawer Lnet/minecraft/class_4597$class_4598;
+	FIELD field_20959 blockBreakingProgressDrawer Lnet/minecraft/class_4597$class_4598;
+	FIELD field_20961 teamColorAwareOutlineDrawer Lnet/minecraft/class_4618;
+	METHOD method_22997 getBlockBuilders ()Lnet/minecraft/class_750;
+	METHOD method_23000 getGeneralVertexConsumers ()Lnet/minecraft/class_4597$class_4598;
+	METHOD method_23001 getBlockBreakingVertexConsumers ()Lnet/minecraft/class_4597$class_4598;
+	METHOD method_23003 getTeamColorAwareOutlineVertexConsumers ()Lnet/minecraft/class_4618;
```
liach: Consider improving the name of this method. I recall you renamed this somewhere else

mappings/net/minecraft/util/math/Box.mapping
```diff
@@ -1,10 +1,10 @@
...
-	FIELD field_1322 minY D
-	FIELD field_1323 minX D
-	FIELD field_1324 maxZ D
-	FIELD field_1325 maxY D
+	FIELD field_1320 x2 D
+	FIELD field_1321 z1 D
+	FIELD field_1322 y1 D
+	FIELD field_1323 x1 D
+	FIELD field_1324 z2 D
+	FIELD field_1325 y2 D
```
liach, line 7: why 1 and 2 instead of 0 and 1?

Runemoro: Why 0 and 1 rather than 1 and 2? There doesn't seem to be a consensus on how these should be named, there are currently many things named `x2`. I opened an issue here: https://github.com/FabricMC/yarn/issues/941

modmuss50 (approved):
Im happy with this, im by no means an expert with rendering so getting more reviews in might not be a bad idea.

PR #945, opened 2019-11-08
Map biome color cache stuff and matrixstack pop

liach:

### DIFF
--- a/mappings/net/minecraft/client/render/WorldRenderer.mapping
+++ b/mappings/net/minecraft/client/render/WorldRenderer.mapping
@@ -72,6 +72,10 @@ CLASS net/minecraft/class_761 net/minecraft/client/render/WorldRenderer
 		ARG 4 maxX
 		ARG 5 maxY
 		ARG 6 maxZ
+	METHOD method_21596 checkBlockRerender (Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Lnet/minecraft/class_2680;)V
+		ARG 1 pos
+		ARG 2 old
+		ARG 3 updated
 	METHOD method_22710 render (Lnet/minecraft/class_4587;FJZLnet/minecraft/class_4184;Lnet/minecraft/class_757;Lnet/minecraft/class_765;Lnet/minecraft/class_1159;)V
 		ARG 1 matrix
 		ARG 2 tickDelta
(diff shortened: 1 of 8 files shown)


mappings/net/minecraft/client/render/WorldRenderer.mapping
```diff
@@ -72,6 +72,10 @@ CLASS net/minecraft/class_761 net/minecraft/client/render/WorldRenderer
 		ARG 4 maxX
 		ARG 5 maxY
 		ARG 6 maxZ
+	METHOD method_21596 checkBlockRerender (Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Lnet/minecraft/class_2680;)V
```
Runemoro, line 75: `Render`

mappings/net/minecraft/world/World.mapping
```diff
@@ -175,8 +178,8 @@ CLASS net/minecraft/class_1937 net/minecraft/world/World
 	METHOD method_8508 updateNeighborsExcept (Lnet/minecraft/class_2338;Lnet/minecraft/class_2248;Lnet/minecraft/class_2350;)V
 		ARG 1 pos
 		ARG 2 sourceBlock
-	METHOD method_8509 setTicksSinceLightning (I)V
-		ARG 1 ticksSinceLightning
+	METHOD method_8509 setLightningBrightTicks (I)V
+		ARG 1 lightningBrightTicks
```
Runemoro: What is this? It's not clear from the new name.

liach: the ticks the world should be rendered bright due to a lightning. on client a lightning sets it to 2. decreases 1 per tick.

mappings/net/minecraft/client/render/WorldRenderer.mapping
liach: it rebuilds render if the old and new states are in different model groups

Runemoro: Oh, I see! I didn't realized it was re-render, it just looked like a typo.

mappings/net/minecraft/world/World.mapping
Runemoro: So this is decreased every tick? If yes, I think `getLightningTicksRemaining` would be better. Otherwise, someone might think it's a constant value for how long all lightnings are.

Runemoro (approved):

Juuxel (approved):

Player3324 (approved):
LGTM, thanks!

PR #946, opened 2019-11-08
Update engima to latest

liach:
Now we can navigate to methods through method references and inner class parameters can be named now.

We also have stats gen ready like this:
image

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
 		}
 	}
 	dependencies {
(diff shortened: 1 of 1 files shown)


Runemoro (approved):

liach:
Working method reference:
image
Working inner class param names:
image

Juuxel (approved):

modmuss50 (approved):

PR #948, opened 2019-11-09
Produce TinyV2 mappings for 1.14.4

natanfudge:
This is the same build.gradle file as the default branch.


liach:
I suggest using the build.gradle at fb663d6f1dab98ccada33cb2f93dd53fc7d1dd93

natanfudge:
oh right modmuss changed his setup

modmuss50:
Yeah, it doesnt make too much diffrence here but keeping them the same might be nice.

Lets make sure its all working well in 1.15 before merging this as well. I want to merge this as it requires changing the order on the maven manually.

build.gradle
```diff
@@ -445,57 +457,56 @@ clean.doFirst {
...
-			"enigma",
-			file("mappings/").getAbsolutePath(),
-			"right"
+				"tiny",
+				composeInput.getAbsolutePath(),
+				"enigma",
+				file("mappings_official/").getAbsolutePath(),
+				"enigma",
+				file("mappings/").getAbsolutePath(),
+				"right"
```
Sollace, line 504: There's a lot of formatting changes in here. IMO I'd try to keep it the same as it was for a cleaner PR.

natanfudge: it's fineee

modmuss50:
Before merging this I do want to check that is isnt changing the auto generated names? As that may have been an issue on the main branch

natanfudge:
I'm going to add a flag to stitch to be able to toggle that.

modmuss50:
Sounds good, sadly if it was a bug we need to keep the bug for the older versions.

natanfudge:
Well, you _can_ decide that it's desired behavior that it will only insert _unmapped_ names.

modmuss50:
No, the dev shoudlnt have this option imo as source code for x version of yarn should work in someone else's dev env with the same version. For already released versions it needs to be kept the same. I dont see an issue in changing it from the current snapshot onwards though.

natanfudge:
No, not the dev, **you**.

liach:
you need to update stitch in another place below as well.

build.gradle
modmuss50: Slighly annoying, but yeah dont worry.

modmuss50 (approved):
Im happy with this, v2 seems to be working well in 1.15 so this should hopefully be all thats needed for 1.14

Player3324 (approved):

PR #949, opened 2019-11-09
Player-suggested changes

Sollace:
#839


liach:
Consider finding other usages of "view" (with grepwin or something similar) and replace them accordingly as well.

Sollace:
I did look for other usages, but decided to leave them because they were paired with things like "BlockView".

Sollace:
I did look for other usages, but decided to leave them because they were paired with things like "BlockView".

Runemoro (approved):

Player3324 (approved):

modmuss50 (approved):

PR #799, opened 2019-07-25
Mapped several things across all the modelling classes and renderers

Sollace:
It's quite a lot, so I can't really cover every little thing I did in a description here, sorry!

PR #951, opened 2019-11-10
Client and miscellaneous mappings

liach:
Entity sneaks, block jump/walk multiplier, client params, bee debug renderer


mappings/net/minecraft/client/render/model/ModelLoader.mapping
```diff
@@ -7,9 +7,22 @@ CLASS net/minecraft/class_1088 net/minecraft/client/render/model/ModelLoader
...
+	FIELD field_20848 DESTROY_STAGES Ljava/util/List;
+	FIELD field_21014 OAK_SIGN Lnet/minecraft/class_2960;
+	FIELD field_21015 SPRUCE_SIGN Lnet/minecraft/class_2960;
+	FIELD field_21016 BIRCH_SIGN Lnet/minecraft/class_2960;
+	FIELD field_21017 ACACIA_SIGN Lnet/minecraft/class_2960;
+	FIELD field_21018 JUNGLE_SIGN Lnet/minecraft/class_2960;
+	FIELD field_21019 DARK_OAK_SIGN Lnet/minecraft/class_2960;
 	FIELD field_21020 BLOCK_BREAKING_STAGES Ljava/util/List;
+	FIELD field_21557 SHIELD_BASE Lnet/minecraft/class_2960;
+	FIELD field_21558 NO_PATTERN_SHIELD_BASE Lnet/minecraft/class_2960;
```
Sollace: You're overlapping with my mappings in PR #799

liach: can have yours merged first and then I fix conflicts

mappings/net/minecraft/block/BeeHiveBlock.mapping
```diff
@@ -1,9 +1,34 @@
 CLASS net/minecraft/class_4481 net/minecraft/block/BeeHiveBlock
+	FIELD field_20418 NATRURAL_FACINGS [Lnet/minecraft/class_2350;
```
Runemoro: "natrual"? Should it be "natural", and if yes, I'm not sure what "natural facings" means...

mappings/net/minecraft/block/Block.mapping
```diff
@@ -77,6 +84,8 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 	FIELD field_19061 SOLID_SMALL_SQUARE_SHAPE Lnet/minecraft/class_265;
 	FIELD field_19312 FULL_CUBE_SHAPE_CACHE Lcom/google/common/cache/LoadingCache;
 	FIELD field_20720 opaque Z
+	FIELD field_21207 velocityMultiplier F
+	FIELD field_21208 jumpVelocityMultiplier F
```
Runemoro, line 88: What blocks override this? If it's slime and honey, I think it should be "bounce velocity multiplier/ratio", since it affects bounces off of the block, not jumps. It could also be called "[restitution](https://en.m.wikipedia.org/wiki/Coefficient_of_restitution)".

mappings/net/minecraft/client/render/block/entity/BlockEntityRenderDispatcher.mapping
```diff
@@ -14,23 +14,33 @@ CLASS net/minecraft/class_824 net/minecraft/client/render/block/entity/BlockEnti
...
 		ARG 2 matrix
 	METHOD method_3554 runReported (Lnet/minecraft/class_2586;Ljava/lang/Runnable;)V
+		ARG 0 blockEntity
+		ARG 1 task
 	METHOD method_3555 render (Lnet/minecraft/class_2586;FLnet/minecraft/class_4587;Lnet/minecraft/class_4597;)V
 		ARG 1 blockEntity
 		ARG 2 tickDelta
-		ARG 3 matrix
+		ARG 3 matrices
+		ARG 4 vertexConsumerProvider
```
Runemoro, line 45: This can be `vertexConsumers`.

mappings/net/minecraft/client/render/block/entity/BlockEntityRenderer.mapping
```diff
@@ -1,6 +1,15 @@
 CLASS net/minecraft/class_827 net/minecraft/client/render/block/entity/BlockEntityRenderer
 	FIELD field_20989 blockEntityRenderDispatcher Lnet/minecraft/class_824;
+	METHOD <init> (Lnet/minecraft/class_824;)V
+		ARG 1 dispatcher
 	METHOD method_23082 getSprite (Lnet/minecraft/class_2960;)Lnet/minecraft/class_1058;
 		ARG 1 id
+	METHOD method_3563 shouldSendToChunkBuilder (Lnet/minecraft/class_2586;)Z
```
Runemoro, line 7: When is this true? If it's for block entities that render as normal blocks (like the furnace), it could 
 be `rendersAsRegularBlock`.

mappings/net/minecraft/client/render/entity/EntityRenderer.mapping
```diff
@@ -2,16 +2,29 @@ CLASS net/minecraft/class_897 net/minecraft/client/render/entity/EntityRenderer
...
 		ARG 1 entity
+		ARG 2 tickDelta
 	METHOD method_3921 hasLabel (Lnet/minecraft/class_1297;)Z
 		ARG 1 entity
 	METHOD method_3926 renderLabelIfPresent (Lnet/minecraft/class_1297;Ljava/lang/String;Lnet/minecraft/class_4587;Lnet/minecraft/class_4597;I)V
 		ARG 1 entity
+		ARG 2 text
+		ARG 3 matrices
+		ARG 4 vertexConsumerProvider
+		ARG 5 lightmapCoordinate
```
Runemoro: lightmapCoordinate**s** (it's the lightmap uv packed into an int). Or just "light", to match the name in VertexConsumer.

mappings/net/minecraft/block/BeeHiveBlock.mapping
liach: like the possible faces for world gen

mappings/net/minecraft/block/Block.mapping
liach: jump one is not bounce. it's what prevents you from jumping.

mappings/net/minecraft/client/render/block/entity/BlockEntityRenderer.mapping
liach: this is true for beacons and what else...

mappings/net/minecraft/block/Block.mapping
liach: @Runemoro To clarify, this has nothing to do with the bouncing factors.

mappings/net/minecraft/client/render/block/entity/BlockEntityRenderDispatcher.mapping
liach: Aren't most of the param usages called `vertexConsumerProvider`?

Runemoro: No, the `VertexConsumerProvider` was renamed to that in my last PR, and all parameters I named are `vertexConsumers`.

mappings/net/minecraft/block/BeeHiveBlock.mapping
Runemoro: Maybe `VALID_FACINGS` would be better. It's not just worldgen, you also can't place a bee hive facing up.

mappings/net/minecraft/client/render/block/entity/BlockEntityRenderer.mapping
Runemoro: What about something like `isBlockModelRendered` (it also renders the regular block as well as custom rendering)?

mappings/net/minecraft/block/BeeHiveBlock.mapping
liach: it won't face north for worldgen, only south east west

Runemoro: Ok then.

mappings/net/minecraft/client/render/block/entity/BlockEntityRenderer.mapping
liach: This looks like block entities that render things beyond itself. Fyi this is true for beacon and structure block.

Runemoro: Based on the current name, this makes me think that if the method returns true, the tile entity block is sent to the chunk builder to render its block model as part of the chunk. So for example beacons, structure blocks, but not chests.

If that's what it does, `shouldRenderBlock` would be a better name.

Otherwise, the name should be improved to make it clear when I should be overriding this method in my mod.

Runemoro: Based on the current name, this makes me think that if the method returns true, the tile entity block is sent to the chunk builder to render its block model as part of the chunk. So for example beacons, structure blocks, but not chests.

If that's what it does, `shouldRenderBlock` would be a better name.

Otherwise, the name should be improved to make it clear when I should be overriding this method in my mod.

liach: This is not block model, but it renders a lot of stuff beyond the block model (like beacon or structure block). You should look into it on your own given I am not that good at rendering...

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -517,7 +527,7 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	METHOD method_5809 isOnFire ()Z
 	METHOD method_5810 isPushable ()Z
 	METHOD method_5812 getPosVector ()Lnet/minecraft/class_243;
-	METHOD method_5814 setPosition (DDD)V
+	METHOD method_5814 initPosition (DDD)V
```
Juuxel, line 530: I'm not sure if `init` is a good verb here. It's also used in ticking.

Sollace: Also what makes it "not a setter"?

liach: it is called in the entity spawn packet for client entity creation, so...

liach:
Need to remind myself to update this.

mappings/net/minecraft/client/render/block/entity/BlockEntityRenderer.mapping
Runemoro: @liach Figured out what this does and mapped it in [my PR](https://github.com/FabricMC/yarn/pull/974). When it's true, the chunk builder will add the block entity to a list of block entities that are always rendered, even if they are in a chunk that's culled (for example, underground or behind the player).

This is true for the beacon and the structure block because they render the beam and the wireframe outside of their bounding box, which are visible even if the block itself isn't.

liach:
Planning to redo this in the snapshot tomorrow.

PR #968, opened 2019-11-19
Improve game rule mappings

ChloeDawn:
- `of` -> `create`: The former implied it was constructing an instance of the class, which is not the case. It constructs a type representation of the rule.
- `value` -> `initialValue`: Better implies the semantics of the given value during construction.
- `notifier` -> `changeCallback`: It is a callback executed when the rule is changed, this name is more appropriate.
- `fromNbt` -> `load`: The method is non-static, so `fromX` naming is a unconventional. It loads the game rules from the given NBT.
- `RULES`, `forEach`, `RuleConsumer` -> `RULE_TYPES`, `forEachType`, `RuleTypeConsumer`: Better clarification between the concept of rules and rule types.
- `setFromString`, `valueToString` -> `deserialize`, `serialize`: These String-based methods are used exclusively for (de)serializing the rules to and from NBT.

### DIFF
--- a/mappings/net/minecraft/world/GameRules.mapping
+++ b/mappings/net/minecraft/world/GameRules.mapping
@@ -3,7 +3,7 @@ CLASS net/minecraft/class_1928 net/minecraft/world/GameRules
 		FIELD field_19411 value Z
 		METHOD <init> (Lnet/minecraft/class_1928$class_4314;Z)V
 			ARG 1 type
-			ARG 2 value
+			ARG 2 initialValue
 		METHOD method_20753 get ()Z
 		METHOD method_20754 (Lnet/minecraft/server/MinecraftServer;Lnet/minecraft/class_1928$class_4310;)V
 			ARG 0 server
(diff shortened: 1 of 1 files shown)


modmuss50 (approved):

mappings/net/minecraft/world/GameRules.mapping
```diff
@@ -65,27 +65,30 @@ CLASS net/minecraft/class_1928 net/minecraft/world/GameRules
 		METHOD method_20776 setFromArgument (Lcom/mojang/brigadier/context/CommandContext;Ljava/lang/String;)V
 			ARG 1 context
 			ARG 2 name
-		METHOD method_20777 setFromString (Ljava/lang/String;)V
+		METHOD method_20777 deserialize (Ljava/lang/String;)V
```
liach, line 68: Eh, `read` and `write` instead?

ChloeDawn: `write` would be a poor name, as it is a no-args method that returns the serialized form. With the proposed name you would call `rule.serialize()` (Technically no-one but Mojang would call it, it is a protected method intended only for serializing the rule for NBT saving).
```java
public CompoundTag toNbt() {
	final CompoundTag tag = new CompoundTag();
	this.rules.forEach((key, rule) -> {
		tag.putString(key.name, rule.serialize());
	});
	return tag;
}
```
```java
public void load(final CompoundTag tag) {
	this.rules.forEach((key, rule) -> {
		if (tag.contains(key.name)) {
			rule.deserialize(tag.getString(key.name));
		}
	});
}
```

Gegy (approved):

liach (approved):

Runemoro (approved):

PR #969, opened 2019-11-19
Debug mappings v2: Electric Boogaloo

Sollace:
Redone based on 19w46b.

PR #970, opened 2019-11-19
class_4706 -> net/minecraft/util/OrderedSet

Sollace:
Enigma did _not_ want to decompile this one. I had to write the mappings by hand.

The class itself is only used in `ChunkTicketManager` as its container for holding tickets, but the class itself is just a generic implementation of `AbstractSet<T>` with an array as its internal storage, so I thought it fit well to go into the utils package instead.

Between this and #969 that is all unknown (19w46) classes named and mapped.


mappings/net/minecraft/util/ArraySet.mapping
```diff
@@ -0,0 +1,26 @@
+CLASS net/minecraft/class_4706 net/minecraft/util/ArraySet
+	CLASS class_4707 Iterator
```
liach: This name will look a bit ugly in regular decompiled code like `class Iterator implements java.util.Iterator`

mappings/net/minecraft/util/ArraySet.mapping
```diff
@@ -0,0 +1,26 @@
...
+		FIELD field_21566 start I
+		FIELD field_21567 index I
+	FIELD field_21562 sortingFunction Ljava/util/Comparator;
```
liach: `comparator` should suffice.

mappings/net/minecraft/util/ArraySet.mapping
```diff
@@ -0,0 +1,26 @@
...
+	FIELD field_21563 entries [Ljava/lang/Object;
+	FIELD field_21564 capacity I
+	METHOD <init> (ILjava/util/Comparator;)V
+		ARG 0 initialCapacity
+		ARG 1 sortingFunction
+	METHOD method_23859 create (I)Lnet/minecraft/class_4706;
+		ARG 0 capacity
```
liach:
```suggestion
		ARG 0 initialCapacity
```

mappings/net/minecraft/util/ArraySet.mapping
```diff
@@ -0,0 +1,26 @@
...
+	METHOD method_23862 addIfNotPresent (Ljava/lang/Object;)Ljava/lang/Object;
```
liach:
```suggestion
	METHOD method_23862 addIfAbsent (Ljava/lang/Object;)Ljava/lang/Object;
```
Also, what is the returned value?

mappings/net/minecraft/util/ArraySet.mapping
```diff
@@ -0,0 +1,26 @@
...
+	METHOD method_23863 spliceAdd (Ljava/lang/Object;I)V
+		ARG 0 element
```
liach: Hmm, what is the second parameter?

mappings/net/minecraft/util/ArraySet.mapping
```diff
@@ -0,0 +1,26 @@
...
+	METHOD method_23865 peek ()Ljava/lang/Object;
```
liach: Does this peek the maximum or the minimum value?

mappings/net/minecraft/util/ArraySet.mapping
```diff
@@ -0,0 +1,26 @@
...
+	METHOD method_23866 toInsertionPoint (I)I
+		ARG 0 wouldBeIndex
+	METHOD method_23868 resizeArray (I)V
+		ARG 0 newCapacity
+	METHOD method_23869 getIndexOf (Ljava/lang/Object;)I
```
liach: `index`

mappings/net/minecraft/util/ArraySet.mapping
Sollace: Comparitor is used for sorting, though. :/

mappings/net/minecraft/util/ArraySet.mapping
Sollace: Returned value is the stored value. It's either the existing value, if present, or the passed in value after being inserted.

mappings/net/minecraft/util/ArraySet.mapping
Sollace: Insertion index.

```
list.spliceAdd(entry, index)
```

Oh, maybe I missed it.

mappings/net/minecraft/util/ArraySet.mapping
Sollace: It uses queue logic. Peek returns `values[0]`.

mappings/net/minecraft/util/ArraySet.mapping
Sollace: ?

Sollace: `wouldBeIndex` is the return value of `Arrays.binarySearch` where the search found nothing.

The documentation notes that value as `(-(insertion point) - 1)`

```
   private int getIndexOf(T element) {
      return Arrays.binarySearch(this.entries, 0, this.capacity, element, this.sortingFunction);
   }

   private static int toInsertionPoint(int wouldBeIndex) {
      return -wouldBeIndex - 1;
   }

   @Override
   public boolean add(T element) {
      int i = this.getIndexOf(element);
      if (i >= 0) {
         return false;
      }
      this.spliceAdd(element, ArraySet.toInsertionPoint(i));
      return true;
   }
```

mappings/net/minecraft/util/ArraySet.mapping
Pyrofab: I would note `java.util.AbstractList` uses `Itr` for its internal Iterator implementation.

mappings/net/minecraft/util/ArraySet.mapping
Pyrofab: Agreed with Sollace, I'd rather have a more descriptive name than redundant information.

mappings/net/minecraft/util/ArraySet.mapping
Pyrofab: `java.util.List` uses `indexOf` and not `getIndexOf` for that operation.

mappings/net/minecraft/util/ArraySet.mapping
Sollace: I don't like that name, tbh. Seems to be standard in the java libs as well, so I guess I'll take it.

mappings/net/minecraft/util/ArraySet.mapping
liach: Isn't this already a **sorting** function? Is this like a navigable set?

mappings/net/minecraft/util/ArraySet.mapping
Sollace: Yes. I named it as such to avoid conflicting with the inherited method.

mappings/net/minecraft/util/ArraySet.mapping
Sollace: It's technically an `OrderedSet`.

In fact, maybe that's a better name.

mappings/net/minecraft/util/ArraySet.mapping
Pyrofab: There should be no inherited method, Sets don't have it.

mappings/net/minecraft/util/ArraySet.mapping
liach: so does this directly impl iterator do does it extend abstractlist itr? iterator is fine if this extends itr

mappings/net/minecraft/util/ArraySet.mapping
liach: this is not exposed, so index can be used as the cleanest name.

mappings/net/minecraft/util/ArraySet.mapping
liach: so should be peekMin

Sollace: And what's a `peekMin`? Queues only have one `peek` method and it always returns from the head of the queue.

mappings/net/minecraft/util/ArraySet.mapping
Pyrofab: `index` is a weird name imo. I would rather have consistency and use the same name as the equivalent List method.

mappings/net/minecraft/util/ArraySet.mapping
Sollace: ~~Alternatively: We could call it `head()`, `getHead()`, or `first()`.~~

mappings/net/minecraft/util/OrderedSet.mapping
```diff
@@ -0,0 +1,27 @@
+CLASS net/minecraft/class_4706 net/minecraft/util/ArraySet
```
liach:
```suggestion
CLASS net/minecraft/class_4706 net/minecraft/util/OrderedSet
```

liach:
Also, after your description, I feel this is more like a priority queue that does not allow duplicate elements. Is this the case?

mappings/net/minecraft/util/OrderedSet.mapping
```diff
@@ -0,0 +1,27 @@
+CLASS net/minecraft/class_4706 net/minecraft/util/OrderedSet
+	CLASS class_4707 Itr
+		FIELD field_21566 start I
+		FIELD field_21567 index I
+	FIELD field_21562 sortingFunction Ljava/util/Comparator;
+	FIELD field_21563 entries [Ljava/lang/Object;
+	FIELD field_21564 capacity I
```
liach, line 7: This is freaking size not capacity!
```suggestion
	FIELD field_21564 size I
```
Look at what the `size()` method returns!

mappings/net/minecraft/util/OrderedSet.mapping
```diff
@@ -0,0 +1,27 @@
+CLASS net/minecraft/class_4706 net/minecraft/util/OrderedSet
+	CLASS class_4707 Itr
+		FIELD field_21566 start I
```
liach, line 3:
```suggestion
		FIELD field_21566 next I
```
The next index of the element for the iterator's next call to return.

mappings/net/minecraft/util/OrderedSet.mapping
```diff
@@ -0,0 +1,27 @@
...
+		FIELD field_21567 index I
```
liach, line 4:
```suggestion
		FIELD field_21567 lastIndex I
```
Like the removal target for the iterator.

mappings/net/minecraft/util/OrderedSet.mapping
```diff
@@ -0,0 +1,27 @@
...
+		ARG 0 element
+		ARG 1 insertionPoint
+	METHOD method_23865 peek ()Ljava/lang/Object;
+	METHOD method_23866 toInsertionPoint (I)I
+		ARG 0 wouldBeIndex
+	METHOD method_23868 resizeArray (I)V
+		ARG 0 newCapacity
+	METHOD method_23869 index (Ljava/lang/Object;)I
+		ARG 0 element
+	METHOD method_23870 spliceRemove (I)V
```
liach, line 24:
```suggestion
	METHOD method_23870 remove (I)V
```
This is like an `O(N)` simple removal logic by copying tail elements to cover the last element; `splice` just make things complex.

mappings/net/minecraft/util/OrderedSet.mapping
```diff
@@ -0,0 +1,27 @@
...
+	FIELD field_21562 sortingFunction Ljava/util/Comparator;
+	FIELD field_21563 entries [Ljava/lang/Object;
+	FIELD field_21564 capacity I
+	METHOD <init> (ILjava/util/Comparator;)V
+		ARG 0 initialCapacity
+		ARG 1 sortingFunction
+	METHOD method_23859 create (I)Lnet/minecraft/class_4706;
+		ARG 0 initialCapacity
+	METHOD method_23862 addIfAbsent (Ljava/lang/Object;)Ljava/lang/Object;
+	METHOD method_23863 spliceAdd (Ljava/lang/Object;I)V
```
liach, line 14:
```suggestion
	METHOD method_23863 add (Ljava/lang/Object;I)V
```
Same as remove argument below.

mappings/net/minecraft/util/OrderedSet.mapping
```diff
@@ -0,0 +1,27 @@
...
+		ARG 0 element
+		ARG 1 insertionPoint
+	METHOD method_23865 peek ()Ljava/lang/Object;
```
liach, line 17:
```suggestion
	METHOD method_23865 first ()Ljava/lang/Object;
```
Similar to https://docs.oracle.com/javase/8/docs/api/java/util/SortedSet.html#first--

mappings/net/minecraft/util/OrderedSet.mapping
```diff
@@ -0,0 +1,27 @@
...
+	METHOD method_23866 toInsertionPoint (I)I
+		ARG 0 wouldBeIndex
```
liach, line 19:
```suggestion
		ARG 0 binarySearchResult
```
or `indexResult`

mappings/net/minecraft/util/OrderedSet.mapping
```diff
@@ -0,0 +1,27 @@
+CLASS net/minecraft/class_4706 net/minecraft/util/OrderedSet
+	CLASS class_4707 Itr
+		FIELD field_21566 start I
+		FIELD field_21567 index I
+	FIELD field_21562 sortingFunction Ljava/util/Comparator;
```
liach, line 5: This is like an internal logic for the array-backed set so that lookup performance can be O(N log(N)). Given there is a `first` method, I suggest calling this class a `PrioritySet` and this field a `comparator`.
```suggestion
	FIELD field_21562 comparator Ljava/util/Comparator;
```

liach (changes requested):
Seriously, my eyes are bleeding after looking at the actual code in mod dev env and comparing it with your mappings! Too many faults. :angry:

liach:
I recommend #973 over this pull request.

PR #973, opened 2019-11-21
SortedArraySet mappings

Runemoro:
Replacement for https://github.com/FabricMC/yarn/pull/970 which currently has many problems (closes https://github.com/FabricMC/yarn/pull/970).

Here are the main differences:
 - `OrderedSet` -> `SortedArraySet`. Ordered just means that it has a certain order (for example a `LinkHashSet` is ordered in the insertion order). Sorted makes it clear that the elements are automatically being sorted as they are added using the given comparator. (*Edit:* As @liach mentioned, there's also `java.util.SortedSet`. Naming this one `SortedArraySet` since it acts exactly like an array-based implementation of `java.util.SortedSet`, but Mojang seems to have not bothered implementing all the methods).
 - `Itr` -> `SetIterator`. Matches JDK convention of naming interators `<type>Iterator`: `ArrayList.ListIterator`, `Map.KeyIterator`, etc.
 - `start` -> `lastIndex`. This is the index of the previously returned element (used by `remove`)
 - `index` -> `nextIndex`. This is the index of the next element to be returned
 - `sortingFunction` -> `comparator`. A "sorting function" could be understood to mean a function that actually does the sorting of the array. Here, it's just a `java.util.Comparator` comparing elements pairwise.
 - `addIfAbsent` -> `addAndGet`. The other `add` method also adds only if absent (as do all sets!). Calling this method is the same as doing an `add` followed by a `get`.
 - `spliceAdd` -> `add`. It's just adding an element at a given position...
 - `getDirect` -> `get`. Not sure what's "direct" about this, especially since there is no other "indirect" `get` method.

### DIFF
--- a/mappings/net/minecraft/util/SortedSet.mapping
+++ b/mappings/net/minecraft/util/SortedSet.mapping
@@ -17,7 +17,7 @@ CLASS net/minecraft/class_4706 net/minecraft/util/SortedSet
 		ARG 2 index
 	METHOD method_23864 cast ([Ljava/lang/Object;)[Ljava/lang/Object;
 		ARG 0 array
-	METHOD method_23865 getFirst ()Ljava/lang/Object;
+	METHOD method_23865 first ()Ljava/lang/Object;
 	METHOD method_23866 insertionPoint (I)I
 		ARG 0 binarySearchResult
 	METHOD method_23868 ensureCapacity (I)V
--- a/mappings/net/minecraft/util/SortedSet.mapping
+++ b/mappings/net/minecraft/util/SortedSet.mapping
@@ -1,0 +1,30 @@
+CLASS net/minecraft/class_4706 net/minecraft/util/SortedSet
+	CLASS class_4707 SetIterator
+		FIELD field_21566 nextIndex I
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/util/SortedSet.mapping
```diff
@@ -0,0 +1,30 @@
+CLASS net/minecraft/class_4706 net/minecraft/util/SortedSet
```
liach: This name is to similar to that of `java.util.SortedSet`. :thinking:

liach (approved):

mappings/net/minecraft/util/SortedSet.mapping
Runemoro: Didn't know that existed... It does act like `java.util.SortedSet`, except it has fewer methods. A possible name could be `SortedArraySet`, though that might make people think it's an implementation of `SortedSet`.

Any other suggestions @liach ?

liach: looks great to me.

Juuxel (approved):

PR #977, opened 2019-11-22
Generate mappings for specialized methods

### LINKED ISSUE #1: Flaw in generic mapping
Synthetic bridges methods, most commonly seen from implementing generic supertypes, are not logically mapped to be deobfuscated. Taking `azb$a` => `class_1867$class_1868` => `ShapelessRecipe$Serializer`  (which extends`ayy` => `class_1862` => `RecipeSerializer`) as an example:
- We start with explicit bridge `a(Lit;Layw;)V` for `a(Lit;Lazb;)V`
- Intermediary skips mapping the bridge (as it overrides `class_1862#method_8124`) and maps the actual method to `method_8143`
- Mapping `method_8124` (as Yarn [does](https://github.com/FabricMC/yarn/blob/18w50a/mappings/net/minecraft/recipe/RecipeSerializer.mapping#L3)) carries over to the mapping ([here](https://github.com/FabricMC/yarn/blob/18w50a/mappings/net/minecraft/recipe/crafting/ShapelessRecipe.mapping#L4))
- The actual method remains unmapped and is left as `method_8143`

This process is masked from decompiling hiding bridge methods by default, but the same effect is visibly present from other methods:
- We start with implicit (ie only synthetic) bridge `b(Lqc;Lit;)Layw;` for `a(Lqc;Lit;)Lazb;`
- Intermediary skips mapping the bridge (`method_8122`) and maps the actual method to `method_8141`
- As before the [naming](https://github.com/FabricMC/yarn/blob/18w50a/mappings/net/minecraft/recipe/RecipeSerializer.mapping#L7) is carried [forward](https://github.com/FabricMC/yarn/blob/18w50a/mappings/net/minecraft/recipe/crafting/ShapelessRecipe.mapping#L8)
- The actual method remains unmapped, resulting in a mapped `read` bouncing to an unmapped `method_8141`

Practically speaking as it stands nothing appears to break from this (after all, Mojang uses different names of `a` and `b`), but hiding the synthetic methods as decompilers often do results in classes appearing to not consistently implement interfaces.
This also leaves the door open for the implementing method name to differ from the interface it is implementing, purely as they have different Intermediary names and nothing forces them to be the same. Enigma will happily let you do this as to it they are different methods. Whilst again this isn't technically wrong from the JVM's point of view, you could never write Java code that would naturally compile like that (as well as the decompile having no chance of recompiling).

If the actual methods that are bridged to shared the same Intermediary name as the bridges (and thus the original interface) this would go away as the whole system would change together. Doesn't fix Yarn's problems from mapping to Notch directly, but that is arguably an Enigma problem.

### LINKED ISSUE #256: CoralFeature.generate not getting mapped in output
CoralFeature.generate not getting mapped in build.

yarn-18w48a-tiny:

`CLASS	btf	net/minecraft/world/gen/feature/CoralFeature	net/minecraft/class_2978`
`METHOD	btf	(Lbat;Lbqa;Ljava/util/Random;Let;Lbvu;)Z	a	method_12865	method_12865`

https://github.com/FabricMC/yarn/blob/18w49a/mappings/net/minecraft/world/gen/feature/CoralFeature.mapping#L1-L4

Runemoro:
- Closes https://github.com/FabricMC/yarn/issues/256
 - Closes https://github.com/FabricMC/intermediary/issues/1

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
 		}
 	}
 	dependencies {
-		classpath "cuchaz:enigma:0.14.2.143"
+		classpath "cuchaz:enigma:0.14.2.144"
 		classpath "net.fabricmc:stitch:0.4.0.67"
 		classpath "commons-io:commons-io:2.6"
 		classpath "com.google.guava:guava:28.0-jre"
@@ -49,7 +49,7 @@ configurations {
 
 dependencies {
 	enigmaRuntime "net.fabricmc:stitch:0.4.0.67"
-	enigmaRuntime "cuchaz:enigma:0.14.2.143"
+	enigmaRuntime "cuchaz:enigma:0.14.2.144"
 }
 
 def setupGroup = "jar setup"
@@ -75,6 +75,7 @@ import com.google.common.collect.Iterables
 import cuchaz.enigma.command.CheckMappingsCommand
 import cuchaz.enigma.command.ComposeMappingsCommand
 import cuchaz.enigma.command.ConvertMappingsCommand
+import cuchaz.enigma.command.MapSpecializedMethodsCommand
 import groovy.io.FileType
 import groovy.json.JsonSlurper
 import net.fabricmc.stitch.commands.CommandMergeTiny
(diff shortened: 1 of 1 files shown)


Juuxel (approved):

liach (approved):
The built binaries look good to me.

liach:
Should be merged once @modmuss50 confirms the built binaries are correct.

PR #975, opened 2019-11-22
Map some mostly world gen stuff

valoeghese:

PR #980, opened 2019-11-23
Some stuff for 1.15 pre 1 and pre2

liach:
wood type or sign type? it is only used by signs.

### DIFF
--- a/mappings/net/minecraft/client/render/AtlasTextureId.mapping
+++ b/mappings/net/minecraft/client/render/AtlasTextureId.mapping
@@ -1,0 +1,15 @@
+CLASS net/minecraft/class_4730 net/minecraft/client/render/AtlasTextureId
+	FIELD field_21769 atlas Lnet/minecraft/class_2960;
+	FIELD field_21770 texture Lnet/minecraft/class_2960;
+	FIELD field_21771 layer Lnet/minecraft/class_1921;
+	METHOD <init> (Lnet/minecraft/class_2960;Lnet/minecraft/class_2960;)V
+		ARG 1 atlas
+		ARG 2 texture
+	METHOD method_24144 getAtlas ()Lnet/minecraft/class_2960;
+	METHOD method_24145 getVertexConsumer (Lnet/minecraft/class_4597;Ljava/util/function/Function;)Lnet/minecraft/class_4588;
+		ARG 1 vertexConsumers
+		ARG 2 layerFactory
+	METHOD method_24146 getRenderLayer (Ljava/util/function/Function;)Lnet/minecraft/class_1921;
+		ARG 1 layerFactory
+	METHOD method_24147 getTexture ()Lnet/minecraft/class_2960;
+	METHOD method_24148 getSprite ()Lnet/minecraft/class_1058;
--- a/mappings/net/minecraft/client/render/VertexConsumerHelper.mapping
+++ b/mappings/net/minecraft/client/render/VertexConsumerHelper.mapping
@@ -1,0 +1,7 @@
+CLASS net/minecraft/class_4720 net/minecraft/client/render/VertexConsumerHelper
+	CLASS class_4589 DelegatingVertexConsumer
+		FIELD field_21685 one Lnet/minecraft/class_4588;
+		FIELD field_21686 two Lnet/minecraft/class_4588;
+	METHOD method_24037 delegate (Lnet/minecraft/class_4588;Lnet/minecraft/class_4588;)Lnet/minecraft/class_4588;
+		ARG 0 one
+		ARG 1 two
--- a/mappings/net/minecraft/client/render/VertexConsumerHelper.mapping
+++ b/mappings/net/minecraft/client/render/VertexConsumerHelper.mapping
@@ -1,7 +1,7 @@
 CLASS net/minecraft/class_4720 net/minecraft/client/render/VertexConsumerHelper
-	CLASS class_4589 DelegatingVertexConsumer
+	CLASS class_4589 Union
 		FIELD field_21685 one Lnet/minecraft/class_4588;
 		FIELD field_21686 two Lnet/minecraft/class_4588;
-	METHOD method_24037 delegate (Lnet/minecraft/class_4588;Lnet/minecraft/class_4588;)Lnet/minecraft/class_4588;
+	METHOD method_24037 union (Lnet/minecraft/class_4588;Lnet/minecraft/class_4588;)Lnet/minecraft/class_4588;
 		ARG 0 one
 		ARG 1 two
--- a/mappings/net/minecraft/client/render/VertexConsumers.mapping
+++ b/mappings/net/minecraft/client/render/VertexConsumers.mapping
@@ -1,7 +1,10 @@
-CLASS net/minecraft/class_4720 net/minecraft/client/render/VertexConsumerHelper
+CLASS net/minecraft/class_4720 net/minecraft/client/render/VertexConsumers
 	CLASS class_4589 Union
 		FIELD field_21685 one Lnet/minecraft/class_4588;
 		FIELD field_21686 two Lnet/minecraft/class_4588;
+		METHOD <init> (Lnet/minecraft/class_4588;Lnet/minecraft/class_4588;)V
+			ARG 1 one
+			ARG 2 two
 	METHOD method_24037 union (Lnet/minecraft/class_4588;Lnet/minecraft/class_4588;)Lnet/minecraft/class_4588;
 		ARG 0 one
 		ARG 1 two
--- a/mappings/net/minecraft/client/render/VertexConsumers.mapping
+++ b/mappings/net/minecraft/client/render/VertexConsumers.mapping
@@ -1,10 +1,10 @@
 CLASS net/minecraft/class_4720 net/minecraft/client/render/VertexConsumers
-	CLASS class_4589 Union
-		FIELD field_21685 one Lnet/minecraft/class_4588;
-		FIELD field_21686 two Lnet/minecraft/class_4588;
+	CLASS class_4589 Dual
+		FIELD field_21685 first Lnet/minecraft/class_4588;
+		FIELD field_21686 second Lnet/minecraft/class_4588;
 		METHOD <init> (Lnet/minecraft/class_4588;Lnet/minecraft/class_4588;)V
-			ARG 1 one
-			ARG 2 two
-	METHOD method_24037 union (Lnet/minecraft/class_4588;Lnet/minecraft/class_4588;)Lnet/minecraft/class_4588;
+			ARG 1 first
+			ARG 2 second
+	METHOD method_24037 dual (Lnet/minecraft/class_4588;Lnet/minecraft/class_4588;)Lnet/minecraft/class_4588;
 		ARG 0 one
 		ARG 1 two
--- a/mappings/net/minecraft/client/render/VertexConsumers.mapping
+++ b/mappings/net/minecraft/client/render/VertexConsumers.mapping
@@ -6,5 +6,5 @@ CLASS net/minecraft/class_4720 net/minecraft/client/render/VertexConsumers
 			ARG 1 first
 			ARG 2 second
 	METHOD method_24037 dual (Lnet/minecraft/class_4588;Lnet/minecraft/class_4588;)Lnet/minecraft/class_4588;
-		ARG 0 one
-		ARG 1 two
+		ARG 0 first
+		ARG 1 second
--- a/mappings/net/minecraft/client/texture/SpriteAtlasTexture.mapping
+++ b/mappings/net/minecraft/client/texture/SpriteAtlasTexture.mapping
@@ -4,11 +4,10 @@ CLASS net/minecraft/class_1059 net/minecraft/client/texture/SpriteAtlasTexture
 		FIELD field_17901 width I
 		FIELD field_17902 height I
 		FIELD field_17903 sprites Ljava/util/List;
-		METHOD <init> (Ljava/util/Set;IILjava/util/List;)V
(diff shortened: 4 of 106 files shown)


mappings/net/minecraft/client/render/AtlasTextureId.mapping
```diff
@@ -0,0 +1,15 @@
+CLASS net/minecraft/class_4730 net/minecraft/client/render/AtlasTextureId
```
Juuxel: Just a note: Mojang seems to call this `Material` based on the toString.

Juuxel (approved):

mappings/net/minecraft/client/render/AtlasTextureId.mapping
liach: however, from usage, it is more of an id for a sprite (as seen in textured render layers class)

grondag: AtlasTextureId is better than material. However, name should probably also refer to Sprite. The atlas texture includes all sprites, and this identifies a specific sprite within a specific atlas texture.

liach: Renamed it to `SpriteIdentifer` given `Sprite` is always in an atlas in minecraft.

Runemoro (approved):

grondag:
I think putting Sign in the name is better because, as you note, it is only used by signs.

liach:
What if mojang merged this with `BoatEntity.Type` in the future?

mappings/net/minecraft/client/render/VertexConsumerHelper.mapping
```diff
@@ -0,0 +1,7 @@
+CLASS net/minecraft/class_4720 net/minecraft/client/render/VertexConsumerHelper
```
grondag: Think this name needs to indicate that this consumer routes the output to two different consumers.  `Helper` doesn't convey that.

I called it `DualVertexConsumer`, but I dunno if that is best.  Maybe `DualOutputVertexConsumer` or `DuplicatingVertexConsumer`?

liach: This is the utility class. We use it like `VertexConsumerHelper.union()`. You don't say `Arrays.asList()`'s `Arrays` need to be `ArrayToList` so that `ArrayToList.asList()` makes sense to you

grondag: It only does one thing and the class that handles the implementation is package-private.   `union` is also a bad name for the method, in that overall the names give more or less no indication of what this thing does or how it is used.

grondag:
> What if mojang merged this with `BoatEntity.Type` in the future?

What if they add a list of their favorite soccer players?  It's entirely possible Mojang could muddy this up but then it would have a significantly different, broader function and should be renamed.  

Absent strong indication of Mojang's plans, classes should be named based on what they currently do, IMO.

mappings/net/minecraft/client/render/VertexConsumerHelper.mapping
liach: from the param type of 2 vertex consumers and the return type of one vertex consumer and `union` name, you can clearly see that these two vertex consumers are unioned together. there is no sign of "duplicating" as the constuctor explicitly prevents duplicate consumers from passed into the argument. dual output maybe, but that is not clear either.

also, in #975, this class is also mapped as `net/minecraft/util/WoodType`. Feel free to vote on that.

grondag: We clearly don't agree. But doesn't matter much. I can continue to use my own mappings.

liach: Yep, that is the advantage loom offered us over forgegradle (more flexible mappings!)

Pyrofab: I find `DuplicatingVertexConsumer` to be the clearest name proposed here, and if it is not used for the helper class I'd want to see it used for the implementation. And I think `VertexConsumerHelper#union` should be renamed as well, its behaviour is far from obvious.

Runemoro:
Sign type is better. What if a mod wants to add more types, like metal signs?

mappings/net/minecraft/client/render/VertexConsumerHelper.mapping
liach: > the class that handles the implementation is package-private

At first sight I thought `VertexConsumerHelper` is package private or the utility method is; turns out only the impl class is, which does not matter. And this package-private class is not exposed in the return type of the static method either, which makes thing safe.

Mojang should have declared it private because that class should be. It is an impl detail and should not be leaked in any way, like `java.util.Arrays.ArrayList` is private as well.

Also this is not "duplicating" @Pyrofab: This is just same vertex data that should be transformed differently by a regular consumer and a color outline consumer or something like that from my previous experience, just like stream concat's result call the close of both base stream when closed isn't duplicating streams.

mappings/net/minecraft/client/texture/SpriteAtlasTexture.mapping
```diff
@@ -11,22 +11,32 @@ CLASS net/minecraft/class_1059 net/minecraft/client/texture/SpriteAtlasTexture
...
 		ARG 1 resourceManager
 		ARG 2 stitcher
 	METHOD method_18163 stitch (Lnet/minecraft/class_3300;Ljava/util/stream/Stream;Lnet/minecraft/class_3695;I)Lnet/minecraft/class_1059$class_4007;
 	METHOD method_18164 loadSprites (Lnet/minecraft/class_3300;Ljava/util/Set;)Ljava/util/Collection;
+	METHOD method_24106 getId ()Lnet/minecraft/class_2960;
 	METHOD method_4601 clear ()V
 	METHOD method_4603 getTexturePath (Lnet/minecraft/class_2960;)Lnet/minecraft/class_2960;
+		ARG 1 id
 	METHOD method_4604 loadSprite (Lnet/minecraft/class_3300;Lnet/minecraft/class_1058$class_4727;IIII)Lnet/minecraft/class_1058;
 		ARG 1 container
```
Juuxel, line 32: This could be changed to `resourceManager` as well.

liach:
```java
    default void quad(MatrixStack.Entry matrixEntry, BakedQuad quad, float[] brightnesses, float red, float green, float blue, int[] lights, int overlay, boolean boolean10) {
        int[] js = quad.getVertexData();
        Vec3i vec3i12 = quad.getFace().getVector();
        Vector3f vector3f13 = new Vector3f((float)vec3i12.getX(), (float)vec3i12.getY(), (float)vec3i12.getZ());
        Matrix4f matrix4f14 = matrixEntry.getModel();
        vector3f13.transform(matrixEntry.getNormal());
        int integer15 = 8;
        int integer16 = js.length / 8;
        try (MemoryStack memoryStack = MemoryStack.stackPush()) {
            ByteBuffer byteBuffer19 = memoryStack.malloc(VertexFormats.POSITION_COLOR_TEXTURE_LIGHT_NORMAL.getVertexSize());
            IntBuffer intBuffer20 = byteBuffer19.asIntBuffer();
            for (int l = 0; l < integer16; ++l) {
                intBuffer20.clear();
                intBuffer20.put(js, l * 8, 8);
                float float22 = byteBuffer19.getFloat(0);
                float float23 = byteBuffer19.getFloat(4);
                float float24 = byteBuffer19.getFloat(8);
                float float25;
                float float26;
                float float27;
                if (boolean10) { // <---- What is this?
                    float float28 = (byteBuffer19.get(12) & 0xFF) / 255.0f;
                    float float29 = (byteBuffer19.get(13) & 0xFF) / 255.0f;
                    float float30 = (byteBuffer19.get(14) & 0xFF) / 255.0f;
                    float float25 = float28 * brightnesses[l] * red;
                    float float26 = float29 * brightnesses[l] * green;
                    float float27 = float30 * brightnesses[l] * blue;
                }
                else {
                    float25 = brightnesses[l] * red;
                    float26 = brightnesses[l] * green;
                    float27 = brightnesses[l] * blue;
                }
                int integer28 = lights[l];
                float float29 = byteBuffer19.getFloat(16);
                float float30 = byteBuffer19.getFloat(20);
                Vector4f vector4f31 = new Vector4f(float22, float23, float24, 1.0f);
                vector4f31.transform(matrix4f14);
                this.elements(vector4f31.getX(), vector4f31.getY(), vector4f31.getZ(), float25, float26, float27, 1.0f, float29, float30, overlay, integer28, vector3f13.getX(), vector3f13.getY(), vector3f13.getZ());
            }
        }
    }
```
@runemoro can you help name `boolean10` here?

Sollace:
@liach `boolean10`
True ===> butter contains colour data, add it and render in colour
False ===> Just use the light value, render in greyscale (or close to it)!

liach:
ah, so like `useQuadColorData`

mappings/net/minecraft/client/render/VertexConsumerHelper.mapping
Pyrofab: I find your `Stream#concat` example ill-fitting. The main characteristic of a concatenated stream is to have the values of one stream coming after (concatenated with) the values of another. The main characteristic of `class_4589` is to send the vertex data to 2 delegate consumers. In doing so, it is effectively duplicating the data, or doing 2 operations at once (hence dual). Union is somewhat correct, but I have to look at the implementation to know what it actually means.
I was recommending renaming the implementation class because most modders look at a method's implementation before using it - `union() {return new Union()}` gives no additional information, whereas `union() {return new DuplicatingVertexConsumer()}` would make me understand very quickly what it is about.

liach: the impl class is already called union. comment on the new util classn i renamed it to vertex consumers.

Sollace: This isn't `concat` or `union`. This is very clearly mirroring output between render layers.

Might have to be something like `Layers.combine(layerOne, layerTwo);`

Also note that it can be chained:

```RenderLayer combined = Layers.combine(Layers.combine(layerOne, layerTwo), layerThree))```

**Review by modmuss50 — APPROVED** [review]

File: mappings/net/minecraft/client/render/VertexConsumerHelper.mapping (hunk 2 above)
Comment (liach) [reply mappings/net/minecraft/client/render/VertexConsumerHelper.mapping]: did you even read this @Sollace , this is freaking vertex consumers not freaking render layer util class! look at the stuff before you speak!

Comment (Pyrofab) [reply mappings/net/minecraft/client/render/VertexConsumerHelper.mapping]: I would rather not have unnecessary aggressive behaviour here. What Sollace said is approximately right, the combining vertex consumer is used to mirror output between 2 layers.

File: mappings/net/minecraft/client/render/VertexConsumers.mapping [hunk 4]
```diff
@@ -0,0 +1,10 @@
+CLASS net/minecraft/class_4720 net/minecraft/client/render/VertexConsumers
+	CLASS class_4589 Union
+		FIELD field_21685 one Lnet/minecraft/class_4588;
+		FIELD field_21686 two Lnet/minecraft/class_4588;
+		METHOD <init> (Lnet/minecraft/class_4588;Lnet/minecraft/class_4588;)V
+			ARG 1 one
```
natanfudge: Should be `first` and `second`

mappings/net/minecraft/client/render/VertexConsumers.mapping
```diff
@@ -0,0 +1,10 @@
...
+			ARG 2 two
+	METHOD method_24037 union (Lnet/minecraft/class_4588;Lnet/minecraft/class_4588;)Lnet/minecraft/class_4588;
+		ARG 0 one
```
natanfudge: same

natanfudge (approved):

liach:
@Sollace @Pyrofab the dual/union thing has been addressed. dual is the shortest, so dual

mappings/net/minecraft/client/texture/SpriteAtlasTexture.mapping
```diff
@@ -11,23 +11,41 @@ CLASS net/minecraft/class_1059 net/minecraft/client/texture/SpriteAtlasTexture
...
 	METHOD method_18164 loadSprites (Lnet/minecraft/class_3300;Ljava/util/Set;)Ljava/util/Collection;
+		ARG 1 resourceManager
+		ARG 2 ids
+	METHOD method_24106 getId ()Lnet/minecraft/class_2960;
 	METHOD method_4601 clear ()V
 	METHOD method_4603 getTexturePath (Lnet/minecraft/class_2960;)Lnet/minecraft/class_2960;
+		ARG 1 id
 	METHOD method_4604 loadSprite (Lnet/minecraft/class_3300;Lnet/minecraft/class_1058$class_4727;IIII)Lnet/minecraft/class_1058;
-		ARG 1 container
+		ARG 1 resoruceManager
```
Juuxel: Typo here.

Juuxel (changes requested):

mappings/net/minecraft/client/texture/SpriteAtlasTexture.mapping
liach:
```suggestion
		ARG 1 resourceManager
```

mappings/net/minecraft/client/render/VertexConsumers.mapping
liach: done

natanfudge: Looks like it's still `one` and `two` in the parameters of `dual`

Juuxel (approved):

PR #983, opened 2019-11-26
more world gen mappings, fix some names, refactor some biome source related stuff

valoeghese:


mappings/net/minecraft/world/biome/DefaultBiomeFeatures.mapping
```diff
@@ -64,10 +81,16 @@ CLASS net/minecraft/class_3864 net/minecraft/world/biome/DefaultBiomeFeatures
 	FIELD field_21180 RED_MUSHROOM_BLOCK Lnet/minecraft/class_2680;
 	FIELD field_21181 BROWN_MUSHROOM_BLOCK Lnet/minecraft/class_2680;
 	FIELD field_21182 MUSHROOM_BLOCK Lnet/minecraft/class_2680;
+	FIELD field_21183 JUNGLE_TREE_2_CONFIG Lnet/minecraft/class_4640;
```
coderbot16: more descriptive name plz? is this for large jungle trees or something?

mappings/net/minecraft/world/gen/feature/RandomPatchFeatureConfig.mapping
```diff
@@ -0,0 +1,14 @@
+CLASS net/minecraft/class_4638 net/minecraft/world/gen/feature/RandomPatchFeatureConfig
+	CLASS class_4639 Builder
+	FIELD field_21237 stateProvider Lnet/minecraft/class_4651;
+	FIELD field_21238 blockPlacer Lnet/minecraft/class_4629;
+	FIELD field_21239 whitelist Ljava/util/Set;
+	FIELD field_21240 blacklist Ljava/util/Set;
+	FIELD field_21241 tries I
+	FIELD field_21242 xSpread I
```
coderbot16: convention appears to be offsetX, not xOffset, same for the other two

mappings/net/minecraft/world/biome/BiomeAccess.mapping
```diff
@@ -1,6 +1,6 @@
 CLASS net/minecraft/class_4543 net/minecraft/world/biome/BiomeAccess
 	CLASS class_4544 Storage
-		METHOD method_16359 getStoredBiome (III)Lnet/minecraft/class_1959;
+		METHOD method_16359 getBiome (III)Lnet/minecraft/class_1959;
```
liach: May be confusing with `world.getBiome` as this xyz is not block xyz, neither are they chunk xyz

mappings/net/minecraft/world/gen/feature/RandomPatchFeatureConfig.mapping
liach: Yeah, favor convention over mojang names

liach (changes requested):

mappings/net/minecraft/world/biome/BiomeAccess.mapping
valoeghese: It's basically the same as the 1.14 `getBiome` method but 3d and not using block xyz

liach: we have another one that use blockpos, people will get confused man

mappings/net/minecraft/world/biome/DefaultBiomeFeatures.mapping
valoeghese: it has no cocoa beans iirc

mappings/net/minecraft/world/biome/BiomeAccess.mapping
valoeghese: but why "get **Stored** biome?"

valoeghese: maybe sampleBiome would be better

liach: because the actual getbiome method goes through the access strategy to get the correct biome for the block

and this 3 int method is like getting the raw biome stored, so it is not sampling either but rather getting the stored biome in the 3d array with the 3d index

valoeghese: Oh! so this is kinda like the getBiomeForNoiseGen method from 1.14 but slightly different

liach: actually if someone can find a way to optimize the get biome call through the biome strategy, they may be able to improve mc performance as currently many mob ai and block logic call getbiome and everytime the biome is get the game goes through all the computations within the biome access types

mappings/net/minecraft/world/biome/DefaultBiomeFeatures.mapping
Runemoro: I'd name the other one `JUNGLE_TREE_1_CONFIG`.

valoeghese: it's the sapling tree. renamed

liach (approved):

Runemoro (approved):

PR #994, opened 2019-12-03
1.15 pre4 1

liach:
Some new stuff for pre4

### DIFF
--- a/mappings/net/minecraft/util/math/Quaternion.mapping
+++ b/mappings/net/minecraft/util/math/Quaternion.mapping
@@ -1,5 +1,10 @@
 CLASS net/minecraft/class_1158 net/minecraft/util/math/Quaternion
+	COMMENT Quaternion is a type of number used to represent a counterclockwise
+	COMMENT rotation around an axis.
 	FIELD field_21493 IDENTITY Lnet/minecraft/class_1158;
+	FIELD field_21582 b F
+	FIELD field_21583 c F
+	FIELD field_21584 d F
 	FIELD field_21585 a F
 	METHOD <init> (FFFF)V
 		ARG 1 b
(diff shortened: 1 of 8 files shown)


coderbot16 (approved):

Juuxel (approved):

mappings/net/minecraft/util/math/Quaternion.mapping
```diff
@@ -1,5 +1,10 @@
 CLASS net/minecraft/class_1158 net/minecraft/util/math/Quaternion
+	COMMENT Quaternion is a type of number used to represent a counterclockwise
+	COMMENT rotation around an axis.
```
modmuss50: Not seeing these show up in the tiny file. and why are there 2?

modmuss50 (changes requested):

mappings/net/minecraft/util/math/Quaternion.mapping
natanfudge: You mean you tried building it yourself?

modmuss50: Yes

natanfudge: uhh i'm gonna need to take a look at it then

Runemoro: I'm not sure what you mean by "counterclockwise rotation", a quaternion isn't stored as an axis plus an angle. The counterclockwise angle of rotation is the angle of the vector (sqrt(b^2 + c^2 + d^2), a).

liach: There are two because there are new lines. I intentionally designed it to be before the commented element and swap on new lines so that it will be easier for pr reviews.

Rune: the counterclockwise rotation means that the constructor that takes a vec3 and an angle is like a rotation counterclockwise around that vec3 axis for that much angle.

liach: Removed for now. Since modmuss points out jd still has some issues, we can delay that a bit and keep them in github issues for now.

modmuss50 (approved):

PR #1008, opened 2019-12-12
Model transformation mode

Sollace:
As Liach suggested, here's a PR renaming `ModelTransformation.Type` to `ModelTransformation.Mode`.

As a preliminary change I've also renamed all parameters of that type to `renderMode` because it makes sense, though I'm open to calling it just `mode` if people take issue with the `render` part.


Sollace:
Updated to 1.15.1-pre1

liach (approved):
Update to merge into 1.15.1.

Sollace:
@liach I see no conflicts, so updated.

modmuss50 (approved):

Issue #1014, opened 2019-12-16
Orphaned datafix (RemovePoiValidTagFix)


quat1024:
`RemovePoiValidTagFix` is in `net.minecraft.datafixer`, but all other Minecraft datafixer things are in `net.minecraft.datafixers` with an `s` at the end.

Pyrofab:
Shouldn't it be the opposite ? Package names are usually singular.

liach:
We indeed should move all of them to `datafixer` package instead. Yarn uses singular names unlike mojang which uses plural names (seen from their foss libraries)

Issue #895, opened 2019-10-04
PortalBlock -> NetherPortalBlock


Juuxel:
It's only used by `minecraft:nether_portal` and it's hardcoded to the nether.

Juuxel:
Looks like there's also some fields and methods in Entity that reference "portals" when they specifically check for nether portals or move between the nether and the overworld.

YanisBft:
@Juuxel not sure about renaming fields and methods referencing portals, as the nbt for portals never contains "nether"

YanisBft:
can be closed

PR #1015, opened 2019-12-16
Fixes for #895 & #1014

YanisBft:
#1014 was discussed a bit on discord, and we agreed that all packages should be singular


liach:
Mind turn `fixes` into `fix` and `schemas` into `schema` instead for subpackages?

i509VCB:
Have we thought of possibly making a subcategory within the data package so it would be `net.minecraft.data.fixer.NbtOps` for example.

liach (approved):
image
Looks great in enigma.

Juuxel:
Note that this isn't a full fix for #895 as it also relates to the relevant method/field names in `Entity`.

i509VCB: Datafixer stuff should not be moved to `net.minecraft.data` as that package only has data generation tools.

Juuxel (approved):

YanisBft:
@Juuxel, as I said on your issue, entity nbt is still named "Portal" and does not contain "NetherPortal", that's why I was not sure to rename these fields

modmuss50 (approved):

Issue #773, opened 2019-07-01
Consistency in client.network.packet


ExtraCrafTX:
In order to be consistent with the other packets:
* [`OpenContainerPacket`](https://github.com/FabricMC/yarn/blob/1.14.3/mappings/net/minecraft/client/network/packet/OpenContainerPacket.mapping) -> `OpenContainerS2CPacket`
* [`SetTradeOffersPacket`](https://github.com/FabricMC/yarn/blob/1.14.3/mappings/net/minecraft/client/network/packet/SetTradeOffersPacket.mapping) -> `SetTradeOffersS2CPacket`

liach:
Fixed in 4a2fe5bd1b2de0e8343f1ffb109717b77849ce78

PR #1016, opened 2019-12-17
Make packet names consistent

Earthcomputer:
It seems most of the work here had been done already (I remember there being more inconsistencies), but here's the rest of them!


liach:
There is an "OpenContainerPacket" to change as well. See #773

Earthcomputer:
Whoops, I must have missed that!

liach (approved):

modmuss50 (approved):

PR #1020, opened 2019-12-23
Better block state flattening mappings

Earthcomputer:
This PR adds and fixes mappings related to the flattening of block state IDs (upgrading worlds from 1.12). Some mappings were missing, some misleading, and some plain wrong.


mappings/net/minecraft/world/chunk/WorldChunk.mapping
```diff
@@ -66,4 +66,7 @@ CLASS net/minecraft/class_2818 net/minecraft/world/chunk/WorldChunk
 		ARG 2 box
 		ARG 3 list
 		ARG 4 predicate
+	METHOD method_20471 activateTickSchedulers (Lnet/minecraft/class_3218;)V
+		ARG 1 world
+	METHOD method_20530 deactivateTickSchedulers ()V
```
liach: enable/disable sounds more clear to me.

JamiesWhiteShirt (approved):

Prospector (approved):
Looks fine to me 👍

mappings/net/minecraft/world/chunk/UpgradeData.mapping
```diff
@@ -1,16 +1,31 @@
...
+	METHOD method_12351 applyAdjacentBlock (Lnet/minecraft/class_2680;Lnet/minecraft/class_2350;Lnet/minecraft/class_1936;Lnet/minecraft/class_2338;Lnet/minecraft/class_2338;)Lnet/minecraft/class_2680;
+		ARG 0 oldState
+		ARG 1 dir
+		ARG 2 world
+		ARG 3 currentPos
+		ARG 4 otherPos
+	METHOD method_12352 upgradeSide (Lnet/minecraft/class_2818;Lnet/minecraft/class_2355;)V
+		ARG 0 chunk
+		ARG 1 side
+	METHOD method_12356 postProcessChunk (Lnet/minecraft/class_2818;)V
```
liach:
```suggestion
	METHOD method_12356 postProcess (Lnet/minecraft/class_2818;)V
```

mappings/net/minecraft/datafixer/fix/ChunkPalettedStorageFix.mapping
```diff
@@ -41,6 +48,10 @@ CLASS net/minecraft/class_3582 net/minecraft/datafixer/fix/ChunkPalettedStorageF
 		METHOD method_15640 get (III)I
 			ARG 1 x
 			ARG 2 y
+		METHOD method_15641 isLowBits (I)Z
```
liach: like `index & (-index) == index`?

Earthcomputer: No, it's whether the value at the index is stored in the low bits of the value in the packed integer array: `(index & 0x1) == 0x0`

mappings/net/minecraft/world/chunk/UpgradeData.mapping
Earthcomputer: Honestly I don't really like the name post process, it doesn't really describe the purpose of the method, only when it's called. Perhaps `fixChunk` (or just `fix`) might be a better name?

Earthcomputer: I renamed it to `upgrade`

mappings/net/minecraft/datafixer/fix/ChunkPalettedStorageFix.mapping
liach: What you wrote is like `isEven` instead, or do you mean `(array[index] & 1) == 0` or `(value & (1 << index)) == 0`? Afaik low bit means something like [`Integer.lowestOneBit()`](https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#lowestOneBit-int-)

Earthcomputer: By "low bits" I mean that each byte in the backing byte array is split into 2 groups of 4 bits, the low bits and the high bits (perhaps "low nibble" would be clearer?). Just because the method returns true when the input is even doesn't mean that it should be called `isEven`, it's used along with the other method `getRawIndex`, which equally shouldn't be called `divideByTwo`. `getRawIndex` gets the index into the backing byte array your index refers to, and this method (currently called `isLowBits`) returns true when your index refers to the low nibble of the byte in that byte array, and false if it's the high nibble.

liach: I'd suggest calling it "low nibble", or just call this method `isLower` (you already have the param named index so it should be clear)

Earthcomputer: :+1: renamed it to `usesLowNibble`

modmuss50 (approved):

PR #1024, opened 2019-12-24
Small change to whitelist field in net.minecraft.server.MinecraftServer

oh-noey:
The field changed is only used to check whether whitelist enforcement is enabled, not if the whitelist itself is enabled. Going with the naming convention of stick with minecraft terms, I've changed it to `enforceWhitelist` however this creates a strangely named getter.


modmuss50 (approved):

liach:
Hmm, should have added some javadoc. I recall this is like `whitelist-enabled` in server properties

oh-noey:
It uses the flag `enforce-whitelist` from the server properties file, which determines if the server should kick invalid players when the whitelist is reloaded. I actually had no idea that javadocs were possible to add, I'll keep that in mind for the future!

PR #1028, opened 2019-12-27
Fix dodgy parameter names

Chocohead:
These have been wonky for 3 days over 14 months, might as well fix them now


modmuss50 (approved):
not sure how this happened.

PR #1032, opened 2020-01-01
Some potion-related names

howtonotwin:


liach (approved):

mappings/net/minecraft/entity/effect/StatusEffectInstance.mapping
```diff
@@ -8,28 +8,38 @@ CLASS net/minecraft/class_1293 net/minecraft/entity/effect/StatusEffectInstance
...
 		ARG 2 duration
+		ARG 3 amplifier
 	METHOD <init> (Lnet/minecraft/class_1291;IIZZ)V
-		ARG 1 effect
+		ARG 1 type
 		ARG 2 duration
 		ARG 3 amplifier
-		ARG 5 showParticles
+		ARG 4 ambient
+		ARG 5 visible
```
Juuxel, line 25: The field is called `showParticles`, so this should be `showParticles` as well (it's called `ShowParticles` in the NBT).

Juuxel (changes requested):

mappings/net/minecraft/entity/effect/StatusEffectInstance.mapping
liach: i think this is a vararg ctor in which this param is passed into both show particle and show icon in inv screen params of another constructor.

Juuxel: Ah, right. Should've looked in the ctor body.

Juuxel (approved):

YanisBft (approved):

modmuss50 (approved):

Issue #860, opened 2019-09-14
Players' StatHandler getters are inconsistent


Juuxel:
`ClientPlayerEntity` has `getStats` but `ServerPlayerEntity` has `getStatHandler`.

PR #858, opened 2019-09-13
Spawn restriction mappings

Juuxel:
Closes #830.

Some classes in this PR have long `canSpawn` names (~~`AnimalEntity.canAnimalSpawn`,~~ `MobEntity.canMobSpawn` and `MagmaCubeEntity.canMagmaCubeSpawn`) because Enigma doesn't let you give the same name to a subclass method with the same signature.

Edit: Looks like #854 maps `AnimalEntity.method_20663` as `isValidNaturalSpawn`. I think I'll adopt it to this PR as well.

PR #1040, opened 2020-01-07
Fixed #838, #860 and unnamed deserializers

### LINKED ISSUE #838: RecipeAlternatesWidget -> RecipeAlternativesWidget
Alternative seems to be a more common variant of the word. "Alternates" makes me think that it renders a recipe with alternating items (like those shown by JEI when several item variants are accepted).

YanisBft:
- #838 - Renamed all "alternate" to "alternative"
- #860 - Renamed `stats` field to `statHandler` and `getStats` method to `getStatHandler` in `ClientPlayerEntity` so they match their equivalent in `ServerPlayerEntity`

### DIFF
--- a/mappings/net/minecraft/world/gen/feature/MegaTreeFeatureConfig.mapping
+++ b/mappings/net/minecraft/world/gen/feature/MegaTreeFeatureConfig.mapping
@@ -4,7 +4,7 @@ CLASS net/minecraft/class_4636 net/minecraft/world/gen/feature/MegaTreeFeatureCo
 	METHOD <init> (Lnet/minecraft/class_4651;Lnet/minecraft/class_4651;Ljava/util/List;III)V
 		ARG 5 heightInterval
 		ARG 6 crownHeight
-	METHOD method_23408 deserialize2 (Lcom/mojang/datafixers/Dynamic;)Lnet/minecraft/class_4636;
+	METHOD method_23408 deserialize (Lcom/mojang/datafixers/Dynamic;)Lnet/minecraft/class_4636;
 	CLASS class_4637 Builder
 		FIELD field_21236 heightInterval I
 		FIELD field_21592 crownHeight I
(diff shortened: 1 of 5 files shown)


Juuxel (approved):

mappings/net/minecraft/world/gen/feature/MegaTreeFeatureConfig.mapping
```diff
@@ -4,6 +4,7 @@ CLASS net/minecraft/class_4636 net/minecraft/world/gen/feature/MegaTreeFeatureCo
 	METHOD <init> (Lnet/minecraft/class_4651;Lnet/minecraft/class_4651;Ljava/util/List;III)V
 		ARG 5 heightInterval
 		ARG 6 crownHeight
+	METHOD method_23408 deserialize2 (Lcom/mojang/datafixers/Dynamic;)Lnet/minecraft/class_4636;
```
liach: Hmm, is there a name conflict?

liach (changes requested):

mappings/net/minecraft/world/gen/feature/MegaTreeFeatureConfig.mapping
YanisBft: Yeah, the parent class already has a `deserialize` method. Tbh, I'm not a big fan of `deserialize2`, but it's already used in BranchedTreeFeatureConfig and I don't have any other idea

liach: if both methods are static, then it's an enigma bug.  try rename the method in a text editor and see if travis builds.

Juuxel: Yeah, I'm pretty sure it's an Enigma bug. I encountered it with #858 but forgot to make an issue.

YanisBft: Done, and travis builds

liach (approved):

modmuss50 (approved):

PR #1044, opened 2020-01-10
Map GameMenuScreen field and parameter to showMenu

haykam821:
This parameter/field is used to determine whether the pause menu with buttons (normal) or overlay (<kbd>F3</kbd>+<kbd>Escape</kbd>) will be shown.


liach:
Should this screen be renamed to a pause screen instead?

Prospector:
No, because it does not only show when the game is paused. See: multiplayer

Prospector (approved):

liach:
Well the minecraft wiki calls it a pause menu. It pauses gameplay input at least. https://minecraft.gamepedia.com/Pause_menu

Prospector:
The minecraft wiki is irrelevant. The text on the screen calls it a Game Menu

liach (approved):

PR #1047, opened 2020-01-11
InfoEnchantment -> EnchantmentState

Juuxel:
`InfoEnchantment` is not an enchantment called "Info".


liach:
this is more like an entry than a state. the entry is like an alt to Map.Entry Enchantment, Integer.

liach:
How about `EnchantmentInfo`? "info enchantment" probably means it is an enchantment in the tooltip info

Juuxel:
It's not really an enchantment info (just an enchantment together with the level). `EnchantmentEntry` could work, but it's a bit vague.

liach:
It's not a state either. A state sounds like a `Map<Enchantment, Integer>` that is bound to an item stack to me.

Juuxel:
Yeah, true. I think I'll open an issue for name suggestions.

PR #1048, opened 2020-01-11
Use auto-generated names for registries

Juuxel:
These names can be auto-generated from the `create()` calls that they are initialized with. Note that there are some changes compared to the previous Yarn names, like `BLOCK_ENTITY` -> `BLOCK_ENTITY_TYPE`.

Three of the original names are left in this PR. `REGISTRIES` can't be automapped, and `CONTAINER` and `STATUS_EFFECT` differ from the Yarn names of those concepts.


liach (approved):
Just make sure all these fields are covered in built binaries.

Juuxel:
I tested this in a devenv, so it should be fine. I can test building a mod to see if that works.

liach:
As long as dev env works, it should be fine. Actual mods are always in intermediary and this change in no way affects intermediary.

mappings/net/minecraft/util/registry/Registry.mapping
```diff
@@ -1,44 +1,7 @@
...
-	FIELD field_11160 ENCHANTMENT Lnet/minecraft/class_2378;
-	FIELD field_16643 CHUNK_STATUS Lnet/minecraft/class_2348;
-	FIELD field_16644 STRUCTURE_FEATURE Lnet/minecraft/class_2378;
-	FIELD field_16645 STRUCTURE_PIECE Lnet/minecraft/class_2378;
-	FIELD field_16792 RULE_TEST Lnet/minecraft/class_2378;
-	FIELD field_16793 STRUCTURE_POOL_ELEMENT Lnet/minecraft/class_2378;
-	FIELD field_16794 STRUCTURE_PROCESSOR Lnet/minecraft/class_2378;
-	FIELD field_17166 VILLAGER_TYPE Lnet/minecraft/class_2348;
-	FIELD field_17167 VILLAGER_PROFESSION Lnet/minecraft/class_2348;
-	FIELD field_17429 CONTAINER Lnet/minecraft/class_2378;
```
modmuss50, line 34: What is this one now called?

modmuss50 (changes requested):

mappings/net/minecraft/util/registry/Registry.mapping
Juuxel: IIRC this is `MENU`, and there was a third name change as well. I'll look it up in Enigma.

modmuss50: Yes, thats what I was worried about. For now keep it as container.

Juuxel: The full list of changed names is:
- `BLOCK_ENTITY` => `BLOCK_ENTITY_TYPE`
- `DIMENSION` => `DIMENSION_TYPE`
- `CONTAINER` => `MENU`
- `STATUS_EFFECT` => `MOB_EFFECT`

I don't think we should give other names to registered objects, as we don't rename blocks/items/entities either (there are also awful names like `area_effect_cloud` there).

If we do rename some registries, I think `CONTAINER` and `STATUS_EFFECT` can have their old names and the two others can have `_TYPE` at the end.

Pyrofab: I'd rather keep `STATUS_EFFECT`. Besides, are there that many registries that it's essential to auto-generate names for them ?

Juuxel: There are 42 registries. Also, I don't see any point in mapping things that can be mapped automatically.

Pyrofab: Well, consistency is the main issue. I think we should override generated names when they conflict with the established ones.

Juuxel: I added back the old names of `CONTAINER` and `STATUS_EFFECT`. (renaming those can be a discussion issue, anyway)

Pyrofab (approved):

modmuss50 (approved):

PR #1050, opened 2020-01-11
LivingEntity#removeStatusEffect clarification

Pyrofab:
While the previous names were technically correct, they imply the most straightforward method you can call is the former `removeStatusEffect`. However, it seems to be an internal method, which doesn't execute any cleanup/sync operation.
Most modders should call tryRemoveStatusEffect in every case.
This has caused confusion in a few mods already ([example 1](https://github.com/Ladysnake/Requiem/blob/master/src/main/java/ladysnake/requiem/mixin/possession/player/PossessorServerPlayerEntityMixin.java#L172), [example 2](https://github.com/CottonMC/PracticalWitchcraft/blob/master/src/main/java/io/github/cottonmc/witchcraft/util/WitchcraftNetworking.java), [example 3](https://github.com/Yoghurt4C/OmenaMaitoTee/blob/05e84d8bce6fe8600f24e9b659801aa8960511bc/src/main/java/mods/omenamaito/potions/ImmunizationEffect.java#L32-L36), [example 4](https://github.com/grondag/adversity/blob/7f41400d7e313b54026b8db39c4058acbbc94fc6/src/main/java/grondag/adversity/item/SalvationPotion.java#L67)).

### DIFF
--- a/mappings/net/minecraft/entity/LivingEntity.mapping
+++ b/mappings/net/minecraft/entity/LivingEntity.mapping
@@ -182,7 +182,7 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_6014 getItemUseTimeLeft ()I
 	METHOD method_6015 setAttacker (Lnet/minecraft/class_1309;)V
 		ARG 1 attacker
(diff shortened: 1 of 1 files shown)


liach (approved):

modmuss50 (approved):

grondag:
Nice catch

Issue #1046, opened 2020-01-10
net.minecraft.server.integrated.IntegratedServer.field_5524 -> paused


engiwengi:
I also think `clientPaused` works as it is describing its client. Otherwise `serverPaused` as the integrated server itself is paused.

Issue #998, opened 2019-12-07
BlockEntity.setWorld named wrong


Runemoro:
It also sets position:

```java
   public void setWorld(World world, BlockPos blockPos) {
      this.world = world;
      this.pos = blockPos.toImmutable();
   }
```

liach:
Looks functionally similar to `Entity setPosition`, both of which set up the (block) entity's initial state.

Runemoro:
I would prefer calling it (and also the one in `Entity`) `setLocation`. Position has a specific meaning and refers to a position in the world. We can use the more general word "location" to avoid using "position" for world + position.

Issue #964, opened 2019-11-17
Entity position setters


Juuxel:
There are multiple methods for setting entity positions, and they need a bit of improving.

- Confusingly, there are both `Entity.setPos()` and `Entity.setPosition()`.
  - `setPos` is the "basic" position setter - it just sets the fields.
  - `setPosition` calls `setPos` and also updates the entity's bounding box.
- There's `setPositionAndAngles` (two overloads, the BlockPos one calls the double one) that calls `method_22862` with the XYZ, sets the yaw and pitch, and calls `method_23311`.
  - `method_22862` calls `setPos`, and sets `prev(Render)X/Y/Z` to the passed XYZ.
  - `method_23311` calls `setPosition` with the entity's current coordinates. I'd assume it's for updating the bounding box.
- There's also `setPositionAnglesAndUpdate`, which calls `setPosition` but also sets the angles and `prevX/Y/Z/Yaw/Pitch`.

Pyrofab:
I would suggest :
- `setPos` -> `setPosition`; `setPosition` -> `updatePosition`
- `method_22862` -> reposition
- `method_23311` -> `refreshPosition`

liach:
Note that the current `setPosition` one also resets the bounding box of the entity.

PR #1051, opened 2020-01-11
(Block) entity position fixes (#964, #998, #1046, #1061)

YanisBft:


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -202,7 +202,7 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	METHOD method_23325 getParticleZ (D)D
 		ARG 1 widthScale
 	METHOD method_23326 getVelocityMultiplier ()F
-	METHOD method_23327 setPos (DDD)V
+	METHOD method_23327 setPosition (DDD)V
```
liach: pos imo is fine (shorter)

mappings/net/minecraft/block/entity/BlockEntity.mapping
```diff
@@ -25,7 +25,7 @@ CLASS net/minecraft/class_2586 net/minecraft/block/entity/BlockEntity
 		ARG 1 x
 		ARG 3 y
 		ARG 5 z
-	METHOD method_11009 setWorld (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;)V
+	METHOD method_11009 setLocation (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;)V
```
liach, line 28: is this more like init location (i.e. only called once in lifetime or after revalidation)

mappings/net/minecraft/entity/Entity.mapping
YanisBft: Do I rename `Position` to `Pos` in all methods?

mappings/net/minecraft/block/entity/BlockEntity.mapping
YanisBft: It's called each time `setBlockEntity` is called, so I think `set` is better than `init` in this case

Juuxel: Isn't `setBlockEntity` only called once per BE?

YanisBft: Nope it's not. Additionally, `setLocation` is called in `loadBlockEntity`

mappings/net/minecraft/entity/Entity.mapping
liach: Hmm, good question. This method compared to others is fairly frequently used as all xyz direct modifications has been migrated to this method in 1.15 snapshots.

liach (approved):

liach:
Lgtm. hope we can get this fix in soon.

Juuxel (approved):

Issue #682, opened 2019-05-06
Drop verbose 'Inv' wording from Inventrory methods.


modmuss50:
Drops 'Inv' from all Inventory methods. What else am i getting a stack on? A carrot?
Named afg(class_3829) -> PasteCallback, its only ever called by things that 'paste' a BlockEntity in world and its not specific to Inventories, although its only implemented by Inventory inheritors and JukeboxBlockEntity, still not 100% on the name imo, feedback would be nice.

Taken from #330

NeunEinser:
> Im for it, but I suggest keeping it on 
> 
>  `getInvSize` 
> `getInvWidth`
>  `canPlayerUseInv`
> `setInvProperty`
> `onInvOpen`
> `isInvEmpty`
> `getInvProperty`
> `onInvClose`
> `getInvPropertyCount`
> `getInvHeight`
> 
> This would leave us with the following changes
> 
> `getInvStack` -> `getStack`
> `takeInvStack` -> `takeStack`
> `setInvStack` -> `setStack`
> `getInvMaxStackAmount` -> `getMaxStackAmount`
> `removeInvStack` -> `removeStack`
> `isValidInvStack` -> `isValidStack`

_Originally posted by @modmuss50 in https://github.com/FabricMC/yarn/pull/330#issuecomment-449591284_

I think this sounds like a good approach. It's not clear that `setHeight` refers to the inventory's height, hence `setInvHeight`, it is however clear that `setStack` is something that happens in the inventory.

liach:
My concern is name clashes with other methods, e.g. that two get world methods with different name+signature in hopper minecart (apparently entity and block entity have get world methods that are named differently)

YanisBft:
Fixed by #1154

PR #1053, opened 2020-01-11
Clarify Inventory#getMaxStackAmount

Pyrofab:
The previous name could be confused as meaning "the maximum amount of stacks in the inventory".

### DIFF
--- a/mappings/net/minecraft/inventory/Inventory.mapping
+++ b/mappings/net/minecraft/inventory/Inventory.mapping
@@ -23,6 +23,7 @@ CLASS net/minecraft/class_1263 net/minecraft/inventory/Inventory
 	METHOD method_5443 canPlayerUseInv (Lnet/minecraft/class_1657;)Z
 		ARG 1 player
 	METHOD method_5444 getInvMaxCountPerStack ()I
+		COMMENT @return the max {@link net.minecraft.item.ItemStack#count count} of item stacks in this inventory
 	METHOD method_5447 setInvStack (ILnet/minecraft/class_1799;)V
 		ARG 1 slot
 		ARG 2 stack
--- a/mappings/net/minecraft/inventory/Inventory.mapping
+++ b/mappings/net/minecraft/inventory/Inventory.mapping
@@ -22,7 +22,7 @@ CLASS net/minecraft/class_1263 net/minecraft/inventory/Inventory
 	METHOD method_5442 isInvEmpty ()Z
 	METHOD method_5443 canPlayerUseInv (Lnet/minecraft/class_1657;)Z
 		ARG 1 player
-	METHOD method_5444 getInvMaxStackAmount ()I
+	METHOD method_5444 getInvMaxCountPerStack ()I
 	METHOD method_5447 setInvStack (ILnet/minecraft/class_1799;)V
 		ARG 1 slot
 		ARG 2 stack
--- a/mappings/net/minecraft/inventory/Inventory.mapping
+++ b/mappings/net/minecraft/inventory/Inventory.mapping
@@ -22,7 +22,7 @@ CLASS net/minecraft/class_1263 net/minecraft/inventory/Inventory
 	METHOD method_5442 isInvEmpty ()Z
 	METHOD method_5443 canPlayerUseInv (Lnet/minecraft/class_1657;)Z
 		ARG 1 player
-	METHOD method_5444 getInvMaxCountPerStack ()I
+	METHOD method_5444 getMaxCountPerStack ()I
 		COMMENT @return the max {@link net.minecraft.item.ItemStack#count count} of item stacks in this inventory
 	METHOD method_5447 setInvStack (ILnet/minecraft/class_1799;)V
 		ARG 1 slot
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/inventory/Inventory.mapping
```diff
@@ -22,7 +22,8 @@ CLASS net/minecraft/class_1263 net/minecraft/inventory/Inventory
 	METHOD method_5442 isInvEmpty ()Z
 	METHOD method_5443 canPlayerUseInv (Lnet/minecraft/class_1657;)Z
 		ARG 1 player
-	METHOD method_5444 getInvMaxStackAmount ()I
+	METHOD method_5444 getInvMaxCountPerStack ()I
```
Juuxel: Do we need `Inv` here? (Related: #682)

Pyrofab: Probably not, but shouldn't the change belong to another PR ?

Juuxel: I think it would be easy to remove `Inv` now that the method is being renamed anyway.

Pyrofab: Fair enough

Juuxel (approved):

mappings/net/minecraft/inventory/Inventory.mapping
```diff
@@ -22,7 +22,8 @@ CLASS net/minecraft/class_1263 net/minecraft/inventory/Inventory
 	METHOD method_5442 isInvEmpty ()Z
 	METHOD method_5443 canPlayerUseInv (Lnet/minecraft/class_1657;)Z
 		ARG 1 player
-	METHOD method_5444 getInvMaxStackAmount ()I
+	METHOD method_5444 getMaxCountPerStack ()I
```
liach, line 25:
```suggestion
	METHOD method_5444 getInvStackMaxCount ()I
```
If we use `count` in item stack, call count than amount.

liach (changes requested):
Imo moving `Max` to before `Amount` shows it applies to amount instead of stack

mappings/net/minecraft/inventory/Inventory.mapping
```diff
@@ -23,6 +23,7 @@ CLASS net/minecraft/class_1263 net/minecraft/inventory/Inventory
 	METHOD method_5443 canPlayerUseInv (Lnet/minecraft/class_1657;)Z
 		ARG 1 player
 	METHOD method_5444 getInvMaxCountPerStack ()I
+		COMMENT @return the max {@link net.minecraft.item.ItemStack#count count} of item stacks in this inventory
```
natanfudge: You don't need to repeat `count` in the link

mappings/net/minecraft/inventory/Inventory.mapping
Juuxel: I think `getMaxCountPerStack` is clearer than `getInvStackMaxCount`.

liach: `per` sounds to be as if this method is stack/slot aware, but it's not.

shedaniel: I vote for `getMaxCount/AmountPerStack` tho.

mappings/net/minecraft/inventory/Inventory.mapping
Pyrofab: `return the max ItemStack#count of item stacks in this inventory` ?

mappings/net/minecraft/inventory/Inventory.mapping
Pyrofab: `getInvStackMaxCount` does not resolve the initial issue, being that you could confuse it as meaning "max count of stacks in the inventory".

mappings/net/minecraft/inventory/Inventory.mapping
liach: since the count field is private, you should refer to the getter method.

mappings/net/minecraft/inventory/Inventory.mapping
liach: changing per to of may be clearer, i guess.

mappings/net/minecraft/inventory/Inventory.mapping
Juuxel: If you refer to `getCount`, you need to repeat "count" in the link for it to look nice (maybe also use `@linkplain`). 🤔

mappings/net/minecraft/inventory/Inventory.mapping
liach:
```suggestion
	METHOD method_5444 getMaxCountOfStack ()I
```

Pyrofab: `getMaxCountOfStacks` ?

Juuxel: `getMaxCountOfStacks` sounds like something that returns how many stacks can be in the inventory.

Pyrofab: True, it's not much of an improvement.

liach: Then just `getMaxCountOfStack` so you know it applies to one stack rather than all the stacks

liach:
Uh, I think we need a vote on this rename. I personally is against this one while juuz is for.

Prospector:
getSlotCapacity

Runemoro (approved):

YanisBft:
Closing this PR, as it's not updated since 1.15.1 and is now entirely covered by #1154

PR #1054, opened 2020-01-11
Improved Window names

comp500:
I added some more names in Window, and changed the name for `setFullscreen` to `swapBuffers`, as it actually calls `RenderSystem.flipFrame` (a blaze3d method, that calls `GLFW.glfwSwapBuffers`), and while it does update the screen to reflect the set fullscreen value, it is called every frame regardless of whether the fullscreen value has changed.

### DIFF
--- a/mappings/net/minecraft/client/util/Window.mapping
+++ b/mappings/net/minecraft/client/util/Window.mapping
@@ -4,6 +4,7 @@ CLASS net/minecraft/class_1041 net/minecraft/client/util/Window
 	FIELD field_5174 windowedWidth I
 	FIELD field_5175 windowedX I
 	FIELD field_5176 eventHandler Lnet/minecraft/class_3678;
+	FIELD field_5177 currentFullscreen Z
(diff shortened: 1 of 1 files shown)


modmuss50 (approved):

mappings/net/minecraft/client/util/Window.mapping
```diff
@@ -31,8 +32,9 @@ CLASS net/minecraft/class_1041 net/minecraft/client/util/Window
 		ARG 5 title
 	METHOD method_15997 setScaleFactor (D)V
 		ARG 1 scaleFactor
-	METHOD method_15998 setFullscreen ()V
+	METHOD method_15998 swapBuffers ()V
 	METHOD method_15999 setFramerateLimit (I)V
+		ARG 1 limit
```
Juuxel:
```suggestion
		ARG 1 framerateLimit
```

Juuxel (changes requested):

Juuxel (approved):

PR #1063, opened 2020-01-17
Map the 1.15.2 changes

modmuss50:
Im no expert on rendering, so please check over this.

### DIFF
--- a/mappings/net/minecraft/client/render/model/BakedModel.mapping
+++ b/mappings/net/minecraft/client/render/model/BakedModel.mapping
@@ -4,7 +4,7 @@ CLASS net/minecraft/class_1087 net/minecraft/client/render/model/BakedModel
 		COMMENT <p>
 		COMMENT True, the model will be lit from the side, like a block.
 		COMMENT <p>
-		COMMENT False, the model will be lit from the font, like an item.
+		COMMENT False, the model will be lit from the front, like an item.
 	METHOD method_4707 getQuads (Lnet/minecraft/class_2680;Lnet/minecraft/class_2350;Ljava/util/Random;)Ljava/util/List;
 		ARG 1 state
 		ARG 2 face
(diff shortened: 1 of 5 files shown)


mappings/net/minecraft/client/render/model/BakedModel.mapping
```diff
@@ -1,4 +1,10 @@
 CLASS net/minecraft/class_1087 net/minecraft/client/render/model/BakedModel
+	METHOD method_24304 isSideLit ()Z
+		COMMENT Allows control of the lighting when rendering a model in a GUI.
+		COMMENT <p>
+		COMMENT True, the model will be lit from the side, like a block.
+		COMMENT <p>
+		COMMENT False, the model will be lit from the font, like an item.
```
Juuxel: Typo here:
```suggestion
		COMMENT False, the model will be lit from the front, like an item.
```

Juuxel (changes requested):

Juuxel (approved):

grondag:
This looks right, ~~except `hasDepth` is the opposite of what it should be. Suggest `isFlat` instead~~.

**Edit: I went over this once more, and the last boolean parameter to the JSON model builder constructor is `false` for generated item model.  I must have misread it as true before.  So `hasDepth` is a good name and I think this can be merged as-is.**

I think I know what happened here.  Previously, there was only one boolean parameter (for JSON models) that was set based on which models were generated item models.  This parameter did control depth lighting in GUI so the prior name was appropriate.

In 1.15.2, Mojang decided to control depth lighting via a JSON attribute but also kept the flag that indicates a generated item model.  The prior flag now controls geometric transformation for item entity rendering, but was still mapped to `depthInGui`.

The name `depthInGui` would be an adequate name for the *new* field/method, but it's better to change both names so that mods can explicitly determine what to return in each case vs having `depthInGui` now point to an entirely different field/method. 

See attached diagram for analysis.
image

modmuss50:
Thanks for looking at it, I’ll make the changes later on

grondag (approved):

liach:
wanted to take a look but sure, problems can be fixed later

modmuss50:
Ah, my mistake. Had 2 reviews, and it was quite important as it clears some things up.

PR #1066, opened 2020-01-19
1.15.2 Profiler mappings

liach:
Missing stuff and (mainly) 3 new classes

### DIFF
--- a/mappings/net/minecraft/util/profiler/ProfileLocationInfo.mapping
+++ b/mappings/net/minecraft/util/profiler/ProfileLocationInfo.mapping
@@ -1,0 +1,9 @@
+CLASS net/minecraft/class_4748 net/minecraft/util/profiler/ProfileLocationInfo
+	COMMENT Profiling information on a specific profiler location.
+	METHOD method_24272 getTotalTime ()J
+		COMMENT Returns the total time spent visiting the profiler location.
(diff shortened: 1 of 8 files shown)


modmuss50 (approved):

mappings/net/minecraft/util/profiler/ProfileLocationInfo.mapping
```diff
@@ -0,0 +1,9 @@
+CLASS net/minecraft/class_4748 net/minecraft/util/profiler/ProfileLocationInfo
+	COMMENT Profiling information on a specific profiler location.
```
Pyrofab, line 2: Wouldn't this be a `section` of the code rather than a precise location ?

liach:
@pyrofab we have been using location as it's indicated by a path like a.b.c (replace . with ascii char 30), which location is a better term than section

Pyrofab:
If it is represented by a path, why not use `path` instead of `location` ? A "code path" makes quite a bit of sense too.

liach:
not gonna refactor on that. Path leads to a location, and it's a location in profiler tree not in code, so location makes sense. please don't bother about this!

modmuss50 (approved):

Juuxel (approved):

PR #1068, opened 2020-01-19
Status effect stufff

liach:


haykam821 (approved):

mappings/net/minecraft/entity/effect/StatusEffectInstance.mapping
```diff
@@ -1,4 +1,8 @@
 CLASS net/minecraft/class_1293 net/minecraft/entity/effect/StatusEffectInstance
+	FIELD field_21830 hiddenEffect Lnet/minecraft/class_1293;
+		COMMENT The effect overwritten when merging effects. Duration decreases with
+		COMMENT this one, and always have a longer duration and lower amplifier compared
+		COMMENT to this effect.
```
Pyrofab: This doesn't seem clear enough. Could you give examples of how this is used ?

liach: this is rather internal for status effects. not really user oriented

liach: just upgrade (merge) one long effect with low amplifier with one short effect with high amplifier

Pyrofab: so when the short effect with a strong amplifier runs out, the long effect with a weak amplifier takes over ? I think you should just say that.

liach: https://minecraft.gamepedia.com/Java_Edition_1.15.2_Pre-Release_1#Gameplay

Pyrofab: I meant saying that in the comment, as I think it's not clear enough right now.

liach: Oh shoot, @Pyrofab please do not block prs for trivial things like this... This is not public, just some impl detail and never used by mods. You are discouraging people from documenting private fields and I am disagreeing over this (you've also asked for too much on profiler mappings)

Pyrofab: Because as everyone knows, my suggestions are legally binding.

Pyrofab (approved):

YanisBft (approved):

Juuxel (approved):

PR #1071, opened 2020-01-21
Generate a mergedv2 distribution

modmuss50:
Generates an artificat that also contains the mojang names. This is the same jar that loom uses at runtime. It is useful for tools that do not want to parse both intermediary and yarn.

and gradle update + dep updates

is `mergedv2` a good classifier?


build.gradle
```diff
@@ -191,6 +191,22 @@ task downloadIntermediary(type: Download) {
 	dest new File(cacheFilesMinecraft, "${minecraft_version}-intermediary.tiny")
 }
 
+task downloadIntermediaryV2(type: Download) {
+	group = buildMappingGroup
+	def url = "https://maven.fabricmc.net/net/fabricmc/intermediary/${minecraft_version}/intermediary-${minecraft_version}-v2.jar"
+	src com.google.common.net.UrlEscapers.urlFragmentEscaper().escape(url)
```
Player3324, line 197: This is not right, but it probably doesn't really break here so I don't mind too much. The URL fragment is the `#asd` part. Escaping should always happen while translating from one "language" (arbitrary String blobs) to another (URL path segment), so on the individual strings.

Player3324 (approved):
Nothing obvious wrong :)

build.gradle
modmuss50: This has worked fine for downloading the v1 mappings, as far as I know it’s only needed when the version contains a space or some other char. (April fools version)

Runemoro (approved):

liach (approved):
Looks good, consider merge into 20w10a branch as well as 1.15.2

Issue #915, opened 2019-10-18
packet name issues


liach:
`GuiOpenS2CPacket`: only applies to horse inventories
`Gui` packets: rename to `Container`/`Menu` packets instead
`EntityS2CPacket`: should be rename to `EntityMotionS2CPacket` `EntityMovementS2CPacket`
`EntityPotionEffectS2CPacket` -> `EntityStatusEffectS2CPacket`

liach:
Thanks @Yanis48! :heart:

PR #1074, opened 2020-01-24
Moved all S2C/C2S packets

### LINKED ISSUE #622: Move packets out of `net.minecraft.server` and `net.minecraft.client`
Packets are shared on both sides (that's their purpose!), so they don't belong in the side-specific packages. They should all be moved into `net.minecraft.network.packet.s2c` and `net.minecraft.network.packet.c2s`.

YanisBft:
Fixes #622 . As Runemoro said, packets are not exclusive to either client or server side, so it makes no sense to let them in client or server packages. According to the thumbs up, many people seem to agree on it.

Additionally, some packets were renamed:
- `EntityPotionEffectS2CPacket` -> `EntityStatusEffectS2CPacket`, as it's the name used everywhere else for now. After this PR, #915 can be closed, as the other issues described there are already fixed
- `RemoveEntityEffectS2CPacket` -> `RemoveEntityStatusEffectS2CPacket`, to match the above
- `GuiActionConfirmC2SPacket` -> `ConfirmGuiActionC2SPacket`, to match the S2C one


liach:
assuming we rarely do qualified reference (i.e. no import), i'd prefer clientbound/serverbound for package names

liach:
Another way to group packets is probably by their recipients (the generic T in packet class). they are more specific than c2s/s2c packets (e.g. server play, client play, client login, server login, etc.)

YanisBft:
> assuming we rarely do qualified reference (i.e. no import), i'd prefer clientbound/serverbound for package names

Hum, aren't they confusing names? I think there's nothing clearer than `s2c`/`c2s`. Waiting for more opinions about your idea tho.

> Another way to group packets is probably by their recipients (the generic T in packet class). they are more specific than c2s/s2c packets (e.g. server play, client play, client login, server login, etc.)

I think it won't be consistent with how we sort classes in other packages (ie we don't sort features by the feature config they use). Again, waiting for more opinions.

Pyrofab:
`s2c` and `c2s` can be confusing for newcomers when they don't know what the abbreviations mean. Maybe keeping them in class names (for brevity) but using full English names in packages would help make things more intuitive.

Player3324:
We already had this *bound discussion multiple times, It is just a bad ambiguous scheme and strictly inferior to c2s/s2c outside some minor stylistic concerns. 

I wouldn't mind a state sub-hierarchy, e.g. `n/m/network/packet/c2s/play/ConfirmGuiActionC2SPacket`, but probably nothing beyond that since direct vanilla packet use is a rather low level facility anyway.

Prospector:
Again, though, it's not actually ambiguous since XBound and Xbound are entirely different in the english language. I don't think yarn should cater to an incorrect understanding of the english language, personally.

At some point one must learn what C2S and S2C means in the first place, and since Clientbound/Serverbound will be intuitive for most english natives and likely many non-natives, anyone who doesn't already know what it means can simply learn it as everyone must for C2S/S2C anyway. Once they have learned it, they will not just have learned a single convention that one project uses, but something valuable to know about the english language for any future encounters of the *bound suffix in the rest of their life.

YanisBft:
Any more opinions on this? What should I do, rename s2c/c2s packages to clientbound/serverbound? Create sub-packages for each listener type?

Pyrofab:
Probably create sub-packages for the listener types, although it's probably not immediately necessary.

YanisBft:
Splitted into subpackages

liach (approved):
Looks good after a fetch locally.

liach:
I hope this one gets decided soon so that I can work on renaming some fields in packet classes (related to networking api v1)

Player3324 (approved):
LGTM conceptually, didn't check details. Thanks for doing this!

liach:
Just need a final round of reviews. Ready to merge imo after that.

Earthcomputer (approved):

Juuxel (approved):

PR #1076, opened 2020-01-25
Motive -> Motif

Sollace:
I realise @Mojang calls them "Motives" but that's incorrect. The feature on a painting is a _motif_, correcting that is my _motive_ for creating this PR.

### DIFF
--- a/mappings/net/minecraft/entity/decoration/AbstractDecorationEntity.mapping
+++ b/mappings/net/minecraft/entity/decoration/AbstractDecorationEntity.mapping
@@ -1,12 +1,19 @@
 CLASS net/minecraft/class_1530 net/minecraft/entity/decoration/AbstractDecorationEntity
+	FIELD field_7097 obstructionCheckCoutner I
 	FIELD field_7098 PREDICATE Ljava/util/function/Predicate;
 	FIELD field_7099 facing Lnet/minecraft/class_2350;
-	FIELD field_7100 blockPos Lnet/minecraft/class_2338;
+	FIELD field_7100 attachmentPos Lnet/minecraft/class_2338;
 	METHOD <init> (Lnet/minecraft/class_1299;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;)V
-		ARG 1 world
+		ARG 1 type
+		ARG 2 world
+		ARG 3 pos
+	METHOD method_6888 canStayAttached ()Z
 	METHOD method_6889 onBreak (Lnet/minecraft/class_1297;)V
+		ARG 1 entity
 	METHOD method_6891 getHeightPixels ()I
 	METHOD method_6892 setFacing (Lnet/minecraft/class_2350;)V
+		ARG 1 facing
 	METHOD method_6894 onPlace ()V
+	METHOD method_6895 updateAttachmentPosition ()V
 	METHOD method_6896 getDecorationBlockPos ()Lnet/minecraft/class_2338;
 	METHOD method_6897 getWidthPixels ()I
--- a/mappings/net/minecraft/entity/decoration/AbstractDecorationEntity.mapping
+++ b/mappings/net/minecraft/entity/decoration/AbstractDecorationEntity.mapping
@@ -1,5 +1,5 @@
 CLASS net/minecraft/class_1530 net/minecraft/entity/decoration/AbstractDecorationEntity
-	FIELD field_7097 obstructionCheckCoutner I
+	FIELD field_7097 obstructionCheckCounter I
 	FIELD field_7098 PREDICATE Ljava/util/function/Predicate;
 	FIELD field_7099 facing Lnet/minecraft/class_2350;
 	FIELD field_7100 attachmentPos Lnet/minecraft/class_2338;
--- a/mappings/net/minecraft/client/network/packet/PaintingSpawnS2CPacket.mapping
+++ b/mappings/net/minecraft/client/network/packet/PaintingSpawnS2CPacket.mapping
@@ -1,12 +1,12 @@
 CLASS net/minecraft/class_2612 net/minecraft/client/network/packet/PaintingSpawnS2CPacket
 	FIELD field_12008 pos Lnet/minecraft/class_2338;
 	FIELD field_12009 uuid Ljava/util/UUID;
-	FIELD field_12010 motive I
+	FIELD field_12010 motif I
 	FIELD field_12011 facing Lnet/minecraft/class_2350;
 	FIELD field_12012 id I
 	METHOD <init> (Lnet/minecraft/class_1534;)V
 		ARG 1 entity
-	METHOD method_11221 getMotive ()Lnet/minecraft/class_1535;
+	METHOD method_11221 getMotif ()Lnet/minecraft/class_1535;
 	METHOD method_11222 getPaintingUuid ()Ljava/util/UUID;
 	METHOD method_11223 getFacing ()Lnet/minecraft/class_2350;
 	METHOD method_11225 getId ()I
--- a/mappings/net/minecraft/client/texture/PaintingManager.mapping
+++ b/mappings/net/minecraft/client/texture/PaintingManager.mapping
@@ -1,5 +1,7 @@
 CLASS net/minecraft/class_4044 net/minecraft/client/texture/PaintingManager
 	FIELD field_18032 PAINTING_BACK_ID Lnet/minecraft/class_2960;
+	METHOD <init> (Lnet/minecraft/class_1060;)V
+		ARG 1 manager
 	METHOD method_18342 getBackSprite ()Lnet/minecraft/class_1058;
 	METHOD method_18345 getPaintingSprite (Lnet/minecraft/class_1535;)Lnet/minecraft/class_1058;
-		ARG 1 motive
+		ARG 1 motif
--- a/mappings/net/minecraft/datafixer/fix/EntityPaintingMotifFix.mapping
+++ b/mappings/net/minecraft/datafixer/fix/EntityPaintingMotifFix.mapping
@@ -1,0 +1,7 @@
+CLASS net/minecraft/class_3607 net/minecraft/datafixer/fix/EntityPaintingMotifFix
+	FIELD field_15898 RENAMED_MOTIFS Ljava/util/Map;
+	METHOD <init> (Lcom/mojang/datafixers/schemas/Schema;Z)V
+		ARG 1 outputSchema
+		ARG 2 changesType
+	METHOD method_15723 renameMotif (Lcom/mojang/datafixers/Dynamic;)Lcom/mojang/datafixers/Dynamic;
+		ARG 1 dynamic
--- a/mappings/net/minecraft/datafixer/fix/EntityPaintingMotiveFix.mapping
+++ b/mappings/net/minecraft/datafixer/fix/EntityPaintingMotiveFix.mapping
@@ -1,5 +1,0 @@
-CLASS net/minecraft/class_3607 net/minecraft/datafixer/fix/EntityPaintingMotiveFix
-	FIELD field_15898 RENAMED_MOTIVES Ljava/util/Map;
-	METHOD <init> (Lcom/mojang/datafixers/schemas/Schema;Z)V
-		ARG 1 outputSchema
-		ARG 2 changesType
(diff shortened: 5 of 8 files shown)


Juuxel:
They're registered in a registry called `minecraft:motive`.

Sollace:
> 
> 
> They're registered in a registry called `minecraft:motive`.

Yeah, but it's still wrong.

https://www.dictionary.com/browse/motif

Sollace:
From the definition of Motif:
image

Juuxel:
We don't rename other bad registered names either, like `Entity.AREA_EFFECT_CLOUD` or `Decorator.HELL_FIRE`. I don't see why registries and the objects registered in them should be any different.

Sollace:
> 
> 
> We don't rename other bad registered names either, like `Entity.AREA_EFFECT_CLOUD` or `Decorator.HELL_FIRE`. I don't see why registries and the objects registered in them should be any different.

Difference it that this one is clearly a typo. You wouldn't leave a field named `Entity.ARAE_EFFECT_CLUOD` in, would you?

Fuyukai:
There's also EntityPaintingMotiveFix. It's clearly deliberate.

Sollace:
> 
> 
> There's also EntityPaintingMotiveFix. It's clearly deliberate.

The choice here is either we use the name Mojang gave it, or we use a name that actually describes the object.

i509VCB:
I agree with the name change since a motive sounds like a cause.

Motif sounds like the correct name for the object even though the registry is incorrectly named. 

Now this isn't an apple -> orange rename, the name is still related to the internal name of the registry.

Earthcomputer:
I was very confused initially when I saw the "motive" registry, I would have been less confused if it were called "motif". Plus, if we use the argument that we should call it "motive" just because Mojang calls it that, means that we should also rename "container" to "menu". Except this is worse than the container/menu debate, as a motif is a real and well defined English word meaning exactly this, and motive means something else and is confusing.

Prospector:
Motive is a synonym for Motif in this context. Not a necessary change. Motive is just as correct as Motif.

image

mappings/net/minecraft/entity/decoration/AbstractDecorationEntity.mapping
```diff
@@ -1,12 +1,19 @@
 CLASS net/minecraft/class_1530 net/minecraft/entity/decoration/AbstractDecorationEntity
+	FIELD field_7097 obstructionCheckCoutner I
```
liach:
```suggestion
	FIELD field_7097 obstructionCheckCounter I
```

liach (changes requested):

Prospector (commented):
This is like voting to change DyeColor to DyeColour. No point, stick with the mojang name since it's just as correct.

i509VCB:
> This is like voting to change DyeColor to DyeColour. No point, stick with the mojang name since it's just as correct.

That is American vs British English in that case.

Motive, similarly to a criminal's motive invokes the idea of an action/cause rather than a representation of an idea (which a motif is).

Prospector:
see my other message. The second definition for motive is literally "a motif"

Earthcomputer:
As a native English speaker, I did not know that "motive" was a synonym of "motif". Maybe I'm weird and just haven't seen it before, but either way there's ambiguity in "motive", and absolutely no ambiguity in "motif". That, along with the fact that "motive" is much more commonly used to mean something else, is a good reason for this change.

Prospector:
Honestly I hadn't heard it either until I saw it in the registry name either but once I looked it up it immediately made sense. I don't think people not knowing a real word is a valid reason to change a name. It's not a name that could possibly be confused with something else. If there was another concept in Minecraft that involved the other definition of motive then I may feel differently. To me this is exactly like the issue with people not understanding what Clientbound/Serverbound mean. I don't think a lack of understanding of English should dictate the direction yarn moves in.

The worst case scenario with the current name is that someone is confused for a moment but then learns a new word. Best case scenario is they already know the word.

Prospector:
That said, I wouldn't be entirely opposed to naming it to like PAINTING_MOTIVE or something just to clarify. However, there's no reason that a mod couldn't expand the motives to something other than paintings if they wanted to so perhaps that is not a great idea after a second thought.

Earthcomputer:
> It's not a name that could possibly be confused with something else

Oh, you'd be hilariously surprised, when I first saw it, I thought the default entry "kebab" was the motive for like a crime, and I thought the registry was just Mojang's test of the registry system, and someone being funny. It's incredible that I actually thought that but it just goes to show how badly known the "motive" synonym is.

Of course I would be in favour of PAINTING_MOTIVE or PAINTING as well as MOTIF, those remove the confusion.

liach:
https://www.dictionary.com/browse/motive

See definition 3 under the noun definitions (you need to click "SEE MORE")

image

Sollace:
> Motive is a synonym for Motif in this context. Not a necessary change. Motive is just as correct as Motif.

I'm siding with @liach here. I write as a side-hobby, and generally have better knowledge of english than the layman, and even _I_ didn't know that Motive could be used a synonym for Motif, and that's most likely because people (Americans) who don't know how to spell hear motif and write motive meaning the former.

The synonym meaning is _incredibly_ obscure, almost more obscure than the word "motif" itself, so nothing changes here.

modmuss50:
I’ve never encountered the word motif before

Sollace:
> 
> 
> > It's not a name that could possibly be confused with something else
> 
> Oh, you'd be hilariously surprised, when I first saw it, I thought the default entry "kebab" was the motive for like a crime, and I thought the registry was just Mojang's test of the registry system, and someone being funny. It's incredible that I actually thought that but it just goes to show how badly known the "motive" synonym is.
> 
> Of course I would be in favour of PAINTING_MOTIVE or PAINTING as well as MOTIF, those remove the confusion.

Looking back, I can probably recount a similar experience to what @Earthcomputer is describing. The first time I saw it I was looking at the registries and thought "MOTIVE? Is this something to do with entity AI or something?"

I didn't bother looking looking into the usage, because it wasn't what I was looking for, but I remember coming out of it believing the registry must have been some kind of test of the registry system by Mojang or something.

liach:
So motive is not a wrong term. We should add javadoc to all of these indicating the word "motive" means like "motif" to clarify. That's why we have javadocs.

Sollace:
> 
> 
> So motive is not a wrong term. We should add javadoc to all of these indicating the word "motive" means like "motif" to clarify. That's why we have javadocs.

It's not _technically_ wrong in the same way I'm _technically_ a monkey.

Earthcomputer:
"Motive" is not technically wrong, apparently, but it's a horribly confusing name. It's different from clientbound/serverbound in that literally every native English speaker in this conversation has had to consult a dictionary. Literally the _only_ argument for keeping it as "motive" is that that's what Mojang calls them, and we haven't exactly followed that reasoning in the past, have we?

What I think might be quite likely is that someone from Mojang knew the word in their native language and consulted a bilingual dictionary to translate it into English, and picked the wrong synonym, without knowledge of what a native English speaker would have chosen to call it. Yarn names shouldn't be bound by this, they should be the least confusing names for English speakers.

Motive is confusing, motif is apparently confusing for some, fine. How about renaming it to "painting", then? Or "painting_motif"?

Prospector:
let's go with PAINTING_MOTIVE so you know what it means and you get to learn a new word.

Sollace:
> 
> 
> let's go with PAINTING_MOTIVE so you know what it means and you get to learn a new word.

You mean `PAINTING_MOTIF` as @Earthcomputer suggested?

Sollace:
We can't exactly call it a painting, because we have `PaintingEntity`. How does a painting have a painting? That doesn't make much sense.

You can say a painting has an _image_. But a painting having a painting is kind of redundant.

Prospector:
No, keep the mojang name just add a clarifying prefix to it since motive has two definitions.

Sollace:
No. The first thing everyone thinks of when they read motive is of a murder.

It's a bad name and shouldn't be involved. At all.

Prospector:
It is just as valid of a word. With a painting prefix there is 0 confusion, and if anyone wasn't aware of the other definition of motive they will learn something.

Sollace:
I would much rather people learn the definition of Motif than learn the wrong definition of Motive. It makes no difference either way, either, because in both cases you have people reaching for a dictionary.

(or possibly not, because they think they know what Motive means, in which case you'll be getting questions of how/why a painting can have a motive)

Prospector:
Why are you so hung up on the fact that a word can mean multiple things? It is factually, objectively, **not wrong**. Therefore, it won't be changed.

coderbot16:
`PAINTING_MOTIVE` would perfectly express that it's for paintings, clearing any confusion.

Sollace:
It's an extremely obscure meaning, and (very clearly) confusing for a lot of people, and I don't think prefixing with 'PAINTING` will entirely clear that up. (if any it's going to raise more questions)

coderbot16:
But motif is just as obscure. Might as well use the obscure word that the registry names use instead of creating another mismatch.

Also, all of the confusion in the first place appears to be in determining that it's related to paintings.

Sollace:
The reason I prefer Motif over Motive is because people aren't going to automatically assume they know what it means and go straight to the first (or second) definition when it's actually used for the third definition.

Sollace:
Anyway, I've argued this enough. You can choose to merge it or not. It won't affect me (as I have never worked with paintings in Minecraft and might not for a while)

Sollace:
A wise man once said:
image

liach:
My points against this rename:
1. Motive is a correct term.
2. The clarification should be made in comments (javadoc)

Other non-rename related changes, including some mappings for paintings, may go into another pr.

Earthcomputer (approved):

Earthcomputer:
I have no real preference between `PAINTING_MOTIVE` vs `PAINTING_MOTIF`, as the prefix gives all the clarification needed. `MOTIF` on its own is better than `MOTIVE` on its own due to the difference in obscurity.

On a recent PR, I had named some methods with an "activate" and "deactivate" prefix, and liach suggested that I rename them to "enable" and "disable", which are synonyms, but it was for clarity. It's the same thing here, just because both words are obscure and that you may not have heard of either, doesn't change the fact that one is far more obscure than the other.

With that in mind, the only reason you have left to object to this change is because it's the same as the Mojang name. In which case we should merge the menu PR.

modmuss50 (approved):
After some thought, I dont see why not. It seems motif is the technically correct name here.

Its a little used part of the game (for mods) so I see no issue in changing it as it wont cause a headace for everyone.

liach:
for the dfu one, i suggest keeping motive name and clarify in jd as motive is used in ldc string.

Earthcomputer:
Javadocs don't exist to fix naming issues, they exist to add extra information about a good name. You wouldn't (or shouldn't) name loop variables "i" and "j" only to add a comment explaining what they mean, you'd rename "i" and "j" to something more meaningful. In this case, I would be in favour of adding a javadoc explaining what a "motif" is to those who don't know, while also doing the rename from "motive" to "motif" to help those who do know.
If I was not a native English speaker, I would prefer to learn a word that English speakers would actually use, than some obscure synonym that even native speakers don't understand.

i509VCB (approved):

liach:
Motive is a valid word for "purpose/theme of a painting". As a result, I hereby oppose this change in where string literals refer to `motive`, e.g. in data fixer fix and registry code.

liach (approved):
Quality-wise this pull request is good, but I still hold by :-1: position on this rename proposal.

Prospector:
Why is `painting_motive` not the best of all worlds? It both keeps the mojang name (which is *entirely* valid an objectively not incorrect) and also clarifies that it is for paintings as that is not clear with the pure mojang name.

Earthcomputer:
+1 for renaming to `painting_motive`, while clarifying further that it's a "motif" in Javadocs.

ghost:
Honestly I've never seen motif before so I'd get cofused when seeing this. Also I like prospectors proposal (painting_motive) better.

mappings/net/minecraft/entity/decoration/AbstractDecorationEntity.mapping
NinjaPhenix: Is this a typo, `Coutner`

Runemoro:
This doesn't make sense. Both motif and motive are correct spellings for the same term. Matching the in-game ID is more important than matching the usual spelling.

It's confusing to have to remember that for commands, you need to use "motive" while for code you need to use "motif".

Prospector:
Yeah, I agree changing to motif makes no sense, though I do think adding a `painting` prefix to it could be a good thing. Unless we think the same motives could be expanded to other things?

Earthcomputer:
This can be closed now since #1100 got merged.

PR #1077, opened 2020-01-26
Tag -> Nbt Redux for 1.15.2

Prospector:
I have made changes to many method names using this set of rules:
* Methods which convert an **object into NBT** will either be named:
  * `toNbt` - when an existing NBT is not passed into the method as a parameter. This method entirely creates the NBT which is returned.
  * `saveToNbt` - when either the method has a void return type or an existing NBT is passed into the method, meaning this method does not create the NBT that is returned.
* Methods which convert **NBT into an object** will either be named:
  * `fromNbt` - when this method creates an object using the NBT data passed in as a parameter.
  * `loadFromNbt` - when this method uses NBT data to modify an existing instance of an object.


liach:
Really, only 84 places?

Prospector:
I literally went through every single reference of `CompoundNbt` and checked every method currently named `toTag`. I'm sure I missed something but this is already way more consistent than it was before (names were sorta all over the place)

liach:
Imo just use `toNbt` and `fromNbt`, like the simple from and load from versions usually don't coexist; same goes for to and save to versions

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -246,10 +246,10 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
...
 	METHOD method_5637 isWet ()Z
 		COMMENT Returns whether this entity is touching water, or is being rained on, or is inside a bubble column...
-		COMMENT 
-		COMMENT @see net.minecraft.entity.Entity#isTouchingWater() 
-		COMMENT @see net.minecraft.entity.Entity#isBeingRainedOn() 
-		COMMENT @see net.minecraft.entity.Entity#isInsideBubbleColumn() 
+		COMMENT
+		COMMENT @see net.minecraft.entity.Entity
+		COMMENT @see net.minecraft.entity.Entity
+		COMMENT @see net.minecraft.entity.Entity
```
i509VCB: Were these supposed to change?

Prospector:
I think it's an important distinction personally. See for example this where one method creates NBT while the other just saves to existing NBT.
image
Minecraft is all over the place with how these methods work so I think it is a good idea to name them consistently depending on function so that people can intuitively tell what a method does just by looking at the name.

mappings/net/minecraft/entity/Entity.mapping
Prospector: must be an enigma bug. I didn't touch anything to do with comments

mappings/net/minecraft/inventory/ContainerLock.mapping
```diff
@@ -11,11 +11,7 @@ CLASS net/minecraft/class_1273 net/minecraft/inventory/ContainerLock
...
 		ARG 1 stack
 			COMMENT the key item stack
-	METHOD method_5473 fromTag (Lnet/minecraft/class_2487;)Lnet/minecraft/class_1273;
-		COMMENT Creates a new {@code ContainerLock} from the {@code Lock} key of the compound tag.
-		COMMENT <p>
-		COMMENT If the {@code Lock} key is not present, returns an empty lock.
-		ARG 0 tag
-	METHOD method_5474 toTag (Lnet/minecraft/class_2487;)V
-		COMMENT Inserts the key string of this lock into the {@code Lock} key of the compound tag.
-		ARG 1 tag
```
i509VCB, line 21: hmm enigma must be tripping balls right now

Prospector: uhhhhh does enigma just nuke comments when you rename a method then?

Prospector: that is super annoying

Prospector: I guess make note of all the times that happens and I will fix them all manually like tomorrow

liach:
> I think it's an important distinction personally. See for example this where one method creates NBT while the other just saves to existing NBT.

Agreed.

Additionally, in my opinion, if the method has the NBT parameter, the method should just be `loadFrom`/`saveTo` as we already know there is NBT (as seen in params).

Prospector:
That's a good call, although my worry with that is someone may be typing `Nbt` hoping for a method autocomplete of either `loadFromNbt` or `fromNbt` and it not coming up if the method doesn't have `Nbt` in it.

mappings/net/minecraft/inventory/ContainerLock.mapping
Juuxel: See FabricMC/Enigma#190.

Juuxel:
I feel like the `*Tag` class names just sound better compared to `*Nbt` or `Nbt*`. I like the distinction between `loadFrom/saveToTag` and the regular `from/toTag`, though. (And they're especially better than `(de)serialize`)

Prospector:
The problem with Tag is the conflict with the other kind of tags. When I was renaming all these I saw a bunch of instances of both types of tags being used lines apart from each other and it was very confusing before I made the distinction. Plus, no one calls them tags colloquially, it's always NBT. So if a modder is searching for a method to save NBT they won't be searching for tag, they will be searching for NBT.

natanfudge:
Pros of NBT:
- Doesn't conflict with block/item tags (it's the only case of a name conflict I think)
- "NBT" sounds like some storage format, "Tag" sounds like just the key you attach to a value, its _tag_. Like tagging a shirt with a price _tag_.   

Pros of Tag:
- Tag sounds better
- ??? @Juuxel @2xsaiko @UntaggableYoghurt4C

Pyrofab:
I like the method renames, as they make things easier to find and understand. I however feel like the class renames are unnecessary. In my opinion, `NBT` should be the format, and `Tag` should be data objects for that format.

- Serialization methods are called and implemented very often by modders, so they need to be unambiguous and easy to search for.
- You can say those methods interact with the *NBT format*, as opposed to specific tags.
- The base NBT `Tag` class can already be distinguished from the registry `Tag<T>` by its lack of generics.
- `Tag` subclasses (eg. `CompoundTag`) are already widely understood and unambiguous. I would dislike renaming them to a more unwieldy variant. (besides, they are reminiscent of other cursed names)

Prospector:
I would argue *Nbt would be perfectly understandable for people who know the existing names because they would already know the classes are NBT classes, but there is a huge gain in intuitiveness for new users who it is now immediately clear what stuff is for NBT and what stuff is for data tags. If you have a method called saveToNbt and you override it and you're given a CompoundNbt, it's pretty obvious, "oh there's the NBT I save to" while if it's called a CompoundTag then, "what do tags have to do with this? Is it a key for the saving? Where to I save it to? Guess I have to ask on the discord or scour the wiki finding out something that should be self explanatory."

kvverti:
@Prospector and I, among others, discussed in the discord renaming the nbt classes to be like the Gson classes

Eg. `NbtList`, `NbtInt`, `NbtByteArray` with the base class being `NbtElement`

ChloeDawn:
`NbtReader` and `NbtReaders` are incorrect names. They are not readers, they are types. The appropriate names would be **`NbtType`** and **`NbtTypes`** respectively.

See for reference:
- `CompoundTag#createCrashReport` (type identification)
- `CompoundTag#read` (type identification)
- `StringNbtReader#readArray` (type referential equality)

Following these name fixes, the utility class in Fabric's API should be named `NbtIds`.
The fields in each tag class will also need migrating from `READER` to `TYPE`.
I would further suggest renaming `TagReaders#of` to `byId`, as, in either situation, an `of` method implies construction rather than lookup and does not read fluently for a plural-named class either (proposing **`TagTypes#byId`**).

Prospector:
Leaving this open for discussion on renaming the Tag classes to using the gson format: NbtElement or NbtTag, NbtObject or NbtCompound, NbtString, NbtList, etc.

@ChloeDawn your suggestion has been implemented in #1078

dblsaiko:
I'm heavily against at least using `Object` as the name for `CompoundTag`, because all the other names that would come with using gson names aren't really json specific (and would actually stay the same).
`Object` is first and foremost `java.lang.Object`, and that doesn't make sense at all to describe `CompoundTag`. It only works in the case of json since that's what it's actually called in the specification, exactly like `CompoundTag`.
You can't just take the terminology used for one thing and apply it to something different that already has a name, only because the function of those two things is similar.
Same as to why I wouldn't write a json library and call json objects compounds, or making `Map` implementations and naming them `*Dictionary` because that's what they're called in C#.
Renaming it to `Object` would make no sense to someone who doesn't know where that name comes from, and makes all the existing documentation harder to use because it uses different names, so instead of learning the names once, you'll have to keep translating between the class names and the ones used in the documentation.

If you really want to rename it, at least use collection names, i.e. `Map`. We already have `ListTag` so there's precedent, `Map` is unambiguous and actually describes the data structure, and everyone who has touched Java knows what it refers to. But I'd still want to keep the names of what the tag types are actually called so that it's consistent with everything else written about NBT (either by Mojang or third parties), and it's really not hard to learn the meaning of one class name.

---

As for where to put the `Nbt` in the name to make it clear that all these classes refer to NBT (even though tbh I already think it's clear enough since they're in the `nbt` package), I'd either use e.g. `StringNbtTag` or `NbtStringTag`, preferring the former. Sure, it would be longer than any of `NbtString` or `StringNbt`, but leaving away the `Tag` feels wrong, since all these classes aren't strings, or integers, or what have you, they're tags that contain/wrap those types.
Take the ZIP format for example. If you'd write a ZIP reader, you'd also call the file/directory entry classes `FileZipEntry` or `DirectoryZipEntry` (with the base class being `ZipEntry`) or even just `FileEntry`/`DirectoryEntry` (that would be the format we have currently, e.g. `StringTag`), and not `ZipFile` or `FileZip`. At least I would. (this might not be how you'd actually write a ZIP reader, but it's just an example)

kvverti:
I would disagree on calling the NBT data elements “tags” since tag is generally used to refer to the keys in a key-value store, not the values. I think using “tag” to refer to int, string, etc. values is a case of bad terminology by Mojang.

kvverti:
Ok so the [original specification](http://web.archive.org/web/20110723210920/http://www.minecraft.net/docs/NBT.txt) (according to Minecraft Wiki) uses tag (specifically Named Tag) to mean a structure of data type, length, and payload. It _also_ uses tag (plain Tag) to refer to only the payload. So take that as you will.

Sollace:
@kvverti It's Tags all the way down!

Pyrofab:
I agree with farfetchd on Object being inappropriate for NBT, however I would dislike making `NbtTag` a thing. I find both `NbtString` and `ZipFile` acceptable concise names.

ChloeDawn:
`NbtTag` is equal to `Named Binary Tag Tag`, which is nonsensical

ChloeDawn:
I disagree that `Nbt` should be a prefix in the naming pattern, as it introduces an ambiguity between data classes (e.g. `NbtString`) and utility classes (e.g. `NbtIo`). Suffixes should be the standard here for the data classes.

dblsaiko:
Not really. NBT refers to the format, so it makes perfect sense, it's a tag in the NBT format. If you expand JsonObject to "JavaScript Object Notation Object" it also sounds dumb

ChloeDawn:
> Not really. NBT refers to the format, so it makes perfect sense, it's a tag in the NBT format. If you expand JsonObject to "JavaScript Object Notation Object" it also sounds dumb

"This naming pattern sounds dumb but we will use it because someone else did"

mappings/net/minecraft/block/entity/ShulkerBoxBlockEntity.mapping
```diff
@@ -16,10 +16,10 @@ CLASS net/minecraft/class_2627 net/minecraft/block/entity/ShulkerBoxBlockEntity
 	METHOD method_11315 getCollisionBox (Lnet/minecraft/class_2350;)Lnet/minecraft/class_238;
 		ARG 1 facing
 	METHOD method_11316 pushEntities ()V
-	METHOD method_11317 serializeInventory (Lnet/minecraft/class_2487;)Lnet/minecraft/class_2487;
+	METHOD method_11317 saveInventoryToNbt (Lnet/minecraft/class_2487;)Lnet/minecraft/class_2487;
```
ChloeDawn, line 19: Could methods such as these just be simplified to `saveInventory`?

mappings/net/minecraft/command/arguments/NbtPathArgumentType.mapping
```diff
@@ -53,28 +53,24 @@ CLASS net/minecraft/class_2203 net/minecraft/command/arguments/NbtPathArgumentTy
...
-			ARG 1 tag
 			ARG 2 source
 		METHOD method_9368 put (Lnet/minecraft/class_2520;Ljava/util/function/Supplier;)I
-			ARG 1 tag
 			ARG 2 source
 		METHOD method_9369 getTerminals (Lnet/minecraft/class_2520;)Ljava/util/List;
 			ARG 1 start
 		METHOD method_9371 forEach (Ljava/util/List;Ljava/util/function/Function;)I
-			ARG 0 tags
+			ARG 0 nbts
```
ChloeDawn, line 62: Collections of NBT can probably still be referred to as `tags` within scopes like this. It is more readable than `nbts` IMO.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -517,8 +517,8 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 		ARG 1 type
 		ARG 2 movement
 	METHOD method_5785 scheduleVelocityUpdate ()V
-	METHOD method_5786 saveSelfToTag (Lnet/minecraft/class_2487;)Z
-		ARG 1 tag
+	METHOD method_5786 saveSelfToNbt (Lnet/minecraft/class_2487;)Z
```
ChloeDawn, line 520: Is `Self` needed here?

dblsaiko:
> "This naming pattern sounds dumb but we will use it because someone else did"

When did I ever say anything like that? I wasn't referring to the naming pattern, but to "hey it has the word 'tag' in there twice if you expand the acronym".

Pyrofab:
On the topic of the naming pattern, I don't think anyone was ever confused by `JsonReader` being a reader for the json format vs `JsonObject` being a data class.

mappings/net/minecraft/item/map/MapState.mapping
```diff
@@ -55,7 +55,7 @@ CLASS net/minecraft/class_22 net/minecraft/item/map/MapState
 		ARG 1 world
 		ARG 2 x
 		ARG 3 z
-	METHOD method_110 addDecorationsTag (Lnet/minecraft/class_1799;Lnet/minecraft/class_2338;Ljava/lang/String;Lnet/minecraft/class_20$class_21;)V
+	METHOD method_110 saveDecorationsToNbt (Lnet/minecraft/class_1799;Lnet/minecraft/class_2338;Ljava/lang/String;Lnet/minecraft/class_20$class_21;)V
```
ChloeDawn, line 58: `addDecorationToNbt`

ChloeDawn (changes requested):

mappings/net/minecraft/structure/Structure.mapping
```diff
@@ -6,14 +6,16 @@ CLASS net/minecraft/class_3499 net/minecraft/structure/Structure
...
+	METHOD method_15175 saveToNbt (Lnet/minecraft/class_2487;)Lnet/minecraft/class_2487;
+		ARG 1 nbt
 	METHOD method_15181 getAuthor ()Ljava/lang/String;
 	METHOD method_15182 place (Lnet/minecraft/class_1936;Lnet/minecraft/class_2338;Lnet/minecraft/class_3492;)V
-	METHOD method_15183 fromTag (Lnet/minecraft/class_2487;)V
-		ARG 1 tag
-	METHOD method_15184 createDoubleListTag ([D)Lnet/minecraft/class_2499;
+	METHOD method_15183 loadFromNbt (Lnet/minecraft/class_2487;)V
+		ARG 1 nbt
+	METHOD method_15184 createDoubleListNbt ([D)Lnet/minecraft/class_2499;
```
ChloeDawn, line 17: `asListNbt`

mappings/net/minecraft/structure/Structure.mapping
```diff
@@ -6,14 +6,16 @@ CLASS net/minecraft/class_3499 net/minecraft/structure/Structure
 	METHOD method_15160 getSize ()Lnet/minecraft/class_2338;
 	METHOD method_15161 setAuthor (Ljava/lang/String;)V
 		ARG 1 name
-	METHOD method_15169 createIntListTag ([I)Lnet/minecraft/class_2499;
-	METHOD method_15175 toTag (Lnet/minecraft/class_2487;)Lnet/minecraft/class_2487;
-		ARG 1 tag
+	METHOD method_15169 createIntListNbt ([I)Lnet/minecraft/class_2499;
```
ChloeDawn, line 9: `asListNbt`

mappings/net/minecraft/stat/ServerStatHandler.mapping
```diff
@@ -13,7 +13,8 @@ CLASS net/minecraft/class_3442 net/minecraft/stat/ServerStatHandler
 		ARG 1 dataFixer
 		ARG 2 json
 	METHOD method_14907 getStatId (Lnet/minecraft/class_3445;)Lnet/minecraft/class_2960;
-	METHOD method_14908 jsonToCompound (Lcom/google/gson/JsonObject;)Lnet/minecraft/class_2487;
+	METHOD method_14908 jsonToCompoundNbt (Lcom/google/gson/JsonObject;)Lnet/minecraft/class_2487;
```
ChloeDawn, line 16: `convertToNbt`

mappings/net/minecraft/server/world/ServerTickScheduler.mapping
```diff
@@ -14,7 +14,7 @@ CLASS net/minecraft/class_1949 net/minecraft/server/world/ServerTickScheduler
 		ARG 3 idToName
 		ARG 4 nameToId
 		ARG 5 tickConsumer
-	METHOD method_20469 serializeScheduledTicks (Ljava/util/function/Function;Ljava/lang/Iterable;J)Lnet/minecraft/class_2499;
+	METHOD method_20469 scheduledTicksToNbt (Ljava/util/function/Function;Ljava/lang/Iterable;J)Lnet/minecraft/class_2499;
```
ChloeDawn, line 17: `toNbt`

mappings/net/minecraft/nbt/AbstractNumberNbt.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_2514 net/minecraft/nbt/AbstractNumberTag
+CLASS net/minecraft/class_2514 net/minecraft/nbt/AbstractNumberNbt
```
ChloeDawn, line 1: `NumberNbt`

ChloeDawn (changes requested):

mappings/net/minecraft/command/arguments/NbtPathArgumentType.mapping
Sollace: Nbt is like Moose. The collective term for Nbt is Nbt.

mappings/net/minecraft/entity/Entity.mapping
Sollace: No. I don't think it is.

mappings/net/minecraft/structure/Structure.mapping
Sollace: `asNbtList`

mappings/net/minecraft/nbt/AbstractNumberNbt.mapping
Sollace: Ignoring the Nbt naming for now, this could have just been `NumberTag`. There's no confusion that `Number` in Java is the common root class to all numeric object types.

PR #1078, opened 2020-01-26
NBT method renames without class renames

Prospector:
Basically #1077 but without the controversial class renames. I renamed a few that made sense and referred to NBT as a format and not a tag.

Edit: the rationale behind the various method name changes, copied from the other PR:

> I have made changes to many method names using this set of rules:
> * Methods which convert an **object into NBT** will either be named:
>   * `toNbt` - when an existing NBT is not passed into the method as a parameter. This method entirely creates the NBT which is returned.
>   * `saveToNbt` - when either the method has a void return type or an existing NBT is passed into the method, meaning this method does not create the NBT that is returned.
> * Methods which convert **NBT into an object** will either be named:
>   * `fromNbt` - when the method creates an object using the NBT data passed in as a parameter.
>   * `loadFromNbt` - when the method uses NBT data to modify an existing instance of an object.


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -246,10 +246,10 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
...
 	METHOD method_5637 isWet ()Z
 		COMMENT Returns whether this entity is touching water, or is being rained on, or is inside a bubble column...
-		COMMENT 
-		COMMENT @see net.minecraft.entity.Entity#isTouchingWater() 
-		COMMENT @see net.minecraft.entity.Entity#isBeingRainedOn() 
-		COMMENT @see net.minecraft.entity.Entity#isInsideBubbleColumn() 
+		COMMENT
+		COMMENT @see net.minecraft.entity.Entity
+		COMMENT @see net.minecraft.entity.Entity
+		COMMENT @see net.minecraft.entity.Entity
```
i509VCB: Yep Engima tripped balls again

mappings/net/minecraft/inventory/ContainerLock.mapping
```diff
@@ -11,11 +11,7 @@ CLASS net/minecraft/class_1273 net/minecraft/inventory/ContainerLock
...
 		ARG 1 stack
 			COMMENT the key item stack
-	METHOD method_5473 fromTag (Lnet/minecraft/class_2487;)Lnet/minecraft/class_1273;
-		COMMENT Creates a new {@code ContainerLock} from the {@code Lock} key of the compound tag.
-		COMMENT <p>
-		COMMENT If the {@code Lock} key is not present, returns an empty lock.
-		ARG 0 tag
-	METHOD method_5474 toTag (Lnet/minecraft/class_2487;)V
-		COMMENT Inserts the key string of this lock into the {@code Lock} key of the compound tag.
-		ARG 1 tag
```
i509VCB, line 21: So I guess enigma needs an update to not nuke comments.

mappings/net/minecraft/entity/Entity.mapping
quat1024: Oi!

mappings/net/minecraft/inventory/ContainerLock.mapping
```diff
@@ -11,11 +11,7 @@ CLASS net/minecraft/class_1273 net/minecraft/inventory/ContainerLock
...
+	METHOD method_5473 fromNbt (Lnet/minecraft/class_2487;)Lnet/minecraft/class_1273;
```
quat1024, line 14: Comments got deleted

i509VCB (approved):
These nuked comments are related to https://github.com/FabricMC/Enigma/issues/190

mappings/net/minecraft/entity/Entity.mapping
Juuxel: Maybe Enigma really wants you to see `Entity`?

quat1024 (approved):
A few doc comments seem to have been screwed up again. Overall I really like this though. I was initially confused to the distinction between `toNbt`/`saveToNbt` but the justification in the other issue makes sense.

mappings/net/minecraft/inventory/ContainerLock.mapping
Prospector: I fixed manually

mappings/net/minecraft/block/entity/BlockEntity.mapping
```diff
@@ -31,12 +33,12 @@ CLASS net/minecraft/class_2586 net/minecraft/block/entity/BlockEntity
...
-	METHOD method_11014 fromTag (Lnet/minecraft/class_2487;)V
-		ARG 1 tag
+	METHOD method_11014 loadFromNbt (Lnet/minecraft/class_2487;)V
+		ARG 1 nbt
 	METHOD method_11015 isRemoved ()Z
 	METHOD method_11016 getPos ()Lnet/minecraft/class_2338;
 	METHOD method_11017 getType ()Lnet/minecraft/class_2591;
 	METHOD method_16886 toUpdatePacket ()Lnet/minecraft/class_2622;
-	METHOD method_16887 toInitialChunkDataTag ()Lnet/minecraft/class_2487;
+	METHOD method_16887 initialChunkDataToNbt ()Lnet/minecraft/class_2487;
```
Pyrofab: This name could be improved. Wouldn't `getInitialChunkData` be more appropriate ? Javadoc could help too.

mappings/net/minecraft/block/entity/LootableContainerBlockEntity.mapping
```diff
@@ -6,14 +6,17 @@ CLASS net/minecraft/class_2621 net/minecraft/block/entity/LootableContainerBlock
...
 	METHOD method_11282 getInvStackList ()Lnet/minecraft/class_2371;
-	METHOD method_11283 deserializeLootTable (Lnet/minecraft/class_2487;)Z
+	METHOD method_11283 lootTableFromNbt (Lnet/minecraft/class_2487;)Z
+		ARG 1 nbt
 	METHOD method_11285 setLootTable (Lnet/minecraft/class_2960;J)V
 		ARG 1 id
 		ARG 2 seed
-	METHOD method_11286 serializeLootTable (Lnet/minecraft/class_2487;)Z
+	METHOD method_11286 lootTableToNbt (Lnet/minecraft/class_2487;)Z
+		ARG 1 nbt
```
Pyrofab: I don't think this is really an improvement. `(de)serializeLootTable` feels much more natural.

mappings/net/minecraft/client/options/HotbarStorageEntry.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS net/minecraft/class_748 net/minecraft/client/options/HotbarStorageEntry
 	FIELD field_3948 delegate Lnet/minecraft/class_2371;
-	METHOD method_3152 fromListTag (Lnet/minecraft/class_2499;)V
-	METHOD method_3153 toListTag ()Lnet/minecraft/class_2499;
+	METHOD method_3152 loadFromListNbt (Lnet/minecraft/class_2499;)V
+		ARG 1 nbt
+	METHOD method_3153 toListNbt ()Lnet/minecraft/class_2499;
```
Pyrofab: Couldn't it be `loadFromNbt`/`toNbt` ? No need to restate the parameter's type.

mappings/net/minecraft/entity/effect/StatusEffectInstance.mapping
```diff
@@ -50,22 +50,22 @@ CLASS net/minecraft/class_1293 net/minecraft/entity/effect/StatusEffectInstance
...
-	METHOD method_24275 fromTag (Lnet/minecraft/class_1291;Lnet/minecraft/class_2487;)Lnet/minecraft/class_1293;
+	METHOD method_24275 fromNbt (Lnet/minecraft/class_1291;Lnet/minecraft/class_2487;)Lnet/minecraft/class_1293;
 		ARG 0 type
-		ARG 1 tag
+		ARG 1 nbt
 	METHOD method_24276 copyFrom (Lnet/minecraft/class_1293;)V
 		ARG 1 that
-	METHOD method_24277 typelessToTag (Lnet/minecraft/class_2487;)V
-		ARG 1 tag
+	METHOD method_24277 saveTypelessToNbt (Lnet/minecraft/class_2487;)V
```
Pyrofab: Not a very clear name, could use doc.

liach:
Can we use `saveTo` instead of `saveToNbt` and `loadFrom` instead of `loadFromNbt`, as nbt can be inferred from param types?

Prospector:
IMO we should keep the Nbt part for reasons I said on the other PR: if someone's looking for a method to save to NBT then they may type `nbt` and look for something that autocompletes. Also just consistency with the other methods.

mappings/net/minecraft/item/ItemStack.mapping
```diff
@@ -31,7 +31,7 @@ CLASS net/minecraft/class_1799 net/minecraft/item/ItemStack
 	METHOD method_7910 finishUsing (Lnet/minecraft/class_1937;Lnet/minecraft/class_1309;)Lnet/minecraft/class_1799;
 		ARG 1 world
 		ARG 2 user
-	METHOD method_7911 getOrCreateSubTag (Ljava/lang/String;)Lnet/minecraft/class_2487;
+	METHOD method_7911 getOrCreateSubNbt (Ljava/lang/String;)Lnet/minecraft/class_2487;
```
Pyrofab: Not sure about this rename. You are working with a sub tag, not some sub NBT anything.

TheGlitch76 (approved):

mappings/net/minecraft/loot/function/CopyNbtLootFunction.mapping
```diff
@@ -60,19 +60,19 @@ CLASS net/minecraft/class_3837 net/minecraft/loot/function/CopyNbtLootFunction
...
 			METHOD method_16866 (Lnet/minecraft/class_2520;Lnet/minecraft/class_2520;)V
-				ARG 1 listTag
+				ARG 1 listNbt
 			METHOD method_16867 (Ljava/util/List;Lnet/minecraft/class_2520;)V
-				ARG 1 foundTag
+				ARG 1 foundNbt
 		CLASS 3
 			METHOD method_16868 (Lnet/minecraft/class_2520;Lnet/minecraft/class_2520;)V
-				ARG 1 compoundTag
+				ARG 1 coumpoundNbt
```
Pyrofab: Still a compoundTag if we don't rename the classes.

mappings/net/minecraft/nbt/NbtHelper.mapping
```diff
@@ -1,17 +1,17 @@
...
-		ARG 0 tag
+		ARG 0 nbt
 	METHOD method_10682 withProperty (Lnet/minecraft/class_2688;Lnet/minecraft/class_2769;Ljava/lang/String;Lnet/minecraft/class_2487;Lnet/minecraft/class_2487;)Lnet/minecraft/class_2688;
 		ARG 0 state
 		ARG 1 property
 		ARG 2 key
-		ARG 3 propertiesTag
-		ARG 4 mainTag
+		ARG 3 propertiesNbt
+		ARG 4 mainNbt
```
Pyrofab: Specifically a tag.

mappings/net/minecraft/nbt/StringNbt.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_2519 net/minecraft/nbt/StringTag
+CLASS net/minecraft/class_2519 net/minecraft/nbt/StringNbt
```
Pyrofab: Isn't this a class rename of a core NBT type ?

mappings/net/minecraft/nbt/StringNbtReader.mapping
```diff
@@ -16,19 +16,19 @@ CLASS net/minecraft/class_2522 net/minecraft/nbt/StringNbtReader
...
 	METHOD method_10725 readString ()Ljava/lang/String;
-	METHOD method_10726 parseTagPrimitiveArray ()Lnet/minecraft/class_2520;
-	METHOD method_10727 parseCompoundTag ()Lnet/minecraft/class_2487;
+	METHOD method_10726 parseNbtPrimitiveArray ()Lnet/minecraft/class_2520;
+	METHOD method_10727 parseCompoundNbt ()Lnet/minecraft/class_2487;
 	METHOD method_10728 readArray (Lnet/minecraft/class_4614;Lnet/minecraft/class_4614;)Ljava/util/List;
 		ARG 1 arrayTypeReader
 		ARG 2 typeReader
-	METHOD method_10729 parseListTag ()Lnet/minecraft/class_2520;
+	METHOD method_10729 parseListNbt ()Lnet/minecraft/class_2520;
```
Pyrofab: Those renames do not make sense outside of the other PR.

mappings/net/minecraft/stat/ServerStatHandler.mapping
```diff
@@ -13,7 +13,8 @@ CLASS net/minecraft/class_3442 net/minecraft/stat/ServerStatHandler
 		ARG 1 dataFixer
 		ARG 2 json
 	METHOD method_14907 getStatId (Lnet/minecraft/class_3445;)Lnet/minecraft/class_2960;
-	METHOD method_14908 jsonToCompound (Lcom/google/gson/JsonObject;)Lnet/minecraft/class_2487;
+	METHOD method_14908 jsonToCompoundNbt (Lcom/google/gson/JsonObject;)Lnet/minecraft/class_2487;
```
Pyrofab: Why not just `jsonToNbt` ?

mappings/net/minecraft/structure/Structure.mapping
```diff
@@ -6,14 +6,16 @@ CLASS net/minecraft/class_3499 net/minecraft/structure/Structure
 	METHOD method_15160 getSize ()Lnet/minecraft/class_2338;
 	METHOD method_15161 setAuthor (Ljava/lang/String;)V
 		ARG 1 name
-	METHOD method_15169 createIntListTag ([I)Lnet/minecraft/class_2499;
-	METHOD method_15175 toTag (Lnet/minecraft/class_2487;)Lnet/minecraft/class_2487;
-		ARG 1 tag
+	METHOD method_15169 createIntListNbt ([I)Lnet/minecraft/class_2499;
```
Pyrofab: Does not make sense outside of the other PR.

mappings/net/minecraft/structure/Structure.mapping
```diff
@@ -6,14 +6,16 @@ CLASS net/minecraft/class_3499 net/minecraft/structure/Structure
...
+	METHOD method_15175 saveToNbt (Lnet/minecraft/class_2487;)Lnet/minecraft/class_2487;
+		ARG 1 nbt
 	METHOD method_15181 getAuthor ()Ljava/lang/String;
 	METHOD method_15182 place (Lnet/minecraft/class_1936;Lnet/minecraft/class_2338;Lnet/minecraft/class_3492;)V
-	METHOD method_15183 fromTag (Lnet/minecraft/class_2487;)V
-		ARG 1 tag
-	METHOD method_15184 createDoubleListTag ([D)Lnet/minecraft/class_2499;
+	METHOD method_15183 loadFromNbt (Lnet/minecraft/class_2487;)V
+		ARG 1 nbt
+	METHOD method_15184 createDoubleListNbt ([D)Lnet/minecraft/class_2499;
```
Pyrofab: Does not make sense outside of the other PR.

mappings/net/minecraft/world/level/LevelProperties.mapping
```diff
@@ -43,15 +43,20 @@ CLASS net/minecraft/class_31 net/minecraft/world/level/LevelProperties
 	FIELD field_190 raining Z
 	FIELD field_191 scheduledEvents Lnet/minecraft/class_236;
 	FIELD field_192 rainTime I
-	FIELD field_193 worldData Ljava/util/Map;
+	FIELD field_193 worldNbt Ljava/util/Map;
```
Pyrofab: Does this rename really make sense here ?

mappings/net/minecraft/block/entity/LootableContainerBlockEntity.mapping
Prospector: I did it for consistency with the other names. Serialize to what? Now it tells you. We didn't want to use serialize/deserialize because Minecraft uses a bunch of different serialization formats.

mappings/net/minecraft/block/entity/BlockEntity.mapping
Prospector: I think no, because this creates an nbt tag, it doesn't return an existing one.

liach:
doesn't ide auto infer what you type as param types when you type? what i said is recommended by effective javs

mappings/net/minecraft/block/entity/LootableContainerBlockEntity.mapping
Pyrofab: Couldn't we find a name that doesn't require a second take every time you read it ?

mappings/net/minecraft/block/entity/BlockEntity.mapping
Pyrofab: I personally think that's just a detail. However, if you want to be consistent, it's still the block entity that's being serialized to some initial chunk data NBT, not an `initialChunkData` property that is being serialized to NBT.

Prospector: ah ok I didn't look too much into what it actually does, just reformatting the names to the new format. What would you suggest?

mappings/net/minecraft/block/entity/LootableContainerBlockEntity.mapping
Prospector: such as?

mappings/net/minecraft/client/options/HotbarStorageEntry.mapping
Prospector: yup you're right that shouldn't be like that

mappings/net/minecraft/entity/effect/StatusEffectInstance.mapping
Prospector: agreed, but again not my name, just reformatted the existing name. I don't think typeless is a good description of what it does but I'm not sure what would be better. This method basically adds all the info of a status effect such as amplifier and all that stuff but without a status effect type.

mappings/net/minecraft/item/ItemStack.mapping
Prospector: hmm NBT should still be somewhere in the name though imo. getOrCreateNbtSubTag?

mappings/net/minecraft/nbt/NbtHelper.mapping
Prospector: as opposed to what other type of nbt? But fair enough this is in the NbtHelper class so using tag is fine here I suppose

mappings/net/minecraft/nbt/StringNbt.mapping
Prospector: yup needs fixing. guess I missed reverting that one

mappings/net/minecraft/nbt/StringNbtReader.mapping
Prospector: indeed

mappings/net/minecraft/stat/ServerStatHandler.mapping
Prospector: good call

mappings/net/minecraft/world/level/LevelProperties.mapping
Prospector: it seems strange that it is referred to as "data" here when everywhere else we call nbt what it is...nbt.

mappings/net/minecraft/block/entity/BlockEntity.mapping
liach: This is like serializing this block entity to be sent to the client, usually used for initial block entity sending in chunk data packet and used by many block entities to have this data sent in update packet as well. Maybe some `toClientboundNbt` or so?

mappings/net/minecraft/block/entity/LootableContainerBlockEntity.mapping
liach: Should be `loadLootTableFromNbt` and `saveLootTableToNbt` according to your (prospector's) rationale, but I'd prefer `loadLootTableFrom` and `saveLootTableTo` (see param types)

mappings/net/minecraft/client/options/HotbarStorageEntry.mapping
liach: So why not just `loadFrom` and `toNbt`

mappings/net/minecraft/entity/effect/StatusEffectInstance.mapping
liach: This is to prevent writing the extraneous status effect type info to the hidden effect tag. Feel free to come up with better names like `saveInfoToNbt`

mappings/net/minecraft/item/ItemStack.mapping
liach: This looks good to me.

mappings/net/minecraft/nbt/NbtHelper.mapping
liach: Why not just `properties` `main`

mappings/net/minecraft/item/ItemStack.mapping
Pyrofab: I'd rather avoid redundancy. If you think `nbt` is necessary, the name you gave is fine.

mappings/net/minecraft/block/entity/LootableContainerBlockEntity.mapping
Prospector: ah yes, that one does pass nbt doesn't it

mappings/net/minecraft/entity/effect/StatusEffectInstance.mapping
Prospector: I went with savePropertiesToNbt since that's a bit more descriptive than info.

mappings/net/minecraft/item/ItemStack.mapping
Prospector: how about getOrCreateNbtSubTag?

mappings/net/minecraft/block/entity/BlockEntity.mapping
Prospector: Resolved as `toUpdateNbt` to match `toUpdatePacket`

Prospector:
Not sure but certainly it can't infer that a `CompoundTag` is in any way related to `Nbt`

mappings/net/minecraft/item/ItemStack.mapping
Pyrofab: That works, I guess.

Prospector: Oops lol I didn't notice I came up with that yesterday too and already asked lol.

liach:
@Prospector Is your updated rationale like this:

- Methods which convert an object into NBT will either be named:
  - `toNbt` - when an existing NBT is not passed into the method as a parameter. This method entirely creates the NBT which is returned.
  - `save` - when either the method has a void return type or an existing NBT is passed into the method, meaning this method does not create the NBT that is returned.
- Methods which convert NBT into an object will either be named:
  - `fromNbt` - when the method creates an object using the NBT data passed in as a parameter.
  - `load` - when the method uses NBT data to modify an existing instance of an object.

Prospector:
Indeed. The only reason I didn't switch `toNbt` and `fromNbt` to `to` and `from` is because I don't think it would be intuitive when overriding methods from a super.

dblsaiko:
> Indeed. The only reason I didn't switch `toNbt` and `fromNbt` to `to` and `from` is because I don't think it would be intuitive when overriding methods from a super.

That definitely is the problem with using `to`, additionally naming it `to` would mean that similar method that create something else instead of NBT can't follow the same naming style (there can only be one `to` with no parameters.

But that's not really the problem with `from`, since those are always static so you can't override them anyway (at least, it would make sense for them to always be static, since otherwise you'd already have to have an instance of the object you're trying to create), and there you have a parameter that you can distinguish the method by.

So I'd suggest having `toNbt()` and `from()`

Player3324:
Relying on the signature is bad for readability, especially when reviewing at a call site in Github or similar. The name alone should describe enough.

It'd be very unfortunate if we dropped the universal convention `to<formatName>`/`from<formatName>` for serialization methods. I am not happy with the trend towards overly short names.

Prospector:
We agree save and load are fine though because there's usually just 1 way to save and load, right?

Player3324:
"We" definitely doesn't include me, I've always been in favor of consistently specifying the format like Yarn usually did.
It makes little difference whether there are multiple ways to serialize something, it is all about informing the reader about both the format and the conversion-like behavior. "save" could easily be understood as self contained as in `void xy() { something.save(); }`, interacting with external storage autonomously. Java wouldn't even complain if save wasn't self contained and did return something while the program ignored it that way.

liach:
Imo method params are always visible in enigma/tiny format and always visible in actual code usage. We are relying on descriptor instead of generic signature (and we indeed should specify when generic signature matters as this is what is supposed to happen in java language)

Player3324:
They are invisible at the call site in code unless you use an IDE/plugin that embeds them into the text, hover over it specifically or go to the declaration with F3. This gets much worse without an IDE (like when reviewing a PR on Github..) or if the serialized form is not identifiable by the data type (JSON String..).

Prospector:
I'm certainly not opposed to reverting it back to `saveToNbt` and `loadFromNbt` but really I think in certain cases like BlockEntities, having just `save` and `load` is perfectly fine and the to/from NBT just adds noise. When something *only* saves to NBT and consumes a CompoundTag (which we may name to NbtCompound or something similar with Nbt in the name very soon), it really seems redundant to have the `to/fromNbt` suffixes.

grondag:
I prefer having Nbt in the names.  It is better in those situations that Player notes when you see method names without parameters or extensive context, but even with IDE support I find it faster to read and understand. It may be that `save` and `load` are unambiguous today, but then  you add, say, packet or JSON serialization for whatever reason.

Player3324:
I wasn't talking about `saveToNbt`, but `toNbt` as it was decided for Yarn a while ago. Specifying the originally intended use case by adding `save` or `load` prefixes is only marginally useful. I don't know how you could get there since the original already was correctly using `to<format>` precisely as just described - the only difference is that `Tag` is now the more precise `Nbt`, which is a good thing.

liach:
So there are 4 cases of s/l afaik:

Action | Save | Load
------- | ---- | ----
Modifies existing object | Modifies passed NBT (and return) | Modifies `this` based on NBT
Creates new object | Creates new NBT and return | (Static) Creates new instance from NBT

So the naming proposal:

Type | Modifies passed NBT (and return) | Modifies `this` based on NBT | Creates new NBT and return | (Static) Creates new instance from NBT
------ | ----------------------------------- | ----------------------------------- | ----------------------------------- | -----------------------------------
Old name | `void toNbt(Tag)` `Tag toNbt(Tag)` | `void fromNbt(Tag)` | `Tag toNbt()` | `static Object loadFromNbt(Tag)`/`static Object readFromNbt(Tag)`
My current proposal | `void save(Tag)` (`void saveTo(Tag)`) | `void load(Tag)` `void loadFrom(Tag)` | `Tag toNbt()` | `static Object fromNbt()`

This naming scheme is getting a bit confusing... Hope we can find a clear way out.

liach:
Another distinguishing suggested by player:
So we will now have:
 - `static Object fromNbt(Tag)`
 - `void fromNbt(Tag)` 
   - Player suggests `setFromNbt` would be a closer name, but will keep this scheme
 - `void toNbt(Tag)`
 - `Tag toNbt()`

Prospector:
I've reverted the change that simplified `saveToNbt` to just `save` since Player brought up a decent point. I would still like to include the save prefix when an existing tag is being modified and the load prefix when an existing instance is being modified.

Prospector:
can we get a current 👍 and 👎 reaction status on how people feel about this?

Juuxel:
Now that I think of it, I think it would be simpler just to have the model that Player suggested in liach's comment.

Prospector:
I will be closing this and re-making it based on 20w15a

PR #1079, opened 2020-01-28
1.15.2 EntityRenderer changes

Sollace:
Fills in a few missing names in `LivingEntityRenderer`.

I only realise now that this might conflict with my other PR, which hasn't been merged, but I can fix that if this one is merged first.

PR #1080, opened 2020-01-29
LivingEntity rendering fix

Pyrofab:
Renamed `Entity#canSeePlayer` to `isInvisibleTo`.
It returns true if an entity is invisible and not revealed by spectator mode or scoreboard team setting, and is used by LivingEntityRenderer to check if an entity should be rendered as translucent.
The other names were previously unmapped.


Pyrofab:
Oops, seems like this overlaps with #1079

liach (approved):

YanisBft (approved):

Juuxel (approved):

modmuss50 (approved):

Issue #997, opened 2019-12-07
ColoredBlock named wrong


Runemoro:
It's implemented by the beacon, stained glass, but not wool or terracotta. It's used to dye the beacon beam.

`TransparentColoredBlock`?

liach:
merging #867 into this one.

Sollace:
What is wrong with calling it `ColoredBlock`?

(typing the murkanization like that hurts my brain)

Runemoro:
> What is wrong with calling it ColoredBlock?

Without having read this issue, what would you have thought this code does when the player walks on some colored wool?

```java
if (blockUnderPlayer instanceof ColoredBlock) {
    player.sendMessage("You're standing on a " + ((ColoredBlock) blockUnderPlayer).getColor() + " block");
}
```

Also, I wasted time implementing this in a mod I'm porting to 1.15, not knowing that I wasn't even supposed to implement it for non-transparent blocks.

liach:
StainedBlock?

Sollace:
> 
> 
> > What is wrong with calling it ColoredBlock?
> 
> Without having read this issue, what would you have thought this code does when the player walks on some colored wool?
> 
> ```java
> if (blockUnderPlayer instanceof ColoredBlock) {
>     player.sendMessage("You're standing on a " + ((ColoredBlock) blockUnderPlayer).getColor() + " block");
> }
> ```
> 
> Also, I wasted time implementing this in a mod I'm porting to 1.15, not knowing that I wasn't even supposed to implement it for non-transparent blocks.

I'd imagine it would print "You're standing on a green block".

Was this added when mojang put in all the coloured glass, clays, and terracottas? Seems like a general-use tinted block to me.

Runemoro:
> I'd imagine it would print "You're standing on a green block".

Except it doesn't, which is exactly the problem! Wool and other non-transparent colored blocks don't implement this class. Only the stained glass blocks do (used by the beacon to get new beam color).

liach:
BeaconBeamStainable?

Runemoro:
> BeaconBeamStainable

That makes it seem like the beacon beam stains it. Also, it could be used by mods to color their own light beam effects. I think the map renderer is also using this to adjust the color of thr blocks under it.

Is anything wrong with ColoredTransparentBlock? Another option could be ColorFilter.

YanisBft:
can be closed

Issue #995, opened 2019-12-05
ConnectedPlantBlock -> ConnectingBlock


Runemoro:
As UpcraftLP said on Discord, this is used for the chorus stem, but there's nothing that's specific to plants.  Putting this in an issue so that we don't forget about it.

It would be an improvement to change Connected to Connecting, since it doesn't always have to be connected.

YanisBft:
can be closed

PR #1083, opened 2020-02-04
A couple of block class renamings

### LINKED ISSUE #867: client.block.ColoredBlock -> block.ColoredBlock
This method is used by beacon's tick, which is present on both sides. as a result, it should not be in the client package.

YanisBft:
Fixes #867 , #997 , #995 

`ColoredBlock` has been renamed to `Stainable`, to match the other interfaces implemented by block classes (`Fertilizable`, `Waterloggable`, etc).

Also mapped missing parameters in `BeehiveBlock`.


Juuxel:
I think `class_2429` could still be `ConnectedBlock` because we also have `HorizontalConnectedBlock`.

modmuss50 (approved):
Looks good.

YanisBft:
> I think `class_2429` could still be `ConnectedBlock` because we also have `HorizontalConnectedBlock`.

As Runemoro said tho, the block can be connected or not, so "connected" is not appropriate. Should I rename `HorizontalConnectedBlock` to be consistent?

liach (approved):

PR #1087, opened 2020-02-05
Biome internal mappings (not new biome types)

liach:

### DIFF
--- a/mappings/net/minecraft/util/math/Vec3d.mapping
+++ b/mappings/net/minecraft/util/math/Vec3d.mapping
@@ -50,5 +50,5 @@ CLASS net/minecraft/class_243 net/minecraft/util/math/Vec3d
 	METHOD method_18806 multiply (Lnet/minecraft/class_243;)Lnet/minecraft/class_243;
 		ARG 1 mult
 	METHOD method_22882 negate ()Lnet/minecraft/class_243;
-	METHOD method_24457 breakDownRgb (I)Lnet/minecraft/class_243;
+	METHOD method_24457 unpackRgb (I)Lnet/minecraft/class_243;
 		ARG 0 rgb
--- a/mappings/net/minecraft/util/math/Vec3d.mapping
+++ b/mappings/net/minecraft/util/math/Vec3d.mapping
@@ -50,3 +50,5 @@ CLASS net/minecraft/class_243 net/minecraft/util/math/Vec3d
 	METHOD method_18806 multiply (Lnet/minecraft/class_243;)Lnet/minecraft/class_243;
 		ARG 1 mult
 	METHOD method_22882 negate ()Lnet/minecraft/class_243;
+	METHOD method_24457 breakDownRgb (I)Lnet/minecraft/class_243;
+		ARG 0 rgb
--- a/mappings/net/minecraft/world/biome/Biome.mapping
+++ b/mappings/net/minecraft/world/biome/Biome.mapping
@@ -1,8 +1,8 @@
 CLASS net/minecraft/class_1959 net/minecraft/world/biome/Biome
 	FIELD field_20335 temperatureCache Ljava/lang/ThreadLocal;
 	FIELD field_21806 skyColor I
-	FIELD field_22039 specialEffects Lnet/minecraft/class_4763;
-	FIELD field_22040 noises Ljava/util/List;
+	FIELD field_22039 effects Lnet/minecraft/class_4763;
+	FIELD field_22040 noisePoints Ljava/util/List;
 	FIELD field_9323 BIOMES Ljava/util/Set;
 	FIELD field_9324 FOLIAGE_NOISE Lnet/minecraft/class_3543;
 	FIELD field_9325 spawns Ljava/util/Map;
@@ -30,7 +30,7 @@ CLASS net/minecraft/class_1959 net/minecraft/world/biome/Biome
 	METHOD method_24374 getNoisePenalty (Lnet/minecraft/class_1959$class_4762;)F
 		ARG 1 noise
 	METHOD method_24376 getFogColor ()I
-	METHOD method_24377 getSpecialEffects ()Lnet/minecraft/class_4763;
+	METHOD method_24377 getEffects ()Lnet/minecraft/class_4763;
 	METHOD method_24378 getParticleConfig ()Ljava/util/Optional;
 	METHOD method_8684 hasStructureFeature (Lnet/minecraft/class_3195;)Z
 	METHOD method_8685 canSetIce (Lnet/minecraft/class_4538;Lnet/minecraft/class_2338;Z)Z
@@ -170,12 +170,13 @@ CLASS net/minecraft/class_1959 net/minecraft/world/biome/Biome
 		COMMENT Represents a point in a multi-dimensional cartesian plane. Mixed-noise
 		COMMENT biome generator picks the closest noise point from its selected point
 		COMMENT and choose the biome associated to that closest point. Another factor,
-		COMMENT influence, allows a farther point
-		FIELD field_22043 heat F
+		COMMENT rarity potential, favors larger differences in values instead, contrary
+		COMMENT to other point values.
+		FIELD field_22043 temperature F
 		FIELD field_22044 humidity F
 		FIELD field_22045 hilliness F
 		FIELD field_22046 style F
-		FIELD field_22047 penaltyPeak F
+		FIELD field_22047 rarityPotential F
 			COMMENT This value awards another point with value farthest from this one; i.e.
 			COMMENT unlike other points where closer distance is better, for this value the
 			COMMENT farther the better. The result of the different values can be
@@ -186,6 +187,6 @@ CLASS net/minecraft/class_1959 net/minecraft/world/biome/Biome
 			ARG 2 humidity
 			ARG 3 hilliness
 			ARG 4 style
-			ARG 5 penaltyPeak
+			ARG 5 rarityPotential
 		METHOD method_24381 getPenalty (Lnet/minecraft/class_1959$class_4762;)F
 			ARG 1 other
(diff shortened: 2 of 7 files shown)


mappings/net/minecraft/world/biome/BiomeSpecialEffects.mapping
```diff
@@ -0,0 +1,28 @@
+CLASS net/minecraft/class_4763 net/minecraft/world/biome/BiomeSpecialEffects
```
modmuss50: Whats so special? Why not BiomeEffects?

mappings/net/minecraft/world/biome/Biome.mapping
```diff
@@ -154,3 +166,9 @@ CLASS net/minecraft/class_1959 net/minecraft/world/biome/Biome
 			ARG 2 weight
 			ARG 3 minGroupSize
 			ARG 4 maxGroupSize
+	CLASS class_4762 Noise
+		FIELD field_22047 reduction F
+		METHOD <init> (FFFFF)V
+			ARG 5 reduction
+		METHOD method_24381 getDistance (Lnet/minecraft/class_1959$class_4762;)F
```
Earthcomputer: `getSquaredDistance`

mappings/net/minecraft/util/math/Vec3d.mapping
```diff
@@ -50,3 +50,5 @@ CLASS net/minecraft/class_243 net/minecraft/util/math/Vec3d
 	METHOD method_18806 multiply (Lnet/minecraft/class_243;)Lnet/minecraft/class_243;
 		ARG 1 mult
 	METHOD method_22882 negate ()Lnet/minecraft/class_243;
+	METHOD method_24457 breakDownRgb (I)Lnet/minecraft/class_243;
```
Earthcomputer: `unpackRgb`

mappings/net/minecraft/world/dimension/Dimension.mapping
```diff
@@ -14,7 +14,8 @@ CLASS net/minecraft/class_2869 net/minecraft/world/dimension/Dimension
 		ARG 1 x
 		ARG 2 z
 		ARG 3 checkMobSpawnValidity
-	METHOD method_12445 getFogColor (IF)Lnet/minecraft/class_243;
+	METHOD method_12445 tweakFogColor (IF)Lnet/minecraft/class_243;
```
Earthcomputer: This sounds like it's changing the fog color of the biome, rather than the biome applying its tweaks. How about something like `applyBiomeColorModifier`?

Earthcomputer (changes requested):

mappings/net/minecraft/world/biome/BiomeSpecialEffects.mapping
liach: moj call them special effects. yep should just use effects, will fix

mappings/net/minecraft/world/biome/Biome.mapping
liach: this is conceptual distance and notice that - before reduction square, so not sqr dist.

mappings/net/minecraft/world/dimension/Dimension.mapping
liach: it is like dim shifting fog color, e.g. overworld shift white and blue by tiny bit, nether keep it same, end shift it black significantly

Earthcomputer: Yeah but the name you have there seems to suggest that this is a weirdly named setter for the fog color. What the method actually does is apply a modifier to the input color, and return the result.

mappings/net/minecraft/world/biome/Biome.mapping
```diff
@@ -1,6 +1,8 @@
 CLASS net/minecraft/class_1959 net/minecraft/world/biome/Biome
 	FIELD field_20335 temperatureCache Ljava/lang/ThreadLocal;
 	FIELD field_21806 skyColor I
+	FIELD field_22039 specialEffects Lnet/minecraft/class_4763;
+	FIELD field_22040 noises Ljava/util/List;
```
valoeghese: oh no

mappings/net/minecraft/world/biome/Biome.mapping
```diff
@@ -25,6 +27,11 @@ CLASS net/minecraft/class_1959 net/minecraft/world/biome/Biome
 	METHOD method_21740 getTemperature (Lnet/minecraft/class_2338;)F
 		ARG 1 blockPos
 	METHOD method_24218 calculateSkyColor ()I
+	METHOD method_24374 getNoisePenalty (Lnet/minecraft/class_1959$class_4762;)F
```
valoeghese: "noise penalty"?

mappings/net/minecraft/world/biome/Biome.mapping
```diff
@@ -25,6 +27,11 @@ CLASS net/minecraft/class_1959 net/minecraft/world/biome/Biome
...
+		ARG 1 noise
+	METHOD method_24376 getFogColor ()I
```
valoeghese, line 32: took them long enough to add this

mappings/net/minecraft/world/biome/Biome.mapping
liach: yeah, like penalty between a generator selected noise point and a biome. lowest penalty got picked by the gen

mappings/net/minecraft/world/biome/Biome.mapping
```diff
@@ -154,3 +166,26 @@ CLASS net/minecraft/class_1959 net/minecraft/world/biome/Biome
...
 			ARG 3 minGroupSize
 			ARG 4 maxGroupSize
+	CLASS class_4762 MixedNoisePoint
+		COMMENT Represents a point in a multi-dimensional cartesian plane. Mixed-noise
+		COMMENT biome generator picks the closest noise point from its selected point
+		COMMENT and choose the biome associated to that closest point. Another factor,
+		COMMENT influence, allows a farther point
+		FIELD field_22043 heat F
+		FIELD field_22044 humidity F
+		FIELD field_22045 hilliness F
```
valoeghese, line 177: if we're basing names off of the biome settings then this should be `scale`

mappings/net/minecraft/world/biome/Biome.mapping
```diff
@@ -154,3 +166,26 @@ CLASS net/minecraft/class_1959 net/minecraft/world/biome/Biome
 			ARG 2 weight
 			ARG 3 minGroupSize
 			ARG 4 maxGroupSize
+	CLASS class_4762 MixedNoisePoint
+		COMMENT Represents a point in a multi-dimensional cartesian plane. Mixed-noise
+		COMMENT biome generator picks the closest noise point from its selected point
+		COMMENT and choose the biome associated to that closest point. Another factor,
+		COMMENT influence, allows a farther point
+		FIELD field_22043 heat F
```
valoeghese: similarly this should be `temperature`, as per the biome settings

mappings/net/minecraft/world/biome/Biome.mapping
valoeghese: yeah but why call it penalty

mappings/net/minecraft/world/biome/Biome.mapping
liach: scale? or depth?

mappings/net/minecraft/world/biome/Biome.mapping
valoeghese: make it something chance related, show it's a relative rarity to the target

mappings/net/minecraft/world/biome/Biome.mapping
Earthcomputer: I see no need to be consistent with biome setting names here. These names represent our best guess on what these values are supposed to mean based on the values that different biomes have them set for. Ultimately they aren't really related to the values in biome settings, and can hold different, even contradictory values (see the badlands biome). These names are fine as they are.

liach:
@Earthcomputer Requesting another review.

mappings/net/minecraft/world/biome/BiomeSpecialEffects.mapping
Prospector: well they're specifically visual effects, which are often called special effects.

Prospector (approved):
Not 100% sure on all the names but they're better than what we have (which is nothing)

Earthcomputer:
I still don't like the name `getNoisePenalty`. It is a distance in 4D space, which is then adjusted by `rarityPotential`. A name like `getDistance`, `getNoiseDistance` or `getVoronoiDistance` is better. If you see it necessary, put the fact that it's adjusted by the `rarityPotential`, which is not intuitive as an axis in the 5th dimension, in the Javadocs instead.

liach:
@Earthcomputer fixed.

Earthcomputer (approved):

mappings/net/minecraft/world/biome/BiomeSpecialEffects.mapping
liach: I will just go with `effects`. Don't think fog color or water color is that "special".

mappings/net/minecraft/world/biome/Biome.mapping
```diff
@@ -154,3 +166,40 @@ CLASS net/minecraft/class_1959 net/minecraft/world/biome/Biome
...
+	CLASS class_4762 MixedNoisePoint
+		COMMENT Represents a point in a multi-dimensional cartesian plane. Mixed-noise
+		COMMENT biome generator picks the closest noise point from its selected point
+		COMMENT and choose the biome associated to that closest point. Another factor,
+		COMMENT rarity potential, favors larger differences in values instead, contrary
+		COMMENT to other point values.
+		FIELD field_22043 temperature F
+		FIELD field_22044 humidity F
+		FIELD field_22045 hilliness F
+		FIELD field_22046 style F
```
Juuxel, line 178: `style`?

liach: Something like a style/theme. Similar theme is favored.

Juuxel: What does it do? Edit: yes, I should read these responses better lol

Juuxel (approved):

mappings/net/minecraft/world/biome/Biome.mapping
liach: Dunno, actually all 5 names are arbitrary guesses from hardcoded values given in mojang code and we aren't sure if any of them is correct at all

PR #1090, opened 2020-02-05
New shulker and Item related mappings [20w06a]

i509VCB:
Adds new mapping for where the Shulker box/entity lid collision box is calculated (mainly used to check if the shulker can physically open) and names item settings which determine if an item should not burn.


liach:
I'd prefer shnup's fireproof over unburnable

i509VCB:
Yea I prefer fireproof. @Shnupbups if you want to take that shulker box name I have go ahead.

i509VCB:
Redudant due to #1091

PR #1113, opened 2020-02-12
Standardize EntityData naming

Pyrofab:
A few implementations of `EntityData` had naming collisions, like `LivingEntity.EntityData` or `HorseEntity.EntityData`.
Other names seemed not descriptive enough, like `ZombieEntity.Data`.
I went for the naming scheme used by `FoxEntity.FoxData`, prefixing every implementation with the entity name.


LambdAurora (approved):

YanisBft (approved):

liach:
just Data should be fine for preventing qualified name usage

Pyrofab:
But Data is not explicit enough when used from within the class.

liach:
or rename it to SpawmParameter as that's its primary usage.

liach:
InitParameter, i mean.

Pyrofab:
So you are suggesting to rename the base interface ? What naming scheme should the implementations use then ?

liach:
The base should probably be like `EntityInitParameter` and the subclasses be `InitParameter` (given this class is used in `Entity#init` method)

Pyrofab:
Honestly, InitParameter sounds like the most vague thing. Naming things based on where they are passed as parameters is a bit counterintuitive.

Juuxel:
Could you update this to 20w07a?

modmuss50 (approved):

liach:
After consideration, though the name `Data` may be inaccurate, we can change it in another pr later. Should've approved this one. Thanks pyrofab!

Pyrofab:
That was indeed my thought, an `EntityData` rename requires more discussion.

PR #1117, opened 2020-02-14
BlockBox names and overloads in BlockPos

magneticflux-:
- method_23627 -> stream
  - New overload for `BlockBox` param  
- blockBox -> box
  - Shortened name  
- method_22874 -> center
  - Returns a `BlockPos` at the average of each axis bounds. Biased to the `min*` corner of the box.

### DIFF
--- a/mappings/net/minecraft/util/math/BlockBox.mapping
+++ b/mappings/net/minecraft/util/math/BlockBox.mapping
@@ -67,5 +67,5 @@ CLASS net/minecraft/class_3341 net/minecraft/util/math/BlockBox
 		ARG 1 x
 		ARG 2 y
 		ARG 3 z
-	METHOD method_22874 center ()Lnet/minecraft/class_2382;
+	METHOD method_22874 getCenter ()Lnet/minecraft/class_2382;
 		COMMENT @implNote Biased toward the minimum bound corner of the box.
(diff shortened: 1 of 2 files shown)


magneticflux-:
Should `method_22874` be named `getCenter` instead of just `center`? I tried to match the pattern of the `empty` method.

Juuxel:
Yeah, I think `getCenter` is better (`empty` is a static factory so it's different).

Juuxel (approved):

liach:
hmm, why close?

magneticflux-:
Another mistake with Cloak, sorry! I thought I could delete it locally but not remotely, but apparently not.

mappings/net/minecraft/util/math/BlockBox.mapping
```diff
@@ -67,3 +67,5 @@ CLASS net/minecraft/class_3341 net/minecraft/util/math/BlockBox
 		ARG 1 x
 		ARG 2 y
 		ARG 3 z
+	METHOD method_22874 center ()Lnet/minecraft/class_2382;
```
liach: or `getCenter`?

liach (approved):

modmuss50 (approved):

PR #1119, opened 2020-02-16
build.gradle less hacky and added 1.2.5 mappings from hacky yarn

hYdos:


hYdos:
did not mean for this to be on the main fabricmc yarn repo oops...

PR #1121, opened 2020-02-20
ChunkGeneratorConfig minY and maxY to getTopBedrockY and getBottomBedrockY

falseresync:


Juuxel:
These are overridden in `CavesChunkGeneratorConfig` to return 0 for min Y and 127 for max Y 🤔

falseresync:
Lol wut

falseresync:
Hmm `getMaxY = 0` in `OverworldChunkGeneratorConfig` too. That is super weird

falseresync:
But my way seems correct (?) In `SurfaceChunkGenerator.buildBedrock` they're not used as in my variant

Juuxel:

Runemoro:
@falseresync I don't see the problem, they're not swapped...

liach:
So this one actually relates to #1123 and are both troubled by this min/max y thing

falseresync:
They're probably not `min`/`max` but rather some sort of `limit`, which can be either. But there're 2 of them. Maybe one should be `limitY` (or `yLimit`?) and the other one is `min` or `max`?

liach:
How are these 2 fields used, just curious?

falseresync:
They’re used as min and max respectively, but they switch their roles in 2(?) cases

liach:
> They’re used as min and max respectively, but they switch their roles in 2(?) cases

Like how? in a loop? Show the actual usage in enigma/decompiled code

falseresync:
I think they’re used in 1 place only, and switch their roles in children classes, aka not on the access side, but on the declaration side

falseresync:
Cannot check rn! Will update when I can, anyone else does that

mappings/net/minecraft/world/gen/chunk/ChunkGeneratorConfig.mapping
```diff
@@ -41,5 +41,5 @@ CLASS net/minecraft/class_2888 net/minecraft/world/gen/chunk/ChunkGeneratorConfi
 		ARG 1 state
 	METHOD method_12572 setDefaultFluid (Lnet/minecraft/class_2680;)V
 		ARG 1 state
-	METHOD method_16400 getMaxY ()I
-	METHOD method_16401 getMinY ()I
+	METHOD method_16400 getMinY ()I
+	METHOD method_16401 getMaxY ()I
```
liach:
```suggestion
	METHOD method_16400 getTopBedrockY ()I
	METHOD method_16401 getBottomBedrockY ()I
```

liach (changes requested):

liach (approved):

Runemoro:
So what are the values of these for the overworld, where there is no top bedrock, and end, where there is no bedrock at all?

liach:
0 top bedrock or 256 bottom bedrock would skip bottom/top bedrock generation.

jaskarth (approved):

falseresync:
See #1157

Issue #1111, opened 2020-02-11
field_14165 in server login network handler


liach:
Should be `serverId` sent in packet, though it's been optimized by vm to be `""` in compiled bytecode.

PR #1123, opened 2020-02-21
Suggest mappings

Martmists-GH:
- field_13224 -> biomeSize
  - name from getter  
- field_13223 -> riverSize
  - name from getter  
- field_13222 -> forcedBiome
  - name from getter  
- field_13221 -> minY
  - name from getter

### DIFF
--- a/mappings/net/minecraft/block/BedBlock.mapping
+++ b/mappings/net/minecraft/block/BedBlock.mapping
@@ -17,6 +17,8 @@ CLASS net/minecraft/class_2244 net/minecraft/block/BedBlock
 	METHOD method_18476 getDirection (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_2350;
 		ARG 0 world
 		ARG 1 pos
+	METHOD method_24163 getOtherBedPartDirection (Lnet/minecraft/class_2680;)Lnet/minecraft/class_2350;
+	METHOD method_24163 getOppositePartDirection (Lnet/minecraft/class_2680;)Lnet/minecraft/class_2350;
 	METHOD method_9484 findWakeUpPosition (Lnet/minecraft/class_1299;Lnet/minecraft/class_4538;Lnet/minecraft/class_2338;I)Ljava/util/Optional;
 		ARG 0 type
 		ARG 2 pos
--- a/mappings/net/minecraft/block/BedBlock.mapping
+++ b/mappings/net/minecraft/block/BedBlock.mapping
@@ -17,7 +17,6 @@ CLASS net/minecraft/class_2244 net/minecraft/block/BedBlock
 	METHOD method_18476 getDirection (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_2350;
 		ARG 0 world
 		ARG 1 pos
-	METHOD method_24163 getOtherBedPartDirection (Lnet/minecraft/class_2680;)Lnet/minecraft/class_2350;
 	METHOD method_24163 getOppositePartDirection (Lnet/minecraft/class_2680;)Lnet/minecraft/class_2350;
 	METHOD method_9484 findWakeUpPosition (Lnet/minecraft/class_1299;Lnet/minecraft/class_4538;Lnet/minecraft/class_2338;I)Ljava/util/Optional;
 		ARG 0 type
--- a/mappings/net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig.mapping
+++ b/mappings/net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig.mapping
@@ -1,4 +1,5 @@
 CLASS net/minecraft/class_2906 net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig
+	FIELD field_13221 minY I
 	FIELD field_13222 forcedBiome I
 	FIELD field_13223 riverSize I
 	FIELD field_13224 biomeSize I
--- a/mappings/net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig.mapping
+++ b/mappings/net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig.mapping
@@ -1,5 +1,4 @@
 CLASS net/minecraft/class_2906 net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig
-	FIELD field_13221 minY I
 	FIELD field_13222 forcedBiome I
 	METHOD method_12614 getBiomeSize ()I
 	METHOD method_12615 getForcedBiome ()I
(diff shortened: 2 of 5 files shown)


mappings/net/minecraft/block/BedBlock.mapping
```diff
@@ -17,6 +17,8 @@ CLASS net/minecraft/class_2244 net/minecraft/block/BedBlock
 	METHOD method_18476 getDirection (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_2350;
 		ARG 0 world
 		ARG 1 pos
+	METHOD method_24163 getOtherBedPartDirection (Lnet/minecraft/class_2680;)Lnet/minecraft/class_2350;
+	METHOD method_24163 getOppositePartDirection (Lnet/minecraft/class_2680;)Lnet/minecraft/class_2350;
```
Martmists-GH, line 20: Not sure what happened here but blame Cloak

mappings/net/minecraft/block/BedBlock.mapping
```diff
@@ -17,6 +17,8 @@ CLASS net/minecraft/class_2244 net/minecraft/block/BedBlock
 	METHOD method_18476 getDirection (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_2350;
 		ARG 0 world
 		ARG 1 pos
+	METHOD method_24163 getOtherBedPartDirection (Lnet/minecraft/class_2680;)Lnet/minecraft/class_2350;
```
liach:
```suggestion
```

liach (changes requested):

liach (approved):

mappings/net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig.mapping
```diff
@@ -1,4 +1,8 @@
 CLASS net/minecraft/class_2906 net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig
+	FIELD field_13221 minY I
+	FIELD field_13222 forcedBiome I
+	FIELD field_13223 riverSize I
+	FIELD field_13224 biomeSize I
```
Juuxel: The getters return constant values instead of these fields. In addition, `getMinY` does not even return the constant value of `field_13221`.

Juuxel (changes requested):

mappings/net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig.mapping
Martmists-GH: so should I just undo those fields?

liach:
@juuxel note javac can optimize final constants

Juuxel:
I don't think javac optimizes *instance* fields, which those are.

liach:
Juuz: See #1111, where a final `String` field with fixed `""` value in `ServerLoginNetworkHandler` has been optimized (its calls in auth call and server hello packet has been inlined to `""`)

mappings/net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig.mapping
Runemoro: @liach Juuxel is right, the fields aren't used in the getters, they were inlined by javac. There's no way for sure the getters are in the same order as the fields, and there's no point in mapping them anyway.

liach: Do the fields have the same values? If they cannot identify by value (like a few private static final fields in piglin/hoglin) then i'd leave them unmapped.

Juuxel:
Oh, hmm... There's still the one field that doesn't match any getter's return value that should be left unmapped.

mappings/net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig.mapping
Juuxel: Three of them match three of the getters, but `field_13221` ("`minY`") does not match the `getMinY` return value.

mappings/net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig.mapping
```diff
@@ -1,4 +1,8 @@
 CLASS net/minecraft/class_2906 net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig
+	FIELD field_13221 minY I
```
Juuxel: `getMinY` does not return the value of this field.

mappings/net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig.mapping
Juuxel: Both of these (and the respective getters) are 4, so it can't be known which one is which.

Juuxel (changes requested):
`forcedBiome` is the only unambiguous field in `OverworldChunkGeneratorConfig`, so the others should be left unmapped.
overworldchunkgeneratorconfig

Runemoro (changes requested):
I agree with Juuxel, you can't assume the fields are in the same order as usages, and they're unused anyway so there's no point in mapping them.

mappings/net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig.mapping
liach:
```suggestion
```

liach (changes requested):
Make this change and things should be good to go.

liach:
Just noticed this pr is still targeting 20w08a, edit the pr to make it merge into 20w09a

mappings/net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig.mapping
liach:
```suggestion
```
Please also remove this unknown mapping.

Martmists-GH:
all done

mappings/net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig.mapping
liach: @modmuss50 If martmists allows project maintainer to edit his pr, please remove this mapping entry and things are good to go.

Juuxel:

Martmists-GH:
done

mappings/net/minecraft/world/gen/chunk/OverworldChunkGeneratorConfig.mapping
```diff
@@ -1,4 +1,6 @@
...
```
liach:
```suggestion
```
Should remove this mapping.

liach (changes requested):

Runemoro (approved):

liach (approved):

haykam821 (approved):

YanisBft (approved):

JosephMillsAtWork:
I disagree with this. If they are Unused then cool. But some naming is better than none. Just I suggest injecting java docs for this so the next person that see's this does not try to "fix" it.

Juuxel:
It's not really possible to name unused fields that don't directly correspond to some constants in methods.

JosephMillsAtWork:
@Juuxel  Let me make a example and  push it.

PR #1134, opened 2020-02-25
minusDiv -> getLerpProgress + other improvements

Earthcomputer:
Someone didn't look at this function very carefully :P

### DIFF
--- a/mappings/net/minecraft/util/math/MathHelper.mapping
+++ b/mappings/net/minecraft/util/math/MathHelper.mapping
@@ -76,10 +76,15 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
 		ARG 0 hue
 		ARG 1 saturation
 		ARG 2 value
-	METHOD method_15370 minusDiv (DDD)D
-		ARG 0 numerator
-		ARG 2 delta
-		ARG 4 denominator
+	METHOD method_15370 inverseLerp (DDD)D
+		COMMENT Essentially does the inverse of the linear interpolation (lerp) function.
+		COMMENT Returns the delta value needed to interpolate between {@code a} and {@code b}.
+		ARG 0 value
+			COMMENT The result of the lerp function
+		ARG 2 a
+			COMMENT The value interpolated from
+		ARG 4 b
+			COMMENT The value interpolated to
 	METHOD method_15371 hashCode (III)J
 		ARG 0 x
 		ARG 1 y
--- a/mappings/net/minecraft/util/math/MathHelper.mapping
+++ b/mappings/net/minecraft/util/math/MathHelper.mapping
@@ -29,6 +29,8 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
 		ARG 0 a
 		ARG 1 b
 	METHOD method_15349 atan2 (DD)D
+		ARG 0 y
+		ARG 2 x
 	METHOD method_15350 clamp (DDD)D
 		ARG 0 value
 		ARG 2 min
@@ -76,10 +78,10 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
 		ARG 0 hue
 		ARG 1 saturation
 		ARG 2 value
-	METHOD method_15370 inverseLerp (DDD)D
-		COMMENT Essentially does the inverse of the linear interpolation (lerp) function.
-		COMMENT Returns the delta value needed to interpolate between {@code start} and {@code end}.
-		COMMENT In other words, {@code inverseLerp(lerp(delta, start, end), start, end) == delta}.
+	METHOD method_15370 getLerpProgress (DDD)D
+		COMMENT Gets the fraction of the way that {@code value} is between {@code start} and {@code end}.
+		COMMENT This is the delta value needed to lerp between {@code start} and {@code end} to get {@code value}.
+		COMMENT In other words, {@code getLerpProgress(lerp(delta, start, end), start, end) == delta}.
 		ARG 0 value
 			COMMENT The result of the lerp function
 		ARG 2 start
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -76,10 +76,15 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
...
-		ARG 0 numerator
-		ARG 2 delta
-		ARG 4 denominator
+	METHOD method_15370 inverseLerp (DDD)D
+		COMMENT Essentially does the inverse of the linear interpolation (lerp) function.
+		COMMENT Returns the delta value needed to interpolate between {@code a} and {@code b}.
+		ARG 0 value
+			COMMENT The result of the lerp function
+		ARG 2 a
+			COMMENT The value interpolated from
```
liach, line 88: Maybe specify this is the lower bound (smaller)?

mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -76,10 +76,15 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
 		ARG 0 hue
 		ARG 1 saturation
 		ARG 2 value
-	METHOD method_15370 minusDiv (DDD)D
-		ARG 0 numerator
-		ARG 2 delta
-		ARG 4 denominator
+	METHOD method_15370 inverseLerp (DDD)D
+		COMMENT Essentially does the inverse of the linear interpolation (lerp) function.
```
liach: Maybe say given delta, a, b, `inverseLerp(lerp(delta, a, b), a, b) == delta`? I thought it's like lerping the inversed `value` at first sight.

liach (commented):
Sorry, but it's me who [named it `minusDiv`](https://github.com/FabricMC/yarn/commit/94b94160eafb7d46518dfd18c114950afb58df68#diff-d8865f4320c6cc9692e2b0048a63712eR90-R93).

Runemoro:
Is `inverseLerp` the best name for this? Maybe something like `getProgress(value, start, end)` would be a better name, since it's calculating the "progress" from the start value to the end value given the current value.

liach:
```java
            this.lastNetherPortalPosition = new BlockPos(pos);
            NetherPortalBlock class_2423 = (NetherPortalBlock)Blocks.NETHER_PORTAL;
            BlockPattern.Result result3 = NetherPortalBlock.findPortal(this.world, this.lastNetherPortalPosition);
            double double4 = (result3.getForwards().getAxis() == Direction.Axis.X) ? result3.getFrontTopLeft().getZ() : ((double)result3.getFrontTopLeft().getX());
            double double6 = Math.abs(MathHelper.minusDiv(((result3.getForwards().getAxis() == Direction.Axis.X) ? this.getZ() : this.getX()) - (double)((result3.getForwards().rotateYClockwise().getDirection() == Direction.AxisDirection.NEGATIVE) ? 1 : 0), double4, double4 - result3.getWidth()));
            double double8 = MathHelper.minusDiv(this.getY() - 1.0, result3.getFrontTopLeft().getY(), result3.getFrontTopLeft().getY() - result3.getHeight());
            this.lastNetherPortalDirectionVector = new Vec3d(double6, double8, 0.0);
            this.lastNetherPortalDirection = result3.getForwards();
```
Fyi this is like the only place where this method is used (in entity's portal code).

Earthcomputer:
inverseLerp is definitely a better name than getProgress, especially since this is the math class. However I approve of the start and end parameter names.

Runemoro:
> inverseLerp is definitely a better name than getProgress

The term "progress" is used in many other places in yarn to refer to this kind of value. I don't see why you think the method should be described as the "inverse of lerp" when it can be directly described as "progress". It would be a bit like calling division "inverse multiplication".

The javadocs also focus on this being "the delta value needed to interpolate between", even though its only usage isn't to find the delta for an interpolation.

> especially since this is the math class

I don't see what you mean by that.

Earthcomputer:
After some more thought and discussion on discord, I have decided to name the function `getLerpProgress`. `getProgress` on its own is too vague, whereas `inverseLerp` is a little difficult to read, especially given that it only inverts for one variable.
I will also update the javadocs to describe the function first without using lerp, then later explain its link to lerp.

Earthcomputer:
I also made a couple of other improvements to MathHelper.

liach:
Can you rename `randomUUID` to `randomUuid` as well @Earthcomputer?

Earthcomputer:
Done

mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -156,6 +195,7 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
 		ARG 0 start
 		ARG 1 end
 		ARG 2 delta
+	METHOD method_22860 fwrapDegrees (D)F
```
liach, line 198: May be `fWrapDegrees` or `wrapDegreesF` (note your casing)

liach (approved):

mappings/net/minecraft/util/math/MathHelper.mapping
Earthcomputer: I was following the pattern of `lfloor`, it differs from `wrapDegrees(D)D` only in return type. Also, this is a deprecated method that people should probably use the other methods in favour of.

Juuxel (approved):

mappings/net/minecraft/util/math/MathHelper.mapping
liach: Not really, we have `nextInt` `nextDouble`, so `floorLong` etc should be more accurate and match the pattern better

liach:
Can be merged once branch is 20w09a

Earthcomputer:
Done

PR #1138, opened 2020-02-26
Cape and Elytra mappings for PlayerEntity

Sollace:
Most of the fields mapped are to do with controlling the trailing position and angle of the cape when displayed.

Also filled in some missing method params/method names.

### DIFF
--- a/mappings/net/minecraft/entity/player/PlayerEntity.mapping
+++ b/mappings/net/minecraft/entity/player/PlayerEntity.mapping
@@ -185,7 +185,8 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 		ARG 2 retainOwnership
 	METHOD method_7329 dropItem (Lnet/minecraft/class_1799;ZZ)Lnet/minecraft/class_1542;
 		ARG 1 stack
-		ARG 2 spreadItems
+		ARG 2 throwRandomly
+			COMMENT If true, the item will be thrown in a random direction from the entity regardless of which direction the tntity is facing
 		ARG 3 retainOwnership
 	METHOD method_7330 updateTurtleHelmet ()V
 	METHOD method_7331 requestRespawn ()V
--- a/mappings/net/minecraft/entity/player/PlayerEntity.mapping
+++ b/mappings/net/minecraft/entity/player/PlayerEntity.mapping
@@ -1,10 +1,13 @@
 CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 	FIELD field_18134 POSE_DIMENSIONS Ljava/util/Map;
 	FIELD field_18135 STANDING_DIMENSIONS Lnet/minecraft/class_4048;
+	FIELD field_19428 shoulderEntityAddedTime J
+	FIELD field_7483 strideDistance F
 	FIELD field_7484 itemCooldownManager Lnet/minecraft/class_1796;
 	FIELD field_7486 enderChestInventory Lnet/minecraft/class_1730;
 	FIELD field_7487 sleepTimer I
 	FIELD field_7488 MAIN_ARM Lnet/minecraft/class_2940;
+	FIELD field_7489 abilityResyncCountdown I
 	FIELD field_7490 isSubmergedInWater Z
 	FIELD field_7491 ABSORPTION_AMOUNT Lnet/minecraft/class_2940;
 	FIELD field_7493 hungerManager Lnet/minecraft/class_1702;
@@ -12,9 +15,13 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 	FIELD field_7495 totalExperience I
 	FIELD field_7496 LEFT_SHOULDER_ENTITY Lnet/minecraft/class_2940;
 	FIELD field_7498 playerContainer Lnet/minecraft/class_1723;
+	FIELD field_7499 capeZ D
+	FIELD field_7500 capeX D
 	FIELD field_7501 spawnPosition Lnet/minecraft/class_2338;
+	FIELD field_7502 prevCapeY D
 	FIELD field_7503 abilities Lnet/minecraft/class_1656;
 	FIELD field_7504 experiencePickUpDelay I
+	FIELD field_7505 prevStrideDistance F
 	FIELD field_7506 RIGHT_SHOULDER_ENTITY Lnet/minecraft/class_2940;
 	FIELD field_7507 gameProfile Lcom/mojang/authlib/GameProfile;
 	FIELD field_7508 lastPlayedLevelUpSoundTime I
@@ -26,7 +33,10 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 	FIELD field_7515 spawnForced Z
 	FIELD field_7518 PLAYER_MODEL_PARTS Lnet/minecraft/class_2940;
 	FIELD field_7520 experienceLevel I
+	FIELD field_7521 capeY D
+	FIELD field_7522 prevCapeZ D
 	FIELD field_7523 reducedDebugInfo Z
+	FIELD field_7524 prevCapeX D
 	FIELD field_7525 selectedItem Lnet/minecraft/class_1799;
 	METHOD <init> (Lnet/minecraft/class_1937;Lcom/mojang/authlib/GameProfile;)V
 		ARG 1 world
@@ -54,6 +64,9 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 	METHOD method_21823 shouldCancelInteraction ()Z
 	METHOD method_21824 shouldDismount ()Z
 	METHOD method_21825 clipAtLedge ()Z
+	METHOD method_23668 checkFallFlying ()Z
+	METHOD method_23669 startFallFlying ()V
+	METHOD method_23670 stopFallFlying ()V
 	METHOD method_7254 unlockRecipes (Ljava/util/Collection;)I
 		ARG 1 recipes
 	METHOD method_7255 addExperience (I)V
@@ -64,9 +77,14 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 	METHOD method_7258 isSpawnForced ()Z
 	METHOD method_7259 incrementStat (Lnet/minecraft/class_3445;)V
 		ARG 1 stat
+	METHOD method_7260 increaseRidingMotionStats (DDD)V
+		ARG 1 x
+		ARG 3 y
+		ARG 5 z
 	METHOD method_7261 getAttackCooldownProgress (F)F
 		ARG 1 baseTime
 	METHOD method_7262 dropShoulderEntities ()V
+	METHOD method_7263 spawnSweepAttackParticles ()V
 	METHOD method_7264 isWithinSleepingRange (Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;)Z
 		ARG 1 sleepPos
 		ARG 2 direction
@@ -95,6 +113,10 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 	METHOD method_7280 getSpawnPosition ()Lnet/minecraft/class_2338;
 	METHOD method_7281 incrementStat (Lnet/minecraft/class_2960;)V
 		ARG 1 stat
+	METHOD method_7282 increaseTravelMotionStats (DDD)V
+		ARG 1 x
+		ARG 3 y
+		ARG 5 z
 	METHOD method_7283 setMainArm (Lnet/minecraft/class_1306;)V
 		ARG 1 arm
 	METHOD method_7284 disableShield (Z)V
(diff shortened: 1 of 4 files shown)


liach (approved):

mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -64,9 +77,14 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 	METHOD method_7258 isSpawnForced ()Z
 	METHOD method_7259 incrementStat (Lnet/minecraft/class_3445;)V
 		ARG 1 stat
+	METHOD method_7260 increaseRidingMotionStats (DDD)V
```
Pyrofab, line 80: I feel like `updateRidingMotionStats` would work a little better, but personal preference.

mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -95,6 +113,10 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 	METHOD method_7280 getSpawnPosition ()Lnet/minecraft/class_2338;
 	METHOD method_7281 incrementStat (Lnet/minecraft/class_2960;)V
 		ARG 1 stat
+	METHOD method_7282 increaseTravelMotionStats (DDD)V
```
Pyrofab, line 116: Same here, I'd personally go for `updateTravelMotionStats`.

mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -64,9 +77,14 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 	METHOD method_7258 isSpawnForced ()Z
 	METHOD method_7259 incrementStat (Lnet/minecraft/class_3445;)V
 		ARG 1 stat
+	METHOD method_7260 increaseRidingMotionStats (DDD)V
+		ARG 1 x
+		ARG 3 y
+		ARG 5 z
```
Pyrofab:
```suggestion
		ARG 1 vx
		ARG 3 vy
		ARG 5 vz
```

mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -95,6 +113,10 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 	METHOD method_7280 getSpawnPosition ()Lnet/minecraft/class_2338;
 	METHOD method_7281 incrementStat (Lnet/minecraft/class_2960;)V
 		ARG 1 stat
+	METHOD method_7282 increaseTravelMotionStats (DDD)V
+		ARG 1 x
+		ARG 3 y
+		ARG 5 z
```
Pyrofab:
```suggestion
		ARG 1 vx
		ARG 3 vy
		ARG 5 vz
```

mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -112,13 +134,18 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
...
+		ARG 3 announceChange
 	METHOD method_7290 dropSelectedItem (Z)Z
 		ARG 1 dropEntireStack
 	METHOD method_7291 openHorseInventory (Lnet/minecraft/class_1496;Lnet/minecraft/class_1263;)V
 	METHOD method_7292 getLuck ()F
 	METHOD method_7293 vanishCursedItems ()V
 	METHOD method_7294 canModifyWorld ()Z
 	METHOD method_7295 updateWaterSubmersionState ()Z
+	METHOD method_7296 dropShoulderEntity (Lnet/minecraft/class_2487;)V
+		ARG 1 compound
```
Pyrofab:
```suggestion
		ARG 1 entityNbt
```

mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -154,8 +182,11 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 	METHOD method_7327 getScoreboard ()Lnet/minecraft/class_269;
 	METHOD method_7328 dropItem (Lnet/minecraft/class_1799;Z)Lnet/minecraft/class_1542;
 		ARG 1 stack
+		ARG 2 retainOwnership
 	METHOD method_7329 dropItem (Lnet/minecraft/class_1799;ZZ)Lnet/minecraft/class_1542;
 		ARG 1 stack
+		ARG 2 spreadItems
```
Pyrofab: Although this is technically used to spread items around when dying, this could be used on a single stack - in which case it sounds like it would split the stack. I think `randomToss` would be a little more explicit maybe ?

mappings/net/minecraft/entity/player/PlayerEntity.mapping
liach: velocityX etc?

Pyrofab: I feel like vx is clear enough, but velocityX works too.

mappings/net/minecraft/entity/player/PlayerEntity.mapping
Sollace: I agree, however I went with this due to other places already using names like `increaseStat`

mappings/net/minecraft/entity/player/PlayerEntity.mapping
Sollace: The method only ever takes one stack. `randomToss` works, I supposed, but I can't help but wonder if there's a better term to describe it. `randomToss` sounds too Engrish.

mappings/net/minecraft/entity/player/PlayerEntity.mapping
liach: This name is fine. The update is strictly incremental.

mappings/net/minecraft/entity/player/PlayerEntity.mapping
liach: `spill` should suffice (with comment)

Pyrofab: Not sure what "too engrish" means here, but if you want conciseness "spill" works fine.

Runemoro: I think `spreadItems` is the clearest out of all of these. I immediately understand what it means, but I wouldn't know what "spill" or "randomToss" mean.

liach: Let's continue using spreadItems and add a jd comment on what it means.

Runemoro (approved):

mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -64,9 +77,14 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 	METHOD method_7258 isSpawnForced ()Z
 	METHOD method_7259 incrementStat (Lnet/minecraft/class_3445;)V
 		ARG 1 stat
+	METHOD method_7260 increaseRidingMotionStats (DDD)V
+		ARG 1 vx
+		ARG 3 vy
+		ARG 5 vz
```
Juuxel: v?

mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -95,6 +113,10 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 	METHOD method_7280 getSpawnPosition ()Lnet/minecraft/class_2338;
 	METHOD method_7281 incrementStat (Lnet/minecraft/class_2960;)V
 		ARG 1 stat
+	METHOD method_7282 increaseTravelMotionStats (DDD)V
+		ARG 1 vx
+		ARG 3 vy
+		ARG 5 vz
```
Juuxel: Same here.

mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -54,6 +64,9 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 	METHOD method_21823 shouldCancelInteraction ()Z
 	METHOD method_21824 shouldDismount ()Z
 	METHOD method_21825 clipAtLedge ()Z
+	METHOD method_23668 checkFallFlying ()Z
+	METHOD method_23669 startFallFlying ()V
+	METHOD method_23670 stopFallFlying ()V
```
Juuxel, line 69: These should be `*ElytraFlying` because it specifically checks for elytras.

Juuxel (changes requested):

mappings/net/minecraft/entity/player/PlayerEntity.mapping
liach: Don't think it's elytra; moving while falling count as well (like there's a stat)

mappings/net/minecraft/entity/player/PlayerEntity.mapping
liach: `velocityX/Y/Z` plz

mappings/net/minecraft/entity/player/PlayerEntity.mapping
Juuxel: `23668` only does anything `if (itemStack2.getItem() == Items.ELYTRA && ElytraItem.isUsable(itemStack2)) {` (where the stack is the item in the chest slot)

Juuxel: Maybe it's only `method_23668` that's hardcoded for elytras 🤔

Edit: `isFallFlying` is also used in `ElytraSoundInstance` and `ElytraEntityModel`. I guess this enum is where "fall flying" comes from:

mappings/net/minecraft/entity/player/PlayerEntity.mapping
Pyrofab: I very much disagree. When I saw `spreadItems`, my first thought was "wait, there is a way to automatically split a stack when dropping it ?".

Runemoro: Wait a minute. So this doesn't split the stack to throw the individual items in random directions? In that case, I agree `spreadItems` is a very bad name for this.

Pyrofab: Nop, it's for adding random velocity when dropping the item (eg. For spreading the stacks around you when you die)

liach: `spew`

Runemoro: `throwRandomly`

Juuxel: I think "scatter" is already used in yarn (`ItemScatterer` etc).

Runemoro: But it doesn't scatter the items in the stack! It only throws the stack, as a single entity, in a random direction.

Juuxel: That's what `ItemScatterer` does as well.

Runemoro: No, `ItemScatterer` takes a list of stacks and scatters them in random directions.

mappings/net/minecraft/entity/player/PlayerEntity.mapping
Sollace: My choice was X/Y/Z imo. I don't see the need to specify that it's velocity, but @Pyrofab suggested this change.

mappings/net/minecraft/entity/player/PlayerEntity.mapping
Sollace: That's correct. If it weren't for the enum, I would have called it "gliding", because that's essentially what it is...

mappings/net/minecraft/entity/player/PlayerEntity.mapping
Sollace: Marking these as "resolved" then, as it applied to both this method and the travel variant below.

mappings/net/minecraft/entity/player/PlayerEntity.mapping
Sollace: I'm partial towards `spill`

mappings/net/minecraft/entity/player/PlayerEntity.mapping
Runemoro: Does this method just increase the stat by sqrt(vx^2 + vy^2 + vz^2)? If yes, the parameters could be named `dx` `dy` `dz`, since they're just coordinate differences.

liach: yes use `dx/y/z` or `deltaX/Y/Z` if it's the case, just `d` preferred

Juuxel:

mappings/net/minecraft/entity/player/PlayerEntity.mapping
Sollace: `ItemScatterer` -> `ItemSpiller` 🤔

Sollace:
@Juuxel Branch updated

Runemoro (approved):

YanisBft (approved):

Juuxel (approved):

PR #1139, opened 2020-02-27
A couple of additions to PointOfInterestStorage

Earthcomputer:


mappings/net/minecraft/world/poi/PointOfInterestStorage.mapping
```diff
@@ -62,6 +63,16 @@ CLASS net/minecraft/class_4153 net/minecraft/world/poi/PointOfInterestStorage
 		ARG 3 pos
 		ARG 4 radius
 		ARG 5 occupationStatus
+	METHOD method_22383 getInSquare (Ljava/util/function/Predicate;Lnet/minecraft/class_2338;ILnet/minecraft/class_4153$class_4155;)Ljava/util/stream/Stream;
```
liach, line 66: I'd keep other methods named `get` as they can be distinguished by param types; for the square and circle one i'd name them `getInManhattan` `getInEuclidean`

Earthcomputer: I would rather name them in terms of the shape, rather than in terms of the distance type, as the method returns a stream of all POIs in the shape, not sorted by distance. (Also, manhattan is a diamond shape, not a square).

I would also prefer the other methods to be more specific in the name, as you don't often call these methods directly, and having a million different overloads of "get" that search different areas is confusing to read.

Runemoro: I agree with @Earthcomputer. If the methods were taking ChunkPos and ChunkSectionPos as parameters, then I'd say we should let the parameters disambiguate them, but here the only difference is an int or a long for the position.

I don't mind the names circle and square, but if you really want to name them by distance type, it would be Chebyshev distance for the square.

mappings/net/minecraft/world/poi/PointOfInterestStorage.mapping
```diff
@@ -10,15 +11,15 @@ CLASS net/minecraft/class_4153 net/minecraft/world/poi/PointOfInterestStorage
 		ARG 2 predicate
 	METHOD method_19118 getDistanceFromNearestOccupied (Lnet/minecraft/class_4076;)I
 		ARG 1 pos
-	METHOD method_19119 get (Ljava/util/function/Predicate;JLnet/minecraft/class_4153$class_4155;)Ljava/util/stream/Stream;
+	METHOD method_19119 getInChunkSection (Ljava/util/function/Predicate;JLnet/minecraft/class_4153$class_4155;)Ljava/util/stream/Stream;
```
liach, line 14: Why not `getInSubchunk` (you said it's called this in the tech community)

Runemoro: I think renaming section to subchunk should be done in another PR. But I disagree with that, since "section" is present in NBT tags, debug screen, exception/log messages, Minecraft wiki.

Earthcomputer: I go by the standard currently set by the yarn that I'm merging into, not by some discussion in an issue that was opened 5 minutes ago.

liach (approved):
Should be good I guess, thought I don't like `InChunk` `InChunkSection` suffixes

Runemoro (approved):

Pyrofab (approved):

PR #1143, opened 2020-02-28
20w09a all classes mapped

liach:

### DIFF
--- a/mappings/net/minecraft/client/sound/ReusedAudioStream.mapping
+++ b/mappings/net/minecraft/client/sound/ReusedAudioStream.mapping
@@ -1,0 +1,11 @@
+CLASS net/minecraft/class_4856 net/minecraft/client/sound/ReusedAudioStream
+	FIELD field_22443 delegateFactory Lnet/minecraft/class_4856$class_4857;
+	FIELD field_22444 delegate Lnet/minecraft/class_4234;
+	FIELD field_22445 inputStream Ljava/io/BufferedInputStream;
+	METHOD <init> (Lnet/minecraft/class_4856$class_4857;Ljava/io/InputStream;)V
+		ARG 1 delegateFactory
+		ARG 2 inputStream
+	CLASS class_4857 DelegateFactory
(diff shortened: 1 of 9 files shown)


mappings/net/minecraft/client/sound/ReusedAudioStream.mapping
```diff
@@ -0,0 +1,11 @@
+CLASS net/minecraft/class_4856 net/minecraft/client/sound/ReusedAudioStream
```
Runemoro: What do you mean by `Resused`? Shouldn't it be `Reusable`?

liach: @Runemoro because it automatically skips back to the start of the delegate audio stream when the delegate audio stream is empty, hence `reused` not `reusable`, maybe `reusing`?

Runemoro (approved):

Runemoro (approved):

mappings/net/minecraft/world/biome/source/BiomeAccess.mapping
```diff
@@ -10,6 +10,10 @@ CLASS net/minecraft/class_4543 net/minecraft/world/biome/source/BiomeAccess
 		ARG 1 source
 	METHOD method_22393 getBiome (Lnet/minecraft/class_2338;)Lnet/minecraft/class_1959;
 		ARG 1 pos
+	METHOD method_24854 getBiomeForNoiseGen (III)Lnet/minecraft/class_1959;
+		ARG 1 biomeX
+		ARG 2 biomeY
+		ARG 3 biomeZ
```
Juuxel, line 16: Why the `biome` prefix?

Juuxel (approved):

mappings/net/minecraft/world/biome/source/BiomeAccess.mapping
liach: Because this is the biome x,y,z which is block xyz /4

PR #1147, opened 2020-03-04
Map smithing table-related identifiers

haykam821:
This pull requests maps/improves mappings for the following things:

* [x] Smithing table block
* [x] Smithing table screen handler
* [x] Smithing table screen
* [x] Combination screen handler
* [x] Combination screen


mappings/net/minecraft/container/SmithingTableContainer.mapping
```diff
@@ -0,0 +1,6 @@
+CLASS net/minecraft/class_4862 net/minecraft/container/SmithingTableContainer
+	FIELD field_22485 smithingRecipes Ljava/util/Map;
```
liach: static final field

liach (changes requested):

mappings/net/minecraft/container/SmithingTableContainer.mapping
haykam821: Fixed in 090296774ccd61114f11805cad4e925942a74744.

mappings/net/minecraft/container/CombinationContainer.mapping
```diff
@@ -0,0 +1,15 @@
...
+	METHOD <init> (Lnet/minecraft/class_3917;ILnet/minecraft/class_1661;Lnet/minecraft/class_3914;)V
+		ARG 1 type
+		ARG 2 syncId
+	METHOD method_24923 onCombine (Lnet/minecraft/class_1657;Lnet/minecraft/class_1799;)Lnet/minecraft/class_1799;
+		ARG 1 combiner
+		ARG 2 result
+	METHOD method_24925 isContainerBlock (Lnet/minecraft/class_2680;)Z
+	METHOD method_24927 canTakeResult (Lnet/minecraft/class_1657;Z)Z
+		ARG 2 hasStack
+	METHOD method_24928 setResult ()V
```
liach: Hmm, why does a setter have no parameter? doesn't make sense at all

mappings/net/minecraft/container/CombinationContainer.mapping
```diff
@@ -0,0 +1,15 @@
+CLASS net/minecraft/class_4861 net/minecraft/container/CombinationContainer
+	FIELD field_22479 resultInv Lnet/minecraft/class_1263;
+	FIELD field_22480 inputsInv Lnet/minecraft/class_1263;
```
liach: Just `input` and `result` should suffice

mappings/net/minecraft/container/CombinationContainer.mapping
```diff
@@ -0,0 +1,15 @@
...
+	FIELD field_22482 player Lnet/minecraft/class_1657;
+	METHOD <init> (Lnet/minecraft/class_3917;ILnet/minecraft/class_1661;Lnet/minecraft/class_3914;)V
+		ARG 1 type
+		ARG 2 syncId
+	METHOD method_24923 onCombine (Lnet/minecraft/class_1657;Lnet/minecraft/class_1799;)Lnet/minecraft/class_1799;
+		ARG 1 combiner
+		ARG 2 result
+	METHOD method_24925 isContainerBlock (Lnet/minecraft/class_2680;)Z
+	METHOD method_24927 canTakeResult (Lnet/minecraft/class_1657;Z)Z
+		ARG 2 hasStack
```
haykam821: This needs a better name that doesn't have the `has` prefix (as per [naming conventions](https://github.com/FabricMC/yarn/blob/1b7aa91a4ddc43fd259bc4011834e7e5112143b8/CONVENTIONS.md))

mappings/net/minecraft/container/CombinationContainer.mapping
haykam821: Should this be plural (`inputsInv`) or singular (`inputInv`)?

ghost (approved):

mappings/net/minecraft/container/CombinationContainer.mapping
Juuxel: I think `inputInv`/`inputInventory` would be better.

haykam821: @Juuxel I didn't use 'inventory' as the suffix since the other containers didn't either, though that should probably be discussed as I think it is more natural.

mappings/net/minecraft/container/CombinationContainer.mapping
liach: Lol this is `onInputChanged` man

liach (changes requested):

mappings/net/minecraft/container/CombinationContainer.mapping
haykam821: Fixed in 5d9006292.

mappings/net/minecraft/container/CombinationContainer.mapping
haykam821: In `CraftingTableContainer`, the names of the inventory fields are `craftingInv` and `resultInv`. Perhaps those should be changed as well.

liach: hmm, I suggest the name because there is no input/result slots.

mappings/net/minecraft/container/CombinationContainer.mapping
liach:
```suggestion
		ARG 2 present
```

mappings/net/minecraft/container/SmithingTableContainer.mapping
```diff
@@ -0,0 +1,6 @@
+CLASS net/minecraft/class_4862 net/minecraft/container/SmithingTableContainer
+	FIELD field_22485 SMITHING_RECIPES Ljava/util/Map;
```
liach:
```suggestion
	FIELD field_22485 RECIPES Ljava/util/Map;
```
this is the smithing table container, so smithing is the context.

mappings/net/minecraft/container/CombinationContainer.mapping
Juuxel: `hasStack` is fine IMO

Prospector: I like present personally

mappings/net/minecraft/block/SmithingTableBlock.mapping
```diff
@@ -1 +1,2 @@
 CLASS net/minecraft/class_3717 net/minecraft/block/SmithingTableBlock
+	FIELD field_22511 SMITHING_TABLE_CONTAINER_NAME Lnet/minecraft/class_2588;
```
Prospector: does it need to specify SMITHING_TABLE? what other container would be defined in SmithingTableBlock?

```suggestion
	FIELD field_22511 CONTAINER_NAME Lnet/minecraft/class_2588;
```

Prospector (approved):
Looks mostly good, but a few notes

mappings/net/minecraft/container/CombinationContainer.mapping
Runemoro: > Perhaps those should be changed as well.

They should.

haykam821: @Runemoro If we do change it, it would have to be in a different PR, perhaps with the `updateResult` changes as well.

Runemoro: Yes, but why give these fields a bad name just because other fields have a bad name?

haykam821: It’s consistency over correctness. They can always be consistent and correct later.

Runemoro: So if the other one had a typo, like `reusltsInv`, you'd copy the typo here too?

haykam821: @Runemoro That’s a spelling error rather than a break in convention. In my opinion, whether to use the existing mappings for another mapping depends on the situation; there is a difference between a typo, an inconsistency, and an innaccuracy.

Runemoro: Shortening "inventory" to three-letters isn't a convention. It breaks the convention of not using uncommon abbreviations and is just as bad as a spelling mistake. Also, there are hundreds of other names that use the word "inventory" rather than "inv", so I don't  see why you're copying that one.

haykam821: @Runemoro It breaks the written convention of not using uncommon abbreviations, but the actual convention (at least within the containers‘ mappings) is to suffix it with `inv` when there are multiple inventories within the same container.

haykam821: Issue #1153 has been opened to track the removal of the `inv` suffix.

jaskarth (approved):

mappings/net/minecraft/container/CombinationContainer.mapping
Sollace: Note in my PR I simply called them `input` and `result`.

YanisBft (approved):

liach:
Hmm, I wonder if this pr has any conflict with mine; in yarn if two classes are mapped with different names there will be no merge conflict but enigma won't be able to open the mappings

haykam821:
@liach, the conflicts were resolved in a8b86a67575268c30fea9ac69bfb8ba8f61e61bc.

liach:
marking this as a refactor as it is after fixing merge conflict. also should add like forging -> combining in pr title

liach:
also i don't think smithing table screen is a good name as the screen handler id is just smithing, without table suffix

haykam821:
The three tables have discrepancies in them, so here are the current names/strings and my suggestion for how to standardize them:

Type | Screen Class | Texture ID | Handler Type ID | Handler Type Field
--- | --- | --- | --- | ---
Conventions (suggested) | `{Type}TableScreen` | `textures/gui/container/{type}_table.png` | `{type}` | `{TYPE}`
Smithing Table (#1147) | `SmithingTableScreen` | `textures/gui/container/smithing.png` | `smithing` | `SMITHING`
Smithing Table (#1148) | `SmithingScreen` | `textures/gui/container/smithing.png` | `smithin` | `SMITHING`
Crafting Table | `CraftingTableScreen` | `textures/gui/container/crafting_table.png` | `crafting` | `CRAFTING`
Cartography Table | `CartographyTableScreen` | `textures/gui/container/cartography_table.png` | `cartography_table` | `CARTOGRAPHY_TABLE`
Enchanting Table | `EnchantingScreen` | `textures/gui/container/enchanting_table.png` | `enchantment` | `ENCHANTMENT`

Unfortunately the identifiers for the smithing table's background texture and the cartography table's screen handler type is out of Yarn's control, but I think we should just ignore it.

liach:
i'd stick to identifiers (as done in my merged pr). mind start a vote?

Prospector:
I'd stick to ids as well.

liach:
Also @haykam821 there is `EnchantingScreen` than `EnchantingTableScreen`

haykam821:
I opened #1164 to resolve the inconsistent screen handler type fields. This is more of a Mojang issue than a Yarn issue so we might have to break matching the identifier in order to maintain consistency.

liach:
See also #1166, which resolves the discrepancy in https://github.com/FabricMC/yarn/pull/1147#issuecomment-596513486 as well

Runemoro:
These seem to be fully mapped in the latest version. I compared the current names with the ones suggested in this PR, and the only changes left in this PR are:

 - `ForgingScreen(Handler)` -> `CombinationScreen(Handler)`
 - `SmithingScreen(Handler)` -> `SmithingTableScreen(Handler)`
 - Adds `Inv` suffixes to some fields

haykam821:
@Runemoro The `inv` suffixes should be removed across all mappings. The first two differences, however, should be implemented into Yarn as they fulfill the conventions better.

PR #1149, opened 2020-03-04
Restore GUI names from 20w09a

Juuxel:
This PR restores many names that were removed in the 20w10a when they became obfuscated.


liach (approved):
Good fixup. Mojang names are ugly, and we need a quick cleanup pr (separate I suggest) after.

modmuss50:
Yes, I think using most of the old names is ok for now, will make tools like migrate mappings useful if names do change later.

modmuss50 (approved):

CDAGaming (approved):
Yes please!

jaskarth (approved):
thank u for your service kind juuz

PR #1156, opened 2020-03-06
Document changed UUID serialization methods

Pyrofab:
This PR names and document the UUID serialization methods added to `CompoundTag` and `NbtHelper` in 20w10a. It also documents the deprecation of the methods they replace.

### DIFF
--- a/mappings/net/minecraft/nbt/CompoundTag.mapping
+++ b/mappings/net/minecraft/nbt/CompoundTag.mapping
@@ -101,8 +101,8 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
 		ARG 1 key
 		ARG 2 value
 	METHOD method_10576 containsUuidOld (Ljava/lang/String;)Z
-		COMMENT @deprecated use {@link #containsUuid} for newly added deserialization methods.
-		COMMENT Other places may keep calling this as well as {@link #containsUuid} to ensure a smooth transition for old worlds.
+		COMMENT @deprecated use {@link #containsUuidNew} for newly added deserialization methods.
+		COMMENT Other places may keep calling this as well as {@link #containsUuidNew} to ensure a smooth transition for old worlds.
 		ARG 1 key
 	METHOD method_10577 getBoolean (Ljava/lang/String;)Z
 		ARG 1 key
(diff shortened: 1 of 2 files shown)


liach (approved):

mappings/net/minecraft/nbt/CompoundTag.mapping
```diff
@@ -60,7 +60,8 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
 		ARG 1 key
 	METHOD method_10559 createCrashReport (Ljava/lang/String;Lnet/minecraft/class_4614;Ljava/lang/ClassCastException;)Lnet/minecraft/class_128;
 		ARG 1 key
-	METHOD method_10560 putUuid (Ljava/lang/String;Ljava/util/UUID;)V
+	METHOD method_10560 putUuidOld (Ljava/lang/String;Ljava/util/UUID;)V
```
Runemoro, line 63: Why "old"? What's the difference between this method and the new one exactly? Also, is it also annotated as `@Deprecated`?

liach: The old puts uuid into a sub compound tag with M and L long tags. the new one puts a sub int array tag with 4 uuid components

Runemoro: Since the format is different, I think it's a very bad idea to rename the new method to `putUuid`. Any mod using the old version would now be silently switched to use the new method when updating mappings, breaking all old worlds when the mod is updated!

The new method should be named `putUuidNew`, and all the javadocs should make it clear that the format has been changed.

liach: or `uuidToSubIntArray`?

Runemoro: I think "new" and "old" is good. We can later drop the "new" once the risk of devs not noticing the change is lower.

Pyrofab: Good idea, suffixed the new methods with `New`. We should probably drop it whenever Mojang removes the old method.

Runemoro (approved):

Juuxel (approved):

YanisBft (approved):

PR #988, opened 2019-11-27
Renderer mappings

Sollace:
This maps all the parameter names in model `render` methods, the `setAngles`, and a lot of the entity renderers/feature renderers.

Explanations in the commit descriptions, because it would have been too long to remember for a PR description and it's easier to write it down as I do it.

90% of the parameters all filter directly down to `VertexConsumer.elements` where the usage is pretty clear:

```java
    default void elements(float x, float y, float z, float red, float green, float blue, float alpha, float textureU, float textureV, int overlay, int light, float normalX, float normalY, float normalZ) {
        this.vertex(x, y, z);
        this.color(red, green, blue, alpha);
        this.texture(textureU, textureV);
        this.overlay(overlay);
        this.light(light);
        this.normal(normalX, normalY, normalZ);
        this.next();
    }
```

The shadow rendering in `EntityRenderDispatcher` I recognised from previous versions, and if there was any doubt they have references to other fields mentioning shadows:

```java
matrix.translate(-pos.getX(), -pos.getY(), -pos.getZ());
if (gameOptions.entityShadows && renderShadows && renderer.shadowRadius > 0 && !entity.isInvisible()) {
    double distance = this.getSquaredDistanceToCamera(entity.getX(), entity.getY(), entity.getZ());
    float opacity = (float)((1 - distance / 256) * renderer.shadowOpacity);
    if (opacity > 0.0f) {
        renderShadow(matrix, provider, entity, opacity, tickDelta, world, renderer.shadowRadius);
    }
}
```

Shadow radius also matched here with code in my own project that was altering the shadow radius before that was remapped by the update job to reference this field.

```java
    @Override
    public void render(AbstractClientPlayerEntity entity, float entityYaw, float tickDelta, MatrixStack stack, VertexConsumerProvider renderContext, int lightUv) {
        field_4673 = renderPony.getShadowScale();
        super.render(entity, entityYaw, tickDelta, stack, renderContext, lightUv);
```

PR #1158, opened 2020-03-07
Drop invalid mappings

Sollace:
As I noticed the main branch contains invalid mappings, and since every PR is going to end up containing them in some way or another, I decided to just make a PR to get rid of them and nothing else.

Someone will have to verify `ThreadedAnvilChunkStorage` because the method is gone, but I can't view the class in Enigma to verify that it wasn't just changed and needs a remap.


Runemoro (approved):

liach (approved):

Juuxel (approved):

YanisBft (approved):
SlimeBlock change will conflict with #988

modmuss50:
Where did these come from? I always drop invalid mappings when updating. I assume it happens when merging?

PR #1164, opened 2020-03-09
Change the cartography table's screen handler type field to CARTOGRAPHY

haykam821:
This matches it with the crafting table's screen handler type identifier and field, as the background texture identifiers and screen class names of both use the 'table' suffix.

### DIFF
--- a/mappings/net/minecraft/screen/ScreenHandlerType.mapping
+++ b/mappings/net/minecraft/screen/ScreenHandlerType.mapping
@@ -1,4 +1,5 @@
 CLASS net/minecraft/class_3917 net/minecraft/screen/ScreenHandlerType
+	FIELD field_17343 CARTOGRAPHY Lnet/minecraft/class_3917;
 	FIELD field_17344 factory Lnet/minecraft/class_3917$class_3918;
 	METHOD <init> (Lnet/minecraft/class_3917$class_3918;)V
 		ARG 1 factory


liach:
The id of this screen handler over network is `minecraft:cartography_table`. Don't change it! Total nonse imo.

liach (changes requested):
Sheer nonsense. Code is like 
```java
public static final ContainerType<CartographyTableContainer> CARTOGRAPHY_TABLE = register("cartography_table", CartographyTableContainer::new);
```
(in older mappings, see that string literal)

shedaniel:
I'd rather do the reverse, naming all the `CRAFTING` to `CRAFTING_TABLE` tbh.

liach:
Uh no this is sheer nonsense. I am closing it. If you want to refute comment down here and I will reopen it.

liach:
@shedaniel read this
```java
public static final ContainerType<CraftingTableContainer> CRAFTING = register("crafting", CraftingTableContainer::new);
```

shedaniel:
You might probably want to look at `Registry` to make it consistent then.

shedaniel:
aka `PAINTING_MOTIVE`

haykam821:
See [this table](https://github.com/FabricMC/yarn/pull/1147#issuecomment-596513486). We either have to fix this one naming issue or have inconsistent mappings.

liach:
i am working on fixing that locally

liach:
Things should've been organized in #1166

PR #1155, opened 2020-03-06
I found the fixme package :tiny_potato:

Sollace:
// TODO:

PR #1172, opened 2020-03-11
20w15a Mapping Fixes 2: The Jumping The Shark Update

Sollace:
Redux of #1155 with a clean history.

This PR fills in some missing names, and changes some ones I saw as inaccurate.

<s>i.e. `ForgingScreen` -> `AbstractForgingScreen` (because the class is abstract)</s>


mappings/net/minecraft/block/Block.mapping
```diff
@@ -337,7 +337,7 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		ARG 1 state
 		ARG 2 world
 		ARG 3 pos
-		ARG 4 random
+		ARG 4 rand
```
Juuxel: `random` is better than `rand` (and it's also used more in Yarn)

Juuxel (changes requested):

mappings/net/minecraft/data/client/model/VariantSettings.mapping
```diff
@@ -2,9 +2,9 @@ CLASS net/minecraft/class_4936 net/minecraft/data/client/model/VariantSettings
 	FIELD field_22885 X Lnet/minecraft/class_4938;
 	FIELD field_22886 Y Lnet/minecraft/class_4938;
 	FIELD field_22887 MODEL Lnet/minecraft/class_4938;
-	FIELD field_22888 UVLOCK Lnet/minecraft/class_4938;
+	FIELD field_22888 UV_LOCK Lnet/minecraft/class_4938;
```
liach: the json key in blockstate variant json is `uvlock`

mappings/net/minecraft/screen/ScreenHandlerListener.mapping
```diff
@@ -1,9 +1,11 @@
 CLASS net/minecraft/class_1712 net/minecraft/screen/ScreenHandlerListener
 	METHOD method_7633 onPropertyUpdate (Lnet/minecraft/class_1703;II)V
 		ARG 1 handler
-		ARG 2 propertyId
+		ARG 2 field
```
liach:
```suggestion
		ARG 2 propertyId
```

mappings/net/minecraft/datafixer/fix/EntityProjectileOwnerFix.mapping
```diff
@@ -1,4 +1,6 @@
 CLASS net/minecraft/class_4860 net/minecraft/datafixer/fix/EntityProjectileOwnerFix
+	METHOD <init> (Lcom/mojang/datafixers/schemas/Schema;)V
+		ARG 1 schema
```
liach:
```suggestion
		ARG 1 parent
```

liach (commented):
If you rename screen to AbstractForgingScreen, make sure you also rename the ScreenHandler!

mappings/net/minecraft/datafixer/fix/EntityProjectileOwnerFix.mapping
i509VCB: Obviously this because it matches the params shown in DFU. Maybe we need someone to look over all the fixes and verify these have the right param names.

mappings/net/minecraft/screen/ScreenHandlerListener.mapping
Sollace: Parameters are a `field`/`value` pair though. If you call it that, then we have to do `propertyId`/`propertyValue` and in that case, why not just make it `id`/`value`? It makes no sense.

mappings/net/minecraft/datafixer/fix/EntityProjectileOwnerFix.mapping
Sollace: Why though? The class name is very clearly `Schema`.

liach: So if you have a string parameter that serves as prefix, we'd rather name it `prefix` than `string`.

Sollace: I'll rephrase it, then: what about this parameter makes you think it should be called `parent`? And if all the use cases of this class refers to it as `parent`, why is the class itself not called `FixParent`?

To me a parent means the parameter is itself the parent node in a tree, and/or the same, or a super-type of `this`, the child element.

mappings/net/minecraft/screen/ScreenHandlerListener.mapping
Juuxel: `property`/`value` would be fine IMO. Just `id` is a bit too vague.

Sollace: I can go with `property` / `values` then, if no one else objects.

mappings/net/minecraft/datafixer/fix/EntityProjectileOwnerFix.mapping
liach: Hmm, I think this should actually be
```suggestion
		ARG 1 outputSchema
```
forgot why i suggested `parent` before

Sollace:
@liach Update to 20w11a

mappings/net/minecraft/client/gui/screen/ingame/SmithingScreen.mapping
```diff
@@ -1,5 +1,5 @@
 CLASS net/minecraft/class_4895 net/minecraft/client/gui/screen/ingame/SmithingScreen
-	FIELD field_22795 TEXTURE Lnet/minecraft/class_2960;
+	FIELD field_22795 BACKGROUND Lnet/minecraft/class_2960;
```
liach: Other screens use `TEXTURE`.

mappings/net/minecraft/client/gui/screen/ingame/AbstractForgingScreen.mapping
```diff
@@ -1,10 +1,10 @@
-CLASS net/minecraft/class_4894 net/minecraft/client/gui/screen/ingame/ForgingScreen
-	FIELD field_22794 texture Lnet/minecraft/class_2960;
+CLASS net/minecraft/class_4894 net/minecraft/client/gui/screen/ingame/AbstractForgingScreen
+	FIELD field_22794 background Lnet/minecraft/class_2960;
```
liach: Same, other screens use texture for now.

mappings/net/minecraft/client/gui/screen/ingame/AbstractForgingScreen.mapping
```diff
@@ -1,10 +1,10 @@
-CLASS net/minecraft/class_4894 net/minecraft/client/gui/screen/ingame/ForgingScreen
-	FIELD field_22794 texture Lnet/minecraft/class_2960;
+CLASS net/minecraft/class_4894 net/minecraft/client/gui/screen/ingame/AbstractForgingScreen
```
liach: If you add `Abstract` to this screen, add `Abstract` to the handler as well.

liach (changes requested):

mappings/net/minecraft/client/gui/screen/ingame/SmithingScreen.mapping
Sollace: Other screens should be changed, then. This texture is used to draw the background of the screen.

liach: hmm, i suggest moving that to a separate pr

mappings/net/minecraft/client/gui/screen/ingame/AbstractForgingScreen.mapping
Sollace: Where is the handler? o.O

Sollace:
There's basically nothing left in this PR now... It was supposed to be the things that could be easily (and quickly) merged before the other bigger PR.

mappings/net/minecraft/client/gui/screen/ingame/AbstractForgingScreen.mapping
liach: `ForgingScreenHandler` in `screen` package

Sollace: I wouldn't change it unless the handler class was also abstract, imo.

liach: maybe mojang just forgot to declare one of the classes abstract, who knows

Juuxel (approved):

liach:
I just wonder if it's appropriate to name the screen `Abstract` while the corresponding screen handler's name lacks `Abstract` prefix. If you think that is fine the pr should be good to go

mappings/net/minecraft/client/gui/screen/ingame/AbstractForgingScreen.mapping
YanisBft: So, what's the decision on this?

liach: Probably needs a vote. I personally see no need to declare this abstract, as calling this `forging` already conveys abstraction.

Sollace:
Base branch changed. Looks to be no conflicts @Yanis48

YanisBft (approved):

liach (changes requested):
I hereby oppose this class rename.

Mojang somehow made this screen class abstract while its corresponding handler is not. Imo both classes should have matching names, not like adding `Abstract` to screen class name because mojang declared it abstract.

modmuss50 (approved):

liach:
thank you very much sollace!

PR #1179, opened 2020-03-13
Map some more DataFixer code.

i509VCB:


liach (approved):

mappings/net/minecraft/datafixer/NbtOps.mapping
```diff
@@ -1,2 +1,3 @@
 CLASS net/minecraft/class_2509 net/minecraft/datafixer/NbtOps
+	COMMENT Used to represent Minecraft's nbt tags as a {@link com.mojang.datafixers.Dynamic} for DataFixerUpper.
```
YanisBft: > Minecraft's nbt tags

It means "non binary tags tags"

liach:
```suggestion
	COMMENT Used to represent a Minecraft NBT as a {@link com.mojang.datafixers.Dynamic} for DataFixerUpper.
```

YanisBft (approved):

mappings/net/minecraft/datafixer/fix/BlockEntityCustomNameToTextFix.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_3567 net/minecraft/datafixer/fix/BlockEntityCustomNameToComponentFix
+CLASS net/minecraft/class_3567 net/minecraft/datafixer/fix/BlockEntityCustomNameToTextFix
```
Juuxel, line 1: This name is given by a string in `makeRule`, so it should be left the way it is.

mappings/net/minecraft/datafixer/fix/EntityCustomNameToTextFix.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_3597 net/minecraft/datafixer/fix/EntityCustomNameToComponentFix
+CLASS net/minecraft/class_3597 net/minecraft/datafixer/fix/EntityCustomNameToTextFix
```
Juuxel, line 1: Same here.

Juuxel (changes requested):

mappings/net/minecraft/datafixer/fix/BlockEntityCustomNameToTextFix.mapping
i509VCB: My reason for the change is that we don't use Component as a name anymore

liach: I support i509VCB here.

Juuxel:
(got some inspiration from #yarn-triage)

I think it would be fine if the classes were renamed with a comment like "This fix is internally called `MojangNameHereFix`".

liach:
Is a javadoc comment still necessary if there is string name?

Runemoro (approved):

PR #1181, opened 2020-03-13
Rename InfoEnchantment to Enchantment

### LINKED ISSUE #1057: InfoEnchantment -> ?
`InfoEnchantment` is not an enchantment called "Info", but a data object that stores an enchantment together with its level.

Suggested names:
- `EnchantmentState`
  - > It's not a state either. A state sounds like a Map<Enchantment, Integer> that is bound to an item stack to me.
    \- @liach
- `EnchantmentEntry`
  - A bit vague IMO, and doesn't really describe what it does.
- `EnchantmentInfo`
  - To me, that sounds like a class that stores the description ("info") of an enchantment.

dblsaiko:
Renames `Enchantment` to `EnchantmentType`, and `InfoEnchantment` to `Enchantment`.

Reasoning for this: `EnchantmentType` fits the function that `Enchantment` has, and `Enchantment` as a name for `InfoEnchantment` fits because you can interpret the combination of type+level, e.g. "Sharpness II" as the actual enchantment.

Closes #1057.

### DIFF
--- a/mappings/net/minecraft/loot/condition/TableBonusLootCondition.mapping
+++ b/mappings/net/minecraft/loot/condition/TableBonusLootCondition.mapping
@@ -1,11 +1,11 @@
 CLASS net/minecraft/class_182 net/minecraft/loot/condition/TableBonusLootCondition
-	FIELD field_1234 type Lnet/minecraft/class_1887;
+	FIELD field_1234 enchantment Lnet/minecraft/class_1887;
 	FIELD field_1235 chances [F
 	METHOD <init> (Lnet/minecraft/class_1887;[F)V
-		ARG 1 type
+		ARG 1 enchantment
 		ARG 2 chances
 	METHOD method_800 builder (Lnet/minecraft/class_1887;[F)Lnet/minecraft/class_4570$class_210;
-		ARG 0 type
+		ARG 0 enchantment
 		ARG 1 chances
 	METHOD test (Ljava/lang/Object;)Z
 		ARG 1 context
--- a/mappings/net/minecraft/loot/condition/TableBonusLootCondition.mapping
+++ b/mappings/net/minecraft/loot/condition/TableBonusLootCondition.mapping
@@ -1,11 +1,11 @@
 CLASS net/minecraft/class_182 net/minecraft/loot/condition/TableBonusLootCondition
-	FIELD field_1234 enchantment Lnet/minecraft/class_1887;
+	FIELD field_1234 type Lnet/minecraft/class_1887;
 	FIELD field_1235 chances [F
 	METHOD <init> (Lnet/minecraft/class_1887;[F)V
-		ARG 1 enchantment
+		ARG 1 type
 		ARG 2 chances
 	METHOD method_800 builder (Lnet/minecraft/class_1887;[F)Lnet/minecraft/class_4570$class_210;
-		ARG 0 enchantment
+		ARG 0 type
 		ARG 1 chances
 	METHOD test (Ljava/lang/Object;)Z
 		ARG 1 context
--- a/mappings/net/minecraft/loot/function/ApplyBonusLootFunction.mapping
+++ b/mappings/net/minecraft/loot/function/ApplyBonusLootFunction.mapping
@@ -1,28 +1,28 @@
 CLASS net/minecraft/class_94 net/minecraft/loot/function/ApplyBonusLootFunction
 	FIELD field_1009 formula Lnet/minecraft/class_94$class_96;
 	FIELD field_1010 FACTORIES Ljava/util/Map;
-	FIELD field_1011 type Lnet/minecraft/class_1887;
+	FIELD field_1011 enchantment Lnet/minecraft/class_1887;
 	METHOD <init> ([Lnet/minecraft/class_4570;Lnet/minecraft/class_1887;Lnet/minecraft/class_94$class_96;)V
 		ARG 1 conditions
-		ARG 2 type
+		ARG 2 enchantment
 		ARG 3 formula
 	METHOD method_453 (Lnet/minecraft/class_1887;[Lnet/minecraft/class_4570;)Lnet/minecraft/class_117;
 		ARG 1 conditions
 	METHOD method_455 oreDrops (Lnet/minecraft/class_1887;)Lnet/minecraft/class_120$class_121;
-		ARG 0 type
+		ARG 0 enchantment
 	METHOD method_456 uniformBonusCount (Lnet/minecraft/class_1887;)Lnet/minecraft/class_120$class_121;
-		ARG 0 type
+		ARG 0 enchantment
 	METHOD method_457 (Lnet/minecraft/class_1887;[Lnet/minecraft/class_4570;)Lnet/minecraft/class_117;
 		ARG 1 conditions
 	METHOD method_459 (Lnet/minecraft/class_1887;IF[Lnet/minecraft/class_4570;)Lnet/minecraft/class_117;
 		ARG 3 conditions
 	METHOD method_461 uniformBonusCount (Lnet/minecraft/class_1887;I)Lnet/minecraft/class_120$class_121;
-		ARG 0 type
+		ARG 0 enchantment
 		ARG 1 bonusMultiplier
 	METHOD method_462 (Lnet/minecraft/class_1887;I[Lnet/minecraft/class_4570;)Lnet/minecraft/class_117;
 		ARG 2 conditions
 	METHOD method_463 binomialWithBonusCount (Lnet/minecraft/class_1887;FI)Lnet/minecraft/class_120$class_121;
-		ARG 0 type
+		ARG 0 enchantment
 		ARG 1 probability
 		ARG 2 extra
 	CLASS class_95 BinomialWithBonusCount
--- a/mappings/net/minecraft/loot/function/ApplyBonusLootFunction.mapping
+++ b/mappings/net/minecraft/loot/function/ApplyBonusLootFunction.mapping
@@ -1,28 +1,28 @@
 CLASS net/minecraft/class_94 net/minecraft/loot/function/ApplyBonusLootFunction
 	FIELD field_1009 formula Lnet/minecraft/class_94$class_96;
 	FIELD field_1010 FACTORIES Ljava/util/Map;
-	FIELD field_1011 enchantment Lnet/minecraft/class_1887;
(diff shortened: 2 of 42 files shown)


Runemoro (approved):

mappings/net/minecraft/loot/function/EnchantRandomlyLootFunction.mapping
```diff
@@ -1,14 +1,14 @@
...
 		ARG 1 conditions
-		ARG 2 enchantments
+		ARG 2 types
 	METHOD method_488 ([Lnet/minecraft/class_4570;)Lnet/minecraft/class_117;
 		ARG 0 conditions
 	METHOD method_489 builder ()Lnet/minecraft/class_120$class_121;
 	CLASS class_110 Factory
 	CLASS class_4954 Builder
-		FIELD field_23090 enchantments Ljava/util/Set;
+		FIELD field_23090 types Ljava/util/Set;
```
Runemoro: I think we should keep calling things like these `enchantments`, even though they're types. Look at biome code, for example, which uses `entities` for lists of `EntityType`s which are allowed in the biome.

dblsaiko: Yeah, wasn't sure if I should keep those. I'll revert that then. Same for parameters and everything, or just collections as in this case?

dblsaiko:
Pretty sure I've got everything

TheGlitch76 (approved):

liach:
enchantment is more like a map of enchantment type to int level than a single entry like this. 👎

liach:
This is literally the worst rename I've ever seen... `LeveledEnchantment` is better

liach (changes requested):
This pr's naming of `Enchantment` -> `EnchantmentType` I can accept, but `InfoEnchantment` -> `Enchantment` is :100:% wrong and should never be done.

liach (changes requested):
People have doubted why I refuse this pr. Here I share my logic.

Look at [property](https://github.com/FabricMC/yarn/blob/20w11a/mappings/net/minecraft/state/property/Property.mapping), which represents a key in the `Properties` Compound tag in NBT format of block states. It has a separate value, just like how enchantment has a separate level.

According to @2xsaiko's logic, the object that holds the key-value pair as a whole should be named as `Enchantment` than just the key, and the key should be renamed to `EnchantmentType`.

Let's apply this logic on Property: the object that holds the key-value pair (available in client data gen code; I named it [PropertiesEntry](https://github.com/FabricMC/yarn/blob/20w11a/mappings/net/minecraft/data/client/model/PropertiesEntry.mapping)) should be renamed to `Propety`, and the old `Property` should be renamed to `PropertyType`.

I find his ideas sarcastic; both `InfoEnchantment` and `PropertiesEntry` are obscure classes that don't get exposed to the main game engine (`InfoEnchantment` is only used by Enchantment helper calculations, not in any way exposed to inventory item users (enchantments only exist on items) in `Item` or `ItemStack`; same for `PropertiesEntry` which is exclusively used in client data generator).

It is time to end this farce; making `InfoEnchantment` into `Enchantment` would only misguide users and is not how this thing will ever work. @Juuxel in #1057 complained that `EnchantmentEntry` is a bit too vague, but it doesn't matter because by design, this class is an internal API that is not exposed anywhere (`Item` or `ItemStack`). It's effectively a `Map.Entry<Enchantment, Integer>`, as I've said before, hence by the same logic how `PropertiesEntry` is named, I suggest `EnchantmentsEntry` (notice the `s` which means the enchantment map than a single enchantment) as the new name of `InfoEnchantment` class.

Prospector:
I'll leave my comments from discord here: 
image
image

liach:
> **Prospector**: This isn't a map

Source code:
image

@Prospector please explain how your logic doesn't apply to state property. We need the two same stuff to follow the same rule than setting different rules for each of them, which will lead to maintenance hells like mojang's comparator level 15 redstone input hell

liach:
So before you guys go on answer this question real quick:

Object | The current class representation's name<br> | Other representations<br>(non-mojang classes in vanilla) | The name according to my logic | The name according to this pr's logic
------- | -----------------| ----------- | -------------- | -----------------------------------
Fortune II | `InfoEnchantment`<br><br>only used when enchantments are added to item stacks via ench books or ench tables | `Map.Entry<Enchantment, Integer>`<br><br>seen in `EnchantmentHelper` | `EnchantmentsEntry` | `Enchantment`
Fortune | `Enchantment` | (none) | `Enchantment` | `EnchantmentType`
`"facing": "east"`<br><br>(in a block state json/nbt) | `PropertiesEntry`<br><br>only used in client data generator | `Map.Entry<Propety, Comparable<?>>`<br><br> seen as just `Map` instead of used individually most of the time | `PropertiesEntry` | ***PAIR***
`"facing"`<br><br>the key of the thing above | `Property` | none | `Property` | ***KEY***

Please give me an answer for ***PAIR*** and ***KEY***. 
I currently assume that you'd put `Property` for ***PAIR*** and `PropertyType` for ***KEY***.

Guys, especially @2xsaiko who opened this pr, please post your answer. I find a logical fallacy in this pr if it doesn't ask for a name change for `net.minecraft.state.property.Property`.
Prospector answered `Property` for ***PAIR*** and `PropertyKey` for ***KEY***.

Prospector:
No, I changed my answer after I thought about it more. **_PAIR_** should be something like `StateComponent` and **_KEY_** should be `Property`

liach:
Then same argument goes: we can have the enchantment pair as `ItemStackComponent` and the enchantment key as `Enchantment`. note that states have not only these key-value pairs but also a field for the block/fluid type's id. the item stack, holding enchantment pairs as one of its components, is similar, albeit it has a few more fields.

dblsaiko:
Key: `Property`. Like a Kotlin property, or the equivalent in Java, which is a field. Close enough.
Pair: Not sure, but I'm not a big fan of `PropertiesEntry`. I can't come up with something that I like right now.

> we can have the enchantment pair as `ItemStackComponent`

No, we can't, since that's not its only use. It's also at least used in enchantment books as a standalone thing. And additionally that name is incredibly misleading.

Prospector:
Not at all, the enchantment pair *is* the enchantment. These are entirely unrelated and there is no reason to group these unrelated but vaguely structurally similar things together.

liach:
image
This should really be `ApplicableEnchantment`

liach:
Another evidence for `EnchantmentEntry`:
image

Runemoro:
@liach good point, but I think that it's `Property` that's named wrong. It should be `PropertyType`, and `Property` should be used for the class that actually stores the property.

liach:
guys, you need to think of the implications when someone puts your `Enchantment` into hashmap keys and cry "oh wtf this doesn't have hashcode/equals while it totally can and should have". Don't merge another #650 and bring another disaster to this project.

Runemoro:
Actually, it makes sense for it to not have an equals method. The enchantments on two different items are different enchantments even if they have the same type and level.

liach:
according to your logic, two empty strings should not be equal if they are used to represent a suffix and a prefix. doesn't make sense to be. also this enchantment entry is not bound to item stacks at all

liach:
Also I have another doubt: what makes `EnchantmentType` better than `EnchantmentKind` or `EnchantmentId`, since the NBT is like `Enchantments:[{id:"minecraft:protection", level: 1}]`

liach:
Another conceptual question: If this enchantment entry becomes enchantment, then `Mending II` is considered different from `Mending` and won't trigger `Mending` logic if you just check the enchantment entry. That just... doesn't make sense.

mappings/net/minecraft/loot/function/EnchantRandomlyLootFunction.mapping
liach: https://minecraft.gamepedia.com/Loot_table#Functions
the json field is `enchantments`

mappings/net/minecraft/loot/condition/TableBonusLootCondition.mapping
```diff
@@ -1,11 +1,11 @@
 CLASS net/minecraft/class_182 net/minecraft/loot/condition/TableBonusLootCondition
-	FIELD field_1234 enchantment Lnet/minecraft/class_1887;
+	FIELD field_1234 type Lnet/minecraft/class_1887;
```
liach: This field is called `enchantment` in the json format. https://minecraft.gamepedia.com/Loot_table#Functions

mappings/net/minecraft/loot/function/ApplyBonusLootFunction.mapping
```diff
@@ -1,28 +1,28 @@
 CLASS net/minecraft/class_94 net/minecraft/loot/function/ApplyBonusLootFunction
 	FIELD field_1009 formula Lnet/minecraft/class_94$class_96;
 	FIELD field_1010 FACTORIES Ljava/util/Map;
-	FIELD field_1011 enchantment Lnet/minecraft/class_1887;
+	FIELD field_1011 type Lnet/minecraft/class_1887;
```
liach: This field is called `enchantment` in json format. https://minecraft.gamepedia.com/Loot_table#Functions

mappings/net/minecraft/predicate/item/EnchantmentPredicate.mapping
```diff
@@ -1,13 +1,15 @@
 CLASS net/minecraft/class_2035 net/minecraft/predicate/item/EnchantmentPredicate
 	FIELD field_20687 ARRAY_OF_ANY [Lnet/minecraft/class_2035;
-	FIELD field_9569 enchantment Lnet/minecraft/class_1887;
+	FIELD field_9569 type Lnet/minecraft/class_1887;
```
liach: https://minecraft.gamepedia.com/Advancements/Conditions/item
This json field is called `enchantment`

liach (changes requested):
You shouldn't rename the json fields in loot conditions or advancement conditions. :angry:

liach (approved):
Leaving my :-1: for merging this pr here, this pr's code looks good.

liach:
Linking another argument from the other pr https://github.com/FabricMC/yarn/pull/1182#issuecomment-599152094
I am sorry for being quick-tempered when I see this pr had a lot of upvotes (because I think this change would be really harmful to modders who migrate)

dblsaiko:
Closing in favor of #1182

PR #1182, opened 2020-03-14
InfoEnchantment -> EnchantmentLevelEntry (extends WeightedPicker.Entry)

### LINKED ISSUE #1057: InfoEnchantment -> ?

liach:
also Enchantment$Weight -> Enchantment$Rarity

Fixes #1181 and fixes #1057

### DIFF
--- a/mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
+++ b/mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
@@ -2,7 +2,7 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
 	METHOD method_17884 (Ljava/util/Map;Lnet/minecraft/class_2487;Lnet/minecraft/class_1887;)V
 		ARG 2 enchantment
 	METHOD method_22445 fromTag (Lnet/minecraft/class_2499;)Ljava/util/Map;
-		COMMENT Loads enchantments from a list named binary tag.
+		COMMENT Loads enchantments from an NBT list.
 		ARG 0 tag
 	METHOD method_24365 chooseEquipmentWith (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;Ljava/util/function/Predicate;)Ljava/util/Map$Entry;
 		COMMENT Returns a pair of an equipment slot and the item stack in the supplied
@@ -51,7 +51,7 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
 		ARG 0 entity
 	METHOD method_8206 getLoyalty (Lnet/minecraft/class_1799;)I
 		ARG 0 stack
-	METHOD method_8209 forEach (Lnet/minecraft/class_1890$class_1891;Ljava/lang/Iterable;)V
+	METHOD method_8209 forEachEnchantment (Lnet/minecraft/class_1890$class_1891;Ljava/lang/Iterable;)V
 		ARG 0 action
 		ARG 1 stacks
 	METHOD method_8210 onUserDamaged (Lnet/minecraft/class_1309;Lnet/minecraft/class_1297;)V
@@ -66,7 +66,10 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
 		COMMENT Sets the enchantments on an item stack.
 		COMMENT
 		COMMENT <p>For enchanted books, it sets the enchantments to the item stack's
-		COMMENT stored than regular enchantments.
+		COMMENT stored enchantments than regular enchantments.
+		COMMENT
+		COMMENT @see net.minecraft.item.ItemStack#getEnchantments()
+		COMMENT @see net.minecraft.item.EnchantedBookItem#getEnchantmentTag(net.minecraft.item.ItemStack)
 		ARG 0 enchantments
 		ARG 1 stack
 	METHOD method_8215 getLure (Lnet/minecraft/class_1799;)I
@@ -81,7 +84,7 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
 	METHOD method_8219 getProtectionAmount (Ljava/lang/Iterable;Lnet/minecraft/class_1282;)I
 		ARG 0 equipment
 		ARG 1 source
-	METHOD method_8220 process (Lnet/minecraft/class_1890$class_1891;Lnet/minecraft/class_1799;)V
+	METHOD method_8220 forEachEnchantment (Lnet/minecraft/class_1890$class_1891;Lnet/minecraft/class_1799;)V
 		ARG 0 action
 		ARG 1 stack
 	METHOD method_8221 hasVanishingCurse (Lnet/minecraft/class_1799;)Z
@@ -91,6 +94,9 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
 		COMMENT
 		COMMENT <p>For enchanted books, it retrieves from the item stack's stored than
 		COMMENT regular enchantments.
+		COMMENT
+		COMMENT @see net.minecraft.item.ItemStack#getEnchantments()
+		COMMENT @see net.minecraft.item.EnchantedBookItem#getEnchantmentTag(net.minecraft.item.ItemStack)
 		ARG 0 stack
 	METHOD method_8223 getLuckOfTheSea (Lnet/minecraft/class_1799;)I
 		ARG 0 stack
--- a/mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
+++ b/mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
@@ -1,9 +1,19 @@
 CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
 	METHOD method_17884 (Ljava/util/Map;Lnet/minecraft/class_2487;Lnet/minecraft/class_1887;)V
 		ARG 2 enchantment
-	METHOD method_22445 getEnchantments (Lnet/minecraft/class_2499;)Ljava/util/Map;
+	METHOD method_22445 fromTag (Lnet/minecraft/class_2499;)Ljava/util/Map;
+		COMMENT Loads enchantments from a list named binary tag.
 		ARG 0 tag
-	METHOD method_24365 getMatchingEnchantedEquipment (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;Ljava/util/function/Predicate;)Ljava/util/Map$Entry;
+	METHOD method_24365 getRandomEnchantedEquipment (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;Ljava/util/function/Predicate;)Ljava/util/Map$Entry;
+		COMMENT Returns a pair of an equipment slot and the item stack in the supplied
+		COMMENT entity's slot, indicating the item stack has the enchantment supplied
+		COMMENT and fulfills the extra condition.
+		COMMENT
+		COMMENT <p>If multiple equipment slots' item stacks are valid, a random pair is
+		COMMENT returned.
+		ARG 0 enchantment
+		ARG 1 entity
+		ARG 2 condition
 	METHOD method_25951 hasSoulSpeed (Lnet/minecraft/class_1309;)Z
 		ARG 0 entity
 	METHOD method_8199 getFireAspect (Lnet/minecraft/class_1309;)I
@@ -18,15 +28,24 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
 	METHOD method_8202 getRiptide (Lnet/minecraft/class_1799;)I
 		ARG 0 stack
 	METHOD method_8203 getEquipmentLevel (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;)I
-		ARG 0 ench
+		COMMENT Returns the highest level of the passed enchantment in the enchantment's
+		COMMENT applicable equipment slots' item stacks.
+		ARG 0 enchantment
 		ARG 1 entity
 	METHOD method_8204 getRandomEnchantedEquipment (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;)Ljava/util/Map$Entry;
+		COMMENT Returns a pair of an equipment slot and the item stack in the supplied
+		COMMENT entity's slot, indicating the item stack has the enchantment supplied.
+		COMMENT
+		COMMENT <p>If multiple equipment slots' item stacks are valid, a random pair is
+		COMMENT returned.
+		ARG 0 enchantment
+		ARG 1 entity
 	METHOD method_8205 getKnockback (Lnet/minecraft/class_1309;)I
 		ARG 0 entity
 	METHOD method_8206 getLoyalty (Lnet/minecraft/class_1799;)I
 		ARG 0 stack
-	METHOD method_8209 accept (Lnet/minecraft/class_1890$class_1891;Ljava/lang/Iterable;)V
-		ARG 0 enchantmentHandler
+	METHOD method_8209 forEach (Lnet/minecraft/class_1890$class_1891;Ljava/lang/Iterable;)V
+		ARG 0 action
 		ARG 1 stacks
 	METHOD method_8210 onUserDamaged (Lnet/minecraft/class_1309;Lnet/minecraft/class_1297;)V
 		ARG 0 user
@@ -37,6 +56,10 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
 		ARG 0 user
 		ARG 1 target
 	METHOD method_8214 set (Ljava/util/Map;Lnet/minecraft/class_1799;)V
+		COMMENT Sets the enchantments on an item stack.
+		COMMENT
+		COMMENT <p>For enchanted books, it sets the enchantments to the item stack's
+		COMMENT stored than regular enchantments.
 		ARG 0 enchantments
 		ARG 1 stack
 	METHOD method_8215 getLure (Lnet/minecraft/class_1799;)I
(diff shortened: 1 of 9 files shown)


mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
```diff
@@ -1,9 +1,19 @@
 CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
 	METHOD method_17884 (Ljava/util/Map;Lnet/minecraft/class_2487;Lnet/minecraft/class_1887;)V
 		ARG 2 enchantment
-	METHOD method_22445 getEnchantments (Lnet/minecraft/class_2499;)Ljava/util/Map;
+	METHOD method_22445 fromTag (Lnet/minecraft/class_2499;)Ljava/util/Map;
+		COMMENT Loads enchantments from a list named binary tag.
 		ARG 0 tag
-	METHOD method_24365 getMatchingEnchantedEquipment (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;Ljava/util/function/Predicate;)Ljava/util/Map$Entry;
+	METHOD method_24365 getRandomEnchantedEquipment (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;Ljava/util/function/Predicate;)Ljava/util/Map$Entry;
```
Earthcomputer: I feel like "Random" shouldn't be part of the name here, as it might be an implementation detail. I would rather go with something like `getEquipmentEnchantedWith`, while keeping the new javadocs.

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
```diff
@@ -18,15 +28,24 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
 	METHOD method_8202 getRiptide (Lnet/minecraft/class_1799;)I
 		ARG 0 stack
 	METHOD method_8203 getEquipmentLevel (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;)I
-		ARG 0 ench
+		COMMENT Returns the highest level of the passed enchantment in the enchantment's
+		COMMENT applicable equipment slots' item stacks.
+		ARG 0 enchantment
 		ARG 1 entity
 	METHOD method_8204 getRandomEnchantedEquipment (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;)Ljava/util/Map$Entry;
```
Earthcomputer: Similarly, this could be renamed to `getEquipmentEnchantedWith`

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
```diff
@@ -51,48 +74,71 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
...
+		ARG 0 possibleEntries
+			COMMENT the possible entries
+		ARG 1 pickedEntry
+			COMMENT the picked entry
 	METHOD method_8232 getDepthStrider (Lnet/minecraft/class_1309;)I
 		ARG 0 entity
 	METHOD method_8233 enchant (Ljava/util/Random;Lnet/minecraft/class_1799;IZ)Lnet/minecraft/class_1799;
+		COMMENT Enchants the {@code target} item stack and returns it.
 		ARG 0 random
+			COMMENT the seed
```
Earthcomputer, line 148: It's not a seed, it's a seeded Random

mappings/net/minecraft/enchantment/EnchantmentsEntry.mapping
```diff
@@ -0,0 +1,12 @@
+CLASS net/minecraft/class_1889 net/minecraft/enchantment/EnchantmentsEntry
```
Earthcomputer: Perhaps use the singular `EnchantmentEntry` instead?

Earthcomputer (changes requested):

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
liach: how about getRandom -> choose?

mappings/net/minecraft/enchantment/EnchantmentsEntry.mapping
liach: enchantments refers to the nbt list Enchantments or StoredEnchantments. This object is part of that list rather than part of the current Enchantment we have, so I named it with an s.

liach: Done, thanks earth!

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
liach: Yep, redocumented

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
liach: renamed to `chooseEquipmentWith`, so you call it like `EnchantmentHelper.chooseEquipmentWith(Enchantments.MENDING, entity)`.

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
liach: Renamed. The new name is `chooseEquipmentWith`

Earthcomputer (approved):

dblsaiko (changes requested):
`EnchantmentEntry` is a bad name, since that's not its only purpose. E.g. it's used as the enchantment an enchanted book holds. This is different from a map entry since it will always be used in the context of a map.

Another name I thought of a few hours ago (but couldn't suggest due to me not being at home) is `LevelledEnchantment`. Similar to `EnchantmentLevel` which was already suggested but doesn't imply being the level alone, but instead an enchantment with a level associated.

liach:
@2xsaiko I shall dismiss your review here. There are plenty of places showing that what we currently have as `Enchantment` is indeed called `Enchantment` in minecraft, evident in advancement item predicates https://minecraft.gamepedia.com/Advancements/Conditions/item 
image
image
and Loot functions
image
image
image
image
Imo these are sufficient evidence for keeping the current `Enchantment`, as all these `enchantment` reference are to the current `Enchantment` than the enchantment - level pair.

Prospector (approved):
Not entirely happy with this name but it does convey that a) it shouldn't be extended and it doesn't define logic and b) that it isn't an enchantment like `InfoEnchantment` appears. It also doesn't require the rather pointless rename of every enchantment class to have a `Type` suffix. Overall, this name is 100x better than `InfoEnchantment` which is what really matters, although I have a feeling this isn't the last time we'll see this class in a yarn refactor PR.

dblsaiko:
> @2xsaiko I shall dismiss your review here. There are plenty of places showing that what we currently have as `Enchantment` is indeed called `Enchantment` in minecraft, evident in advancement item predicates https://minecraft.gamepedia.com/Advancements/Conditions/item
> [snip]
> Imo these are sufficient evidence for keeping the current `Enchantment`, as all these `enchantment` reference are to the current `Enchantment` than the enchantment - level pair.

How is that relevant to anything I've said? Not really sure what you're trying to convey here.

liach:
It shows that your pr which renames `Enchantment` to `EnchantmentType` and `InfoEnchantment` to `Enchantment` is wrong, as the current name for `Enchantment` is right. This `InfoEnchantment` class just appears to be one of that left from the suffixation and was most likely intended to be `EnchantmentInfo`.

@2xsaiko just a quick note why this pair isn't really "enchantment":
> This is usually used where multiple enchantment to level mappings can exist, such as in enchanting logic. In other cases, vanilla prefers `Map<Enchantment, Integer>` over `List<EnchantmentLevelEntry>`.

Also I named it `EnchantmentLevelEntry` now as it does carry the level information, and in enchanting logic's generated entry list, there can be multiple entries, like `Fortune II` and `Fortune III` in the enchantment level entries list together. If you name `InfoEnchantment` as `Enchantment`, it doesn't make sense for `Fortune II` and `Fortune III` to be in the list of enchantments together.

> This is different from a map entry since it will always be used in the context of a map.

Also about the usage of this `InfoEnchantment`: It is not intended to be exposed to massive users. See https://github.com/FabricMC/yarn/blob/9517c35cdfa06024e423182f2097b43008ac2403/mappings/net/minecraft/enchantment/EnchantmentHelper.mapping#L151-L154
If mojang really intend to use this `InfoEnchantment` as the enchantment, then this method would not accept `Enchantment, int` but would just accept `InfoEnchantment`.
Most mods adding enchantments may not even use `InfoEnchantment` at all, unlike `StatusEffectInstance` or `ItemStack`, which are used way more frequently.

dblsaiko:
Right, but why not write that in the other PR, it would fit better there. This has nothing to do with renaming `Enchantment` so especially those screenshots take up unnecessary space on this page imo. (I also think that those keys aren't really sufficient because 1. once picked they can't be changed because of backwards compatibility; 2. if I wrote this and called that class EnchantmentType, I'd probably still name these keys 'enchantment' because that's descriptive unlike 'type,' and I wouldn't take 'enchantment_type' because it's longer. I think the registry for these being called 'enchantment' unlike 'entity_type' is a much better argument.)

EnchantmentLevelEntry is actually okay, I guess. Still not 100% sold on the Entry because of what I already said, but fine.

dblsaiko (approved):

mappings/net/minecraft/enchantment/EnchantmentLevelEntry.mapping
```diff
@@ -0,0 +1,16 @@
...
+	COMMENT Represents an entry in an enchantments array, i.e. a pair between
+	COMMENT enchantment and level. This is also a weighted entry and is available
+	COMMENT for the weighted picker to pick from a list.
+	COMMENT
+	COMMENT <p>This is usually used where multiple enchantment to level mappings can
+	COMMENT exist, such as in enchanting logic. In other cases, vanilla prefers
+	COMMENT {@code Map<Enchantment, Integer>} over {@code List<EnchantmentLevelEntry>}.
+	COMMENT
+	COMMENT <p>This class, though effectively immutable, does not override hashCode
+	COMMENT or equals.
```
Juuxel: I don't think defining immutability has anything to do with `hashCode` and `equals`.

Juuxel (approved):

mappings/net/minecraft/enchantment/EnchantmentLevelEntry.mapping
liach: Well, hashCode and equals are unreliable if an object isn't immutbale, i.e. immutability is a prerequisite for useful hashCode/equals

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
```diff
@@ -1,9 +1,22 @@
 CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
 	METHOD method_17884 (Ljava/util/Map;Lnet/minecraft/class_2487;Lnet/minecraft/class_1887;)V
 		ARG 2 enchantment
-	METHOD method_22445 getEnchantments (Lnet/minecraft/class_2499;)Ljava/util/Map;
+	METHOD method_22445 fromTag (Lnet/minecraft/class_2499;)Ljava/util/Map;
+		COMMENT Loads enchantments from a list named binary tag.
```
Pyrofab:
```suggestion
		COMMENT Loads enchantments from a NBT list.
```
I initially read "a list named binary tag" as "a list named 'binary tag'"

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
```diff
@@ -1,9 +1,22 @@
...
 		ARG 0 tag
-	METHOD method_24365 getMatchingEnchantedEquipment (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;Ljava/util/function/Predicate;)Ljava/util/Map$Entry;
+	METHOD method_24365 chooseEquipmentWith (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;Ljava/util/function/Predicate;)Ljava/util/Map$Entry;
```
Pyrofab, line 7:
```suggestion
	METHOD method_24365 findEquipmentWith (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;Ljava/util/function/Predicate;)Ljava/util/Map$Entry;
```
I think `find` is more appropriate than `choose` here.

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
```diff
@@ -18,15 +31,28 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
...
 		ARG 0 stack
 	METHOD method_8203 getEquipmentLevel (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;)I
-		ARG 0 ench
+		COMMENT Returns the highest level of the passed enchantment in the enchantment's
+		COMMENT applicable equipment slots' item stacks.
+		ARG 0 enchantment
+			COMMENT the enchantment
+		ARG 1 entity
+			COMMENT the entity whose equipment slots are checked
+	METHOD method_8204 chooseEquipmentWith (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;)Ljava/util/Map$Entry;
```
Pyrofab, line 40:
```suggestion
	METHOD method_8204 findEquipmentWith (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;)Ljava/util/Map$Entry;
```

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
```diff
@@ -18,15 +31,28 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
...
 		ARG 1 entity
-	METHOD method_8204 getRandomEnchantedEquipment (Lnet/minecraft/class_1887;Lnet/minecraft/class_1309;)Ljava/util/Map$Entry;
+			COMMENT the entity to choose equipments from
 	METHOD method_8205 getKnockback (Lnet/minecraft/class_1309;)I
 		ARG 0 entity
 	METHOD method_8206 getLoyalty (Lnet/minecraft/class_1799;)I
 		ARG 0 stack
-	METHOD method_8209 accept (Lnet/minecraft/class_1890$class_1891;Ljava/lang/Iterable;)V
-		ARG 0 enchantmentHandler
+	METHOD method_8209 forEach (Lnet/minecraft/class_1890$class_1891;Ljava/lang/Iterable;)V
```
Pyrofab:
```suggestion
	METHOD method_8209 forEachEnchantment (Lnet/minecraft/class_1890$class_1891;Ljava/lang/Iterable;)V
```
This method is basically applying an action to every enchantment of every stack in the list.

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
```diff
@@ -51,48 +81,71 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
 	METHOD method_8219 getProtectionAmount (Ljava/lang/Iterable;Lnet/minecraft/class_1282;)I
 		ARG 0 equipment
 		ARG 1 source
-	METHOD method_8220 accept (Lnet/minecraft/class_1890$class_1891;Lnet/minecraft/class_1799;)V
-		ARG 0 enchantmentHandler
+	METHOD method_8220 process (Lnet/minecraft/class_1890$class_1891;Lnet/minecraft/class_1799;)V
```
Pyrofab:
```suggestion
	METHOD method_8220 forEachEnchantment (Lnet/minecraft/class_1890$class_1891;Lnet/minecraft/class_1799;)V
```
This method is quite similar to the one above, applying an action to every enchantment on the stack.

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
```diff
@@ -51,48 +81,71 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
...
+		ARG 0 action
 		ARG 1 stack
 	METHOD method_8221 hasVanishingCurse (Lnet/minecraft/class_1799;)Z
 		ARG 0 stack
-	METHOD method_8222 getEnchantments (Lnet/minecraft/class_1799;)Ljava/util/Map;
+	METHOD method_8222 get (Lnet/minecraft/class_1799;)Ljava/util/Map;
```
Pyrofab, line 92: Not sure about this change. `EnchantmentHelper.get(stack)` does not make it immediately clear that we are getting the enchantments on the stack.

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
```diff
@@ -51,48 +81,71 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
...
+		COMMENT Gets the enchantments on an item stack.
+		COMMENT
+		COMMENT <p>For enchanted books, it retrieves from the item stack's stored than
+		COMMENT regular enchantments.
```
Pyrofab, line 96:
```suggestion
		COMMENT <p>For enchanted books, it retrieves from the stack's stored enchantments
		COMMENT rather than from the regular enchantments data.
		COMMENT @see EnchantedBookItem#getEnchantmentTag
```

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
dblsaiko: I think 'an NBT list' would be more correct

Pyrofab: I hesitated on which to take. "an NBT list" does sound better if you say it as an acronym.

YanisBft (approved):

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
liach: If you use "find", I'd call it `findRandom` (see java stream), which... I like `choose` better

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
liach: private helper method, also note the first parameter is explicit `EnchantmentHelper.Consumer`, so like for each enchantment consumer in iterable

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
liach: It is like for only one item stack i think 🤔

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
liach: This is on par with `EnchantmentHelper.set` which already exists.

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
liach: Don't see the point of rephrasing the first part, see tag might be worthy

mappings/net/minecraft/enchantment/EnchantmentLevelEntry.mapping
liach: Updated to show that these two have no relationship

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
liach: Done

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
liach: Done.

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
liach: Done.

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
liach: Only added new see tags for both set and get methods.

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
liach: `find` fails to indicate the random aspect of this method, while `choose` can have `null` result if none of the options can be chosen.

Pyrofab: If you want to look at java streams, you could call it findAny, it has basically the same behaviour.

Pyrofab (approved):

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
liach: Java's stream `findAny` is more like entity selector's `sort=arbitrary`, while this method behaves like `sort=random`

Runemoro (approved):
Better than the current name, but I still think we should have `EnchantmentType` in the future.

modmuss50 (approved):

PR #1171, opened 2020-03-11
Moved a lot of "util" classes to more appropriate packages

Sollace:
A lot of classes have just been dumped into `net/minecraft/util` without much thought about whether they actually belong there. This PR just cleans them up a little and moves classes to their own packages themed after what they do.

Classes dealing with json data and serialization are in `net/minecraft/util/dynamic`, logging classes are in `net/minecraft/util/logging`, etc.

PR #1186, opened 2020-03-15
Mappings relating to code that uses CsvWriter

Sollace:
These are names I found whilst working on #1171 around where `ServerWorld` works with the CsvWriter.

- Changed the `exportX` methods to `dumpX` methods since that is the terminology used elsewhere, and fits with what Mojang seems to be doing.
- `BakedQuad` was missing mappings that I'm sure were mapped in 1.15, but weren't in this version

- Some names as per Mojang:

ServerChunkManager
    `method_21694` -> `getPendingTasks`
ServerTickScheduler
    `method_20825`  -> `getTicks`

```
writer.write(String.format("entities: %d\n", this.entitiesById.size()));
writer.write(String.format("block_entities: %d\n", this.field_9231.size()));
writer.write(String.format("block_ticks: %d\n", this.method_14196().getTicks()));
writer.write(String.format("fluid_ticks: %d\n", this.method_14179().getTicks()));
writer.write("distance_manager: " + lv.getTicketManager().getDebugString() + "\n");
writer.write(String.format("pending_tasks: %d\n", this.method_14178().getPendingTasks()));
```

`method_14196`, `method_14179`, `method_14178` don't show up as mappable in Enigma, so I couldn't confirm whether they have names or not.

That's about it, really.


mappings/net/minecraft/client/render/model/BakedQuad.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS net/minecraft/class_777 net/minecraft/client/render/model/BakedQuad
+	FIELD field_22441 directionalLighting Z
```
liach: Is this like the `"gui_light": "side"` in item model? I feel this name itself isn't very clear and might get some javadocs.

mappings/net/minecraft/server/world/ChunkTicketManager.mapping
```diff
@@ -69,6 +69,7 @@ CLASS net/minecraft/class_3204 net/minecraft/server/world/ChunkTicketManager
 		ARG 4 argument
 	METHOD method_21623 getTicket (J)Ljava/lang/String;
 		ARG 1 pos
+	METHOD method_21683 getDebugString ()Ljava/lang/String;
```
liach: I'd suggest `getDebugMessage` or `getDumpMessage` if this is only used for dumping. `getString` doesn't sound very informative...

mappings/net/minecraft/test/StructureTestUtil.mapping
```diff
@@ -38,3 +44,14 @@ CLASS net/minecraft/class_4525 net/minecraft/test/StructureTestUtil
...
+		ARG 0 altitude
+		ARG 1 pos
+		ARG 2 world
+	METHOD method_22369 createStructure (Ljava/lang/String;Lnet/minecraft/class_3218;)Lnet/minecraft/class_3499;
+		ARG 0 structureId
+		ARG 1 world
+	METHOD method_23646 createArea (Lnet/minecraft/class_2338;Lnet/minecraft/class_2338;I)Lnet/minecraft/class_3341;
+		ARG 0 pos1
+		ARG 1 pos2
+		ARG 2 horSize
```
liach:
```suggestion
		ARG 0 startPos
		ARG 1 endPos
		ARG 2 horizontalSize
```

mappings/net/minecraft/world/BlockRenderView.mapping
```diff
@@ -6,6 +6,9 @@ CLASS net/minecraft/class_1920 net/minecraft/world/BlockRenderView
 	METHOD method_23752 getColor (Lnet/minecraft/class_2338;Lnet/minecraft/world/level/ColorResolver;)I
 		ARG 1 pos
 		ARG 2 colorResolver
+	METHOD method_24852 getBrightness (Lnet/minecraft/class_2350;Z)F
+		ARG 1 direction
+		ARG 2 hasDirectionalLighting
```
liach:
```suggestion
		ARG 2 directionalLighting
```

liach (commented):
Looks :+1: good!

mappings/net/minecraft/server/world/ChunkTicketManager.mapping
Juuxel: +1 for `getDebugMessage`

Sollace: In this case the string it returns is describing the contents of the tick manager. So really, it's a bit like a `toString` method for dumps. `toDumpString`?

liach:
```suggestion
	METHOD method_21683 toDumpString ()Ljava/lang/String;
```

mappings/net/minecraft/client/render/model/BakedQuad.mapping
Sollace: That's a very good question. I tried to trace where it was coming from, and other parts refer to it as "shade/d" (which I will take responsibility for)

Neither name is idea, so i may look into it a little further to see if I can trace it back to an actual value in the json resources.

Sollace: The value traces back to `ModelElement$Deserializer` where the member name is called "shade" so I've changed everything to use that term.

```
            if (jsonObject5.has("shade") && !JsonHelper.hasBoolean(jsonObject5, "shade")) {
                throw new JsonParseException("Expected shade to be a Boolean");
            }
            boolean boolean10 = JsonHelper.getBoolean(jsonObject5, "shade", true);
```

Sollace: Perhaps it still needs a javadoc, because shade isn't exactly a clear meaning...

Juuxel (approved):

mappings/net/minecraft/server/world/ChunkTicketManager.mapping
```diff
@@ -69,6 +69,7 @@ CLASS net/minecraft/class_3204 net/minecraft/server/world/ChunkTicketManager
 		ARG 4 argument
 	METHOD method_21623 getTicket (J)Ljava/lang/String;
 		ARG 1 pos
+	METHOD method_21683 toDumpString ()Ljava/lang/String;
```
Runemoro, line 72: What's a dump string? How is different from `getDebugString`?

liach: a dumped string, as shown in debug command dump.

mappings/net/minecraft/test/StructureTestUtil.mapping
```diff
@@ -38,3 +44,14 @@ CLASS net/minecraft/class_4525 net/minecraft/test/StructureTestUtil
...
+		ARG 0 altitude
+		ARG 1 pos
+		ARG 2 world
+	METHOD method_22369 createStructure (Ljava/lang/String;Lnet/minecraft/class_3218;)Lnet/minecraft/class_3499;
+		ARG 0 structureId
+		ARG 1 world
+	METHOD method_23646 createArea (Lnet/minecraft/class_2338;Lnet/minecraft/class_2338;I)Lnet/minecraft/class_3341;
+		ARG 0 startPos
+		ARG 1 endPos
+		ARG 2 horizontalSize
```
Runemoro: What does "area" mean here? What's the "horizontal size" and why isn't it determined by start and end positions?

Sollace: The area is a `BlockBox` used for positioning structures. I've revised the parameters a bit, but basically the values are:

`pos` = Pos1 of the box. This is the `pos` parameter passed down from other methods.
`size` = Width/Height/Depth of the box (passed down as the size of the structure)
`margin` = Extra margin to expend the area by along the X+Z axis.

The final box's dimensions are calculated as follows:
`min = pos - [margin, 3, margin]`
`max = pos + size + [margin - 1, 30, margin - 1]`

liach (approved):
Just change branch to 20w12a and this pr should be good to go.

Sollace:
@liach Branch updated.

YanisBft (approved):

PR #1189, opened 2020-03-16
Update BambooSaplingBlock -> BambooShootBlock to match Mojang fixing …

dexman545:
…the incorrect naming in the lang file


liach (approved):

YanisBft:
As far as I know, we preferably follow the IDs rather than the default translation. We didn't named `HayBlock` "HayBaleBlock", `RedstoneBlock` "BlockOfRedstoneBlock" or `WritableBookItem` "BookAndQuillItem" for instance.

modmuss50 (changes requested):
No, I dont think this should be changed. It should follow the registry name as that is constant between all languages.

Issue #1185, opened 2020-03-14
Match Material Names to Those of the MC Wiki; Name Improvements


dexman545:
[As discussed in the Discord](https://discordapp.com/channels/507304429255393322/521545796882006027/688152058515161225), using names from the wiki would allow for easier referencing and other such utility.  Certain Yarn names, and those on the wiki, could do with some updating as well. Below is some naming conventions, and a table for the names.

Proposed Naming Conventions:
  1. When possible, match the wiki name for the material, otherwise change wiki name

		Creating a fabric/yarn wiki page that lists the relation between block and material would be useful if the above is not done
  2. Materials are descriptors, and as such, they should use adjective form of their name

		ie WOOD -> WOODEN
  3. When multiple materials should share a name, they are differentiated by the unique property that they differ by unless clearer with another name

		ie PLANT <=> REPLACEABLE_PLANT
  4. Unless the material only contains one block that can be described by its name, material names are not to be named using a block that it contains
	
2 is iffy, as it would likely add extra characters to each name without much benefit for several instances; as such, it is left off from the current suggestions (names should all be consistent though).
Some effort was also made to keeping old names.

| Current Yarn       | Wiki                    | Suggestion                                         | Reason                                                                                                                                                        |
|--------------------|-------------------------|----------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Underwater\_Plant  | Water Plant             | Underwater Plant, Water Plant                         | Match EntityGroup "Aquatic", or Match wiki                                                                                                                    |
| Earth              | Dirt                    | Topsoil, Earth                                     | Current yarn name is fine, but doesn't include grass or mycelium which is normally considered to be "earth"                                                   |
| Snow\_Block        | Snow                    | Snow, Snow Block                                   | Match Naming conventions                                                                                                                                      |
| Seagrass           | Replaceable Water Plant | Replaceable Underwater Plant, Replaceable Water Plant | Breaks convention set by replaceable plant, match wiki                                                                                                        |
| Snow               | Top Snow                | Replaceable Snow, Snow Layer                       | Match naming conventions                                                                                                                                      |
| Anvil              | Heavy Metal             | Repair Station, Heavy Metal                        | Has grindstone                                                                                                                                                |
| Pumpkin            | Vegetable               | Gourd                                              | Has melons, this name is more general                                                                                                                         |
[...]

PR #1190, opened 2020-03-16
Material Name Improvements

dexman545:
Begin work resolving #1185 
Other names could be improved.

Sand -> Aggregate
Part -> Supported
Packed Ice -> Dense Ice
Organic -> Solid Organic
Pumpkin -> Gourd
Clay -> Organic Product
Anvil -> Repair Station
Seagrass -> Replaceable Underwater Plant
Snow -> Snow Layer
Earth -> Soil

### DIFF
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -36,7 +36,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 		COMMENT Material for full sized snow blocks.
 	FIELD field_15935 PLANT Lnet/minecraft/class_3614;
 		COMMENT Material for plants such as flowers and crops
-	FIELD field_15936 MOB_DERIVED Lnet/minecraft/class_3614;
+	FIELD field_15936 ORGANIC_PRODUCT Lnet/minecraft/class_3614;
 		COMMENT Material for blocks that are a product of mob interactions such as honey or slime, or the infested blocks. Includes clay but not bone blocks.
 	FIELD field_15937 CAKE Lnet/minecraft/class_3614;
 	FIELD field_15938 BAMBOO_SAPLING Lnet/minecraft/class_3614;
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -3,7 +3,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15912 breakByHand Z
 	FIELD field_15913 COBWEB Lnet/minecraft/class_3614;
 	FIELD field_15914 STONE Lnet/minecraft/class_3614;
-		COMMENT Hard solid nonmetallic mineral matter of which rock is made, especially as a building material.
+		COMMENT Material for blocks that are stone or made from it, and generally prefer to be broken by a pickaxe.
 	FIELD field_15915 BUBBLE_COLUMN Lnet/minecraft/class_3614;
 	FIELD field_15916 AGGREGATE Lnet/minecraft/class_3614;
 		COMMENT A material or structure formed from a loosely compacted mass of fragments or particles.
@@ -16,7 +16,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 		COMMENT Not in use, but has foliage color.
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-	FIELD field_15924 SUB_BLOCK Lnet/minecraft/class_3614;
+	FIELD field_15924 PARTIAL_BLOCK Lnet/minecraft/class_3614;
 		COMMENT Material for non-full blocks such as redstone components and torches.
 	FIELD field_15925 CACTUS Lnet/minecraft/class_3614;
 	FIELD field_15926 REPLACEABLE_UNDERWATER_PLANT Lnet/minecraft/class_3614;
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -16,7 +16,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 		COMMENT Not in use, but has foliage color.
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-	FIELD field_15924 SUBBLOCK Lnet/minecraft/class_3614;
+	FIELD field_15924 SUB_BLOCK Lnet/minecraft/class_3614;
 		COMMENT Material for non-full blocks such as redstone components and torches.
 	FIELD field_15925 CACTUS Lnet/minecraft/class_3614;
 	FIELD field_15926 REPLACEABLE_UNDERWATER_PLANT Lnet/minecraft/class_3614;
@@ -65,7 +65,6 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 		COMMENT Material for gourds. Includes the carved pumpkin and jack o' lantern.
 	FIELD field_15955 TNT Lnet/minecraft/class_3614;
 	FIELD field_15956 REPLACEABLE_PLANT Lnet/minecraft/class_3614;
-		COMMENT Material for plant blocks that are replaceable.
 	FIELD field_15957 CARPET Lnet/minecraft/class_3614;
 	FIELD field_15958 ICE Lnet/minecraft/class_3614;
 		COMMENT Material for ice that can melt. See {@link #DENSE_ICE} for unmeltable ice.
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -56,7 +56,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15948 SNOW_LAYER Lnet/minecraft/class_3614;
 		COMMENT Material for non-full blocks of snow. Has the replaceable property.
 	FIELD field_15949 REPAIR_STATION Lnet/minecraft/class_3614;
-		COMMENT Material for blocks that can repair tools, ie. grindstone and anvils.
+		COMMENT Material for blocks that can repair tools, including grindstone and anvils.
 	FIELD field_15950 replaceable Z
 	FIELD field_15951 burnable Z
 	FIELD field_15952 BARRIER Lnet/minecraft/class_3614;
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -72,7 +72,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15960 pistonBehavior Lnet/minecraft/class_3619;
 	FIELD field_17008 SHULKER_BOX Lnet/minecraft/class_3614;
 	FIELD field_22223 NETHER_WOOD Lnet/minecraft/class_3614;
-		COMMENT Material for blocks crafted from stems (Nether logs).
+		COMMENT Material for blocks crafted from Nether stems and hyphae.
 	METHOD <init> (Lnet/minecraft/class_3620;ZZZZZZZLnet/minecraft/class_3619;)V
 		ARG 1 color
 		ARG 2 liquid
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -30,7 +30,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15931 WOOL Lnet/minecraft/class_3614;
 		COMMENT Material for wool and bed blocks.
 	FIELD field_15932 WOOD Lnet/minecraft/class_3614;
-		COMMENT Material for wood logs, and things crafted from them. Includes the nether stems (nether wood logs).
+		COMMENT Material for wood logs, and things crafted from them.
 	FIELD field_15933 PISTON Lnet/minecraft/class_3614;
 	FIELD field_15934 SNOW_BLOCK Lnet/minecraft/class_3614;
 		COMMENT Material for full sized snow blocks.
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -48,8 +48,8 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 		COMMENT Material for glass and glass-like blocks (includes sea lanterns and conduits).
 	FIELD field_15943 FIRE Lnet/minecraft/class_3614;
 	FIELD field_15944 color Lnet/minecraft/class_3620;
-	FIELD field_15945 FODDER Lnet/minecraft/class_3614;
-		COMMENT Food, especially dried hay or feed, for cattle and other livestock. Includes mycelium.
+	FIELD field_15945 SOLID_ORGANIC Lnet/minecraft/class_3614;
+		COMMENT Organic blocks that are solid, eg. hay, target, or grass.
 	FIELD field_15946 BAMBOO Lnet/minecraft/class_3614;
 	FIELD field_15947 UNDERWATER_PLANT Lnet/minecraft/class_3614;
 		COMMENT Material for underwater plants without the replaceable property.
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -37,7 +37,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15935 PLANT Lnet/minecraft/class_3614;
 		COMMENT Material for plants such as flowers and crops
 	FIELD field_15936 ORGANIC_PRODUCT Lnet/minecraft/class_3614;
-		COMMENT Material for blocks that are a product of mob interactions such as honey or slime, or the infested blocks. Includes clay but not bone blocks.
+		COMMENT Material for blocks that come from mobs such as honey, slime, or infested blocks. Includes clay but not bone blocks.
 	FIELD field_15937 CAKE Lnet/minecraft/class_3614;
 	FIELD field_15938 BAMBOO_SAPLING Lnet/minecraft/class_3614;
 	FIELD field_15939 blocksLight Z
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -68,7 +68,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 		COMMENT Material for plant blocks that are replaceable.
 	FIELD field_15957 CARPET Lnet/minecraft/class_3614;
 	FIELD field_15958 ICE Lnet/minecraft/class_3614;
-		COMMENT Material for ice that can melt. {@see DENSE_ICE} for unmeltable ice.
+		COMMENT Material for ice that can melt. See {@link #DENSE_ICE} for unmeltable ice.
 	FIELD field_15959 AIR Lnet/minecraft/class_3614;
 	FIELD field_15960 pistonBehavior Lnet/minecraft/class_3619;
 	FIELD field_17008 SHULKER_BOX Lnet/minecraft/class_3614;
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -49,7 +49,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15943 FIRE Lnet/minecraft/class_3614;
 	FIELD field_15944 color Lnet/minecraft/class_3620;
 	FIELD field_15945 SOLID_ORGANIC Lnet/minecraft/class_3614;
-		COMMENT Organic blocks that are solid, eg. hay, target, or grass blocks.
+		COMMENT Organic blocks that are solid, including hay, target, and grass blocks.
 	FIELD field_15946 BAMBOO Lnet/minecraft/class_3614;
 	FIELD field_15947 UNDERWATER_PLANT Lnet/minecraft/class_3614;
 		COMMENT Material for underwater plants without the replaceable property.
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -35,7 +35,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15934 SNOW_BLOCK Lnet/minecraft/class_3614;
 		COMMENT Material for full sized snow blocks.
 	FIELD field_15935 PLANT Lnet/minecraft/class_3614;
-		COMMENT Material for plants such as flowers and crops
+		COMMENT Material for plants such as flowers and crops.
 	FIELD field_15936 MOB_DERIVED Lnet/minecraft/class_3614;
 		COMMENT Material for blocks that are a product of mob interactions such as honey or slime, or the infested blocks. Includes clay but not bone blocks.
 	FIELD field_15937 CAKE Lnet/minecraft/class_3614;
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -42,7 +42,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15938 BAMBOO_SAPLING Lnet/minecraft/class_3614;
 	FIELD field_15939 blocksLight Z
 	FIELD field_15940 blocksMovement Z
-	FIELD field_15941 TOPSOIL Lnet/minecraft/class_3614;
+	FIELD field_15941 SOIL Lnet/minecraft/class_3614;
 		COMMENT Material for the top layer of soil. Path, dirt, podzol, soul soil, farmland and similar.
 	FIELD field_15942 GLASS Lnet/minecraft/class_3614;
 		COMMENT Material for glass and glass-like blocks (includes sea lanterns and conduits).
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -72,7 +72,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15960 pistonBehavior Lnet/minecraft/class_3619;
 	FIELD field_17008 SHULKER_BOX Lnet/minecraft/class_3614;
 	FIELD field_22223 NETHER_WOOD Lnet/minecraft/class_3614;
-		COMMENT Material for blocks crafted from stems (Nether wood logs).
+		COMMENT Material for blocks crafted from stems (Nether logs).
 	METHOD <init> (Lnet/minecraft/class_3620;ZZZZZZZLnet/minecraft/class_3619;)V
 		ARG 1 color
 		ARG 2 liquid
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -16,8 +16,8 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 		COMMENT Not in use, but has foliage color.
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-	FIELD field_15924 PARTIAL_BLOCK Lnet/minecraft/class_3614;
-		COMMENT Material for non-full blocks such as redstone components, torches and skulls.
+	FIELD field_15924 SUPPORTED Lnet/minecraft/class_3614;
+		COMMENT Material for blocks that require a supporting block such as redstone components, torches, flower pots, and skulls.
 	FIELD field_15925 CACTUS Lnet/minecraft/class_3614;
 	FIELD field_15926 REPLACEABLE_UNDERWATER_PLANT Lnet/minecraft/class_3614;
 	FIELD field_15927 STRUCTURE_VOID Lnet/minecraft/class_3614;
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -23,7 +23,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15927 STRUCTURE_VOID Lnet/minecraft/class_3614;
 		COMMENT Material for structure void block.
 	FIELD field_15928 DENSE_ICE Lnet/minecraft/class_3614;
-		COMMENT Material for ice blocks that do not melt. {@see ICE} for meltable ice.
+		COMMENT Material for ice blocks that do not melt. See {@link #ICE} for meltable ice.
 	FIELD field_15929 liquid Z
 	FIELD field_15930 EGG Lnet/minecraft/class_3614;
 		COMMENT Material for egg blocks, such as dragon and turtle eggs.
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -17,7 +17,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
 	FIELD field_15924 PARTIAL_BLOCK Lnet/minecraft/class_3614;
-		COMMENT Material for non-full blocks such as redstone components and torches.
+		COMMENT Material for non-full blocks such as redstone components, torches and skulls.
 	FIELD field_15925 CACTUS Lnet/minecraft/class_3614;
 	FIELD field_15926 REPLACEABLE_UNDERWATER_PLANT Lnet/minecraft/class_3614;
 	FIELD field_15927 STRUCTURE_VOID Lnet/minecraft/class_3614;
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -49,7 +49,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15943 FIRE Lnet/minecraft/class_3614;
 	FIELD field_15944 color Lnet/minecraft/class_3620;
 	FIELD field_15945 SOLID_ORGANIC Lnet/minecraft/class_3614;
-		COMMENT Organic blocks that are solid, eg. hay, target, or grass.
+		COMMENT Organic blocks that are solid, eg. hay, target, or grass blocks.
 	FIELD field_15946 BAMBOO Lnet/minecraft/class_3614;
 	FIELD field_15947 UNDERWATER_PLANT Lnet/minecraft/class_3614;
 		COMMENT Material for underwater plants without the replaceable property.
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -24,7 +24,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
 		COMMENT Material for leaves.
 	FIELD field_15924 SUBBLOCK Lnet/minecraft/class_3614;
-		COMMENT Nonfull blocks such as redstone components and torches.
+		COMMENT Material for non-full blocks such as redstone components and torches.
 	FIELD field_15925 CACTUS Lnet/minecraft/class_3614;
 		COMMENT Material for cactus.
 	FIELD field_15926 REPLACEABLE_UNDERWATER_PLANT Lnet/minecraft/class_3614;
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -14,7 +14,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
 	FIELD field_15924 DECORATION Lnet/minecraft/class_3614;
 	FIELD field_15925 CACTUS Lnet/minecraft/class_3614;
-	FIELD field_15926 REPLACEABLE_AQUATIC_PLANT Lnet/minecraft/class_3614;
+	FIELD field_15926 REPLACEABLE_UNDERWATER_PLANT Lnet/minecraft/class_3614;
 	FIELD field_15927 STRUCTURE_VOID Lnet/minecraft/class_3614;
 	FIELD field_15928 DENSE_ICE Lnet/minecraft/class_3614;
 	FIELD field_15929 liquid Z
@@ -35,7 +35,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15944 color Lnet/minecraft/class_3620;
 	FIELD field_15945 FODDER Lnet/minecraft/class_3614;
 	FIELD field_15946 BAMBOO Lnet/minecraft/class_3614;
-	FIELD field_15947 AQUATIC_PLANT Lnet/minecraft/class_3614;
+	FIELD field_15947 UNDERWATER_PLANT Lnet/minecraft/class_3614;
 	FIELD field_15948 SNOW_LAYER Lnet/minecraft/class_3614;
 	FIELD field_15949 REPAIR_STATION Lnet/minecraft/class_3614;
 	FIELD field_15950 replaceable Z
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -74,7 +74,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15950 replaceable Z
 	FIELD field_15951 burnable Z
 	FIELD field_15952 BARRIER Lnet/minecraft/class_3614;
-		COMMENT 
+		COMMENT Material for barrier blocks.
 	FIELD field_15953 METAL Lnet/minecraft/class_3614;
 		COMMENT Material for barrier blocks.
 	FIELD field_15954 GOURD Lnet/minecraft/class_3614;
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -4,7 +4,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15913 COBWEB Lnet/minecraft/class_3614;
 	FIELD field_15914 STONE Lnet/minecraft/class_3614;
 	FIELD field_15915 BUBBLE_COLUMN Lnet/minecraft/class_3614;
-	FIELD field_15916 SAND Lnet/minecraft/class_3614;
+	FIELD field_15916 AGGREGATE Lnet/minecraft/class_3614;
 	FIELD field_15917 SPONGE Lnet/minecraft/class_3614;
 	FIELD field_15918 REDSTONE_LAMP Lnet/minecraft/class_3614;
 	FIELD field_15919 PORTAL Lnet/minecraft/class_3614;
@@ -12,11 +12,11 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15921 UNUSED_PLANT Lnet/minecraft/class_3614;
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-	FIELD field_15924 PART Lnet/minecraft/class_3614;
+	FIELD field_15924 DECORATION Lnet/minecraft/class_3614;
 	FIELD field_15925 CACTUS Lnet/minecraft/class_3614;
-	FIELD field_15926 SEAGRASS Lnet/minecraft/class_3614;
+	FIELD field_15926 REPLACEABLE_AQUATIC_PLANT Lnet/minecraft/class_3614;
 	FIELD field_15927 STRUCTURE_VOID Lnet/minecraft/class_3614;
-	FIELD field_15928 PACKED_ICE Lnet/minecraft/class_3614;
+	FIELD field_15928 DENSE_ICE Lnet/minecraft/class_3614;
 	FIELD field_15929 liquid Z
 	FIELD field_15930 EGG Lnet/minecraft/class_3614;
 	FIELD field_15931 WOOL Lnet/minecraft/class_3614;
@@ -24,25 +24,25 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15933 PISTON Lnet/minecraft/class_3614;
 	FIELD field_15934 SNOW_BLOCK Lnet/minecraft/class_3614;
 	FIELD field_15935 PLANT Lnet/minecraft/class_3614;
-	FIELD field_15936 CLAY Lnet/minecraft/class_3614;
+	FIELD field_15936 MOB_DERIVED Lnet/minecraft/class_3614;
 	FIELD field_15937 CAKE Lnet/minecraft/class_3614;
-	FIELD field_15938 BAMBOO_SAPLING Lnet/minecraft/class_3614;
+	FIELD field_15938 BAMBOO_SHOOT Lnet/minecraft/class_3614;
 	FIELD field_15939 blocksLight Z
 	FIELD field_15940 blocksMovement Z
-	FIELD field_15941 EARTH Lnet/minecraft/class_3614;
+	FIELD field_15941 TOPSOIL Lnet/minecraft/class_3614;
 	FIELD field_15942 GLASS Lnet/minecraft/class_3614;
 	FIELD field_15943 FIRE Lnet/minecraft/class_3614;
 	FIELD field_15944 color Lnet/minecraft/class_3620;
-	FIELD field_15945 ORGANIC Lnet/minecraft/class_3614;
+	FIELD field_15945 FODDER Lnet/minecraft/class_3614;
 	FIELD field_15946 BAMBOO Lnet/minecraft/class_3614;
-	FIELD field_15947 UNDERWATER_PLANT Lnet/minecraft/class_3614;
-	FIELD field_15948 SNOW Lnet/minecraft/class_3614;
-	FIELD field_15949 ANVIL Lnet/minecraft/class_3614;
+	FIELD field_15947 AQUATIC_PLANT Lnet/minecraft/class_3614;
+	FIELD field_15948 SNOW_LAYER Lnet/minecraft/class_3614;
+	FIELD field_15949 REPAIR_STATION Lnet/minecraft/class_3614;
 	FIELD field_15950 replaceable Z
 	FIELD field_15951 burnable Z
 	FIELD field_15952 BARRIER Lnet/minecraft/class_3614;
 	FIELD field_15953 METAL Lnet/minecraft/class_3614;
-	FIELD field_15954 PUMPKIN Lnet/minecraft/class_3614;
+	FIELD field_15954 GOURD Lnet/minecraft/class_3614;
 	FIELD field_15955 TNT Lnet/minecraft/class_3614;
 	FIELD field_15956 REPLACEABLE_PLANT Lnet/minecraft/class_3614;
 	FIELD field_15957 CARPET Lnet/minecraft/class_3614;
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -2,99 +2,99 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15911 solid Z
 	FIELD field_15912 breakByHand Z
 	FIELD field_15913 COBWEB Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for cobweb-like blocks.</p>
+		COMMENT Material for cobweb-like blocks.
 	FIELD field_15914 STONE Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Hard solid nonmetallic mineral matter of which rock is made, especially as a building material.</p>
+		COMMENT Hard solid nonmetallic mineral matter of which rock is made, especially as a building material.
 	FIELD field_15915 BUBBLE_COLUMN Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for water bubble streams.</p>
+		COMMENT Material for water bubble streams.
 	FIELD field_15916 AGGREGATE Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: A material or structure formed from a loosely compacted mass of fragments or particles.</p>
+		COMMENT A material or structure formed from a loosely compacted mass of fragments or particles.
 	FIELD field_15917 SPONGE Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for sponges.</p>
+		COMMENT Material for sponges.
 	FIELD field_15918 REDSTONE_LAMP Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for redstone lamps.</p>
+		COMMENT Material for redstone lamps.
 	FIELD field_15919 PORTAL Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for the various portal blocks.</p>
+		COMMENT Material for the various portal blocks.
 	FIELD field_15920 WATER Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for the water block.</p>
+		COMMENT Material for the water block.
 	FIELD field_15921 UNUSED_PLANT Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Not in use, but has foliage color.</p>
+		COMMENT Not in use, but has foliage color.
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for the lava block.</p>
+		COMMENT Material for the lava block.
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for leaves.</p>
+		COMMENT Material for leaves.
 	FIELD field_15924 SUBBLOCK Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Nonfull blocks such as redstone components and torches.</p>
+		COMMENT Nonfull blocks such as redstone components and torches.
 	FIELD field_15925 CACTUS Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for cactus.</p>
+		COMMENT Material for cactus.
 	FIELD field_15926 REPLACEABLE_UNDERWATER_PLANT Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for underwater plants that have the replaceable property. </p>
+		COMMENT Material for underwater plants that have the replaceable property. 
 	FIELD field_15927 STRUCTURE_VOID Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for structure void block.</p>
+		COMMENT Material for structure void block.
 	FIELD field_15928 DENSE_ICE Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for ice blocks that do not melt.</p>
+		COMMENT Material for ice blocks that do not melt.
 	FIELD field_15929 liquid Z
 	FIELD field_15930 EGG Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for egg blocks, such as dragon and turtle eggs.</p>
+		COMMENT Material for egg blocks, such as dragon and turtle eggs.
 	FIELD field_15931 WOOL Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for wool and bed blocks.</p>
+		COMMENT Material for wool and bed blocks.
 	FIELD field_15932 WOOD Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for wood logs, and things crafted from them. Includes the nether stems (nether wood logs).</p>
+		COMMENT Material for wood logs, and things crafted from them. Includes the nether stems (nether wood logs).
 	FIELD field_15933 PISTON Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for piston blocks.</p>
+		COMMENT Material for piston blocks.
 	FIELD field_15934 SNOW_BLOCK Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for full sized snow blocks.</p>
+		COMMENT Material for full sized snow blocks.
 	FIELD field_15935 PLANT Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for plants such as flowers and crops</p>
+		COMMENT Material for plants such as flowers and crops
 	FIELD field_15936 MOB_DERIVED Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for blocks that are a product of mob interactions such as honey or slime, or the infested blocks. Includes clay but not bone blocks.</p>
+		COMMENT Material for blocks that are a product of mob interactions such as honey or slime, or the infested blocks. Includes clay but not bone blocks.
 	FIELD field_15937 CAKE Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for cake.</p>
+		COMMENT Material for cake.
 	FIELD field_15938 BAMBOO_SAPLING Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for bamboo saplings.</p>
+		COMMENT Material for bamboo saplings.
 	FIELD field_15939 blocksLight Z
 	FIELD field_15940 blocksMovement Z
 	FIELD field_15941 TOPSOIL Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: The top layer of soil. Path, dirt, podzol, soul soil, farmland and similar.</p>
+		COMMENT The top layer of soil. Path, dirt, podzol, soul soil, farmland and similar.
 	FIELD field_15942 GLASS Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for glass and glass-like blocks (Includes sea lanterns and conduits).</p>
+		COMMENT Material for glass and glass-like blocks (Includes sea lanterns and conduits).
 	FIELD field_15943 FIRE Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for fire blocks.</p>
+		COMMENT Material for fire blocks.
 	FIELD field_15944 color Lnet/minecraft/class_3620;
 	FIELD field_15945 FODDER Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Food, especially dried hay or feed, for cattle and other livestock. Includes mycelium.</p>
+		COMMENT Food, especially dried hay or feed, for cattle and other livestock. Includes mycelium.
 	FIELD field_15946 BAMBOO Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for bamboo.</p>
+		COMMENT Material for bamboo.
 	FIELD field_15947 UNDERWATER_PLANT Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for underwater plants without the replaceable property.</p>
+		COMMENT Material for underwater plants without the replaceable property.
 	FIELD field_15948 SNOW_LAYER Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for non-full blocks of snow. Has the replaceable property.</p>
+		COMMENT Material for non-full blocks of snow. Has the replaceable property.
 	FIELD field_15949 REPAIR_STATION Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for blocks that can repair tools, ie. grindstone and anvils.</p>
+		COMMENT Material for blocks that can repair tools, ie. grindstone and anvils.
 	FIELD field_15950 replaceable Z
 	FIELD field_15951 burnable Z
 	FIELD field_15952 BARRIER Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: </p>
+		COMMENT 
 	FIELD field_15953 METAL Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for barrier blocks.</p>
+		COMMENT Material for barrier blocks.
 	FIELD field_15954 GOURD Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for gourds. Includes the carved pumpkin and jack o' lantern.</p>
+		COMMENT Material for gourds. Includes the carved pumpkin and jack o' lantern.
 	FIELD field_15955 TNT Lnet/minecraft/class_3614;
-		COMMENT <p>DEFINITION: Material for TNT</p>
+		COMMENT Material for TNT
 	FIELD field_15956 REPLACEABLE_PLANT Lnet/minecraft/class_3614;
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/block/Material.mapping
```diff
@@ -4,45 +4,45 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
-	FIELD field_15949 ANVIL Lnet/minecraft/class_3614;
+	FIELD field_15947 AQUATIC_PLANT Lnet/minecraft/class_3614;
+	FIELD field_15948 SNOW_LAYER Lnet/minecraft/class_3614;
+	FIELD field_15949 REPAIR_STATION Lnet/minecraft/class_3614;
 	FIELD field_15950 replaceable Z
 	FIELD field_15951 burnable Z
 	FIELD field_15952 BARRIER Lnet/minecraft/class_3614;
 	FIELD field_15953 METAL Lnet/minecraft/class_3614;
-	FIELD field_15954 PUMPKIN Lnet/minecraft/class_3614;
+	FIELD field_15954 GOURD Lnet/minecraft/class_3614;
```
modmuss50, line 64: Ive not heard of this before, not sure what other people think?

mappings/net/minecraft/block/Material.mapping
```diff
@@ -4,45 +4,45 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15943 FIRE Lnet/minecraft/class_3614;
 	FIELD field_15944 color Lnet/minecraft/class_3620;
-	FIELD field_15945 ORGANIC Lnet/minecraft/class_3614;
+	FIELD field_15945 FODDER Lnet/minecraft/class_3614;
 	FIELD field_15946 BAMBOO Lnet/minecraft/class_3614;
-	FIELD field_15947 UNDERWATER_PLANT Lnet/minecraft/class_3614;
-	FIELD field_15948 SNOW Lnet/minecraft/class_3614;
-	FIELD field_15949 ANVIL Lnet/minecraft/class_3614;
+	FIELD field_15947 AQUATIC_PLANT Lnet/minecraft/class_3614;
+	FIELD field_15948 SNOW_LAYER Lnet/minecraft/class_3614;
```
modmuss50, line 56: Whats the reasoning behind this? SNOW seems fine?

mappings/net/minecraft/block/Material.mapping
```diff
@@ -4,45 +4,45 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15935 PLANT Lnet/minecraft/class_3614;
-	FIELD field_15936 CLAY Lnet/minecraft/class_3614;
+	FIELD field_15936 MOB_DERIVED Lnet/minecraft/class_3614;
 	FIELD field_15937 CAKE Lnet/minecraft/class_3614;
-	FIELD field_15938 BAMBOO_SAPLING Lnet/minecraft/class_3614;
+	FIELD field_15938 BAMBOO_SHOOT Lnet/minecraft/class_3614;
 	FIELD field_15939 blocksLight Z
 	FIELD field_15940 blocksMovement Z
-	FIELD field_15941 EARTH Lnet/minecraft/class_3614;
+	FIELD field_15941 TOPSOIL Lnet/minecraft/class_3614;
```
modmuss50: This is used by dirt as well. I dont think this is correct

mappings/net/minecraft/block/Material.mapping
```diff
@@ -4,45 +4,45 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+	FIELD field_15928 DENSE_ICE Lnet/minecraft/class_3614;
 	FIELD field_15929 liquid Z
 	FIELD field_15930 EGG Lnet/minecraft/class_3614;
 	FIELD field_15931 WOOL Lnet/minecraft/class_3614;
 	FIELD field_15932 WOOD Lnet/minecraft/class_3614;
 	FIELD field_15933 PISTON Lnet/minecraft/class_3614;
 	FIELD field_15934 SNOW_BLOCK Lnet/minecraft/class_3614;
 	FIELD field_15935 PLANT Lnet/minecraft/class_3614;
-	FIELD field_15936 CLAY Lnet/minecraft/class_3614;
+	FIELD field_15936 MOB_DERIVED Lnet/minecraft/class_3614;
```
modmuss50: Not sure on the reasoning behind this, its only used by clay?

mappings/net/minecraft/block/Material.mapping
```diff
@@ -4,45 +4,45 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-	FIELD field_15924 PART Lnet/minecraft/class_3614;
+	FIELD field_15924 DECORATION Lnet/minecraft/class_3614;
 	FIELD field_15925 CACTUS Lnet/minecraft/class_3614;
-	FIELD field_15926 SEAGRASS Lnet/minecraft/class_3614;
+	FIELD field_15926 REPLACEABLE_AQUATIC_PLANT Lnet/minecraft/class_3614;
 	FIELD field_15927 STRUCTURE_VOID Lnet/minecraft/class_3614;
-	FIELD field_15928 PACKED_ICE Lnet/minecraft/class_3614;
+	FIELD field_15928 DENSE_ICE Lnet/minecraft/class_3614;
```
modmuss50, line 25: Why the change here, is called Packed ice in vanilla?

mappings/net/minecraft/block/Material.mapping
```diff
@@ -4,45 +4,45 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15919 PORTAL Lnet/minecraft/class_3614;
 	FIELD field_15920 WATER Lnet/minecraft/class_3614;
 	FIELD field_15921 UNUSED_PLANT Lnet/minecraft/class_3614;
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-	FIELD field_15924 PART Lnet/minecraft/class_3614;
+	FIELD field_15924 DECORATION Lnet/minecraft/class_3614;
 	FIELD field_15925 CACTUS Lnet/minecraft/class_3614;
-	FIELD field_15926 SEAGRASS Lnet/minecraft/class_3614;
+	FIELD field_15926 REPLACEABLE_AQUATIC_PLANT Lnet/minecraft/class_3614;
```
modmuss50: Again, not sure the resaoning behind this?

mappings/net/minecraft/block/Material.mapping
```diff
@@ -4,45 +4,45 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+	FIELD field_15916 AGGREGATE Lnet/minecraft/class_3614;
 	FIELD field_15917 SPONGE Lnet/minecraft/class_3614;
 	FIELD field_15918 REDSTONE_LAMP Lnet/minecraft/class_3614;
 	FIELD field_15919 PORTAL Lnet/minecraft/class_3614;
 	FIELD field_15920 WATER Lnet/minecraft/class_3614;
 	FIELD field_15921 UNUSED_PLANT Lnet/minecraft/class_3614;
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-	FIELD field_15924 PART Lnet/minecraft/class_3614;
+	FIELD field_15924 DECORATION Lnet/minecraft/class_3614;
```
modmuss50: No, this is used by more than just decroation blocks?

modmuss50 (changes requested):
Im really not sure about some of these changes.

mappings/net/minecraft/block/Material.mapping
Juuxel: I agree, especially since the block is `minecraft:snow`.

mappings/net/minecraft/block/Material.mapping
Juuxel: It's also used for blue ice, which is another type of dense ice.

mappings/net/minecraft/block/Material.mapping
Juuxel: Yeah, it's only used for both types of seagrass.

mappings/net/minecraft/block/Material.mapping
Juuxel: `GOURD` is fine. We also have a `GourdBlock` class.

mappings/net/minecraft/block/Material.mapping
```diff
@@ -4,45 +4,45 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15913 COBWEB Lnet/minecraft/class_3614;
 	FIELD field_15914 STONE Lnet/minecraft/class_3614;
 	FIELD field_15915 BUBBLE_COLUMN Lnet/minecraft/class_3614;
-	FIELD field_15916 SAND Lnet/minecraft/class_3614;
+	FIELD field_15916 AGGREGATE Lnet/minecraft/class_3614;
```
Juuxel, line 8: This sounds a bit weird. It's used for powder-like falling blocks (sands, soul sand, gravel and concrete powder), so maybe `POWDER`?

Juuxel (changes requested):

mappings/net/minecraft/block/Material.mapping
```diff
@@ -4,45 +4,45 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15931 WOOL Lnet/minecraft/class_3614;
 	FIELD field_15932 WOOD Lnet/minecraft/class_3614;
 	FIELD field_15933 PISTON Lnet/minecraft/class_3614;
 	FIELD field_15934 SNOW_BLOCK Lnet/minecraft/class_3614;
 	FIELD field_15935 PLANT Lnet/minecraft/class_3614;
-	FIELD field_15936 CLAY Lnet/minecraft/class_3614;
+	FIELD field_15936 MOB_DERIVED Lnet/minecraft/class_3614;
 	FIELD field_15937 CAKE Lnet/minecraft/class_3614;
-	FIELD field_15938 BAMBOO_SAPLING Lnet/minecraft/class_3614;
+	FIELD field_15938 BAMBOO_SHOOT Lnet/minecraft/class_3614;
```
Juuxel: The block is `minecraft:bamboo_sapling`. (see also https://github.com/FabricMC/yarn/pull/1189#issuecomment-599498412)

mappings/net/minecraft/block/Material.mapping
```diff
@@ -4,45 +4,45 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15939 blocksLight Z
 	FIELD field_15940 blocksMovement Z
-	FIELD field_15941 EARTH Lnet/minecraft/class_3614;
+	FIELD field_15941 TOPSOIL Lnet/minecraft/class_3614;
 	FIELD field_15942 GLASS Lnet/minecraft/class_3614;
 	FIELD field_15943 FIRE Lnet/minecraft/class_3614;
 	FIELD field_15944 color Lnet/minecraft/class_3620;
-	FIELD field_15945 ORGANIC Lnet/minecraft/class_3614;
+	FIELD field_15945 FODDER Lnet/minecraft/class_3614;
```
Juuxel:
```suggestion
	FIELD field_15945 ORGANIC Lnet/minecraft/class_3614;
```
I wouldn't call mycelium or dried kelp "fodder".

Juuxel (changes requested):

mappings/net/minecraft/block/Material.mapping
dexman545: To make it clearer between SNOW_BLOCK and SNOW

mappings/net/minecraft/block/Material.mapping
dexman545: Dirt is considered topsoil as well

mappings/net/minecraft/block/Material.mapping
dexman545: It's not, it's used by infested blocks, honey blocks, honeycomb blocks, and slime blocks. Clay is the odd man out

mappings/net/minecraft/block/Material.mapping
dexman545: And ANVIL used to only be used for anvils, and now the name is misleading. This material only differs from AQUATIC_PLANT by being replaceable, so changing this is to make it clearer to devs who want to get any of the plants that exist underwater.
This also brings it more inline with what PLANT was already doing in terms of naming

mappings/net/minecraft/block/Material.mapping
dexman545: I'm open to suggestions, but PART isn't very descriptive

mappings/net/minecraft/block/Material.mapping
dexman545: Dried kelp can be used for food, and I'm ignoring the odd man out of mycelium.
The goal is to make the name clearer as to what blocks it represents. Suggestions are welcome.

mappings/net/minecraft/block/Material.mapping
liach: what is the property of the material?

mappings/net/minecraft/block/Material.mapping
```diff
@@ -4,45 +4,45 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15946 BAMBOO Lnet/minecraft/class_3614;
-	FIELD field_15947 UNDERWATER_PLANT Lnet/minecraft/class_3614;
-	FIELD field_15948 SNOW Lnet/minecraft/class_3614;
-	FIELD field_15949 ANVIL Lnet/minecraft/class_3614;
+	FIELD field_15947 AQUATIC_PLANT Lnet/minecraft/class_3614;
+	FIELD field_15948 SNOW_LAYER Lnet/minecraft/class_3614;
+	FIELD field_15949 REPAIR_STATION Lnet/minecraft/class_3614;
```
liach, line 58: `FORGING_STATION` like

mappings/net/minecraft/block/Material.mapping
liach: or `FALLABLE`?

dexman545: I prefer `AGGREGATE` over `POWDER`. `FALLABLE` is not in keeping with the rest of the materials being some descriptor of what the blocks are/are made from

mappings/net/minecraft/block/Material.mapping
dexman545: Has the same properties as EARTH, SAND, NETHER_WOOD, and a few others, only differing in the color

liach: just say "xxx colored earth"

mappings/net/minecraft/block/Material.mapping
dexman545: This material does not include the smithing table, so FORGING_STATION seems odd

mappings/net/minecraft/block/Material.mapping
dexman545: `CLAY` can also be considered decayed organic material from animals, so it somewhat fits with this name.

mappings/net/minecraft/block/Material.mapping
liach: compared to clay, this one is like raw biomass/living biomass while clay/honey block etc are non-living bio-derived stuff

mappings/net/minecraft/block/Material.mapping
liach: good

mappings/net/minecraft/block/Material.mapping
dexman545: the fodder is that most of the blocks in it are a food source, or storage of one
I'm implying that mooshrooms eat mycelium with that name, but fodder encompasses most of the blocks in that material
since organic is far too broad

mappings/net/minecraft/block/Material.mapping
dexman545: I also want to clarify this since 1.17 (iirc the right version) will have "snowier snow," so I want to clarify this as much as possible so that it doesn't become out of date soon

mappings/net/minecraft/block/Material.mapping
dexman545: Used for skulls, torches, potted plants, ladders, redstone components, and rails.
Perhaps `SUBBLOCK`?

liach: like fluid loggable block? redstone components and rails are circuits

dexman545: None of the blocks in this material are fluid loggable (except maybe ladders).
If `SUBBLOCK` is already in use for fluid logging then that won't work for a material name in this case. `PART` seems like a multipart or part of something else, which these are not. Initially chose `DECORATION` as that's what's on the wiki (per the original issue), and `PART` is just too vague.

mappings/net/minecraft/block/Material.mapping
bluebear94: `SQUISHY`

mappings/net/minecraft/block/Material.mapping
Sollace: _Aggregate_ is vaguely familliar. Looking it up, I get the definition "a natural mixture of sand, gravel, and crushed stone" in the context of geology, though I agree it's a little too technical. If you use it, this will require a Javadoc.

Sollace: image
> https://en.wikipedia.org/wiki/Aggregate_(geology)

mappings/net/minecraft/block/Material.mapping
Sollace: `ORGANIC` -> `BIO` or `BIOMASS`

I'm not sure what fodder means, other than something you throw dock for pigs to eat.

mappings/net/minecraft/block/Material.mapping
Sollace: What about clay makes it `MOB_DERIVED`? What does that even mean?

dexman545: Realistically, clay is made of organic material, usually from animals (mobs).
Every other block with this material is from some action or produce of a mob, ie derived from a mob.

mappings/net/minecraft/block/Material.mapping
dexman545: Aggregate is a fairly common term. Regardless, it is more descriptive of these blocks than "sand" or "powder."

mappings/net/minecraft/block/Material.mapping
dexman545: Fodder is the storage of food for later use. This or a synonym would function well. Biomass is far too general for this group.
image

YanisBft:
I think adding some temporary docs with the old names would help mods to update

liach (changes requested):
Single line jd don't need `<p>` tags and don't need `DESCRIPTION:` prefix.

mappings/net/minecraft/block/Material.mapping
```diff
@@ -2,55 +2,99 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+		COMMENT Material for the water block.
 	FIELD field_15921 UNUSED_PLANT Lnet/minecraft/class_3614;
+		COMMENT Not in use, but has foliage color.
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
+		COMMENT Material for the lava block.
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-	FIELD field_15924 PART Lnet/minecraft/class_3614;
+		COMMENT Material for leaves.
+	FIELD field_15924 SUBBLOCK Lnet/minecraft/class_3614;
+		COMMENT Nonfull blocks such as redstone components and torches.
```
liach:
```suggestion
		COMMENT Material for non-full blocks such as redstone components and torches.
```

mappings/net/minecraft/block/Material.mapping
```diff
@@ -2,55 +2,99 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+		COMMENT Material for blocks that are a product of mob interactions such as honey or slime, or the infested blocks. Includes clay but not bone blocks.
 	FIELD field_15937 CAKE Lnet/minecraft/class_3614;
+		COMMENT Material for cake.
 	FIELD field_15938 BAMBOO_SAPLING Lnet/minecraft/class_3614;
+		COMMENT Material for bamboo saplings.
 	FIELD field_15939 blocksLight Z
 	FIELD field_15940 blocksMovement Z
-	FIELD field_15941 EARTH Lnet/minecraft/class_3614;
+	FIELD field_15941 TOPSOIL Lnet/minecraft/class_3614;
+		COMMENT The top layer of soil. Path, dirt, podzol, soul soil, farmland and similar.
```
liach:
```suggestion
		COMMENT Material for the top layer of soil. Path, dirt, podzol, soul soil, farmland and similar.
```

mappings/net/minecraft/block/Material.mapping
```diff
@@ -2,55 +2,99 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15942 GLASS Lnet/minecraft/class_3614;
+		COMMENT Material for glass and glass-like blocks (Includes sea lanterns and conduits).
```
liach:
```suggestion
		COMMENT Material for glass and glass-like blocks (includes sea lanterns and conduits).
```

mappings/net/minecraft/block/Material.mapping
```diff
@@ -2,55 +2,99 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
-	FIELD field_15949 ANVIL Lnet/minecraft/class_3614;
+		COMMENT Material for underwater plants without the replaceable property.
+	FIELD field_15948 SNOW_LAYER Lnet/minecraft/class_3614;
+		COMMENT Material for non-full blocks of snow. Has the replaceable property.
+	FIELD field_15949 REPAIR_STATION Lnet/minecraft/class_3614;
+		COMMENT Material for blocks that can repair tools, ie. grindstone and anvils.
 	FIELD field_15950 replaceable Z
 	FIELD field_15951 burnable Z
 	FIELD field_15952 BARRIER Lnet/minecraft/class_3614;
+		COMMENT
```
liach:
```suggestion
		COMMENT Material for barrier blocks.
```

mappings/net/minecraft/block/Material.mapping
```diff
@@ -2,55 +2,99 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+		COMMENT Material for non-full blocks of snow. Has the replaceable property.
+	FIELD field_15949 REPAIR_STATION Lnet/minecraft/class_3614;
+		COMMENT Material for blocks that can repair tools, ie. grindstone and anvils.
 	FIELD field_15950 replaceable Z
 	FIELD field_15951 burnable Z
 	FIELD field_15952 BARRIER Lnet/minecraft/class_3614;
+		COMMENT 
 	FIELD field_15953 METAL Lnet/minecraft/class_3614;
-	FIELD field_15954 PUMPKIN Lnet/minecraft/class_3614;
+		COMMENT Material for barrier blocks.
```
liach:
```suggestion
		COMMENT Material for metal blocks.
```

mappings/net/minecraft/block/Material.mapping
```diff
@@ -2,55 +2,99 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+	FIELD field_15954 GOURD Lnet/minecraft/class_3614;
+		COMMENT Material for gourds. Includes the carved pumpkin and jack o' lantern.
 	FIELD field_15955 TNT Lnet/minecraft/class_3614;
+		COMMENT Material for TNT
```
liach:
```suggestion
		COMMENT Material for TNT.
```

mappings/net/minecraft/block/Material.mapping
```diff
@@ -2,55 +2,99 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15956 REPLACEABLE_PLANT Lnet/minecraft/class_3614;
+		COMMENT Material for plant blocks that are replaceable.
 	FIELD field_15957 CARPET Lnet/minecraft/class_3614;
+		COMMENT Material for carpets.
 	FIELD field_15958 ICE Lnet/minecraft/class_3614;
+		COMMENT Material for ice that can melt.
 	FIELD field_15959 AIR Lnet/minecraft/class_3614;
+		COMMENT Material for air.
 	FIELD field_15960 pistonBehavior Lnet/minecraft/class_3619;
+		COMMENT
```
liach:
```suggestion
```

liach (approved):
Looks good aside from a few controversial field names.

mappings/net/minecraft/block/Material.mapping
Sollace: .>Pumpkin
.>material for -~~barrier~~- metal blocks.

Uuuh....

mappings/net/minecraft/block/Material.mapping
```diff
@@ -2,55 +2,98 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+	FIELD field_15916 AGGREGATE Lnet/minecraft/class_3614;
+		COMMENT A material or structure formed from a loosely compacted mass of fragments or particles.
 	FIELD field_15917 SPONGE Lnet/minecraft/class_3614;
+		COMMENT Material for sponges.
 	FIELD field_15918 REDSTONE_LAMP Lnet/minecraft/class_3614;
+		COMMENT Material for redstone lamps.
 	FIELD field_15919 PORTAL Lnet/minecraft/class_3614;
+		COMMENT Material for the various portal blocks.
 	FIELD field_15920 WATER Lnet/minecraft/class_3614;
+		COMMENT Material for the water block.
```
Sollace: Is this really necessary? It's not enough that the name of the field is `WATER`, we have to specify that it's for `water` too?

Sollace: Likewise applies to other jds in this PR.

mappings/net/minecraft/block/Material.mapping
```diff
@@ -2,55 +2,98 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+	FIELD field_15954 GOURD Lnet/minecraft/class_3614;
+		COMMENT Material for gourds. Includes the carved pumpkin and jack o' lantern.
 	FIELD field_15955 TNT Lnet/minecraft/class_3614;
+		COMMENT Material for TNT.
 	FIELD field_15956 REPLACEABLE_PLANT Lnet/minecraft/class_3614;
+		COMMENT Material for plant blocks that are replaceable.
 	FIELD field_15957 CARPET Lnet/minecraft/class_3614;
+		COMMENT Material for carpets.
 	FIELD field_15958 ICE Lnet/minecraft/class_3614;
+		COMMENT Material for ice that can melt.
```
Sollace: All ice can melt (irl). Rather specify on the other one that it's for ice that _can't_ melt, or specify here which one to use for ice that cannot melt instead.

dexman545: Not all ice in MC can melt. That is included in the JD because it isn't included in the name.

mappings/net/minecraft/block/Material.mapping
dexman545: Did it for consistency, and because blocks could be added to this material at any time, so would need the JD/rename at that point.

mappings/net/minecraft/block/Material.mapping
Juuxel: (The comment is for `METAL`)

liach: Yes, we should start using split than unified diffs. unified diffs are confusing!

liach: image
This is how it actually looks in split diff @sollace

mappings/net/minecraft/block/Material.mapping
kvverti: No, the names should reflect how they are prgramatically used. This material has nothing to do with entities

mappings/net/minecraft/block/Material.mapping
Sollace: All ice can melt (In Real Life) is what I mean. Meltable is the default state and implied by the name "ice".

Saying that this is ice that can melt is redundant. You should rather mention where the behaviour is a divergence from the expected behaviour.

Hence, don't say this is ice that can melt, say the other one is ice that cannot melt and link to it from this one.

mappings/net/minecraft/block/Material.mapping
Sollace: Doesn't address the issue. These comments are all redundant.

mappings/net/minecraft/block/Material.mapping
dexman545: Simply keep with how they are programmatically used is not in keeping with how Materials have ever been named, and to switch to such a system would only add confusion.

mappings/net/minecraft/block/Material.mapping
dexman545: yes, the name ice implies that it is meltable, but it also implies that it is all ice, which it is not.
A link to the unmeltable kind can be added, but keeping what is already there in the JD will enhance clarity

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,86 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+		COMMENT Material for the various portal blocks.
 	FIELD field_15920 WATER Lnet/minecraft/class_3614;
 	FIELD field_15921 UNUSED_PLANT Lnet/minecraft/class_3614;
+		COMMENT Not in use, but has foliage color.
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-	FIELD field_15924 PART Lnet/minecraft/class_3614;
+	FIELD field_15924 SUBBLOCK Lnet/minecraft/class_3614;
+		COMMENT Material for non-full blocks such as redstone components and torches.
+		COMMENT Previously called PART
```
liach:
```suggestion
```
people can just look up intermediary names.

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,86 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15925 CACTUS Lnet/minecraft/class_3614;
-	FIELD field_15926 SEAGRASS Lnet/minecraft/class_3614;
+	FIELD field_15926 REPLACEABLE_UNDERWATER_PLANT Lnet/minecraft/class_3614;
+		COMMENT Previously called SEAGRASS
```
liach:
```suggestion
```
People look up intermediary names.

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,86 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15912 breakByHand Z
 	FIELD field_15913 COBWEB Lnet/minecraft/class_3614;
 	FIELD field_15914 STONE Lnet/minecraft/class_3614;
+		COMMENT Hard solid nonmetallic mineral matter of which rock is made, especially as a building material.
 	FIELD field_15915 BUBBLE_COLUMN Lnet/minecraft/class_3614;
-	FIELD field_15916 SAND Lnet/minecraft/class_3614;
+	FIELD field_15916 AGGREGATE Lnet/minecraft/class_3614;
+		COMMENT A material or structure formed from a loosely compacted mass of fragments or particles.
+		COMMENT Previously called SAND
```
liach:
```suggestion
```

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,86 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15925 CACTUS Lnet/minecraft/class_3614;
-	FIELD field_15926 SEAGRASS Lnet/minecraft/class_3614;
+	FIELD field_15926 REPLACEABLE_UNDERWATER_PLANT Lnet/minecraft/class_3614;
+		COMMENT Previously called SEAGRASS
 	FIELD field_15927 STRUCTURE_VOID Lnet/minecraft/class_3614;
-	FIELD field_15928 PACKED_ICE Lnet/minecraft/class_3614;
+		COMMENT Material for structure void block.
+	FIELD field_15928 DENSE_ICE Lnet/minecraft/class_3614;
+		COMMENT Material for ice blocks that do not melt. {@see ICE} for meltable ice.
+		COMMENT Previously called PACKED_ICE
```
liach:
```suggestion
```

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,86 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+		COMMENT Material for wood logs, and things crafted from them. Includes the nether stems (nether wood logs).
 	FIELD field_15933 PISTON Lnet/minecraft/class_3614;
 	FIELD field_15934 SNOW_BLOCK Lnet/minecraft/class_3614;
+		COMMENT Material for full sized snow blocks.
 	FIELD field_15935 PLANT Lnet/minecraft/class_3614;
-	FIELD field_15936 CLAY Lnet/minecraft/class_3614;
+		COMMENT Material for plants such as flowers and crops
+	FIELD field_15936 MOB_DERIVED Lnet/minecraft/class_3614;
+		COMMENT Material for blocks that are a product of mob interactions such as honey or slime, or the infested blocks. Includes clay but not bone blocks.
+		COMMENT Previously called CLAY
```
liach:
```suggestion
```

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,86 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15937 CAKE Lnet/minecraft/class_3614;
 	FIELD field_15938 BAMBOO_SAPLING Lnet/minecraft/class_3614;
 	FIELD field_15939 blocksLight Z
 	FIELD field_15940 blocksMovement Z
-	FIELD field_15941 EARTH Lnet/minecraft/class_3614;
+	FIELD field_15941 TOPSOIL Lnet/minecraft/class_3614;
+		COMMENT Material for the top layer of soil. Path, dirt, podzol, soul soil, farmland and similar.
+		COMMENT Previously called EARTH
```
liach:
```suggestion
```

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,86 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15942 GLASS Lnet/minecraft/class_3614;
+		COMMENT Material for glass and glass-like blocks (includes sea lanterns and conduits).
 	FIELD field_15943 FIRE Lnet/minecraft/class_3614;
 	FIELD field_15944 color Lnet/minecraft/class_3620;
-	FIELD field_15945 ORGANIC Lnet/minecraft/class_3614;
+	FIELD field_15945 FODDER Lnet/minecraft/class_3614;
+		COMMENT Food, especially dried hay or feed, for cattle and other livestock. Includes mycelium.
+		COMMENT Previously called ORGANIC
```
liach:
```suggestion
```

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,86 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15946 BAMBOO Lnet/minecraft/class_3614;
 	FIELD field_15947 UNDERWATER_PLANT Lnet/minecraft/class_3614;
-	FIELD field_15948 SNOW Lnet/minecraft/class_3614;
-	FIELD field_15949 ANVIL Lnet/minecraft/class_3614;
+		COMMENT Material for underwater plants without the replaceable property.
+	FIELD field_15948 SNOW_LAYER Lnet/minecraft/class_3614;
+		COMMENT Material for non-full blocks of snow. Has the replaceable property.
+	FIELD field_15949 REPAIR_STATION Lnet/minecraft/class_3614;
+		COMMENT Material for blocks that can repair tools, ie. grindstone and anvils.
+		COMMENT Previously called ANVIL
```
liach:
```suggestion
```

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,86 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15950 replaceable Z
 	FIELD field_15951 burnable Z
 	FIELD field_15952 BARRIER Lnet/minecraft/class_3614;
 	FIELD field_15953 METAL Lnet/minecraft/class_3614;
-	FIELD field_15954 PUMPKIN Lnet/minecraft/class_3614;
+	FIELD field_15954 GOURD Lnet/minecraft/class_3614;
+		COMMENT Material for gourds. Includes the carved pumpkin and jack o' lantern.
+		COMMENT Previously called PUMPKIN
```
liach:
```suggestion
```

liach (changes requested):
Please remove these `Previously called` stuff from javadocs. These information are trivial to the material themselves. There is intermediary names which people can utilize to track the changes.

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,78 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+		COMMENT Not in use, but has foliage color.
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-	FIELD field_15924 PART Lnet/minecraft/class_3614;
+	FIELD field_15924 SUBBLOCK Lnet/minecraft/class_3614;
+		COMMENT Material for non-full blocks such as redstone components and torches.
 	FIELD field_15925 CACTUS Lnet/minecraft/class_3614;
-	FIELD field_15926 SEAGRASS Lnet/minecraft/class_3614;
+	FIELD field_15926 REPLACEABLE_UNDERWATER_PLANT Lnet/minecraft/class_3614;
+		COMMENT Previously called SEAGRASS
```
liach: Please remove this piece of useless javadocs
```suggestion
```

liach (changes requested):

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,86 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+		COMMENT Material for wool and bed blocks.
 	FIELD field_15932 WOOD Lnet/minecraft/class_3614;
+		COMMENT Material for wood logs, and things crafted from them. Includes the nether stems (nether wood logs).
 	FIELD field_15933 PISTON Lnet/minecraft/class_3614;
 	FIELD field_15934 SNOW_BLOCK Lnet/minecraft/class_3614;
+		COMMENT Material for full sized snow blocks.
 	FIELD field_15935 PLANT Lnet/minecraft/class_3614;
-	FIELD field_15936 CLAY Lnet/minecraft/class_3614;
+		COMMENT Material for plants such as flowers and crops
+	FIELD field_15936 MOB_DERIVED Lnet/minecraft/class_3614;
```
Sollace: @dexman545 
> Realistically, clay is made of organic material

No. No it is not.
image

mappings/net/minecraft/block/Material.mapping
Sollace: Dirt is not _exclusively_ topsoil.

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,77 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15953 METAL Lnet/minecraft/class_3614;
-	FIELD field_15954 PUMPKIN Lnet/minecraft/class_3614;
+	FIELD field_15954 GOURD Lnet/minecraft/class_3614;
+		COMMENT Material for gourds. Includes the carved pumpkin and jack o' lantern.
 	FIELD field_15955 TNT Lnet/minecraft/class_3614;
 	FIELD field_15956 REPLACEABLE_PLANT Lnet/minecraft/class_3614;
+		COMMENT Material for plant blocks that are replaceable.
 	FIELD field_15957 CARPET Lnet/minecraft/class_3614;
 	FIELD field_15958 ICE Lnet/minecraft/class_3614;
+		COMMENT Material for ice that can melt. {@see DENSE_ICE} for unmeltable ice.
```
liach:
```suggestion
		COMMENT Material for ice that can melt. See {@link #DENSE_ICE} for unmeltable ice.
```

mappings/net/minecraft/block/Material.mapping
Sollace: You already mention anvils in the list of examples. You don't need this.

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,86 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15951 burnable Z
 	FIELD field_15952 BARRIER Lnet/minecraft/class_3614;
 	FIELD field_15953 METAL Lnet/minecraft/class_3614;
-	FIELD field_15954 PUMPKIN Lnet/minecraft/class_3614;
+	FIELD field_15954 GOURD Lnet/minecraft/class_3614;
+		COMMENT Material for gourds. Includes the carved pumpkin and jack o' lantern.
+		COMMENT Previously called PUMPKIN
 	FIELD field_15955 TNT Lnet/minecraft/class_3614;
 	FIELD field_15956 REPLACEABLE_PLANT Lnet/minecraft/class_3614;
+		COMMENT Material for plant blocks that are replaceable.
```
Sollace: Redundant. The name already says everything you need to know.

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,86 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15918 REDSTONE_LAMP Lnet/minecraft/class_3614;
 	FIELD field_15919 PORTAL Lnet/minecraft/class_3614;
+		COMMENT Material for the various portal blocks.
 	FIELD field_15920 WATER Lnet/minecraft/class_3614;
 	FIELD field_15921 UNUSED_PLANT Lnet/minecraft/class_3614;
+		COMMENT Not in use, but has foliage color.
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-	FIELD field_15924 PART Lnet/minecraft/class_3614;
+	FIELD field_15924 SUBBLOCK Lnet/minecraft/class_3614;
```
Sollace: Why does this need to change? `PART` was fine, and if you're going to change it at least make it `SUB_BLOCK`.

mappings/net/minecraft/block/Material.mapping
Sollace: You don't need this.

mappings/net/minecraft/block/Material.mapping
Sollace:
```suggestion
```

mappings/net/minecraft/block/Material.mapping
Sollace:
```suggestion
```

mappings/net/minecraft/block/Material.mapping
Sollace:
```suggestion
```

mappings/net/minecraft/block/Material.mapping
Sollace:
```suggestion
```

mappings/net/minecraft/block/Material.mapping
Sollace:
```suggestion
```

mappings/net/minecraft/block/Material.mapping
Sollace:
```suggestion
```

mappings/net/minecraft/block/Material.mapping
Sollace:
```suggestion
```

mappings/net/minecraft/block/Material.mapping
Sollace:
```suggestion
```

mappings/net/minecraft/block/Material.mapping
Sollace:
```suggestion
```

mappings/net/minecraft/block/Material.mapping
dexman545: yes, but everything with this material can be described as topsoil. The purpose of this name is to narrow down the overgeneralized name of "earth" to make it more descriptive. "Earth" is far too broad, so if you have an alternative name I'm open to suggestions.

mappings/net/minecraft/block/Material.mapping
Prospector: perhaps `PARTICLES`?

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15915 BUBBLE_COLUMN Lnet/minecraft/class_3614;
-	FIELD field_15916 SAND Lnet/minecraft/class_3614;
+	FIELD field_15916 AGGREGATE Lnet/minecraft/class_3614;
+		COMMENT A material or structure formed from a loosely compacted mass of fragments or particles.
 	FIELD field_15917 SPONGE Lnet/minecraft/class_3614;
 	FIELD field_15918 REDSTONE_LAMP Lnet/minecraft/class_3614;
 	FIELD field_15919 PORTAL Lnet/minecraft/class_3614;
+		COMMENT Material for the various portal blocks.
 	FIELD field_15920 WATER Lnet/minecraft/class_3614;
 	FIELD field_15921 UNUSED_PLANT Lnet/minecraft/class_3614;
```
Prospector, line 15: A more descriptive name for UNUSED_PLANT would be nice for people wanting to use it

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+		COMMENT Not in use, but has foliage color.
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-	FIELD field_15924 PART Lnet/minecraft/class_3614;
+	FIELD field_15924 SUB_BLOCK Lnet/minecraft/class_3614;
```
Prospector: Sub block seems like a bit of an odd name to me because they're still just normal blocks. I think PART is just fine personally.

mappings/net/minecraft/block/Material.mapping
Prospector: This one definitely doesn't work since clay isn't "MOB_DERIVED" and even for the other blocks that are, it's a *very* strange name. `SQUISHY` is pretty good, as suggested above.

mappings/net/minecraft/block/Material.mapping
Prospector: Not really an improvement

mappings/net/minecraft/block/Material.mapping
Prospector: Not an improvement because it loses accuracy for certain of the blocks with the material

mappings/net/minecraft/block/Material.mapping
dexman545: "Earth" is too broad of a name. "Topsoil" is closer to what the blocks are.

mappings/net/minecraft/block/Material.mapping
dexman545: "Particles" is used elsewhere by the game, and you can't really call gravel particles. "Aggregate" is more descriptive.

mappings/net/minecraft/block/Material.mapping
dexman545: "Part" to me seems like a part of something, and since MC doesn't have multiparts or anything really that is assembled from "parts" it seems odd to me.

mappings/net/minecraft/block/Material.mapping
dexman545: I think javadoc would be better in this case, as something like "plant_not_used_by_vanilla" is a bit awkward.

mappings/net/minecraft/block/Material.mapping
dexman545: Another suggestion was "moldable." Will hold off editing for more feedback.

mappings/net/minecraft/block/Material.mapping
dexman545: Javadoc for it was added, if a rename is desired it can be done

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
-	FIELD field_15941 EARTH Lnet/minecraft/class_3614;
+	FIELD field_15941 TOPSOIL Lnet/minecraft/class_3614;
+		COMMENT Material for the top layer of soil. Path, dirt, podzol, soul soil, farmland and similar.
 	FIELD field_15942 GLASS Lnet/minecraft/class_3614;
+		COMMENT Material for glass and glass-like blocks (includes sea lanterns and conduits).
 	FIELD field_15943 FIRE Lnet/minecraft/class_3614;
 	FIELD field_15944 color Lnet/minecraft/class_3620;
-	FIELD field_15945 ORGANIC Lnet/minecraft/class_3614;
+	FIELD field_15945 SOLID_ORGANIC Lnet/minecraft/class_3614;
+		COMMENT Organic blocks that are solid, eg. hay, target, or grass.
```
YanisBft: grass or grass blocks?

dexman545: Updated.

mappings/net/minecraft/block/Material.mapping
YanisBft: `PARTIAL_BLOCK`?

dexman545: How is "partial" any different from "sub"?

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15955 TNT Lnet/minecraft/class_3614;
 	FIELD field_15956 REPLACEABLE_PLANT Lnet/minecraft/class_3614;
 	FIELD field_15957 CARPET Lnet/minecraft/class_3614;
 	FIELD field_15958 ICE Lnet/minecraft/class_3614;
+		COMMENT Material for ice that can melt. See {@link #DENSE_ICE} for unmeltable ice.
 	FIELD field_15959 AIR Lnet/minecraft/class_3614;
 	FIELD field_15960 pistonBehavior Lnet/minecraft/class_3619;
 	FIELD field_17008 SHULKER_BOX Lnet/minecraft/class_3614;
 	FIELD field_22223 NETHER_WOOD Lnet/minecraft/class_3614;
+		COMMENT Material for blocks crafted from stems (Nether wood logs).
```
Juuxel: The nether fungi aren't wood.

Juuxel (changes requested):

mappings/net/minecraft/block/Material.mapping
dexman545: Unfortunately, mojang considers them wood.
image

Juuxel: Well, they're wood-like but not "nether wood logs".

dexman545: The tags call them wood, they belong to the logs category, they can be used to craft wood items such as chests and sticks - they are functionally wood.

liach: Just "nether logs" should suffice; logs are type of wood

dexman545: logs will not, as the planks and other nether wood blocks use this material.
image

liach:
```suggestion
		COMMENT Material for blocks crafted from stems (Nether logs).
```
what's wrong with this description

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 	FIELD field_15912 breakByHand Z
 	FIELD field_15913 COBWEB Lnet/minecraft/class_3614;
 	FIELD field_15914 STONE Lnet/minecraft/class_3614;
+		COMMENT Hard solid nonmetallic mineral matter of which rock is made, especially as a building material.
```
Juuxel:
```suggestion
```
I don't think we need to describe the chemical composition of stone in Yarn javadocs (especially here).

liach: Rather than mentioning chemical composition, we may consider mentioning their breaking property (i.e. blocks that general can only be harvested with pickaxes and be broken faster with pickaxes)

mappings/net/minecraft/block/Material.mapping
liach: sub usually indicates an inclusion, so people would assume it to belong to a bigger block or a multiblock. partial doesn't have this ambiguity

mappings/net/minecraft/block/Material.mapping
liach: Mention *Minecraft* features of these blocks, such as explosion resistance, color, breaking information, harvest information, etc.

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+		COMMENT Material for structure void block.
+	FIELD field_15928 DENSE_ICE Lnet/minecraft/class_3614;
+		COMMENT Material for ice blocks that do not melt. {@see ICE} for meltable ice.
 	FIELD field_15929 liquid Z
 	FIELD field_15930 EGG Lnet/minecraft/class_3614;
+		COMMENT Material for egg blocks, such as dragon and turtle eggs.
 	FIELD field_15931 WOOL Lnet/minecraft/class_3614;
+		COMMENT Material for wool and bed blocks.
 	FIELD field_15932 WOOD Lnet/minecraft/class_3614;
+		COMMENT Material for wood logs, and things crafted from them. Includes the nether stems (nether wood logs).
```
liach: Really? Isn't nether stem stuff under `NETHER_WOOD`?

mappings/net/minecraft/block/Material.mapping
dexman545: CHanged to PARTIAL

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+	FIELD field_15924 PARTIAL_BLOCK Lnet/minecraft/class_3614;
+		COMMENT Material for non-full blocks such as redstone components and torches.
 	FIELD field_15925 CACTUS Lnet/minecraft/class_3614;
-	FIELD field_15926 SEAGRASS Lnet/minecraft/class_3614;
+	FIELD field_15926 REPLACEABLE_UNDERWATER_PLANT Lnet/minecraft/class_3614;
 	FIELD field_15927 STRUCTURE_VOID Lnet/minecraft/class_3614;
-	FIELD field_15928 PACKED_ICE Lnet/minecraft/class_3614;
+		COMMENT Material for structure void block.
+	FIELD field_15928 DENSE_ICE Lnet/minecraft/class_3614;
+		COMMENT Material for ice blocks that do not melt. {@see ICE} for meltable ice.
```
liach:
```suggestion
		COMMENT Material for ice blocks that do not melt. See {@link #ICE} for meltable ice.
```

liach: See https://travis-ci.com/github/FabricMC/yarn/jobs/320793104#L366

mappings/net/minecraft/block/Material.mapping
dexman545: FODDER -> SOLID_ORGANIC to better encompass all the blocks that use it, and to narrow down the blocks it applies to

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15931 WOOL Lnet/minecraft/class_3614;
+		COMMENT Material for wool and bed blocks.
 	FIELD field_15932 WOOD Lnet/minecraft/class_3614;
+		COMMENT Material for wood logs, and things crafted from them.
 	FIELD field_15933 PISTON Lnet/minecraft/class_3614;
 	FIELD field_15934 SNOW_BLOCK Lnet/minecraft/class_3614;
+		COMMENT Material for full sized snow blocks.
 	FIELD field_15935 PLANT Lnet/minecraft/class_3614;
-	FIELD field_15936 CLAY Lnet/minecraft/class_3614;
+		COMMENT Material for plants such as flowers and crops
```
liach, line 38:
```suggestion
		COMMENT Material for plants such as flowers and crops.
```

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
-	FIELD field_15941 EARTH Lnet/minecraft/class_3614;
+	FIELD field_15941 TOPSOIL Lnet/minecraft/class_3614;
+		COMMENT Material for the top layer of soil. Path, dirt, podzol, soul soil, farmland and similar.
 	FIELD field_15942 GLASS Lnet/minecraft/class_3614;
+		COMMENT Material for glass and glass-like blocks (includes sea lanterns and conduits).
 	FIELD field_15943 FIRE Lnet/minecraft/class_3614;
 	FIELD field_15944 color Lnet/minecraft/class_3620;
-	FIELD field_15945 ORGANIC Lnet/minecraft/class_3614;
+	FIELD field_15945 SOLID_ORGANIC Lnet/minecraft/class_3614;
+		COMMENT Organic blocks that are solid, eg. hay, target, or grass blocks.
```
liach:
```suggestion
		COMMENT Organic blocks that are solid, including hay, target, and grass blocks.
```
Or javadoc summary would be `Organic blocks that are solid, eg.`

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15946 BAMBOO Lnet/minecraft/class_3614;
 	FIELD field_15947 UNDERWATER_PLANT Lnet/minecraft/class_3614;
-	FIELD field_15948 SNOW Lnet/minecraft/class_3614;
-	FIELD field_15949 ANVIL Lnet/minecraft/class_3614;
+		COMMENT Material for underwater plants without the replaceable property.
+	FIELD field_15948 SNOW_LAYER Lnet/minecraft/class_3614;
+		COMMENT Material for non-full blocks of snow. Has the replaceable property.
+	FIELD field_15949 REPAIR_STATION Lnet/minecraft/class_3614;
+		COMMENT Material for blocks that can repair tools, ie. grindstone and anvils.
```
liach:
```suggestion
		COMMENT Material for blocks that can repair tools, including grindstone and anvils.
```

liach (approved):

liach (approved):

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15919 PORTAL Lnet/minecraft/class_3614;
+		COMMENT Material for the various portal blocks.
 	FIELD field_15920 WATER Lnet/minecraft/class_3614;
 	FIELD field_15921 UNUSED_PLANT Lnet/minecraft/class_3614;
+		COMMENT Not in use, but has foliage color.
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-	FIELD field_15924 PART Lnet/minecraft/class_3614;
+	FIELD field_15924 PARTIAL_BLOCK Lnet/minecraft/class_3614;
+		COMMENT Material for non-full blocks such as redstone components and torches.
```
YanisBft:
```suggestion
		COMMENT Material for non-full blocks such as redstone components, torches and skulls.
```

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15955 TNT Lnet/minecraft/class_3614;
 	FIELD field_15956 REPLACEABLE_PLANT Lnet/minecraft/class_3614;
 	FIELD field_15957 CARPET Lnet/minecraft/class_3614;
 	FIELD field_15958 ICE Lnet/minecraft/class_3614;
+		COMMENT Material for ice that can melt. See {@link #DENSE_ICE} for unmeltable ice.
 	FIELD field_15959 AIR Lnet/minecraft/class_3614;
 	FIELD field_15960 pistonBehavior Lnet/minecraft/class_3619;
 	FIELD field_17008 SHULKER_BOX Lnet/minecraft/class_3614;
 	FIELD field_22223 NETHER_WOOD Lnet/minecraft/class_3614;
+		COMMENT Material for blocks crafted from stems (Nether logs).
```
YanisBft:
```suggestion
		COMMENT Material for blocks crafted from Nether stems and hyphae.
```

YanisBft (commented):
This PR is more than fine now imo, except for the CLAY -> MOB_DERIVED change. Both are wrong, but can't think of a right name.

liach:
the clay sounds like organic product/organic sediment

YanisBft (approved):
I don't think we will get a better name for mob_derived, take my approval

mappings/net/minecraft/block/Material.mapping
immibis: Does it make sense to call *anything* "moldable" when molding isn't a thing you can do in minecraft?

mappings/net/minecraft/block/Material.mapping
Sollace: Shouldn't "topsoil" be two words? `TOP_SOIL`

mappings/net/minecraft/block/Material.mapping
Sollace: We don't yet know what Mojang is going to do. I'd rather change it later than change it now and later find out the change was in the wrong direction.

mappings/net/minecraft/block/Material.mapping
dexman545: Topsoil is one word.

mappings/net/minecraft/block/Material.mapping
dexman545: As it stands now, this name still clarifies the blocks that use this material.

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
+		COMMENT Material for wool and bed blocks.
 	FIELD field_15932 WOOD Lnet/minecraft/class_3614;
+		COMMENT Material for wood logs, and things crafted from them.
 	FIELD field_15933 PISTON Lnet/minecraft/class_3614;
 	FIELD field_15934 SNOW_BLOCK Lnet/minecraft/class_3614;
+		COMMENT Material for full sized snow blocks.
 	FIELD field_15935 PLANT Lnet/minecraft/class_3614;
-	FIELD field_15936 CLAY Lnet/minecraft/class_3614;
+		COMMENT Material for plants such as flowers and crops.
+	FIELD field_15936 MOB_DERIVED Lnet/minecraft/class_3614;
```
Juuxel: `MOB_DERIVED` is not a good name. I like `SQUISHY`, and `ORGANIC_PRODUCT` could also work (inspired by @liach's comment: https://github.com/FabricMC/yarn/pull/1190#issuecomment-615967554).

Juuxel (changes requested):

mappings/net/minecraft/block/Material.mapping
LoganDark: I like `SQUISHY`.

NeusFear: Could just be `ORGANIC` in general

dexman545: `ORGANIC` itself will not work, as that is too broad. If `ORGANGIC_PRODUCT` is what is wanted, sure.

Juuxel (approved):
Not the biggest fan of `AGGREGATE`, but it works.

dblsaiko (approved):

liach:
On a side note, this fixes #1191. I personally don't see much problem with this pr, guess this's good to go?

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15918 REDSTONE_LAMP Lnet/minecraft/class_3614;
 	FIELD field_15919 PORTAL Lnet/minecraft/class_3614;
+		COMMENT Material for the various portal blocks.
 	FIELD field_15920 WATER Lnet/minecraft/class_3614;
 	FIELD field_15921 UNUSED_PLANT Lnet/minecraft/class_3614;
+		COMMENT Not in use, but has foliage color.
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-	FIELD field_15924 PART Lnet/minecraft/class_3614;
+	FIELD field_15924 PARTIAL_BLOCK Lnet/minecraft/class_3614;
```
Earthcomputer: Not sure this one is descriptive. I'd consider a slab a "partial block"

Earthcomputer (changes requested):

mappings/net/minecraft/block/Material.mapping
dblsaiko: What's this material for?

dexman545: well, would be more common to assume slabs share material with their full block counterparts.

This material is for things like redstone components, rails, torches, ladders, etc.

Earthcomputer: A material does not describe a block shape. It describes a subset of its properties. PARTIAL_BLOCK is not a good name because it describes its shape rather than its properties.

dexman545: That is not how any material is named, except in cases of clarifying it from another similar property.

dblsaiko: What about `MOUNTED`? `SUPPORTED`?

dexman545: `SUPPORTED` would be fine, if we're ok ignoring flower pots and the potted plants.

Prospector: That's a bug, so SUPPORTED seems good to me https://bugs.mojang.com/browse/MC-127036

dexman545: Not confirmed as a bug, but we'll go with it then.

YanisBft: Skulls tho?

liach: i think skull needs a block to attach to as well, so `SUPPORTED` works very well

liach (approved):

liach:
Think this is good, 4 approvals should suffice

Earthcomputer (approved):

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15918 REDSTONE_LAMP Lnet/minecraft/class_3614;
 	FIELD field_15919 PORTAL Lnet/minecraft/class_3614;
+		COMMENT Material for the various portal blocks.
 	FIELD field_15920 WATER Lnet/minecraft/class_3614;
 	FIELD field_15921 UNUSED_PLANT Lnet/minecraft/class_3614;
+		COMMENT Not in use, but has foliage color.
 	FIELD field_15922 LAVA Lnet/minecraft/class_3614;
 	FIELD field_15923 LEAVES Lnet/minecraft/class_3614;
-	FIELD field_15924 PART Lnet/minecraft/class_3614;
+	FIELD field_15924 SUPPORTED Lnet/minecraft/class_3614;
```
Prospector, line 19: Since this is used for skulls, supported doesn't work. I'd go with PART or COMPONENT.

mappings/net/minecraft/block/Material.mapping
Prospector: Still don't think this is an improvement.

Prospector (changes requested):

mappings/net/minecraft/block/Material.mapping
Earthcomputer: @Prospector https://github.com/FabricMC/yarn/pull/1190#discussion_r415417695

Earthcomputer (changes requested):
My specific problem with this PR is it uses too many overly technical terms that many people won't be familiar with, making it hard to actually choose which material to apply to their block:
- Aggregate: I don't know where this is a fairly common term, but I've only ever heard of it when doing a geology AS level and I don't think many people have done that.
- Topsoil: it's obvious this would apply to grass even if you didn't know that this was a well-defined term (again, I've only ever heard this mentioned in geology and higher-level geography). People may be confused as to whether dirt should be included.
- Gourd: never heard this word before in my life.

Also remember that what a Material actually is, what's important is not where the block generates but what properties it has (and sometimes what group it falls in).

Juuxel:
> Gourd: never heard this word before in my life.

Gourd is already used in Yarn for those exact blocks so I think that's fine.

mappings/net/minecraft/block/Material.mapping
Prospector: That was before I realized skulls were part of it, which aren't "supported." Also, support isn't a property of materials so tbh thinking about it more it really doesn't make sense.

Prospector:
I agree with Earthcomputer (except for gourd, that's a very common word especially around Halloween time), some of these are just not things you would would ever intuitively think are materials without looking through them and googling a few. I think in most cases, an example is far more valuable. If I'm adding a block that's similar in material to clay, then I'd probably give it the clay material instead of making my own, which seems to be how mojang operates as well.

mappings/net/minecraft/block/Material.mapping
dexman545: Changed to SOIL as per feedback

Earthcomputer:
Apparently I'm alone on gourd (my parents never let me celebrate halloween that's my excuse), topsoil has been fixed and for the other similar ones, I don't like them but can't think of a better name. Good names should come first and foremost, but in absence of a better name good javadocs suffice. When picking a material, the javadocs are easy to see in the Material class.

mappings/net/minecraft/block/Material.mapping
```diff
@@ -3,54 +3,76 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
...
 	FIELD field_15932 WOOD Lnet/minecraft/class_3614;
+		COMMENT Material for wood logs, and things crafted from them.
 	FIELD field_15933 PISTON Lnet/minecraft/class_3614;
 	FIELD field_15934 SNOW_BLOCK Lnet/minecraft/class_3614;
+		COMMENT Material for full sized snow blocks.
 	FIELD field_15935 PLANT Lnet/minecraft/class_3614;
-	FIELD field_15936 CLAY Lnet/minecraft/class_3614;
+		COMMENT Material for plants such as flowers and crops
+	FIELD field_15936 ORGANIC_PRODUCT Lnet/minecraft/class_3614;
+		COMMENT Material for blocks that are a product of mob interactions such as honey or slime, or the infested blocks. Includes clay but not bone blocks.
```
Earthcomputer: I think the wording of the javadoc could be simplified a little here: "blocks that come from mobs"?

dexman545: Done

Prospector (approved):
I still don't think supported is great but I guess it's good enough. If anyone thinks up a better name we can change it later.

Earthcomputer (approved):
6 approvals now, I don't see anyone with major objections to anything anymore. If we come up with better names for stuff we can always change them later.

liach:
This pr is up to date as of 20w20b. @modmuss50 mind check it out again?

modmuss50 (approved):

liach:
:tada: congrats

PR #1154, opened 2020-03-06
Remap inventories :D

Sollace:
A bit of a refactor to clean up the inventory method name.

- Removed the instances of "inv" all over the inventory classes
- Added javadocs where applicable
- Corrected some inaccurate names/misleading names

PR #1197, opened 2020-03-17
Inventory -> ItemContainer

### LINKED ISSUE #1107: Inventory -> ItemContainer
**This issue is assuming that #1106 is merged. Discuss in this issue as if it has been.**

The word "inventory" in the game, both internally and externally, is used to refer to the player's inventory only. Many strings in the game refer to this `Inventory` class as a container, and many places in yarn already call this a container, in fact (see `LootableContainerBlockEntity`, `LockableContainerBlockEntity`, etc.).

I am proposing adding an Item prefix to it, for multiple reasons. Firstly, `Container` itself is too vague to be used alone for any class name. There are numerous classes already suffixed with `Container` that bear no relation to this class. Adding an Item prefix to it clarifies that it is something that contains items. Secondly, prefixing with `Item` leaves ample room for modders to make `FluidContainer`s or `EnergyContainer`s or whatever and not having confusing terminology with the item one being simply, `Container`.

Juuxel:
Renames `Inventory` to `ItemContainer`.

"Inventory" in vanilla only refers to entity inventories, such as players or traders. Blocks are *item containers* instead, and the term also includes inventories. The name "(item) container" seems to also be used by Mojang internally, for example with `<C extends ItemContainer>` in recipe generics.

I've left the entity inventory classes suffixed with `Inventory` as they are inventories. `ItemContainer` fields are either `container` or `items`, with `container` mostly being used for references to container blocks (such as chests). Internal item lists of container block entities are called `items`.

Closes #1107, and also includes #1154 by @Sollace (partially modified to fit the new name).


modmuss50 (changes requested):
Im not a big fan of changing this at all. I think inventory is perfectally fine, the google definition almost perfectally reflects it.

jaskarth:
:concern:

Ueaj-Kerman:
Container not ItemContainer :irritatered:

liach:
Item container doesn't specify how many items it can contain. What if it can only contain one item? Inventory, however, indicates that it has the potential to contain multiple items.

Prospector:
What? This class can hold as many or as few items as it wants. Container doesn't specify anything about count, and neither does this class.

Prospector (approved):
Agree with premise, haven't looked at specifics of the changes yet though.

Prospector:
Would one of the five people who :-1: this PR please explain, without using an argument involving "change is bad," what is wrong with this rename?

liach:
> Would one of the five people who 👎 this PR please explain, without using an argument involving "change is bad," what is wrong with this rename?

We have `VillagerDataContainer` which only contains one `VillagerData`, like
```java
public interface VillagerDataContainer {
  VillagerData getVillagerData();
}
```
I would expect `ItemContainer` to be like
```java
public interface ItemContainer {
  Item getItem();
}
```
than an inventory of item stacks that supports some primitive operations.

Also imo #1154 itself is a poor pr that potentially introduce name conflicts. Shouldn't include that

mappings/net/minecraft/server/command/LootCommand.mapping
```diff
@@ -49,7 +49,7 @@ CLASS net/minecraft/class_3039 net/minecraft/server/command/LootCommand
 	METHOD method_13206 addTargetArguments (Lcom/mojang/brigadier/builder/ArgumentBuilder;Lnet/minecraft/class_3039$class_3042;)Lcom/mojang/brigadier/builder/ArgumentBuilder;
 		ARG 0 rootArgument
 		ARG 1 sourceConstructor
-	METHOD method_13207 getBlockInventory (Lnet/minecraft/class_2168;Lnet/minecraft/class_2338;)Lnet/minecraft/class_1263;
+	METHOD method_13207 getBlockItemContainer (Lnet/minecraft/class_2168;Lnet/minecraft/class_2338;)Lnet/minecraft/class_1263;
```
liach, line 52: Really `getBlockInventory` as this item container is within the block

mappings/net/minecraft/screen/LoomScreenHandler.mapping
```diff
@@ -1,7 +1,7 @@
 CLASS net/minecraft/class_1726 net/minecraft/screen/LoomScreenHandler
 	FIELD field_17316 context Lnet/minecraft/class_3914;
 	FIELD field_17317 selectedPattern Lnet/minecraft/class_3915;
-	FIELD field_17318 inventoryChangeListener Ljava/lang/Runnable;
+	FIELD field_17318 itemContainerChangeListener Ljava/lang/Runnable;
```
liach: `itemsChangeListener`

mappings/net/minecraft/container/CraftingResultItemContainer.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS net/minecraft/class_1731 net/minecraft/container/CraftingResultItemContainer
```
liach, line 1: Just `CraftingResultContainer` as crafting result is already an item.

Prospector:
The word container itself has nothing to do with only holding one thing. It just contains things, in this case, items. Just because there is one class that happens to contain one thing doesn't mean the word container suddenly is locked into this strange definition. Besides, what does VillagerData have? I would assume multiple pieces of data, otherwise that would be a strange class to exist at all. Just like how an item container would contain things in a single list or something.

mappings/net/minecraft/block/entity/ShulkerBoxBlockEntity.mapping
```diff
@@ -16,10 +16,10 @@ CLASS net/minecraft/class_2627 net/minecraft/block/entity/ShulkerBoxBlockEntity
 	METHOD method_11315 getCollisionBox (Lnet/minecraft/class_2350;)Lnet/minecraft/class_238;
 		ARG 1 facing
 	METHOD method_11316 pushEntities ()V
-	METHOD method_11317 serializeInventory (Lnet/minecraft/class_2487;)Lnet/minecraft/class_2487;
+	METHOD method_11317 serializeItems (Lnet/minecraft/class_2487;)Lnet/minecraft/class_2487;
```
liach: itemsToTag

mappings/net/minecraft/block/entity/ShulkerBoxBlockEntity.mapping
```diff
@@ -16,10 +16,10 @@ CLASS net/minecraft/class_2627 net/minecraft/block/entity/ShulkerBoxBlockEntity
...
 		ARG 1 tag
 	METHOD method_11318 updateAnimation ()V
-	METHOD method_11319 deserializeInventory (Lnet/minecraft/class_2487;)V
+	METHOD method_11319 deserializeItems (Lnet/minecraft/class_2487;)V
```
liach: itemsFromTag

liach (changes requested):
A few places to optimize if you do change `Inventory` to `ItemContainer`.

Runemoro:
> Would one of the five people who 👎 this PR please explain, without using an argument involving "change is bad," what is wrong with this rename?

It should just be `Container`. `ItemContainer` is long, doesn't match the Mojang name, and isn't clearer than `Container` in the context it's used.

Having `ItemContainer` would be like renaming `Block` to `WorldBlock` to avoid people thinking it's used as part of a Java AST for the block statement.

liach:
> Having `ItemContainer` would be like renaming `Block` to `WorldBlock` to avoid people thinking it's used as part of a Java AST for the block statement.

magnificent argument

mappings/net/minecraft/server/command/LootCommand.mapping
Juuxel: This PR uses "inventory" for entities, not blocks (like vanilla).

Sollace:
I'm against this change. It seems to be in poor taste. Also isn't the whole reason we removed `Container` because it was a name we appropriated from MCP?

Not to mention the confusion when this hits the masses, and all of a sudden you have `Container` -> `ScreenHandler` and `Inventory` -> `Container` / `ItemContainer` at the same time.

liach:
`container` isn't property of other mapping tools, but it is present in the mojang's resource ids for client textures. yet I still oppose this change.

Runemoro:
@Sollace No, the reason is that it's an incorrect name. And you just told everyone who didn't know the the MCP name what it was.

`Container` accurately describes what `Inventory` is, and it's a Mojang name (this was known before they released mappings, see https://github.com/FabricMC/yarn/issues/649)

TheGlitch76:
Can we just casually throw around MCP names now?

Prospector:
No, but it is a fair point that the only reason Container and Inventory were named as they were originally in Yarn was due to MCP. If it wasn't for mcp, following the code would logically lead you to the names Menu and Container.

Juuxel:
I'm closing this because I don't really have time for maintaining this, and `Inventory` is less of a problem than the cursed hand names for menus.

PR #1200, opened 2020-03-18
Quick entity fixes

### LINKED ISSUE #1145: Methods conflicts when classes are recompiled
These conflicts only happen when these classes are decompiled and then recompiled, after removing decompilation artifacts

* `void Entity.teleport(double, double, double)` (`method_20620`) conflicts with `boolean EndermanEntity.teleport(double, double, double)` (`method_7024`)
* `Properties AbstractPropertiesHandler.load(Path)` (`method_16727`) conflicts with `ServerPropertiesHandler ServerPropertiesHandler.load(Path)` (`method_16714`)

*I may have forgotten some, but I'm sure these conflict*

### LINKED ISSUE #1178: SlimeEntity#isBig -> canAttack or something
`isBig` currently:

* returns false on the client side, regardless of the size
* overridden in magma cubes to return true, regardless of the size

This makes me think a name closer to `canAttack` would be better.

### LINKED ISSUE #1193: Class named EntityRenderDispatcher but field named entityRenderManager in MinecraftClient

### LINKED ISSUE #1195: Entity.getSenseCenterPos -> getBlockPos
Shortly before the 20w10a cringe, we had `Entity.getBlockPos` and `Entity.getSenseCenterPos` where the sense center pos was used by piglin and hoglin for their brain to scan blocks. now the new `getSenseCenterPos` as matched by modmuss is like the feet position and no longer a sense center position. time to rename before this name enters 1.16 pre-release mappings.

YanisBft:
Not much to say.
- Closes #1145 
- Closes #1178 
- Closes #1193 
- Closes #1195

### DIFF
--- a/mappings/net/minecraft/entity/mob/EndermanEntity.mapping
+++ b/mappings/net/minecraft/entity/mob/EndermanEntity.mapping
@@ -6,7 +6,7 @@ CLASS net/minecraft/class_1560 net/minecraft/entity/mob/EndermanEntity
 	FIELD field_7255 ANGRY Lnet/minecraft/class_2940;
 	FIELD field_7256 ATTACKING_SPEED_BOOST_UUID Ljava/util/UUID;
 	FIELD field_7257 CARRIED_BLOCK Lnet/minecraft/class_2940;
-	METHOD method_7024 teleport (DDD)Z
+	METHOD method_7024 teleportTo (DDD)Z
 		ARG 1 x
 		ARG 3 y
 		ARG 5 z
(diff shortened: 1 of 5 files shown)


mappings/net/minecraft/entity/mob/EndermanEntity.mapping
```diff
@@ -6,7 +6,7 @@ CLASS net/minecraft/class_1560 net/minecraft/entity/mob/EndermanEntity
 	FIELD field_7255 ANGRY Lnet/minecraft/class_2940;
 	FIELD field_7256 ATTACKING_SPEED_BOOST_UUID Ljava/util/UUID;
 	FIELD field_7257 CARRIED_BLOCK Lnet/minecraft/class_2940;
-	METHOD method_7024 teleport (DDD)Z
+	METHOD method_7024 teleportTo (DDD)Z
```
Earthcomputer, line 9: Is this `to` suffix really the right way to go about this conflict? I would prefer to name it something like `endermanTeleport`, as (iirc) the function does other specific things to do with endermen, like play the teleport sound.

YanisBft: Named it `teleportTo` to match another method named teleportTo, which teleports to an entity instead of coordinates

Runemoro (approved):

liach (approved):

Juuxel (approved):

jaskarth (approved):

PR #1205, opened 2020-03-18
Fix swapped tick names

modmuss50:
I think I accidenly got these names wrong, can someone please validate this as I have got my self confused. :P


LemmaEOF (approved):
Yes, it looks like they did indeed get swapped around.

PR #1211, opened 2020-03-20
Clarify Minecart sounds, fixes #918

### LINKED ISSUE #918: Minecart moving sound names messed up
net/minecraft/client/sound/MinecartSoundInstance (class_1107) and net/minecraft/client/sound/RidingMinecartSoundInstance (class_1108) are mismatched. 1107 is actually the sound when player is riding, and 1108 is the ambient minecart sound.

Confirmed in both 1.14.4 and 19w42a.

liach:
Closes #918

The sound event `entity.minecart.riding` used by `class_1108` is the sound of minecart moving, created when a minecart is spawned on client.
the sound event `entity.minecart.inside` used by `class_1107` is the sound when player is inside minecart, created when the controlled client player starts riding a minecart.

Renamed the classes based on their actual usage than their sound event names.

### DIFF
--- a/mappings/net/minecraft/client/sound/MinecartInsideSoundInstance.mapping
+++ b/mappings/net/minecraft/client/sound/MinecartInsideSoundInstance.mapping
@@ -1,5 +1,5 @@
 CLASS net/minecraft/class_1107 net/minecraft/client/sound/MinecartInsideSoundInstance
-	COMMENT A sound instance played when a player is riding a Minecart.
+	COMMENT A sound instance played when a player is riding a minecart.
 	FIELD field_5456 minecart Lnet/minecraft/class_1688;
 	FIELD field_5457 player Lnet/minecraft/class_1657;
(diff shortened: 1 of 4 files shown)


mappings/net/minecraft/client/sound/MinecartInsideSoundInstance.mapping
```diff
@@ -1,4 +1,5 @@
-CLASS net/minecraft/class_1107 net/minecraft/client/sound/MinecartSoundInstance
+CLASS net/minecraft/class_1107 net/minecraft/client/sound/MinecartInsideSoundInstance
+	COMMENT A sound instance played when a player is riding a Minecart.
```
YanisBft: Doesn't need to be uppercase
```suggestion
	COMMENT A sound instance played when a player is riding a minecart.
```

YanisBft (approved):

Juuxel (approved):
I guess the IDs are like that because
- `inside` = player is inside
- `riding` = the minecart is riding on the rails

PR #1213, opened 2020-03-20
EntityContext -> ShapeContext, EntityContextImpl -> EntityShapeContext

### LINKED ISSUE #908: EntityContext -> ShapeContext
The so-called entity context is only used for collision/outline shape checks. Moreover, its "sneaking" detection is not plain sneaking check but rather `method_21752`, which delegates to `isSneaking` but is only called by shape context checks.

Also see #669 where this name was initially introduced.

liach:
Fixes #908.

This context is exclusively used by block's 3 shape checking methods.
For its usage outside of blocks, it is used to be eventually passed to call one of the shape checking methods (as in CollisionView#canPlace or RayTraceContext)

### DIFF
--- a/mappings/net/minecraft/block/EntityShapeContext.mapping
+++ b/mappings/net/minecraft/block/EntityShapeContext.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_3727 net/minecraft/entity/EntityContextImpl
+CLASS net/minecraft/class_3727 net/minecraft/block/EntityShapeContext
 	FIELD field_16450 minY D
 	FIELD field_16451 descending Z
 	FIELD field_17593 ABSENT Lnet/minecraft/class_3726;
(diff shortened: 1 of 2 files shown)


Juuxel (approved):

YanisBft:
Looks a bit weird in the block package tho. Wouldn't it be better in util?

liach:
These are only used by blocks to conditionally return different collision shapes than used by any collision calculations.

On another note, `CollisionView` and `RayTraceContext` are both in `world` package.

YanisBft:
Why not make a subpackage `context` in `block` then? Just feels out of place imo among block classes

liach:
uhh, these classes are more like the role of `AbstractBlock.OffsetType`or `Material`, which I consider proper under `block` package.

YanisBft (approved):

PR #1216, opened 2020-03-23
BiomeAccessType -> BiomeInterpolator

### LINKED ISSUE #1062: VoronoiBiomeAccessType and .calcChance/distribute naming
The class `net.minecraft.world.biome.source.VoronoiBiomeAccessType` has a two improvable method names besides its own name:

`calcChance(long, int, int, int, double, double, double)` doesn't compute a chance, but a random position around a given point. It also carries the biome rng initialization+advancing steps (manually inlined).

`net.minecraft.world.biome.source.VoronoiBiomeAccessType.distribute(long)` doesn't distribute, but does the standard biome rng nextInt(1024) followed by some scaling to obtain an uniformly distributed random value in [0.45..0.45).

The `storage` parameter naming across the entire subsystem is not very accurate, it is just the parent data source with a little FIFO cache.

VoronoiBiomeAccessType effectively distorts the 3D grid holding biome IDs randomly (locked to the seed), zooms in 4x (for each dim), selects the closest grid point and finally samples it without interpolation. This grid distortion is up to +- 0.45 in every dimension.

The closest analogy to Voronoi diagrams is that biome boundaries more or less follow the same "closest point" rules on a micro level - ignoring quantization. This name however doesn't describe how the grid got actively distorted or zoomed in. A more appropriate name should mention grid distortion, nearest point sampling and the 4x zooming.

liach:
Fixes #1062

### DIFF
--- a/mappings/net/minecraft/world/biome/source/VoronoiBiomeAccessType.mapping
+++ b/mappings/net/minecraft/world/biome/source/VoronoiBiomeAccessType.mapping
@@ -1,13 +1,40 @@
 CLASS net/minecraft/class_4546 net/minecraft/world/biome/source/VoronoiBiomeAccessType
+	COMMENT The voronoi biome access type distorts the biome volume's voxel points
+	COMMENT and finds the closest distorted biome voxel point from a block position
+	COMMENT as if in a euclidean voronoi diagram and returns the biome stored in that
+	COMMENT biome voxel point.
 	METHOD method_22397 sqr (D)D
+		COMMENT Returns the square of {@code d}.
 		ARG 0 d
-	METHOD method_22398 distribute (J)D
+	METHOD method_22398 calculateShift (J)D
+		COMMENT Calculates the shift, in the biome voxel's axis unit, of the weight
+		COMMENT center of a biome position from the regular grid position in an axis.
+		COMMENT
+		COMMENT <p>It takes the 25th to 34th lowest bits of the seed, divides it by
+		COMMENT 1024, minuses it by 0.5, and multiply it by 0.9 to get a shift between
+		COMMENT -0.45 and 0.45 in the biome voxel's axis unit.
 		ARG 0 seed
-	METHOD method_22399 calcChance (JIIIDDD)D
+			COMMENT the seed
+	METHOD method_22399 calculateSquaredDistance (JIIIDDD)D
+		COMMENT Calculates the squared euclidean distance from an accessed (voxel) point
+		COMMENT to the distorted biome (voxel) point based on an original biome point.
+		COMMENT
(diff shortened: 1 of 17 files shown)


jaskarth (approved):
Looks good, I would also add some javadoc to `HorizontalVoronoiBiomeAccessType` saying that it manipulates the biome voxels only along the x and z axis, just to make it a bit clearer.

YanisBft (approved):

mappings/net/minecraft/world/biome/source/VoronoiBiomeAccessType.mapping
```diff
@@ -1,13 +1,45 @@
...
 	METHOD method_22397 sqr (D)D
+		COMMENT Returns the square of {@code d}.
 		ARG 0 d
-	METHOD method_22398 distribute (J)D
+	METHOD method_22398 calculateShift (J)D
+		COMMENT Calculates the shift, in the biome voxel's axis unit, of the weight
+		COMMENT center of a biome position from the regular grid position in an axis.
+		COMMENT
+		COMMENT <p>It takes the 25th to 34th lowest bits of the seed, divides it by
+		COMMENT 1024, minuses it by 0.5, and multiply it by 0.9 to get a shift between
```
Juuxel:
```suggestion
		COMMENT 1024, minuses it by 0.5, and multiplies it by 0.9 to get a shift between
```

Prospector:
```suggestion
		COMMENT 1024, subtracts 0.5 from it, and multiply it by 0.9 to get a shift between
```

liach: why?

Prospector: `minuses` isn't an operation, it's subtraction.

liach: Done

jaskarth (changes requested):
`SkewZoomHorizontalCrossSectionBiomeAccessType` is far too long imo, why not just `Horizontal` like before?

jaskarth (approved):

liach:
@modmuss50 please trigger a rebuild, `libraries.minecraft.net` glitched

YanisBft (approved):

liach:
So should we rename this to `BiomeInterpolator` and `SkewedBiomeInterpolator`/`ColumnFixedSkewedBiomeInterpolator`

liach:
Closing. Will make a new one when people agree there is a better name.

Issue #1120, opened 2020-02-17
Arguments don't match results for PositionedSoundInstance.master


ghost:
The volume and pitch names were swapped in the PositionedSoundInstance.master method, meaning that if the first argument (which is currently "volume") is changed, the pitch would be changed instead, same thing with the second argument ("pitch"), which changes the volume.

liach:
Thanks for your pull request!

PR #1220, opened 2020-03-24
Fix PositionedSoundInstance's swapped parameter names

ghost:
I decided to solve #1120  myself and i fixed the cases where filling the supposed "volume" parameter would actually change the pitch and vice-versa, also, it turns out PositionedSoundInstance.ambient was affected by that mistake too, which is fixed here.


liach (approved):
That indeed is swapped. Thank you for this fix!

Juuxel (approved):

YanisBft (approved):

PR #1232, opened 2020-03-26
World modification fix

Earthcomputer:
`PlayerEntity.canMine` was inverted.
"Restricted" comes from a string in `ServerPlayerInteractionManager.processBlockBreakingAction`.

### DIFF
--- a/mappings/net/minecraft/entity/player/PlayerEntity.mapping
+++ b/mappings/net/minecraft/entity/player/PlayerEntity.mapping
@@ -57,7 +57,7 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 		ARG 2 category
 		ARG 3 volume
 		ARG 4 pitch
-	METHOD method_21701 isWorldModificationRestricted (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_1934;)Z
+	METHOD method_21701 isBlockBreakingRestricted (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_1934;)Z
 		ARG 1 world
 		ARG 2 pos
 		ARG 3 gameMode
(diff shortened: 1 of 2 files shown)


YanisBft:
Hum, this method name looks like it belongs to a World class. What about `isPreventedFromWorldModification`?

kvverti:
`canMine` -> `cannotMine`
`shouldLimitWorldModification` -> `shouldAllowWorldModification`

Earthcomputer:
> this method name looks like it belongs to a World class

Does it really? Looks obvious to me.

> `canMine` -> `cannotMine`

I don't like putting negatives into names, it gets confusing as soon as a `!` introduces a double negative. "Restricted" is just as clear as "cannot mine", and "not restricted" is 100x clearer than "not cannot mine".

> `shouldLimitWorldModification` -> `shouldAllowWorldModification`

No, that would be an incorrect name.

kvverti:
Ah I thought both had changed, since you renamed both methods. In that case `shouldLimitWorldModification` should remain unchanged because it is out of scope.

liach:
`isBlockBreakingRestricted`? `hasRestrictedInteraction`?

Earthcomputer:
People seem to be randomly suggesting alternative names without explaining why they are better than mine

liach:
This is only affecting block breaking. `world modification` is too broad for this, like I'd argue killing mobs is world modification as well.

Runemoro:
@liach Just block breaking, or placing too?

Also, this method doesn't seem to take a `player` parameter. If the player is in adventure mode and they're holding a tool with an NBT tag that allows breaking certain blocks, this method would return true, but breaking blocks is still allowed.

liach:
block breaking only, used in player interaction manager or server play network handler

YanisBft (approved):

mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -134,7 +134,7 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 		ARG 2 inventory
 	METHOD method_7292 getLuck ()F
 	METHOD method_7293 vanishCursedItems ()V
-	METHOD method_7294 canModifyWorld ()Z
+	METHOD method_7294 canModifyBlocks ()Z
```
liach, line 137:
```suggestion
	METHOD method_7294 canBreakBlocks ()Z
```
unless this is used for interaction checks like axe right click/shovel right click

Earthcomputer: It's used for interaction checks where the interaction may modify the block, like inverting a daylight sensor, as well as for block breaking

liach (approved):

hYdos (approved):
looks good. naming looks fine

Juuxel (approved):

Issue #978, opened 2019-11-23
Vector3f is in client.util.math


Juuxel:
It's not client-only (as of 1.15-pre1).

liach:
Hmm, these `Vector` stuff etc. are featured by their mutable fields and overwhelmingly client usage compared to their util.math counterparts. Interesting thing to look into.

Runemoro:
The `Vec` and `Vector` naming is also very weird. It doesn't make sense to name two classes the same but with one name abbreviated. What's the important difference between `Vector3d` and `Vec3d` (why doesn't Mojang use a single class?).

liach:
we have `Vec` for immutable types and `Vector` for mutable types.

Runemoro:
Then I think `Vector` Should be `MutableVec`.

liach:
@Yanis48 Actually I am not very sure on handling this yet; should we rename this to `Vector3f` to distinguish from the immutable vec3d in minecraft util math package and move to minecraft util math, or?

YanisBft:
I think we should rename it to `Vec3f`, and move it to nm.util.math

PR #1235, opened 2020-03-27
Move direction transformation and matrix3f to util math (need same package)

### LINKED ISSUE #947: Move util.ChunkPosDistanceLevelPropagator to world package
Now, we have LevelPropagator and SectionDistanceLevelPropagator in net/minecraft/world/chunk and net/minecraft/util packages. We should look into grouping them into more similar packages. Since @runemoro named them, I will leave this to him.

liach:
image

move two matrices to non-client (as they are no longer client exclusive, see #978);
move two transformations to util math;
move globalpos to dynamic;
move propagator to world, fixes #947;
a few other util additions

### DIFF
--- a/mappings/net/minecraft/util/dynamic/GlobalPos.mapping
+++ b/mappings/net/minecraft/util/dynamic/GlobalPos.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_4208 net/minecraft/util/GlobalPos
+CLASS net/minecraft/class_4208 net/minecraft/util/dynamic/GlobalPos
 	FIELD field_18790 dimension Lnet/minecraft/class_2874;
 	FIELD field_18791 pos Lnet/minecraft/class_2338;
 	METHOD equals (Ljava/lang/Object;)Z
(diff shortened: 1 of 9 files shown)


mappings/net/minecraft/util/dynamic/GlobalPos.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_4208 net/minecraft/util/GlobalPos
+CLASS net/minecraft/class_4208 net/minecraft/util/dynamic/GlobalPos
```
Sollace, line 1: Why is `GlobalPos` going into the dynamic package? Does it work with dynamics?

liach: Yes, this is like dynamic serializable uuid (used by villager brains) and does not implement `Position` or `WorldPosition`

Sollace: Maybe a different class name is in order, then. Because `GlobalPos` tells me that it works with `BlockPos` and nothing else.

Juuxel (approved):

Sollace (approved):

YanisBft (approved):

PR #1244, opened 2020-03-30
block bad names and glstatemanager bad stuff, rotation3 -> affine transformation

### LINKED ISSUE #1224: Block.getRenderingState -> postProcessState
This method is nowhere used by rendering. It is indirectly called by post processors in world, piston, and structures.

### LINKED ISSUE #1233: AbstractBlockState.hasInWallOverlay -> hasEmissiveLighting
Looks like a mismatch when someone tries to port names. Look at magma blocks.

### LINKED ISSUE #1234: method_21914 -> genTextures
https://github.com/FabricMC/yarn/blob/c1d0d9db1f580a757ee43390038784d5228d4c44/mappings/com/mojang/blaze3d/platform/GlStateManager.mapping#L36
Calls `glGenTextures`

### LINKED ISSUE #1241: Rotation3 -> ?
`class_4590`, aka `Rotation3`, added in 19w39a (1.15 blaze3d), in client util math package is pretty much identical to 
https://github.com/MinecraftForge/MinecraftForge/blob/1.14.x/src/main/java/net/minecraftforge/common/model/TRSRTransformation.java
both of which are a wrapper around a 4 by 4 matrix.

Forge calls it translation, rotation, scale, rotation as seen in the javadocs there.
However, in actual code, only one getter of those field is not removed by proguard; and the only usage of that single getter is in the creation logic of the `IDENTITY` transformation instance. This class becomes a meaningless wrapper.

I personally dislike that name because this is simply a 3d transformation. A stackoverflow answer for reference:
https://stackoverflow.com/questions/2465116/understanding-opengl-matrices/2465290#2465290

Finally, if `class_4590` is renamed, `class_4609`(`Rotation3Helper`) needs a rename, too.

liach:
close #1224, close #1233, close #1234, close #1241

### DIFF
--- a/mappings/com/mojang/blaze3d/platform/GlStateManager.mapping
+++ b/mappings/com/mojang/blaze3d/platform/GlStateManager.mapping
@@ -33,7 +33,7 @@ CLASS net/minecraft/class_4493 com/mojang/blaze3d/platform/GlStateManager
 	METHOD method_21912 disableTexture ()V
 	METHOD method_21913 enableClientState (I)V
 		ARG 0 cap
-	METHOD method_21914 getTexLevelParameter ()I
+	METHOD method_21914 genTextures ()I
 	METHOD method_21920 enableRescaleNormal ()V
 	METHOD method_21921 getString (I)Ljava/lang/String;
 		ARG 0 name
@@ -220,7 +220,7 @@ CLASS net/minecraft/class_4493 com/mojang/blaze3d/platform/GlStateManager
 		ARG 1 internalFormat
 		ARG 2 width
 		ARG 3 height
-	METHOD method_21989 texEnv (ILjava/nio/FloatBuffer;)V
+	METHOD method_21989 getFloat (ILjava/nio/FloatBuffer;)V
 		ARG 0 pname
 		ARG 1 params
 	METHOD method_21990 getUniformLocation (ILjava/lang/CharSequence;)I
@@ -474,6 +474,8 @@ CLASS net/minecraft/class_4493 com/mojang/blaze3d/platform/GlStateManager
 		ARG 1 source0Color
 	METHOD method_23282 setupOutline ()V
 	METHOD method_23283 teardownOutline ()V
+	METHOD method_24221 setupGuiFlatDiffuseLighting ()V
+	METHOD method_24222 setUpGui3dDiffuseLighting ()V
 	CLASS class_1010 FBOMode
 	CLASS class_1016 AlphaTestState
 		FIELD field_5042 capState Lnet/minecraft/class_4493$class_1018;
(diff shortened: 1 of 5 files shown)


mappings/net/minecraft/client/util/math/AffineTransformation.mapping
```diff
@@ -7,7 +7,7 @@ CLASS net/minecraft/class_4590 net/minecraft/client/util/math/Rotation3
 	FIELD field_20905 rotation1 Lnet/minecraft/class_1158;
```
Runemoro, line 7: What does the `1` mean here?

And why is there a rotation field at all, can't the rotation be included in the matrix?

Runemoro (approved):

mappings/net/minecraft/client/util/math/AffineTransformation.mapping
liach: See #1241 and that forge class

mappings/com/mojang/blaze3d/platform/GlStateManager.mapping
```diff
@@ -474,6 +474,8 @@ CLASS net/minecraft/class_4493 com/mojang/blaze3d/platform/GlStateManager
 		ARG 1 source0Color
 	METHOD method_23282 setupOutline ()V
 	METHOD method_23283 teardownOutline ()V
+	METHOD method_24221 setupGuiFlatDiffuseLighting ()V
+	METHOD method_24222 setUpGui3dDiffuseLighting ()V
```
YanisBft:
```suggestion
	METHOD method_24222 setupGui3dDiffuseLighting ()V
```

YanisBft (approved):

mappings/net/minecraft/client/util/math/AffineTransformation.mapping
Earthcomputer: The Forge class calls them `leftRot` and `rightRot`. Tbh a lot of these names in this class are a mess and warrant a PR of their own.

> And why is there a rotation field at all, can't the rotation be included in the matrix?

It is. It's a cached version that may be recomputed in the getter.

liach: These fields are really only initialized but never used... The only `getRotation2` is retained because when `IDENTITY` is constructed, `getRotation2` is called to initialize the components

Earthcomputer: There were probably such getters in the original source which proguard stripped because they weren't used elsewhere.

liach: Can I leave this field rename till later because these fields don't really matter?

jaskarth (approved):

modmuss50 (approved):

PR #1106, opened 2020-02-09
Container -> ScreenHandler

Prospector:
This is the culmination of many issues, PRs, and discord chats. While I completely understand those in favor of using the mojang name `Menu` (I supported it for a long time too!), the opinion against it is too strong. I hope that we can settle on this being more preferable to the existing name, `Container`, though.

PR #1247, opened 2020-04-01
Use serialized names for Biome.MixedNoisePoint fields

Juuxel:
A `serialize` method was added in 20w14∞, and we should use the names in it to match serialization output.

Serialized name | Current Yarn name
----------------|--------------------
`temperature`   | `temperature`
`humidity`        | `humidity`
`altitude`          | `hilliness`
`weight`           | `rarityPotential`
`weirdness`      | `style`

### DIFF
--- a/mappings/net/minecraft/world/biome/Biome.mapping
+++ b/mappings/net/minecraft/world/biome/Biome.mapping
@@ -172,20 +172,20 @@ CLASS net/minecraft/class_1959 net/minecraft/world/biome/Biome
 		COMMENT to other point values.
 		FIELD field_22043 temperature F
 		FIELD field_22044 humidity F
-		FIELD field_22045 hilliness F
-		FIELD field_22046 style F
-		FIELD field_22047 rarityPotential F
+		FIELD field_22045 altitude F
+		FIELD field_22046 weirdness F
+		FIELD field_22047 weight F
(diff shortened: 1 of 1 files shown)


jaskarth (approved):

liach:
`weight` and `weirdness` imo doesn't make much sense... 
that weight is actually a buff so as to decrease the effective distance from a sampled point to the target biome noise point, which rarity potential conveys better.
weirdness isn't as good as style. There is nothing weird, but the style can be a scalar that fluctuates and biomes with more consistent styles can reasonably be grouped together in generation.

YanisBft (approved):

Earthcomputer:
What I find impressive is how well mine and @liach 's original guesses match the apparent actual meaning, using only our guesses based on what biomes had which values :P

Earthcomputer:
I would approve of a 13b backport if it kept `rarityPotential` as it is.

modmuss50 (approved):

Earthcomputer:
Why was this merged? We didn't all agree and it isn't even a complete PR (there's other stuff around the world gen code that need renaming now because of this)

Juuxel:
> We didn't all agree

*looks at #1106*

> there's other stuff around the world gen code that need renaming now because of this

Hm, I didn't know that... Sorry.

PR #1253, opened 2020-04-02
Level file mappings

YanisBft:
Small PR to map the new synchronous thing in 20w14a.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
 		}
 	}
 	dependencies {
-		classpath "cuchaz:enigma:0.15.1+build.154"
+		classpath "cuchaz:enigma:0.15.153"
 		classpath "net.fabricmc:stitch:0.4.3.71"
 		classpath "commons-io:commons-io:2.6"
 		classpath "com.google.guava:guava:28.0-jre"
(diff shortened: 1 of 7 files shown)


mappings/net/minecraft/world/storage/RegionFile.mapping
```diff
@@ -11,9 +11,11 @@ CLASS net/minecraft/class_2861 net/minecraft/world/storage/RegionFile
 	METHOD <init> (Ljava/io/File;Ljava/io/File;Z)V
 		ARG 1 file
 		ARG 2 directory
+		ARG 3 synchronous
 	METHOD <init> (Ljava/nio/file/Path;Ljava/nio/file/Path;Lnet/minecraft/class_4486;Z)V
 		ARG 1 file
 		ARG 2 directory
+		ARG 4 synchronous
```
liach: This actually triggers [dsync](https://docs.oracle.com/javase/8/docs/api/java/nio/file/StandardOpenOption.html#DSYNC) standard open options flag than sync (only content is guaranteed, file metadata is not), dunno if this had better be `synchronous` or `dsync`

liach (approved):

build.gradle
```diff
@@ -49,7 +49,7 @@ configurations {
 
 dependencies {
 	enigmaRuntime "net.fabricmc:stitch:0.4.3.71"
```
modmuss50: I would put this in a separate PR, as ideally other deps need updating as well for the asm 8 move

mappings/net/minecraft/world/storage/RegionFile.mapping
YanisBft: updated

build.gradle
YanisBft: reverted enigma update

modmuss50 (approved):

PR #1255, opened 2020-04-03
BlockWithEntity -> AbstractEntityBlock

Prospector:
BlockWithEntity is a bit of an odd name because it doesn't follow the usual pattern for class names. Plus, it includes a javadoc clarifying that this isn't necessarily the class that a modder may want to use, even though it seems like it is at first.

### DIFF
--- a/mappings/net/minecraft/block/AbstractEntityBlock.mapping
+++ b/mappings/net/minecraft/block/AbstractEntityBlock.mapping
@@ -1,0 +1,5 @@
+CLASS net/minecraft/class_2237 net/minecraft/block/AbstractEntityBlock
+	COMMENT This class is *not* necessary for BlockEntities; all you need is {@link BlockEntityProvider}.
+	COMMENT
+	COMMENT In fact, extending this class will have potentially unwanted side effects such as setting
+	COMMENT the render type to INVISIBLE, rather than the default of MODEL.
--- a/mappings/net/minecraft/block/AbstractEntityBlock.mapping
+++ b/mappings/net/minecraft/block/AbstractEntityBlock.mapping
@@ -1,5 +1,5 @@
 CLASS net/minecraft/class_2237 net/minecraft/block/AbstractEntityBlock
-	COMMENT This class is *not* necessary for BlockEntities; all you need is {@link BlockEntityProvider}.
+	COMMENT This class is <b>not</b> necessary for BlockEntities; all you need is {@link BlockEntityProvider}.
 	COMMENT
 	COMMENT In fact, extending this class will have potentially unwanted side effects such as setting
 	COMMENT the render type to INVISIBLE, rather than the default of MODEL.
--- a/mappings/net/minecraft/block/BlockWithEntity.mapping
+++ b/mappings/net/minecraft/block/BlockWithEntity.mapping
@@ -1,1 +1,0 @@
-CLASS net/minecraft/class_2237 net/minecraft/block/BlockWithEntity


mappings/net/minecraft/block/AbstractEntityBlock.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS net/minecraft/class_2237 net/minecraft/block/AbstractEntityBlock
+	COMMENT This class is *not* necessary for BlockEntities; all you need is {@link BlockEntityProvider}.
```
liach:
```suggestion
	COMMENT This class is <b>not</b> necessary for BlockEntities; all you need is {@link BlockEntityProvider}.
```

liach (approved):
Oh my suggested change just now was making it bold, but note that `*` is a markdown thing; javadoc is html

Prospector:
really I just wanted to emphasize it, not make it bold, but bold works too so I added it anyway

Earthcomputer (changes requested):
No, I really don't like this name change. `AbstractEntityBlock` just looks like 3 random words without meaning to me, whereas `BlockWithEntity` makes it very clear. I suggest either `EntityBlock`, `BlockEntityBlock` (both going along with `BlockItem`), or `ContainerBlock` (possibly being renamed to `ScreenHandlerBlock` in the future).

YanisBft:
> I suggest either `EntityBlock`, `BlockEntityBlock` (both going along with `BlockItem`), or `ContainerBlock` (possibly being renamed to `ScreenHandlerBlock` in the future).

`BlockEntityBlock` is quite good, except the repetition of "Block", but I still prefer `BlockWithEntity`. `ContainerBlock`/`ScreenHandlerBlock` is bad though, as not all block entities are containers/screen handlers (see `AbstractBannerBlock`, extending `BlockWithEntity` but not having a screen handler).
Also, the container -> screen handler renaming was already merged a few weeks ago, I don't understand the last part of your message.

Prospector:
- `EntityBlock` or `BlockEntityBlock` - the `Abstract` part really emphasizes how this class has no specific purpose, and it's just sorta a base class that BlockEntity blocks *can* use. Also, if it looks like 3 random words with no meaning, that perfectly sums up what this class is, while `BlockWithEntity` seems like it is more important than it is.
- `ContainerBlock` - no, this has nothing to do with containers
- `ScreenHandlerBlock` - no, this only deals with screen handlers if they have an additional interface

Earthcomputer:
Prefixing this class name with `Abstract` goes against the paragraph in the bottom of [conventions.md](https://github.com/FabricMC/yarn/blob/20w14a/CONVENTIONS.md), and is unnecessary and makes it confusing. It doesn't need to be called `AbstractEntityBlock` especially if there is no `EntityBlock`. The word is superfluous and distracts from the actual function of the class. I can already immediately tell it's abstract without the name.

Prospector:
> unnecessary and makes it confusing

It's a confusing class, hence the javadoc. If anything, making someone skeptical of using the class and checking the doc first is a good thing. However, it actually does add meaning as the class is not just literally an Abstract class in the java sense, but also in the sense that it has no specific theme other than it being an abstract class for EntityBlocks

Earthcomputer:
Speaking of your javadocs, you seem to imply the class shouldn't be used at all. Do you think this is correct, or should you at least state the advantage of extending this class (the creation of the screen handler)?

The class is not confusing. It's the base class for vanilla block entity blocks. Vanilla blocks do not implement `BlockEntityProvider` directly, but rather extend this class, which adds extra default implementations for block entity provider blocks.

liach:
Then AbstractBlockEntityProvider?

Earthcomputer:
No because it's also a Block, it extends Block

liach:
Well this interface does nothing if implemented by a class that is not subclass of block

mappings/net/minecraft/block/AbstractEntityBlock.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS net/minecraft/class_2237 net/minecraft/block/AbstractEntityBlock
+	COMMENT This class is <b>not</b> necessary for BlockEntities; all you need is {@link BlockEntityProvider}.
+	COMMENT
+	COMMENT In fact, extending this class will have potentially unwanted side effects such as setting
+	COMMENT the render type to INVISIBLE, rather than the default of MODEL.
```
Juuxel, line 5: "This class is not necessary for BlockEntities" is not a good javadoc opening line -- it doesn't describe what the class does. (which is providing an implementation of `Block.createMenuProvider` and making instances BER-only by default)

Juuxel (changes requested):

mappings/net/minecraft/block/AbstractEntityBlock.mapping
Prospector: Yes, and it also causes a big source of confusion for people who think their entity blocks need to use it, and then can't figure out why their model doesn't work anymore. I initially had a whole extra paragraph explaining what this class did but removed it as it was obvious from looking at the implementation, and instead just left the warning bit.

liach: You can add a `<strong>Note:</strong>` to attract people's attention in the raw and built javadocs

Runemoro:
> doesn't follow the usual pattern for class names

I don't see what the problem with the current name is. As long as it's a noun phrase it's ok. The suggested name isn't as clear. Also, even if we'd change this, why add `Abstract` to the name and not just use `EntityBlock`?

liach:
> 
> 
> > doesn't follow the usual pattern for class names
> 
> I don't see what the problem with the current name is. As long as it's a noun phrase it's ok. The suggested name isn't as clear. Also, even if we'd change this, why add `Abstract` to the name and not just use `EntityBlock`?

We don't really like `XxxWithYyy` when we can use `YyyXxx`, like `ScreenWithHandler` is not as good as `HandledScreen`. 

We add `Abstract` to indicate that this class is **optional** if you want to implement a block with entity, like this is handy but is not necessary (like how `AbstractList` is handy but not necessary for `List` implementations)

Runemoro (approved):

Runemoro:
I think this should be merged. It's not the perfect name (`AbstractEntityBlock` makes it seem like we have an interface called `EntityBlock`), but it clarifies that this class doesn't need to be implemented by entity blocks.

PR #1256, opened 2020-04-03
20w14a Attributes

liach:

### DIFF
--- a/mappings/net/minecraft/Bootstrap.mapping
+++ b/mappings/net/minecraft/Bootstrap.mapping
@@ -11,4 +11,4 @@ CLASS net/minecraft/class_2966 net/minecraft/Bootstrap
 	METHOD method_12851 initialize ()V
 	METHOD method_12852 setOutputStreams ()V
 	METHOD method_17597 getMissingTranslations ()Ljava/util/Set;
-	METHOD method_17598 logMissingTranslations ()V
+	METHOD method_17598 logMissing ()V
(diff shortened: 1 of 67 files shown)


mappings/net/minecraft/Bootstrap.mapping
```diff
@@ -11,4 +11,4 @@ CLASS net/minecraft/class_2966 net/minecraft/Bootstrap
 	METHOD method_12851 initialize ()V
 	METHOD method_12852 setOutputStreams ()V
 	METHOD method_17597 getMissingTranslations ()Ljava/util/Set;
-	METHOD method_17598 logMissingTranslations ()V
+	METHOD method_17598 logMissing ()V
```
YanisBft, line 14: Not only translations I guess?

YanisBft (approved):

mappings/net/minecraft/Bootstrap.mapping
liach: yes, now it calls [`DefaultAttributeRegistry.checkMissing()`](https://github.com/FabricMC/yarn/pull/1256/files#diff-8355421ab6feca78702e0ae01b0db30aR4)

Runemoro (approved):

liach:
@Juuxel For your concern, static methods in a subclass can be named the same as static methods in its superclass, it's just that static methods in a class cannot have same descriptor and name as an instance method (including inherited ones)

Juuxel:
Oh, they're static. That's fine then.

PR #1265, opened 2020-04-05
Random fixes

### LINKED ISSUE #852: Block.isSolidMediumSquare -> hasTopRim or similar
`isSolidMediumSquare` is similarly named to `isSolidSmallSquare`, but this is a bit misleading.

`isSolidSmallSquare` is true for the top of a fence, for example. `isSolidMediumSquare` makes me think of a comparable larger centered square, but that's not the case, considering it's false for anvils but true for hoppers, cauldrons, and composters.

Also, `isSolidMediumSquare` doesn't take a direction, it only checks the top face of the block. Rails and redstone gates use it to determine valid placement.

Considering these facts, it seems as though the intent of this function is to check for an outer rim on the top face of a block. Indeed, `SOLID_MEDIUM_SQUARE_SHAPE` is created by taking a block shape and subtracting the inner portion.

### LINKED ISSUE #856: ServerWorld.insideTick/ticking
`field_13953`: `insideTick` -> `inBlockTick`
`field_18264`: `ticking` -> `inEntityTick`
`method_14177`: `isInsideTick` -> `isInBlockTick` (used by piston)

### LINKED ISSUE #887: BlockEntity#shouldNotCopyTagFromItem wrong name
method_11011 currently has a wrong mapping BlockEntity#shouldNotCopyTagFromItem.
This method is actually only allowing tag copying for players with level 2 permissions, much like the functionality of method_5833, or Entity#entityDataRequiresOperator

### LINKED ISSUE #1239: Quaternion.set parameter names mismatch the values they actually set
yarn version: 1.15.2+build.14

### LINKED ISSUE #1259: UpgradeData.BulitinLogic -> BuiltinLogic

YanisBft:
- Closes #852 
- Closes #856
- Closes #887 
- Closes #1239 
- Closes #1259

### DIFF
--- a/mappings/net/minecraft/util/math/Quaternion.mapping
+++ b/mappings/net/minecraft/util/math/Quaternion.mapping
@@ -1,14 +1,14 @@
 CLASS net/minecraft/class_1158 net/minecraft/util/math/Quaternion
 	FIELD field_21493 IDENTITY Lnet/minecraft/class_1158;
-	FIELD field_21582 a F
-	FIELD field_21583 b F
-	FIELD field_21584 c F
-	FIELD field_21585 d F
+	FIELD field_21582 x F
+	FIELD field_21583 y F
+	FIELD field_21584 z F
+	FIELD field_21585 w F
 	METHOD <init> (FFFF)V
-		ARG 1 a
-		ARG 2 b
-		ARG 3 c
-		ARG 4 d
+		ARG 1 x
+		ARG 2 y
+		ARG 3 z
+		ARG 4 w
 	METHOD <init> (FFFZ)V
 		ARG 1 x
 		ARG 2 y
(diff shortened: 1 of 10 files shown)


mappings/net/minecraft/util/math/Quaternion.mapping
```diff
@@ -32,10 +35,10 @@ CLASS net/minecraft/class_1158 net/minecraft/util/math/Quaternion
...
 		ARG 3 c
 		ARG 4 d
-	METHOD method_4921 getB ()F
-	METHOD method_4922 getC ()F
-	METHOD method_4923 getD ()F
-	METHOD method_4924 getA ()F
+	METHOD method_4921 getA ()F
+	METHOD method_4922 getB ()F
+	METHOD method_4923 getC ()F
+	METHOD method_4924 getD ()F
```
liach: Uh... no

mappings/net/minecraft/util/math/Quaternion.mapping
```diff
@@ -1,11 +1,14 @@
 CLASS net/minecraft/class_1158 net/minecraft/util/math/Quaternion
 	FIELD field_21493 IDENTITY Lnet/minecraft/class_1158;
-	FIELD field_21585 a F
+	FIELD field_21582 a F
+	FIELD field_21583 b F
+	FIELD field_21584 c F
+	FIELD field_21585 d F
```
liach: hmm, we should start naming them x y z w if people are really confused. the first three are multipliers on 3 imaginary components and the last one is the multiplier on the real component. I don't know why but these mappings lost again.

liach (changes requested):

mappings/net/minecraft/util/math/Quaternion.mapping
Earthcomputer: Yes, call them x, y, z, w please, that's what they're called.

liach: appears that some modders may agree with the xyzw names, such as https://github.com/jellysquid3/sodium-fabric/commit/fcfccc3d534c866f64c6e24e4f5407a2339cff37#diff-807d8c9858c64d22dc8ed37a70b4280cR88-R91

YanisBft: done

liach (approved):
Hope you can include #856, but it's entirely optional.

Juuxel (approved):

dblsaiko (approved):

Earthcomputer (approved):

PR #1269, opened 2020-04-08
Map some 20w15a new classes (entities, enchantments)

liach:

### DIFF
--- a/mappings/net/minecraft/client/render/entity/feature/HorseMarkingRenderer.mapping
+++ b/mappings/net/minecraft/client/render/entity/feature/HorseMarkingRenderer.mapping
@@ -1,2 +1,0 @@
-CLASS net/minecraft/class_5167 net/minecraft/client/render/entity/feature/HorseMarkingRenderer
-	FIELD field_23938 TEXTURES Ljava/util/Map;
--- a/mappings/net/minecraft/client/render/entity/feature/HorseMarkingRenderer.mapping
+++ b/mappings/net/minecraft/client/render/entity/feature/HorseMarkingRenderer.mapping
@@ -1,0 +1,2 @@
+CLASS net/minecraft/class_5167 net/minecraft/client/render/entity/feature/HorseMarkingRenderer
+	FIELD field_23938 TEXTURES Ljava/util/Map;
--- a/mappings/net/minecraft/item/Vanishable.mapping
+++ b/mappings/net/minecraft/item/Vanishable.mapping
@@ -1,1 +1,3 @@
 CLASS net/minecraft/class_5150 net/minecraft/item/Vanishable
+	COMMENT Represents a type of item that can have the "curse of vanishing", or the
+	COMMENT {@code minecraft:vanishing_curse}, enchantment.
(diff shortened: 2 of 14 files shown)


mappings/net/minecraft/item/Vanishable.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS net/minecraft/class_5150 net/minecraft/item/Vanishable
+	COMMENT Represents a type of item that can have the "curse of vanishing", or the
+	COMMENT {@code minecraft:vanishing_curse}, enchantment.
```
YanisBft: Sounds like you said the same thing twice

mappings/net/minecraft/item/Wearable.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS net/minecraft/class_5151 net/minecraft/item/Wearable
+	COMMENT Represents a type of item that is wearable in an armor equipment slot.
+	COMMENT
+	COMMENT <p>This type of item can be targeted by the "curse of bindng", or the
+	COMMENT {@code minecraft:binding_curse}, enchantment.</p>
```
YanisBft: Again, sounds like you said the same thing twice

mappings/net/minecraft/item/Vanishable.mapping
liach: the displayed name isn't the same as the id, so I said the thing twice. [And someone think it is actually unintuitive.](https://minecraft.gamepedia.com/Official_pages/Java_Edition_unintuitive_ID_list#Enchantment_IDs)

mappings/net/minecraft/item/Wearable.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS net/minecraft/class_5151 net/minecraft/item/Wearable
+	COMMENT Represents a type of item that is wearable in an armor equipment slot.
+	COMMENT
+	COMMENT <p>This type of item can be targeted by the "curse of bindng", or the
```
YanisBft:
```suggestion
	COMMENT <p>This type of item can be targeted by the "curse of binding", or the
```

mappings/net/minecraft/item/Vanishable.mapping
Earthcomputer: It's on that list because it's inconsistent, not necessarily unintuitive. No-one finds that name unintuitive. And besides, you don't need to reference something by its ID if it already has a name. On the rare occasion you need an ID you can look it up in the `Enchantments` class.

YanisBft (approved):

Earthcomputer (approved):
Meh, would have preferred the enchantment name rather than the ID but it's good enough.

mappings/net/minecraft/client/render/entity/feature/HorseMarkingRenderer.mapping
```diff
@@ -0,0 +1,2 @@
+CLASS net/minecraft/class_5167 net/minecraft/client/render/entity/feature/HorseMarkingRenderer
```
Shnupbups: Shouldn't it be `HorseMarkingFeatureRenderer`?

liach: Done.

PR #1273, opened 2020-04-08
Misc Strider mappings

Sollace:
There was not a lot to really do, so I just filled in the names that were missing and re-mapped some names that seemed to go missing between 20w13b and this snapshot.

### DIFF
--- a/mappings/net/minecraft/entity/Mountable.mapping
+++ b/mappings/net/minecraft/entity/Mountable.mapping
@@ -1,4 +1,5 @@
-CLASS net/minecraft/class_5146
+CLASS net/minecraft/class_5146 net/minecraft/entity/Mountable
 	METHOD method_6576 setSaddled (Lnet/minecraft/class_3419;)V
+		ARG 1 soundCategory
 	METHOD method_6725 isSaddled ()Z
 	METHOD method_6765 canBeSaddled ()Z
--- a/mappings/net/minecraft/entity/Mountable.mapping
+++ b/mappings/net/minecraft/entity/Mountable.mapping
@@ -1,5 +1,0 @@
-CLASS net/minecraft/class_5146 net/minecraft/entity/Mountable
-	METHOD method_6576 setSaddled (Lnet/minecraft/class_3419;)V
-		ARG 1 soundCategory
-	METHOD method_6725 isSaddled ()Z
-	METHOD method_6765 canBeSaddled ()Z
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -5,8 +5,10 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	FIELD field_18066 standingEyeHeight F
 	FIELD field_18276 velocity Lnet/minecraft/class_243;
 	FIELD field_19271 inLava Z
+	FIELD field_22417 POSE_HEIGHTS Lcom/google/common/collect/ImmutableMap;
 	FIELD field_22467 pos Lnet/minecraft/class_243;
 	FIELD field_22468 blockPos Lnet/minecraft/class_2338;
+	FIELD field_23807 inanimate Z
 	FIELD field_5951 ridingCooldown I
 	FIELD field_5952 onGround Z
 	FIELD field_5953 firstUpdate Z
--- a/mappings/net/minecraft/entity/LivingEntity.mapping
+++ b/mappings/net/minecraft/entity/LivingEntity.mapping
@@ -170,6 +170,8 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 		COMMENT
 		COMMENT <p>This checks both the entity's main and off hand.
 	METHOD method_24831 getPoses ()Lcom/google/common/collect/ImmutableList;
+	METHOD method_26319 canWalkOnLava ()Z
+	METHOD method_26825 getAttribute (Lnet/minecraft/class_1320;)D
 	METHOD method_26827 createLivingAttributes ()Lnet/minecraft/class_5132$class_5133;
 	METHOD method_5973 canTarget (Lnet/minecraft/class_1299;)Z
 		ARG 1 type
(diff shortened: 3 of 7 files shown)


YanisBft (approved):

mappings/net/minecraft/entity/Mountable.mapping
```diff
@@ -1,4 +1,5 @@
-CLASS net/minecraft/class_5146
+CLASS net/minecraft/class_5146 net/minecraft/entity/Mountable
```
liach: More like saddleable, see #1269

Sollace: Entities with that interface are a Mount, which makes them _mountable_. I find this name more grammatical.

Prospector: Yeah, `Saddleable` is incorrect because not all entities with it can be "saddled" which sounds wrong anyway. `Mountable` or even just `Mount` is a more accurate name.

Prospector (approved):

mappings/net/minecraft/entity/Mountable.mapping
liach: No, this is just saddle logic and is mainly used by dispenser to install saddles. The method `setSaddled` is just installing saddle and passing the sound category for the saddling sound to play in (i.e. block sound category for dispensers and player sound category for players)

the other two methods checks if the entity already has saddle or if it can have saddle installed at current state

Prospector: If canBeSaddled is false, then that means the entity is not saddleable and therefore the class name is incorrect. See how llamas always return false, an entity that is *not* saddleable and implements the interface.

liach: Yes because one of its parent classes is saddleable. Show me how this interface ever relates to mounting logic. Also this class is already mapped in latest yarn

Prospector: It's an interface that every mountable animal in the game implements, and the ones that are saddleable return true on canBeSaddled. It doesn't get much simpler than that.

liach: image
As far as I see from decompiled code this interface is 100% saddle related. It has nothing to do with "mounting".

@Prospector please read the code. How about minecarts boats, which are also mountable? If your theory is right, how about ravagers, which you consider mountable, but cannot have saddles installed (because they always come with one)?

Prospector: You keep dodging the fact that an interface you named `Saddleable` is used by entities which factually are not saddleable. Therefore, it's an incorrect name. The common link between the things that implement it are not whether or not they are saddleable, it's whether or not they are mountable. It's like naming `BlockWithEntity` something like `ScreenHandlerEntityBlock`. While yes, the interface does enable screen handler functionality within some subclasses, it is still implemented by blocks which do not use screen handlers, and therefore cannot be named after that part of its functionality.

You did bring up a good point, and that is that there are mobs that are mountable by other entities such as ravagers, chickens, and spiders. Therefore, I'd propose the name `PlayerMountable`.

liach: > You keep dodging the fact that an interface you named Saddleable is used by entities which factually are not saddleable.

Because mojang doesn't want to refactor llama out of the horse inheritance tree just because of that saddleable interface. End of the story. Nowhere in code is this ever related to player mounting at all. If you name it, show evidence in code.

Again, again, again, we need _**evidence**_.

> The common link between the things that implement it are not whether or not they are saddleable, it's whether or not they are mountable.

There is no mounting logic touched in this interface at all. It is just a coincidence that most tameable + mountable entities need saddles to mount, and the only one that does not need saddles extends from one that need saddles (Llama extends Donkey).

Your logic of basing this off this coincidence is just like [attributing the correlation between ice cream sale and crime rate as causation](https://slate.com/news-and-politics/2013/07/warm-weather-homicide-rates-when-ice-cream-sales-rise-homicides-rise-coincidence.html).

liach: This class is already mapped. This change is wrong, as it renames from a nonexistent file. (Should rename from `Saddleable`)

liach (changes requested):

mappings/net/minecraft/entity/Mountable.mapping
Earthcomputer: > How about minecarts boats, which are also mountable?

You don't "mount" a minecart or boat, you "board" them. However you can "ride" both horses and minecarts.

Earthcomputer (changes requested):
It's `Saddleable`, not `Mountable`. Your assertion that it's used by all entities the player can mount is incorrect. It's ONLY implemented by entities that can have a saddle. The fact that llamas indirectly inherit that from horses is a coincidence, and a result of Mojang's weird code and not the purpose of the interface.

If I was a modder and deciding whether I want to make my mod implement this interface, I would much rather it be called `Saddleable` than `Mountable`.

Please remove this class rename.

mappings/net/minecraft/entity/Mountable.mapping
Sollace: This class was most definitely not mapped when I was working on this PR.

liach: > 
> 
> This class was most definitely not mapped when I was working on this PR.

This isn't an excuse for you not to fix merge conflicts. There will be a format error, so please merge the current 20w15a branch and delete the `Saddleable` mapping to make the pr right. I am not asking you to take the `Saddleable` name, read my last comment again.

Earthcomputer: It should be named Saddleable, or something relating to saddles. It's ONLY implemented by entities that can have a saddle, and is only used by the saddle item and saddle dispenser behaviour.

Sollace: > 
> 
> > This class was most definitely not mapped when I was working on this PR.
> 
> This isn't an excuse for you not to fix merge conflicts. There will be a format error, so please merge the current 20w15a branch and delete the `Saddleable` mapping to make the pr right. I am not asking you to take the `Saddleable` name, read my last comment again.

Why didn't you say so? Sure, I'll fix the conflict.

Earthcomputer (approved):
Approve now that "Mountable" is gone

liach (approved):

YanisBft:
Please update Sollace :)

Sollace:
@modmuss50 Was everything okay? I thought I had to rebase this before it could be merged.

modmuss50:
Yes, it was fine. When there are no conflicts its 2 clicks to change the target branch in github.

Sollace:
@modmuss50 I just thought since @Yanis48 specifically requested a rebase on discord.

PR #1274, opened 2020-04-09
getDropTableID -> getLootTableId

Prospector:
This must've made it through in the AbstractBlock changes

### DIFF
--- a/mappings/net/minecraft/block/AbstractBlock.mapping
+++ b/mappings/net/minecraft/block/AbstractBlock.mapping
@@ -37,7 +37,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		ARG 4 context
 	METHOD method_26160 asBlock ()Lnet/minecraft/class_2248;
 	METHOD method_26161 hasBlockEntity ()Z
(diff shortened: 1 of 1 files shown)


liach:
Should this be renamed to `loot table` than `drop table`?

Prospector:
I do agree with you there, though really this was just the single bugfix

liach:
A bug is still a refactor
Also I suggest just calling it "loot table". `drop table` is really a legacy name https://github.com/FabricMC/yarn/commit/6d3075daebae22e7938460743ba82a5f6618e329#diff-0fc22c45ea8a7363223df24d31b7f7caR202 and no longer applies because we call this type of things loot tables now (before it's loot supplier)

Prospector:
Okay, updated. I hope this doesn't turn into another massive bikeshedding. Just wanted to get a simple bugfix in :P

liach (approved):

Juuxel (approved):

YanisBft (approved):

Juuxel:
Hm, maybe `AbstractBlock.Settings.dropTableId` should also be changed.

Earthcomputer (approved):

PR #1275, opened 2020-04-09
NBT Method Rename v2 for 20w15a

Prospector:
Before diving in for reviews, I would recommend reading #1078 for more context on past discussion. Essentially, the changes are as follows:
- Serialization methods that modify an existing NBT Tag object (whether they return it or not) are named `saveToNbt`.
- Serialization methods that return a new NBT Tag object are named `toNbt`.
- Deserialization methods that modify an existing instance of an object using data from NBT Tag objects are named `loadFromNbt`.
- Deserialization methods that create a new instance of an object using data from the NBT Tag objects are named `fromNbt`.

The distinction between modification versus creation is likely to seem odd to you at first, especially without having the code in front of you. Let me briefly explain:
- All over the Minecraft codebases are examples such as this: in `LootableContainerBlockEntity` there are methods to save just their loot tables to NBT. `saveLootTableToNbt` is just far more readable and it is immediately understood what is going on versus `lootTableToNbt`, which looks like a conversion.
- Methods named `xToNbt` imply that the return value will be exactly `x` in NBT format. However, when really this method appends additional NBT onto it, it is important to make the distinction that NBT is being `save`d to an existing NBT tag.
- There are actually many places in the code where these different types of methods are used together and using all `toNbt` and `fromNbt` would cause conflicting names. Such places are in BlockEntity (static `fromNbt` as well as the `fromNbt` where NBT data is loaded for an instance).
- Including `save` and `load` in the names of the methods used by Block Entities and Item Stacks has the added benefit of it being an easy searchable term for if a new modder wanted to **save** some data to their Block Entity or Item. Obviously, this is just a coincidence due to how Mojang has it currently implemented, but it is a nice added benefit.

Also, I don't think Enigma nuked any comments like in previous PRs, thankfully.

### DIFF
--- a/mappings/net/minecraft/block/entity/BannerBlockEntity.mapping
+++ b/mappings/net/minecraft/block/entity/BannerBlockEntity.mapping
@@ -1,8 +1,8 @@
 CLASS net/minecraft/class_2573 net/minecraft/block/entity/BannerBlockEntity
 	FIELD field_11769 patterns Ljava/util/List;
-	FIELD field_11770 patternListTagRead Z
+	FIELD field_11770 patternsNbtRead Z
 	FIELD field_11772 customName Lnet/minecraft/class_2561;
-	FIELD field_11773 patternListTag Lnet/minecraft/class_2499;
+	FIELD field_11773 patternsNbt Lnet/minecraft/class_2499;
 	FIELD field_11774 baseColor Lnet/minecraft/class_1767;
 	METHOD <init> (Lnet/minecraft/class_1767;)V
 		ARG 1 baseColor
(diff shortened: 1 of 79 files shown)


modmuss50:
These names seem to get changed back and forth every other day.

modmuss50 (changes requested):
Personnaly I like tag more than nbt, I know it clashes with the other tags, but I dont think its a major issue.

Prospector:
If anyone else wishes to continue this PR, feel free. Changing names in yarn is too tedious a process and I'm no longer interested.

PR #1279, opened 2020-04-14
Fix wrong NativeImage method names

### LINKED ISSUE #719: NativeImage get/set pixel method names are encorrect
NativeImage uses the same pixel format as OpenCV/OpenGL. Found this whilst testing why pixel values in MineLP aren't being interpreted correctly.

To reflect this both the methods should be renamed:

NativeImage.getPixelRGBA -> NativeImage.getPixelABGR
NativeImage.setPixelRGBA -> NativeImage.setPixelABGR

i509VCB:
Fixes #719

### DIFF
--- a/mappings/net/minecraft/client/texture/NativeImage.mapping
+++ b/mappings/net/minecraft/client/texture/NativeImage.mapping
@@ -39,6 +39,7 @@ CLASS net/minecraft/class_1011 net/minecraft/client/texture/NativeImage
 	METHOD method_24030 getAlpha (I)I
 		ARG 0 color
 	METHOD method_24031 getAbgrColor (IIII)I
+		COMMENT The resulting color of this operation is stored as least to most significant bits.
 		ARG 0 alpha
 		ARG 1 blue
 		ARG 2 green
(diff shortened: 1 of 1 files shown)


i509VCB:
Note the current names are incorrect since NativeImages do not use ARGB, it's just that mojang's strings specified in exceptions are wrong.

liach:
uhh let's stick to rgba, just note this is from least to most significant bits.
It's a common practice to put values from least to most significant bits (especially when the value is array-esque), such as jvm class file versions, where major versions are in the lower bits while minor versions are in higher bits

liach:
Also see my uv example in that issue above

i509VCB:
> uhh let's stick to rgba

This has the same energy as using `motive` because the registry says `motive`.

I am putting my foot and and I am saying we should use the proper color format for the method name, or change to `get/setPixelColor`

liach:
We should probably call it big/little endian rgba instead.

mappings/net/minecraft/client/texture/NativeImage.mapping
```diff
@@ -169,6 +191,8 @@ CLASS net/minecraft/class_1011 net/minecraft/client/texture/NativeImage
 		METHOD method_4339 setPackAlignment ()V
 		METHOD method_4340 setUnpackAlignment ()V
 	CLASS class_1013 GLFormat
```
Runemoro, line 196: This is a weird name, shouldn't it be something like `ColorFormat` or `PixelFormat`?

YanisBft (approved):

liach (approved):

mappings/net/minecraft/client/texture/NativeImage.mapping
i509VCB: I'll address this in a later PR

YanisBft (approved):

modmuss50 (approved):

PR #1290, opened 2020-04-18
Map Structure PalettedBlockInfoList

liach:

### DIFF
--- a/mappings/net/minecraft/structure/Structure.mapping
+++ b/mappings/net/minecraft/structure/Structure.mapping
@@ -1,5 +1,5 @@
 CLASS net/minecraft/class_3499 net/minecraft/structure/Structure
-	FIELD field_15586 blocks Ljava/util/List;
+	FIELD field_15586 blockInfoLists Ljava/util/List;
 	FIELD field_15587 size Lnet/minecraft/class_2338;
 	FIELD field_15588 author Ljava/lang/String;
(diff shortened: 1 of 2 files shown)


Runemoro (approved):

Runemoro:
PR looks good, but maybe we should find a better name than "block info".

liach:
It holds a list of structure block infos (each is like an info about a block in a structure). The Structure class holds a list of these lists. These lists have the same block arrangements, but they have different palettes for each arrangement (i.e. in one palette the block may be stone, the other blackstone)

mappings/net/minecraft/structure/Structure.mapping
```diff
@@ -1,5 +1,5 @@
 CLASS net/minecraft/class_3499 net/minecraft/structure/Structure
-	FIELD field_15586 blocks Ljava/util/List;
+	FIELD field_15586 blockInfoLists Ljava/util/List;
```
Juuxel, line 2: Is this like `List<List<...>>`?

liach: Effectively yes, but it is `List<PalettedBlockInfoList>`, where `PalettedBlockInfoList` contains a `List<StructureBlockInfo>`, so it's like `List<List<StructureBlockInfo>>`

Juuxel (approved):

YanisBft (approved):

PR #1287, opened 2020-04-18
Update dependencies

liach:
Jd still builds without warning, should be good

Waiting on FabricMC/Enigma#225

Signed-off-by: liach <liach@users.noreply.github.com>

PR #1291, opened 2020-04-18
Update stitch and enigma to fix dependency problems

liach:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,6 @@
 buildscript {
 	repositories {
 		jcenter()
-		mavenLocal()
 		maven {
 			name "Fabric Repository"
 			url 'https://maven.fabricmc.net'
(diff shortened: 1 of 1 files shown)


Runemoro (approved):

liach:
wait this one has some problem, let me actually look into it again

build.gradle
```diff
@@ -7,8 +7,10 @@ buildscript {
 		}
 	}
 	dependencies {
-		classpath "cuchaz:enigma:0.15.1+build.164"
-		classpath "net.fabricmc:stitch:0.4.5+build.73"
+		classpath "cuchaz:enigma:0.15.2+build.165"
+		classpath("net.fabricmc:stitch:0.4.5+build.73") {
+			exclude group: "cuchaz", module: "enigma"
```
modmuss50: do we actaully need this?

liach: Turns out if the enigma above fails to resolve, gradle will auto fall back to that one with stitch, which happened in that last #1287 pr

Juuxel (approved):

PR #1296, opened 2020-04-22
Fix method name in VoronoiBiomeAccessType

Kroppeb:
Changes the `calcChance` to `calcSqrDistance` as the old name didn't make sense.

### DIFF
--- a/mappings/net/minecraft/world/biome/source/VoronoiBiomeAccessType.mapping
+++ b/mappings/net/minecraft/world/biome/source/VoronoiBiomeAccessType.mapping
@@ -3,7 +3,7 @@ CLASS net/minecraft/class_4546 net/minecraft/world/biome/source/VoronoiBiomeAcce
 		ARG 0 d
 	METHOD method_22398 distribute (J)D
 		ARG 0 seed
-	METHOD method_22399 calcChance (JIIIDDD)D
+	METHOD method_22399 calcSqrDistance (JIIIDDD)D
 		ARG 0 seed
 		ARG 2 x
 		ARG 3 y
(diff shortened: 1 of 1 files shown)


YanisBft:
You mean `sqrt`?

Kroppeb:
No `sqrt` would mean `sqrt(distance)` but `sqr` is `distance²` which is commonly used to compare multiple distances as it's much faster which is what this is used for.

liach:
Just call it `calcuateSquaredDistance` I guess, this is not really chance

Kroppeb:
Oh, I just assumed the standard was `sqr` as that's how the private function in the class is called. But I noticed that all other places use `squareDistance`. Should I rename `sqr` too then?

liach:
Sure, both are private methods so renaming them has no mod impact. Also see #1062 (consider renaming distribute to show that this is assigning a point to the closest point out of the 4 near it)

Kroppeb:
* Voronoi is quite a good name. Even though it's not exactly correct as it only checks the 8 surrounding biome sections, ignoring the fact that seeds of other sections further away could technically be closer. It still helps a lot with understanding what is going on.
* `distribute` gives a "random" value in the range `[-0.45, 0.45[`. I feel the name is descriptive enough.

YanisBft (approved):

Juuxel (approved):
Looks good (though `sqr` is just as good as `square` IMO)

liach (approved):

Player3324 (approved):

modmuss50 (approved):

liach:
So can #1062 be considered fixed now?

PR #1305, opened 2020-04-29
item property getter

liach:


modmuss50:
What about dropping the Getter?
so just ItemProperty and ItemPropertyRegistry?

liach:
`ItemPropertyGetter` is still a getter. You don't have a property that is a functional interface, only a property getter that is a functional interface.

The registry class can probably be named `ItemProperties` instead. How do you think?

liach:
I am considering moving this class away from `Item`. `ItemProperty` is way too much like `ResourceLocation`, bringing extra confusing because it is totally different from what it actually does. I propose renaming it to like `ModelPredicateProvider` as it provides values for `predicates` entries in `overrides` of a model.

modmuss50 (approved):

Juuxel (approved):

mappings/net/minecraft/client/item/ModelPredicateProviderRegistry.mapping
```diff
@@ -0,0 +1,45 @@
...
+		ARG 0 id
+		ARG 1 provider
+	CLASS 1
+		FIELD field_7910 step D
+		FIELD field_7911 time D
+		FIELD field_7913 lastTick J
+		METHOD method_7736 getTime (Lnet/minecraft/class_1937;D)D
+			ARG 1 world
+			ARG 2 skyAngle
+	CLASS 2
```
YanisBft, line 25: You don't name these classes?

Juuxel: They're anonymous classes if their name is just a number.

YanisBft (approved):

Runemoro (approved):

PR #1307, opened 2020-04-29
Bring back the parameters

YanisBft:
Brings back the parameters dropped when matching the snapshot, especially in particle classes


liach (approved):

mappings/net/minecraft/block/Block.mapping
```diff
@@ -23,6 +23,8 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		ARG 3 side
 	METHOD method_23349 getVelocityMultiplier ()F
 	METHOD method_23350 getJumpVelocityMultiplier ()F
+	METHOD method_27839 isBlock (Lnet/minecraft/class_2248;)Z
+		ARG 1 block
```
Juuxel, line 27: This seems like a weird name... What does an `isBlock(Block)` method do?

YanisBft: Checks if the block equals the block in parameter. Should I rename it to `equals`?

liach: why can't they use `==`?

liach: Yeah they just uses `this == block` in implementation. I guess this is their effort to move away from hardcoding; and I personally suggest just calling the method `is`.

Runemoro (approved):

mappings/net/minecraft/block/Block.mapping
```diff
@@ -23,6 +23,8 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 		ARG 3 side
 	METHOD method_23349 getVelocityMultiplier ()F
 	METHOD method_23350 getJumpVelocityMultiplier ()F
+	METHOD method_27839 isBlock (Lnet/minecraft/class_2248;)Z
```
liach:
```suggestion
	METHOD method_27839 is (Lnet/minecraft/class_2248;)Z
```

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -522,6 +522,8 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		METHOD method_26234 isFullCube (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Z
 			ARG 1 world
 			ARG 2 pos
+		METHOD method_27852 isBlock (Lnet/minecraft/class_2248;)Z
```
```suggestion
		METHOD method_27852 is (Lnet/minecraft/class_2248;)Z
```

liach (commented):
Optional name simplification.

Juuxel (changes requested):
The two `isBlock` methods, `AbstractBlock.method_27852` and `Block.method_27839`, should have different Yarn names since they're different by their intermediary names.

liach:
@Juuxel `method_27852` is in `AbstractBlockState`, not `AbstractBlock`

Juuxel:
Oh, nevermind then. `isBlock` is still a bad name, though.

liach (approved):

Juuxel (approved):

liach:
Just making sure that modmuss gets a ping on this ready-to-merge pr.

PR #1310, opened 2020-04-30
Advancement new stuff and refactor old bad names

liach:

### DIFF
--- a/mappings/net/minecraft/advancement/Advancement.mapping
+++ b/mappings/net/minecraft/advancement/Advancement.mapping
@@ -53,7 +53,7 @@ CLASS net/minecraft/class_161 net/minecraft/advancement/Advancement
 			ARG 8 hidden
 		METHOD method_692 fromJson (Lcom/google/gson/JsonObject;Lnet/minecraft/class_5257;)Lnet/minecraft/class_161$class_162;
 			ARG 0 obj
-			ARG 1 predicateLoader
+			ARG 1 predicateDeserializer
 		METHOD method_693 display (Lnet/minecraft/class_185;)Lnet/minecraft/class_161$class_162;
 			ARG 1 display
 		METHOD method_694 build (Ljava/util/function/Consumer;Ljava/lang/String;)Lnet/minecraft/class_161;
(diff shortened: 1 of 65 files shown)


mappings/net/minecraft/advancement/Advancement.mapping
```diff
@@ -70,7 +72,9 @@ CLASS net/minecraft/class_161 net/minecraft/advancement/Advancement
 			ARG 8 hidden
 		METHOD method_698 toJson ()Lcom/google/gson/JsonObject;
 		METHOD method_699 toPacket (Lnet/minecraft/class_2540;)V
+			ARG 1 buf
 		METHOD method_700 findParent (Ljava/util/function/Function;)Z
```
Runemoro, line 76: Maybe this should just be `getParent`?

mappings/net/minecraft/predicate/entity/EntityPredicate.mapping
```diff
@@ -75,3 +78,38 @@ CLASS net/minecraft/class_2048 net/minecraft/predicate/entity/EntityPredicate
...
+	CLASS class_5258 Extended
+		FIELD field_24388 EMPTY Lnet/minecraft/class_2048$class_5258;
+		FIELD field_24389 conditions [Lnet/minecraft/class_4570;
+		FIELD field_24390 combinedCondition Ljava/util/function/Predicate;
+		METHOD <init> ([Lnet/minecraft/class_4570;)V
+			ARG 1 conditions
+		METHOD method_27803 ofLegacy (Lnet/minecraft/class_2048;)Lnet/minecraft/class_2048$class_5258;
+			ARG 0 predicate
+		METHOD method_27804 toJson (Lnet/minecraft/class_5267;)Lcom/google/gson/JsonElement;
+			ARG 1 saver
```
Runemoro: Shouldn't `saver` be `serializer`? Saver sounds like it's writing the json to the file or something, but the method returns a JsonElement.

mappings/net/minecraft/advancement/Advancement.mapping
liach: it is trying to find the parent and returns a boolean indicating if the parent is found.

mappings/net/minecraft/predicate/entity/EntityPredicate.mapping
liach: the saver is an object that has a method that converts a list of loot conditions to a json element. 
the saver is a singleton, and i am not very sure why mojang even made the saver object, as the only used method is final and only a singleton is ever passed; the used method is only used in one place as well

Runemoro: > converts a list of loot conditions to a json element

> the saver is a singleton

That sounds exactly like a serializer...

liach: so loader/saver to deserializer/serializer?

Runemoro (approved):

YanisBft (approved):

Juuxel (approved):

Issue #1133, opened 2020-02-23
IWorld -> GenView


liach:
A followup of ancient #448. #1098 gives me the idea that interfaces can be named `view`, and this interface is a view for potentially async world generation view, hence `GenView` should suffice. Please vote.

i509VCB:
Did we want to use the `Gen` abbreviation or `Generation`

On the case of IWorld currently, it is right where the split between regular `World`, Client and Server worlds and ChunkRegion (very commonly used in World gen) begins.

Whether we should use `World` or `View` is another question as IWorld starts to define some worldly functions such as the existence of state scheduling, difficulties and chunk managers. 

The mentions of IWorld's use in world gen is a good argument for why Gen/GenerationView could work, however I feel a name like IntermediaryWorld could work but that also doesn't compensate for the possible world gen context this would be used within.

GenerationView or IntermediaryWorld are my two choices but I would like to see some more arguments before a firm decision.

valoeghese:
iapprove

Juuxel:
I like @i509VCB's `GenerationView`.

jaskarth:
I vote for GenerationView as well.

i509VCB:
Would like to put some steam back into the discussion here.

The listed options I've seen mentioned in this issue  so far:

GenView
GenerationView
IntermediaryWorld
GeneratableWorld (basically #448 with new naming standard)

My opinions on the above:
GenView is nice and short but Gen does not very explicit with the abbrivation's meaning (General, Generation, Genji?)

GenerationView and GeneratableWorld both fall in the same category of being much more explicit than GenView

IntermediaryWorld is a compromise as not all world types are actually generatable (Client worlds). 
Also due to the fact this class is an interface, the intermediary portion of the name says this represents a sort of "partial definition" for a world or a worldly functions.

Some other ideas:
ProtoWorld (inspired by the sponge api name and based on the Greek word proto, which in this context would be used to mean basic or prototype). 

PrimativeWorld (assuming we use primative in the sense of a simplistic object), issue I see is the connotation with socital definition of primitive which makes no sense with a type of world

YanisBft:
Closed by #1350

PR #1317, opened 2020-05-01
IWorld -> GenerationView

i509VCB:
This resolves the years old conflict of what `IWorld` should become.

Why `GenerationView`?

Commonly used in world gen

Implemented by `ChunkRegion`, which is used in world gen

Could represent a view of a world that is thread safe within the confines of world gen

(This is related to #1133)

### DIFF
--- a/mappings/net/minecraft/world/GenerationView.mapping
+++ b/mappings/net/minecraft/world/GenerationView.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_1936 net/minecraft/world/IWorld
+CLASS net/minecraft/class_1936 net/minecraft/world/GenerationView
 	METHOD method_20290 playLevelEvent (ILnet/minecraft/class_2338;I)V
 		ARG 1 eventId
 		ARG 2 pos


liach:
Are you sure there is no other obsolete reference of `IWorld` in the mappings?

i509VCB:
Nothing I see with parameters still being named `IWorld` atleast

jaskarth (approved):
iapprove

liach (approved):
Should be good, assuming there is no other trash leftover (most likely not)

YanisBft (approved):

Earthcomputer:
`GenerationView` sounds like it's something that's supposed to be used for world generation, but it's used in a lot of places elsewhere.

liach:
It is not specific for generation, but like something something generation compatible

Runemoro:
> It is not specific for generation, but like something something generation compatible

That would be the same as calling `World` `NetherWorld` even though it's not specific for the nether, because it's nether compatible.

Runemoro (changes requested):
As Earthcomputer said, this is used for things that have nothing to do with worldgen.

i509VCB:
`WorldAccess` was brought up in the yarn channel on discord. Thoughts on that instead?

i509VCB:
Susperseeded by https://github.com/FabricMC/yarn/pull/1350

PR #1323, opened 2020-05-02
Block Action -> Synced Block Event. Level Event -> Synced World Event

### LINKED ISSUE #1319: Block Action -> ?
What yarn currently calls "block actions" are used by:
- Pistons (for pushing)
- Note blocks (for playing a sound)
- Bells (for ringing)
- Chests (for the opening animation)
- Ender chests (for the opening animation)
- Shulker boxes (for the opening animation)
- End gateways (for the beam)
- Mob spawners (triggered when the mob spawns to reset the mob spinning speed)

"Block actions" are triggered by the server, and processed by both the server and clients within range of the block (the "block action" is sent to these clients). Basically, it is used for events that also need to be processed on the client.

The yarn name is confusing and does not capture its actual purpose. It looks like a method you would want to call to simulate a right-click action or something. What should we call it instead?

Earthcomputer:
Closes #1319, please read that issue before commenting on this pull request. Also renames level event to synced world event for consistency.

You may notice a difference between `addSyncedBlockEvent` and `syncWorldEvent`. This is because the former adds it to a queue, to be processed on both the client and the server, while the latter is only processed on the client.

### DIFF
--- a/mappings/net/minecraft/block/AbstractBlock.mapping
+++ b/mappings/net/minecraft/block/AbstractBlock.mapping
@@ -363,7 +363,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		METHOD method_26176 getComparatorOutput (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;)I
 			ARG 1 world
 			ARG 2 pos
-		METHOD method_26177 onBlockAction (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;II)Z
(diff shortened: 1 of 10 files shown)


Juuxel:
Could you rename `BlockEntity.onBlockAction` as well?

Earthcomputer:
Oh whoops, missed that one! Will do

Juuxel (approved):

liach:
so world events are only synced, never processed on logical server in addition when the method is called?

Earthcomputer:
correct

liach (approved):
Looks good then.

YanisBft (approved):

PR #1329, opened 2020-05-03
Basic LevelProperties Mappings

TheBrokenRail:
There not the best, but I think they are OK. Please comment.


liach:
They are more like world-specific/client specific interfaces that are exposed. `Basic` `Writable` fail to make that distinction.

TheBrokenRail:
Any better name ideas?

Pyrofab:
My thoughts:
- The `LevelProperties` name seems a bit outdated. Currently it seems to reflect the state of a single world/dimension. As such, I would recommend naming `class_5219` (the base interface) "WorldProperties".
- `class_5269` is a specialization of `class_5219` with setters. I would suggest naming it "MutableWorldProperties".
- `class_5271` and `class_5268` are the specializations of `class_5269` for `ClientWorld` and `ServerWorld`, respectively. As such, I would suggest simply naming them `ClientWorldProperties` and `ServerWorldProperties`.
- `UnmodifiableLevelProperties` is a specialization of `class_5268`, used by `SecondaryServerWorld`. It adds the property `"Derived"` to crash reports, which leads me to suggest `DerivedWorldProperties`.
- `class_5219` is the interface where every global method from the old `LevelProperties` went (eg. `getLastPlayed` or `getEnabledDataPacks`). If we consider `level` to mean "the whole save", this interface is the one that should be called `LevelProperties`.
- The class that is currently named `LevelProperties` implements `class_5268` and `class_5219`. It does not seem unlikely that it gets split in the future. I guess we could call it `PrimaryWorldProperties`.

PR #1342, opened 2020-05-09
Some minecart-related mappings

immibis:


mappings/net/minecraft/entity/passive/HorseBaseEntity.mapping
```diff
@@ -21,6 +21,8 @@ CLASS net/minecraft/class_1496 net/minecraft/entity/passive/HorseBaseEntity
 	FIELD field_6975 soundTicks I
 	FIELD field_6976 jumpStrength F
 	METHOD method_26899 createBaseHorseAttributes ()Lnet/minecraft/class_5132$class_5133;
+	METHOD method_27930 findPassengerDismountLocationInternal (Lnet/minecraft/class_243;Lnet/minecraft/class_1309;)Lnet/minecraft/class_243;
```
liach:
```suggestion
	METHOD method_27930 findDismountLocationInternal (Lnet/minecraft/class_243;Lnet/minecraft/class_1309;)Lnet/minecraft/class_243;
```

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
```diff
@@ -1,12 +1,15 @@
 CLASS net/minecraft/class_1688 net/minecraft/entity/vehicle/AbstractMinecartEntity
+	FIELD field_24464 DISMOUNT_FREE_Y_SPACES_NEEDED Lcom/google/common/collect/ImmutableMap;
 	FIELD field_7655 clientYVelocity D
 	FIELD field_7656 clientZVelocity D
 	FIELD field_7657 clientPitch D
 	FIELD field_7658 clientXVelocity D
 	FIELD field_7659 clientYaw D
+	FIELD field_7660 isYawFlipped Z
```
liach:
```suggestion
	FIELD field_7660 yawFlipped Z
```

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
```diff
@@ -1,12 +1,15 @@
...
 	FIELD field_7661 CUSTOM_BLOCK_OFFSET Lnet/minecraft/class_2940;
 	FIELD field_7662 clientZ D
 	FIELD field_7663 DAMAGE_WOBBLE_TICKS Lnet/minecraft/class_2940;
+	FIELD field_7664 RAIL_SHAPE_TO_DIRECTION_OFFSETS Ljava/util/Map;
```
liach: :x:
not really direction offsets, but like two ends for lerping.

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
```diff
@@ -22,13 +25,24 @@ CLASS net/minecraft/class_1688 net/minecraft/entity/vehicle/AbstractMinecartEnti
 		ARG 7 z
 	METHOD method_18803 willHitBlockAt (Lnet/minecraft/class_2338;)Z
 		ARG 1 pos
+	METHOD method_22864 getDirectionOffsetsForRailShape (Lnet/minecraft/class_2768;)Lcom/mojang/datafixers/util/Pair;
```
liach: Same, more like getting ends for lerping

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
```diff
@@ -22,13 +25,24 @@ CLASS net/minecraft/class_1688 net/minecraft/entity/vehicle/AbstractMinecartEnti
...
 	METHOD method_7504 getMaxOffRailSpeed ()D
+	METHOD method_7505 snapPositionToRailWithOffset (DDDD)Lnet/minecraft/class_243;
```
liach, line 30: Just calculates the visual position of the minecart on a rail, not really snapping.
Also is this offset y or x or z offset?

mappings/net/minecraft/util/DismountVehicleUtils.mapping
```diff
@@ -0,0 +1,4 @@
+CLASS net/minecraft/class_5275 net/minecraft/util/DismountVehicleUtils
```
liach: `entity/Dismounting` imo

mappings/net/minecraft/world/World.mapping
```diff
@@ -64,6 +64,15 @@ CLASS net/minecraft/class_1937 net/minecraft/world/World
 		ARG 1 pos
 		ARG 2 entity
 		ARG 3 direction
+	METHOD method_26096 getCeilingLevel (Lnet/minecraft/class_2338;D)D
```
liach:
```suggestion
	METHOD method_26096 getCeilingHeight (Lnet/minecraft/class_2338;D)D
```
We usually call these `height` or `y` instead

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
immibis: Any other suggested name? `RAIL_SHAPE_TO_END_COORDINATES`? `RAIL_SHAPE_TO_DIRECTION_VECTORS`?

liach: `LERP_ENDS` or `LERP_OFFSETS` imo, interpolation ends (xyz offsets) are like unique to each rail direction

liach: Also what do you think of my analysis

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
immibis: It's an offset along the rail. It's unspecified which way is + and which way is -. The renderer calls it with both +0.3 and -0.3 so it doesn't matter which way is which.

It applies an offset along the rail direction and then calls `snapPositionToRail`. The returned position is always snapped.

mappings/net/minecraft/util/DismountVehicleUtils.mapping
immibis: `entity/vehicle/Dismounting`?

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
liach: how is the offset used within this method?

immibis: The offset is how far along the rail's direction vector the coordinates should be adjusted by.

The function is basically: `return snapPositionToRail(xyz + offset*getRailDirection(getBlockAt(xyz)))`

In the rendering it used to compute the on-rail position of each end of the cart separately, and then the cart is drawn using those two points, which is supposed to let it turn smoothly around corners instead of suddenly changing its rotation when it passes over a block boundary.

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
immibis: I don't think lerping is a good name. That's an algorithm which uses these, not what they mean.

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
liach: That is not an offset! That is a progress from 0 to 1 indicating what percent of the track should the cart be on.

immibis: It's not 0 to 1; it can be negative. ~~I think this implementation only works for numbers between -0.5 and 0.5.~~ (those numbers are wrong; I'm not sure what the exact limit is)

immibis: ~~Let me re-check this code to make sure that's actually right...~~ actually I know that since the argument values are -0.3 and 0.3. They don't depend on the cart position.

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
liach: That is exactly lerping logic, i.e. interpolation logic.

mappings/net/minecraft/util/DismountVehicleUtils.mapping
liach: Does this apply to dismounting from any entity or just vehicles

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
liach: They are used to get the turn progress of the minecart. Think 0.3 is like the distance in which the cart start to show turn before entering a curvature.

immibis: Using the names in Enigma with Procyon:

* First, integer10/11/12 are the minecart's integer coordinates.
* For some reason if the minecart is on a slope then its position is in the block above the slope, so correct for that.
* If the minecart isn't on a rail, return null.
* Reset the Y coordinate to an integer. I'm not sure this needs to be fully analyzed since the purpose of the function is clear from what it does with the x/z coordinates.
* Get the vector from one end of the rail block to the other and normalize it. This is double18/20.
* Then offset the passed in x/z by double18/20 * offset. i.e. move it that far along the rail's direction vector. <- this is the only step where the offset is used. Before this, the x/z are still the minecart's x/z. **<- this is the part where this function actually does something useful** (at least something that `snapPositionToRail` doesn't already do)
* Adjust the y somehow for slopes
* ~~Snap this position to~~ Find the nearest valid rail position.

mappings/net/minecraft/entity/passive/HorseBaseEntity.mapping
Prospector: I agree, only a passenger would be dismounting.

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
immibis: Can you describe minecart lerping?

mappings/net/minecraft/util/DismountVehicleUtils.mapping
Prospector: Either works for me

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
liach: It is just used for interpolating a minecart's position on a certain type of rail shape, simple as that

immibis: Then maybe "minecart position lerping" would be more specific

liach: your "minecart" and "position" are both useless words that can be inferred from context. maybe `SNAPPING_INTERPOLATIONS` if you really like your `snapToRail` name

immibis: As I see it the main purpose is to define which direction the minecart moves on each rail shape. `SNAPPING_INTERPOLATIONS` does not convey that information - nor does the word "lerping". `snapToRail` can be renamed to `getClosestPositionOnRail` if you don't like the verb `snap`.

liach: it is the direction of the rail shape as defined by two points in 3d-space.

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
liach: This name should be fine.

liach (approved):

mappings/net/minecraft/entity/passive/HorseBaseEntity.mapping
```diff
@@ -21,6 +21,8 @@ CLASS net/minecraft/class_1496 net/minecraft/entity/passive/HorseBaseEntity
 	FIELD field_6975 soundTicks I
 	FIELD field_6976 jumpStrength F
 	METHOD method_26899 createBaseHorseAttributes ()Lnet/minecraft/class_5132$class_5133;
+	METHOD method_27930 findDismountLocationInternal (Lnet/minecraft/class_243;Lnet/minecraft/class_1309;)Lnet/minecraft/class_243;
```
YanisBft:
```suggestion
	METHOD method_27930 findInternalDismountLocation (Lnet/minecraft/class_243;Lnet/minecraft/class_1309;)Lnet/minecraft/class_243;
```

liach: more like `internalFindDismountLocation` as the find action is internal than the location

immibis:
This needs to be updated to 20w20b now, right?

liach:
Yes, change the branch to merge into.

liach:
Superseded by #1555

PR #1351, opened 2020-05-13
20w20a stuff

Shnupbups:
new criteria, gamemode switcher screen, and datafixes

### DIFF
--- a/mappings/com/mojang/realmsclient/client/Errable.mapping
+++ b/mappings/com/mojang/realmsclient/client/Errable.mapping
@@ -1,3 +1,3 @@
-CLASS net/minecraft/class_5221
+CLASS net/minecraft/class_5221 com/mojang/realmsclient/client/Errable
 	METHOD method_21067 error (Lnet/minecraft/class_2561;)V
 	METHOD method_27453 error (Ljava/lang/String;)V
--- a/mappings/net/minecraft/advancement/criterion/ItemUsedOnBlockCriterion.mapping
+++ b/mappings/net/minecraft/advancement/criterion/ItemUsedOnBlockCriterion.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_4711 net/minecraft/advancement/criterion/BlockUsedCriterion
+CLASS net/minecraft/class_4711 net/minecraft/advancement/criterion/ItemUsedOnBlockCriterion
 	FIELD field_21576 id Lnet/minecraft/class_2960;
 	METHOD method_23888 (Lnet/minecraft/class_2680;Lnet/minecraft/class_3222;Lnet/minecraft/class_2338;Lnet/minecraft/class_1799;Lnet/minecraft/class_4711$class_4712;)Z
 		ARG 4 conditions
@@ -8,8 +8,13 @@ CLASS net/minecraft/class_4711 net/minecraft/advancement/criterion/BlockUsedCrit
 		ARG 3 stack
 	CLASS class_4712 Conditions
 		FIELD field_21579 item Lnet/minecraft/class_2073;
+		FIELD field_24495 location Lnet/minecraft/class_2090;
+		METHOD <init> (Lnet/minecraft/class_2048$class_5258;Lnet/minecraft/class_2090;Lnet/minecraft/class_2073;)V
+			ARG 2 location
+			ARG 3 item
 		METHOD method_23892 test (Lnet/minecraft/class_2680;Lnet/minecraft/class_3218;Lnet/minecraft/class_2338;Lnet/minecraft/class_1799;)Z
 			ARG 1 state
 			ARG 2 world
 			ARG 3 pos
 			ARG 4 stack
+		METHOD method_27981 create (Lnet/minecraft/class_2090$class_2091;Lnet/minecraft/class_2073$class_2074;)Lnet/minecraft/class_4711$class_4712;
--- a/mappings/net/minecraft/advancement/criterion/ItemUsedOnBlockCriterion.mapping
+++ b/mappings/net/minecraft/advancement/criterion/ItemUsedOnBlockCriterion.mapping
@@ -1,5 +1,5 @@
 CLASS net/minecraft/class_4711 net/minecraft/advancement/criterion/ItemUsedOnBlockCriterion
-	FIELD field_21576 id Lnet/minecraft/class_2960;
+	FIELD field_21576 ID Lnet/minecraft/class_2960;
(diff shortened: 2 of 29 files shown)


mappings/com/mojang/realmsclient/client/Errable.mapping
```diff
@@ -1,3 +1,3 @@
-CLASS net/minecraft/class_5221
+CLASS net/minecraft/class_5221 com/mojang/realmsclient/client/Errable
```
dblsaiko, line 1: Not sure that's what that actually is. I have ErrorReceiver in #1353 which is also bad though.

dblsaiko (approved):

liach (approved):
Needs to change branch to 20w12b in order to have this merged.

mappings/net/minecraft/advancement/criterion/ItemUsedOnBlockCriterion.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_4711 net/minecraft/advancement/criterion/BlockUsedCriterion
+CLASS net/minecraft/class_4711 net/minecraft/advancement/criterion/ItemUsedOnBlockCriterion
 	FIELD field_21576 id Lnet/minecraft/class_2960;
```
YanisBft:
```suggestion
	FIELD field_21576 ID Lnet/minecraft/class_2960;
```
It's a static final

mappings/net/minecraft/advancement/criterion/PlayerGeneratesContainerLootCriterion.mapping
```diff
@@ -0,0 +1,14 @@
...
+		ARG 2 id
+	CLASS class_5283 Conditions
+		FIELD field_24503 lootTableId Lnet/minecraft/class_2960;
+		METHOD <init> (Lnet/minecraft/class_2048$class_5258;Lnet/minecraft/class_2960;)V
+			ARG 1 entity
+			ARG 2 lootTableId
+		METHOD method_27995 create (Lnet/minecraft/class_2960;)Lnet/minecraft/class_5282$class_5283;
+			ARG 0 lootTableId
+		METHOD method_27996 test (Lnet/minecraft/class_2960;)Z
+			ARG 1 lootTableId
```
YanisBft: `lootTable` instead of `lootTableId` to match the string

mappings/net/minecraft/client/gui/screen/ingame/GameModeSwitcherScreen.mapping
```diff
@@ -0,0 +1,22 @@
...
+		FIELD field_24581 name Lnet/minecraft/class_2561;
+		FIELD field_24582 command Ljava/lang/String;
+		FIELD field_24583 icon Lnet/minecraft/class_1799;
+		METHOD <init> (Ljava/lang/String;ILnet/minecraft/class_2561;Ljava/lang/String;Lnet/minecraft/class_1799;)V
+			ARG 3 name
+			ARG 4 command
+			ARG 5 icon
+		METHOD method_28070 getName ()Lnet/minecraft/class_2561;
+		METHOD method_28074 renderIcon (Lnet/minecraft/class_918;II)V
+			ARG 1 itemRenderer
```
YanisBft, line 32: Mind name the two other parameters? Seems to be x and y

mappings/net/minecraft/advancement/criterion/PlayerGeneratesContainerLootCriterion.mapping
```diff
@@ -0,0 +1,14 @@
+CLASS net/minecraft/class_5282 net/minecraft/advancement/criterion/PlayerGeneratesContainerLootCriterion
+	FIELD field_24502 ID Lnet/minecraft/class_2960;
+	METHOD method_27993 test (Lnet/minecraft/class_3222;Lnet/minecraft/class_2960;)V
+		ARG 1 player
+		ARG 2 id
+	CLASS class_5283 Conditions
+		FIELD field_24503 lootTableId Lnet/minecraft/class_2960;
```
YanisBft:
```suggestion
		FIELD field_24503 lootTable Lnet/minecraft/class_2960;
```

mappings/com/mojang/realmsclient/client/Errable.mapping
YanisBft: `Errable` doesn't look like anything to me

mappings/net/minecraft/client/gui/screen/ingame/GameModeSwitcherScreen.mapping
```diff
@@ -0,0 +1,53 @@
...
+	METHOD method_28063 drawText (Lnet/minecraft/class_4587;Ljava/lang/String;Ljava/lang/String;II)V
+		ARG 1 matrices
+		ARG 2 pressF4
+		ARG 3 selectNext
+		ARG 4 y
+		ARG 5 x
+	METHOD method_28064 switchGameMode (Lnet/minecraft/class_310;Ljava/util/Optional;)V
+		ARG 0 client
+		ARG 1 gameMode
+	METHOD method_28068 onClosed ()V
```
YanisBft:
```suggestion
	METHOD method_28068 onClose ()V
```

Shnupbups:
`Errable` is an interface that has two methods, both named `error` (one takes a `Text`, and the other takes a `String` and has a default implementation of converting the string into `LiteralText` and calling the other).
It is client side only and only implemented in realms based classes, hence where I put it.
'Errable' is simply a word that means something can error. I figured it was fitting.

YanisBft (approved):
Okay, just never heard of this word

mappings/com/mojang/realmsclient/client/Errable.mapping
Shnupbups: Errable means able to error.

liach, line 1: Are you sure that when the error methods are called, the object itself errors with the given message than listen to error from other things?

dblsaiko: Yes, that's exactly what I was thinking about that name. In one of my enigma branches I have a `Validatable` class because I added a validation framework, which does something very similar as far this as I can see, but it has the slight difference in meaning that it doesn't create the errors itself. That might actually be a good name for this interface. Or `ErrorReceiver` as in my PR but I still think that name is also kinda meh, definitely better than `Errable` though, imo.

liach: In fact `Errable` is the right name, as this object has the error when the `error` method is called. Imo `error` should be like `errored` instead

Juuxel (approved):

PR #1356, opened 2020-05-14
Name color channel helper methods in NativeImage

kb-1000:


liach:
maybe just extractXxx than extractXxxChannel? that seems confusing when i see writeToChannel

falseresync:
maybe `getRed`, `writeToChannel` & `asInt`?

liach:
I agree with `getRed` and `writeToChannel`. `asInt` is too generic and isn't very indicative on how 4 ints are converted to one; `packColor` would be better, and the `Int` suffix is not necessary imo.

YanisBft:
Closing this as this class is mapped as of f80e52b. Please reopen and update this PR if you want to backport the mappings in 1.15.2.

PR #1363, opened 2020-05-16
onBlockRemoved -> onStateReplaced

falseresync:
Currently `WorldChunk.setBlockState()`:
1. Sets the state
2. If new state equals old state does nothing
3. Otherwise it updates lighting, heightmaps and calls onBlockRemoved
4. If the block at said position equals block that is being set, then it calls onBlockAdded

That's ignoring BlockEntity refreshing code
`onBlockRemoved` is being called on any state replacement, not only when the block is removed (replaced with air). Vanilla implementations explicitly check for removal

Other suggested name is `onStateRemoved`, but that also implies that state is being replaced with air

### DIFF
--- a/mappings/net/minecraft/block/AbstractBlock.mapping
+++ b/mappings/net/minecraft/block/AbstractBlock.mapping
@@ -92,7 +92,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		ARG 1 state
 		ARG 2 pos
 	METHOD method_9536 onStateReplaced (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Z)V
-		COMMENT Check if block is removed and perform cleaning operations here. Drop inventories here. Don't forget to call parent when you're done
+		COMMENT Called in setBlockState() if newState is different from state. Vanilla blocks perform removal cleanups here.
 		ARG 1 state
 		ARG 2 world
 		ARG 3 pos
--- a/mappings/net/minecraft/block/AbstractBlock.mapping
+++ b/mappings/net/minecraft/block/AbstractBlock.mapping
@@ -91,7 +91,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 	METHOD method_9535 getRenderingSeed (Lnet/minecraft/class_2680;Lnet/minecraft/class_2338;)J
 		ARG 1 state
 		ARG 2 pos
-	METHOD method_9536 onBlockRemoved (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Z)V
+	METHOD method_9536 onStateReplaced (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Z)V
 		ARG 1 state
 		ARG 2 world
 		ARG 3 pos
@@ -434,7 +434,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		METHOD method_26196 createScreenHandlerFactory (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;)Lnet/minecraft/class_3908;
 			ARG 1 world
 			ARG 2 pos
-		METHOD method_26197 onBlockRemoved (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Z)V
+		METHOD method_26197 onStateReplaced (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Z)V
 			ARG 1 world
 			ARG 2 pos
 			ARG 3 state
--- a/mappings/net/minecraft/block/AbstractBlock.mapping
+++ b/mappings/net/minecraft/block/AbstractBlock.mapping
@@ -92,7 +92,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		ARG 1 state
 		ARG 2 pos
 	METHOD method_9536 onStateReplaced (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Z)V
-		COMMENT Called in {@link World#setBlockState()} if {@code newState} is different from {@code state}. Vanilla blocks perform removal cleanups here.
+		COMMENT Called in {@link net.minecraft.world.chunk.WorldChunk#setBlockState(BlockPos, BlockState, boolean)} if {@code newState} is different from {@code state}. Vanilla blocks perform removal cleanups here.
 		ARG 1 state
 		ARG 2 world
 		ARG 3 pos
--- a/mappings/net/minecraft/block/AbstractBlock.mapping
+++ b/mappings/net/minecraft/block/AbstractBlock.mapping
@@ -92,6 +92,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		ARG 1 state
 		ARG 2 pos
 	METHOD method_9536 onStateReplaced (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Z)V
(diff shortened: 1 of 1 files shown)


YanisBft (approved):

natanfudge:
I think `onStateReplaced` is not very descriptive, it should either be `onBlockStateReplaced`  or have some javadocs explaining what it does.

falseresync:
> I think `onStateReplaced` is not very descriptive, it should either be `onBlockStateReplaced` or have some javadocs explaining what it does.

It is only inside BlockState and Block, so that is implied

Juuxel:
I agree with Fudge about the javadocs, they could be useful here.

natanfudge:
This is not the same thing as `getInventoryStack` where inventory and stack are separate words and inventory is inferred.
`BlockState` is a term and ‘block’, ‘state’ , and even ‘a block’s state’ mean different things. There is no such thing as an ‘inventoryStack’. 
You don’t automatically prepend the name of the class to every method to make it create new terms.

natanfudge:
Like, if you had a class named `AP` (short for annotation processor) you can’t have a method called `ply` and say the `ap` is inferred and the developer is supposed to understand that means `apply`. 
In this case including the full name may be too long for fabric’s standards so you could compromise to just explain the true meaning of the method in the javadoc.

falseresync:
> This is not the same thing as getInventoryStack where inventory and stack are separate words and inventory is inferred.
BlockState is a term and ‘block’ and ‘state’ mean different things. There is no such thing as an ‘inventoryStack’.
> You don’t automatically prepend the name of the class to every method to make it create new terms.

It is still implied because there’s no other type of state in block

Juuxel:
Using just `state` on its own in `Block` methods implies block states, though - see `getPlacementState`, `getStateForNeighborUpdate` etc.

natanfudge:
> Using just state on its own in Block methods implies block states, though - see getPlacementState, getStateForNeighborUpdate etc.
 
This is just because yarn insists on cutting names short, not because it makes sense to a developer

falseresync:
> This is just because yarn insists on cutting name short, not because it makes sense to a developer

If you can suggest a situation where some state that is not a BlockState changes in a block, I will change my opinion

Juuxel (approved):

natanfudge:
I can’t do that, nor can I give you a situation in which the BlockState changes in a block, because `Block`s don’t have state.

If by “the state changes in a block” you mean “a “”block””, in the sense a normal Minecraft user would mean, like something that exists multiple of around the world, changes in some way through time”, it could also change because of a blockentity being changed or replaced, not specifically because of what we call a “blockstate”.

I think the root cause of the confusion between “a block’s state” and “a `BlockState`” is that `Block`s aren’t really “”blocks“” in the way people naturally refer to them, but rather ‘block types’, and `BlockState` aren’t really a “”block””’s state, but just an instance of a ‘block type’ in the world.

dblsaiko (approved):

Virtuoel (approved):

liach (approved):
Rename is great. The javadoc had better specifiy on which condition the state replacement will drop stuff

Prospector (approved):
Assuming all your explanations are accurate, the rename looks good

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -91,7 +91,8 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 	METHOD method_9535 getRenderingSeed (Lnet/minecraft/class_2680;Lnet/minecraft/class_2338;)J
 		ARG 1 state
 		ARG 2 pos
-	METHOD method_9536 onBlockRemoved (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Z)V
+	METHOD method_9536 onStateReplaced (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Z)V
+		COMMENT Check if block is removed and perform cleaning operations here. Drop inventories here. Don't forget to call parent when you're done
```
Juuxel:
```suggestion
		COMMENT Check if block is removed and perform cleaning operations here. Drop inventories here. Don't forget to call super when you're done.
```

YanisBft (approved):

modmuss50 (approved):

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -91,7 +91,8 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 	METHOD method_9535 getRenderingSeed (Lnet/minecraft/class_2680;Lnet/minecraft/class_2338;)J
 		ARG 1 state
 		ARG 2 pos
-	METHOD method_9536 onBlockRemoved (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Z)V
+	METHOD method_9536 onStateReplaced (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Z)V
+		COMMENT Check if block is removed and perform cleaning operations here. Drop inventories here. Don't forget to call super when you're done.
```
modmuss50: Could possibly improve this doc tbh, its not a turtorial.

modmuss50 (changes requested):
This doesnt really read like a javadoc imo, its more of a comment.

mappings/net/minecraft/block/AbstractBlock.mapping
dblsaiko: It does say `COMMENT` and not `JAVADOC` ;)

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -91,7 +91,8 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 	METHOD method_9535 getRenderingSeed (Lnet/minecraft/class_2680;Lnet/minecraft/class_2338;)J
 		ARG 1 state
 		ARG 2 pos
-	METHOD method_9536 onBlockRemoved (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Z)V
+	METHOD method_9536 onStateReplaced (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Z)V
+		COMMENT Called in setBlockState() if newState is different from state. Vanilla blocks perform removal cleanups here.
```
```suggestion
		COMMENT Called in {@link World#setBlockState()} if {@code newState} is different from {@code state}. Vanilla blocks perform removal cleanups here.
```

Yanis48:
```suggestion
		COMMENT Called in {@link World#setBlockState()} if the new state is different from the current state. Vanilla blocks perform removal cleanups here.
```

YanisBft (approved):

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -91,7 +91,8 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 	METHOD method_9535 getRenderingSeed (Lnet/minecraft/class_2680;Lnet/minecraft/class_2338;)J
 		ARG 1 state
 		ARG 2 pos
-	METHOD method_9536 onBlockRemoved (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Z)V
+	METHOD method_9536 onStateReplaced (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Z)V
+		COMMENT Called in {@link World#setBlockState()} if {@code newState} is different from {@code state}. Vanilla blocks perform removal cleanups here.
```
```suggestion
		COMMENT Called in {@link net.minecraft.world.chunk.WorldChunk#setBlockState(BlockPos, BlockState, boolean)} if {@code newState} is different from {@code state}. Vanilla blocks perform removal cleanups here.
```
Sorry, my bad! https://travis-ci.com/github/FabricMC/yarn/jobs/336377975#L365
Those classes should've been imported in block class' method descriptors, so guess don't need fully qualified names

liach (approved):
Seems the javadoc link is fixed this time. Thanks for your help!

modmuss50 (approved):

modmuss50:
Looks good, thanks for this.

PR #1364, opened 2020-05-17
BasicInventory -> SimpleInventory

i509VCB:
This is because this is a simple implementation of an inventory rather than a "basic" implementation which sounds like an incomplete/barebones implementation.


Prospector (approved):

shedaniel (approved):

YanisBft (approved):

Earthcomputer (approved):

liach (commented):
Err sure, I don't think Basic stuff is required to be abstract. Simple Inventory isn't a worse name either.

modmuss50:
Removed ready to merge as this is a large impact rename, and I think it needs some more reviews.

modmuss50 (approved):

Player3324 (approved):
IMO an improvement

PR #1372, opened 2020-05-19
Various misc changes

### LINKED ISSUE #1344: WololoGoal purpleSheepPredicate is named wrongly
The `purpleSheepPredicate` actually checks if the sheep is blue.

### LINKED ISSUE #1361: ItemRenderer.renderGuiItem-related names
`renderGuiItem` (`method_4026`) was made private in some 20w snapshot (probably 20w20*). There's now an exact copy of it that calls the private method, as well as some others:

- `renderGuiItem` (`method_4026`) -> `renderGuiItemInner`
- `method_27951` -> `renderGuiItem` (literally a public wrapper method)
- `method_27953` -> `renderGuiItem` (a wrapper with no entity parameter)

### LINKED ISSUE #1369: AbstractBlock[State].onStateReplaced(..., notify) -> moved
It's not a "notification" flag, it tells whether the state was moved. It's only ever passed the `moved` value from `WorldChunk.setBlockState`, which is checked from the `MOVED` (64) flag of `ModifiableWorld.setBlockState`.

Juuxel:
- `AbstractBlock(State).onStateReplaced(..., notify) -> moved`. Fixes #1369.
  - It gets its value from the `MOVED` flag of `setBlockState`.
- Mapped `ItemRenderer.renderGuiItem`-related names. Closes #1361.
- `CriteriaMerger` -> `CriterionMerger`. No issue, but a singular sounds nicer than a plural.
- `WololoGoal.purpleSheepPredicate` -> `convertibleSheepPredicate`. Closes #1344.
  - Suggested by @liach.
- `MinecraftClient.(block/item)ColorMap` -> `(block/item)Colors`
  - Changed to match the class names.


mappings/net/minecraft/client/render/item/ItemRenderer.mapping
```diff
@@ -93,7 +107,7 @@ CLASS net/minecraft/class_918 net/minecraft/client/render/item/ItemRenderer
 		ARG 2 stack
 		ARG 3 x
 		ARG 4 y
-	METHOD method_4026 renderGuiItem (Lnet/minecraft/class_1309;Lnet/minecraft/class_1799;II)V
+	METHOD method_4026 renderGuiItemInner (Lnet/minecraft/class_1309;Lnet/minecraft/class_1799;II)V
```
dblsaiko: "inner"? What's this for?

Juuxel: It follows the naming convention of `blitInner` (from 1.15.2). It's a private implementation method for `renderGuiItem`. The public one with the same signature (`method_27951`) just passes through to this method without doing anything extra, as do all the overloads.

liach: `doRenderInScreen` `innerRenderInScreen` sounds better. We tend ton use screen over gui and just call it in gui/screen than gui/screen item imo

liach (commented):
Just curious, didn't we decide to use screen over GUI?

Juuxel:
IIRC this is used for non-screen GUIs (the HUD).

mappings/net/minecraft/client/render/item/ItemRenderer.mapping
Prospector: screen is just a type of GUI that... well... takes up your whole screen

YanisBft (approved):

haykam821 (approved):

mappings/net/minecraft/client/render/item/ItemRenderer.mapping
```diff
@@ -54,6 +54,19 @@ CLASS net/minecraft/class_918 net/minecraft/client/render/item/ItemRenderer
...
 		ARG 6 vertices
+	METHOD method_27951 renderInGui (Lnet/minecraft/class_1309;Lnet/minecraft/class_1799;II)V
+		COMMENT Renders an item in a GUI with an attached entity.
+		COMMENT
+		COMMENT <p>The entity is used to calculate model overrides for the item.
+		ARG 1 entity
+		ARG 2 stack
+		ARG 3 x
+		ARG 4 y
+	METHOD method_27953 renderInGuiWithoutEntity (Lnet/minecraft/class_1799;II)V
```
shedaniel: would recommend this to be renderGuiItem and the player one with renderGuiItemWithPlayer

shedaniel: or I meant renderItemInGuiWithPlayer, depends which one you prefer.

imo this is because calling with player is a "special case" for it, and calling without an entity is not

Juuxel: I feel like the player one is the default one that should be used in GUIs, since for example `HandledScreen` does that.

shedaniel: HandledScreen is player specific then I guess, I don't feel like we should attach the player without the withPlayer thing at least, the player can be null, which makes this the same as no entity. So let the modder know by adding the withPlayer thing

Juuxel: I can run a usage search in Enigma and see which one Mojang considers the "default" one (by number of usages).

Juuxel: With the player:
search results

Without an entity:
search results

With a manually specified entity (which is the client player in all cases):
search results

liach: Does the renderer apply model overrides with the player given, or what is the player used for?

shedaniel: It is for the overrides predicate, for example, the one for `lefthanded`, I would say that the player one is mainly for handled screens because it is to detect properties for the player, so it would be essential to add `WithPlayer` at the end to inform the modder that it will use the player.

For things like the advancement tabs, they are not player dependent, you don't want to see the icon change due to the player doing some weird things.

liach: then `WithOverrides` suffix is better than `WithPlayer` suffix.

ChloeDawn: Is this distinction necessary? Can they both not just be `renderInGui`? The arguments and documentation can describe other behaviours.

ChloeDawn (changes requested):

mappings/net/minecraft/client/render/item/ItemRenderer.mapping
liach: this is, as this method does not allow item model's overrides to be checked or applied.

liach (approved):

mappings/net/minecraft/client/render/item/ItemRenderer.mapping
Juuxel: @ChloeDawn There are conflicting signatures (`ItemStack, int, int`) with two of them (the player one and the no-entity one), so they have to have different names.

liach (approved):

YanisBft (approved):

modmuss50 (approved):

PR #1374, opened 2020-05-20
Map custom worlds

haykam821:


YanisBft:
Has been entirely mapped already, except two methods which are now in #1535

PR #1353, opened 2020-05-13
20w20a mappings

dblsaiko:
By @i509VCB, @Yanis48, @TheGlitch76, @Earthcomputer, @haykam821 and me (unless I forgot someone)

Closes #1337.

PR #1377, opened 2020-05-20
20w21a mappings based on strings

Kroppeb:
Depends on #1353
Small overlap with #1374

I based these mappings on the string constants.


liach:
The fact that you base this off #1353 makes it very hard and almost pointless to review

dblsaiko:
Also looks like you forgot to commit your changes. I'd make this a draft until #1353 is merged

Kroppeb:
Oof, this was supposed to be a draft and yes, I forgot to commit, I'm an idiot.
I'm gonna split this up anyway so.

PR #1379, opened 2020-05-20
Registry mappings

Earthcomputer:


mappings/net/minecraft/util/registry/Registry.mapping
```diff
@@ -74,6 +74,11 @@ CLASS net/minecraft/class_2378 net/minecraft/util/registry/Registry
 	FIELD field_25095 DIMENSION_TYPE_KEY Lnet/minecraft/class_5321;
 	FIELD field_25096 BIOME_SOURCE Lnet/minecraft/class_2378;
 	FIELD field_25097 CHUNK_GENERATOR Lnet/minecraft/class_2378;
+	FIELD field_25098 registryKey Lnet/minecraft/class_5321;
+		COMMENT The {@code RegistryKey} representing the ID of the actual registry.
+	FIELD field_25099 lifecycle Lcom/mojang/serialization/Lifecycle;
```
liach, line 79: woo, are they starting to deprecate registries

mappings/net/minecraft/util/registry/Registry.mapping
```diff
@@ -74,6 +74,11 @@ CLASS net/minecraft/class_2378 net/minecraft/util/registry/Registry
 	FIELD field_25095 DIMENSION_TYPE_KEY Lnet/minecraft/class_5321;
 	FIELD field_25096 BIOME_SOURCE Lnet/minecraft/class_2378;
 	FIELD field_25097 CHUNK_GENERATOR Lnet/minecraft/class_2378;
+	FIELD field_25098 registryKey Lnet/minecraft/class_5321;
+		COMMENT The {@code RegistryKey} representing the ID of the actual registry.
```
YanisBft:
```suggestion
		COMMENT The {@linkplain RegistryKey} representing the ID of the actual registry.
```

haykam821 (approved):

mappings/net/minecraft/util/registry/Registry.mapping
Earthcomputer: Would that actually link it?

mappings/net/minecraft/util/registry/RegistryKey.mapping
```diff
@@ -0,0 +1,16 @@
...
+	FIELD field_25137 registryId Lnet/minecraft/class_2960;
+	FIELD field_25138 valueId Lnet/minecraft/class_2960;
+	METHOD <init> (Lnet/minecraft/class_2960;Lnet/minecraft/class_2960;)V
+		ARG 1 registryId
+		ARG 2 valueId
+	METHOD method_29177 getValueId ()Lnet/minecraft/class_2960;
+	METHOD method_29179 getOrCreate (Lnet/minecraft/class_5321;Lnet/minecraft/class_2960;)Lnet/minecraft/class_5321;
+		ARG 0 registryId
+		ARG 1 valueId
+	METHOD method_29180 getOrCreateRegistryKey (Lnet/minecraft/class_2960;)Lnet/minecraft/class_5321;
```
liach:
```suggestion
	METHOD method_29180 getOrCreate (Lnet/minecraft/class_2960;)Lnet/minecraft/class_5321;
```
The rename is probably fine as the descriptor differs

mappings/net/minecraft/util/registry/Registry.mapping
YanisBft: It should yes

liach (approved):

mappings/net/minecraft/util/registry/RegistryKey.mapping
Earthcomputer: This is specifically an overload where a registry key for Registry.ROOT is created (i.e. an ID for a registry). I'll rename it to "root key"

mappings/net/minecraft/util/registry/Registry.mapping
liach: It would. The text will just render as plain text (way less discoverable), but you can click it. Doesn't prompt people to click as much.

mappings/net/minecraft/util/registry/RegistryKey.mapping
```diff
@@ -0,0 +1,16 @@
+CLASS net/minecraft/class_5321 net/minecraft/util/registry/RegistryKey
```
shedaniel, line 1: I really want to use RegistryIdentifier instead, since at the end of the day, it is still an identifier

mappings/net/minecraft/util/registry/RegistryKey.mapping
```diff
@@ -0,0 +1,16 @@
...
+	FIELD field_25136 INSTANCES Ljava/util/Map;
+	FIELD field_25137 registryId Lnet/minecraft/class_2960;
```
shedaniel, line 3: would name this parentId imo, but up for debate

mappings/net/minecraft/util/registry/RegistryKey.mapping
Earthcomputer: It's not an identifier though, it does not extend the identifier class. It's a registry id + identifier pair.

mappings/net/minecraft/util/registry/RegistryKey.mapping
Earthcomputer: registryId is clearer

YanisBft (approved):

mappings/net/minecraft/util/registry/RegistryKey.mapping
shedaniel: good point, however it is used for finding the entry, it is almost like it is an identifier for an entry. I am happy with RegistryKey tho. just concerned about the consistency of it since identifier is technically a key as well if you put it in the same context

Earthcomputer: Yeah, tbh this class is a little confusing, as the registry id part of it doesn't seem to be used anywhere. However it's useful to have this distinction so we can have `containsId` vs `containsKey` and the likes.

mappings/net/minecraft/util/registry/RegistryKey.mapping
shedaniel: you can have a registry in both the key and the value id, it is more about which one is the parent imo

dblsaiko (approved):

mappings/net/minecraft/util/registry/RegistryKey.mapping
shedaniel: that sounds fair enough

liach: This is more analogous to that object which is a compound of two identifier for locating an atlas and then a sprite in the atlas. I forgot what that class is named, might be a good idea to make two have similar names.

mappings/net/minecraft/util/registry/RegistryKey.mapping
Earthcomputer: What isn't clear to me though with that is that how do we know "parent" refers to a registry?

shedaniel: parentRegistryId? sounds long and weird tho

but you don't really have to know if that's a registry, because it is just defining a parent for that entry imo

liach: No, just `registryId`

liach: Or even just `registry` as this is just a pojo and holds no references.

modmuss50 (approved):

modmuss50:
thanks

PR #1385, opened 2020-05-20
Added names for fields based on strings in class_5315

Kroppeb:
Seems to store level metadata?

### DIFF
--- a/mappings/net/minecraft/class_5315.mapping
+++ b/mappings/net/minecraft/class_5315.mapping
@@ -1,0 +1,18 @@
+CLASS net/minecraft/class_5315
+	FIELD field_25024 version I
+	FIELD field_25025 lastPlayed J
+	FIELD field_25026 versionName Ljava/lang/String;
+	FIELD field_25027 worldVersion I
+	FIELD field_25028 stable Z
+	METHOD <init> (IJLjava/lang/String;IZ)V
+		ARG 1 version
+		ARG 2 lastPlayed
+		ARG 4 versionName
+		ARG 5 worldVersion
+		ARG 6 stable
+	METHOD method_29022 getVersion ()I
+	METHOD method_29023 fromDynamic (Lcom/mojang/serialization/Dynamic;)Lnet/minecraft/class_5315;
+	METHOD method_29024 getLastPlayed ()J
+	METHOD method_29025 getVersionName ()Ljava/lang/String;
+	METHOD method_29026 getWorldVersion ()I
+	METHOD method_29027 isStable ()Z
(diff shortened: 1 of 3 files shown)


mappings/net/minecraft/class_5315.mapping
```diff
@@ -0,0 +1,18 @@
...
+		ARG 1 version
+		ARG 2 lastPlayed
+		ARG 4 versionName
+		ARG 5 worldVersion
+		ARG 6 stable
+	METHOD method_29022 getVersion ()I
+	METHOD method_29023 fromDynamic (Lcom/mojang/serialization/Dynamic;)Lnet/minecraft/class_5315;
+	METHOD method_29024 getLastPlayed ()J
+	METHOD method_29025 getVersionName ()Ljava/lang/String;
+	METHOD method_29026 getWorldVersion ()I
```
liach: should we prefer "data version" or "world version" name?

mappings/net/minecraft/class_5315.mapping
```diff
@@ -0,0 +1,18 @@
+CLASS net/minecraft/class_5315
+	FIELD field_25024 version I
```
liach: Might specify this is the version of level.dat

mappings/net/minecraft/class_5315.mapping
```diff
@@ -0,0 +1,18 @@
+CLASS net/minecraft/class_5315
```
liach: This is like `SaveVersionInfo` imo

mappings/net/minecraft/class_5315.mapping
Kroppeb: What do you mean exactly?

mappings/net/minecraft/class_5315.mapping
i509VCB: Hmm, what is it called in `MinecraftVersion`?

mappings/net/minecraft/class_5315.mapping
liach: this is the level.dat format version, not the data version of the save represented by the level.dat

mappings/net/minecraft/class_5315.mapping
liach: `world_version` in `version.json`, but it is called `DataVersion` in like persistent data, level.dat, etc, and the community uses "data version" as a convention already. Imo the `world_version` is not accurate, as it should rather be "save version" if we follow this logic.

liach:
Mind edit this pr to merge into 1.16-pre2

liach (approved):

modmuss50 (approved):

YanisBft (approved):

PR #1388, opened 2020-05-20
Named the Codecs where possible

Kroppeb:
Only in files where there is a single static Codec.
Also named 2 (I think) loggers to "LOGGER". should have kept the separate maybe but /shrug

(only 74 files)

### DIFF
--- a/mappings/net/minecraft/class_5299.mapping
+++ b/mappings/net/minecraft/class_5299.mapping
@@ -1,0 +1,3 @@
+CLASS net/minecraft/class_5299
+	CLASS class_5300
+		FIELD field_24648 CODEC Lcom/mojang/serialization/Codec;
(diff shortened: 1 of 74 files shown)


Juuxel (approved):

liach (approved):

Earthcomputer:
Low quality >:(
image

image

Earthcomputer (approved):

Kroppeb:
Yeah I just searched on the builder `create` method, I'll do more tomorrow.

liach:
fyi earth: some of those are not static, like the codec for brain is created per entity with specific set of memory/sensor

YanisBft (approved):

Earthcomputer:
@liach I searched for the static ones, as you can see in the search pattern.

modmuss50 (approved):

PR #1389, opened 2020-05-21
HandledScreen parameters

Juuxel:
- Mapped a lot of parameters in `HandledScreen`
  - Fixed `mouseX` being called `mouseY` in `drawBackground`

### DIFF
--- a/mappings/net/minecraft/client/gui/screen/ingame/HandledScreen.mapping
+++ b/mappings/net/minecraft/client/gui/screen/ingame/HandledScreen.mapping
@@ -41,6 +41,9 @@ CLASS net/minecraft/class_465 net/minecraft/client/gui/screen/ingame/HandledScre
 		ARG 7 pointY
 	METHOD method_2379 calculateOffset ()V
 	METHOD method_2380 drawMouseoverTooltip (Lnet/minecraft/class_4587;II)V
+		ARG 1 matrices
+		ARG 2 x
+		ARG 3 y
 	METHOD method_2381 isClickOutsideBounds (DDIII)Z
 		ARG 1 mouseX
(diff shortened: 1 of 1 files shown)


liach:
Hmm, other screen elements are still using `draw`, like `drawSlot`. Should we change or keep those

dblsaiko:
I would prefer `draw` over `render` with 2D stuff, tbh.

mappings/net/minecraft/client/gui/screen/ingame/HandledScreen.mapping
```diff
@@ -62,13 +65,21 @@ CLASS net/minecraft/class_465 net/minecraft/client/gui/screen/ingame/HandledScre
 		ARG 1 keyCode
 		ARG 2 scanCode
 	METHOD method_2385 drawSlot (Lnet/minecraft/class_4587;Lnet/minecraft/class_1735;)V
```
liach, line 67: Just need to ensure the draw vs render consistency. Why draw than render here?

Juuxel: I think I'll revert the `renderBackground/Foreground` renames for now because I don't feel like standardising it across all screens (someone should still do that, though).

YanisBft (approved):

liach (approved):

dblsaiko (approved):
I don't like `matrices` because it misses the point of what `MatrixStack` actually is but I think that's a general yarn thing, not specific to this PR.

liach:
No longer a refactor since this no longer change method names.

modmuss50 (approved):

PR #1390, opened 2020-05-21
Standardize attribute modifier mappings

haykam821:
This pull request standardizes attribute modifier fields:

- IDs: `[MODIFIER]_ID`
- Modifiers: `[MODIFIER]`

In addition, if an attribute modifier is considered a boost, it is suffixed with `_BOOST`.


Earthcomputer:
What about standardizing whether or not to have a `ATTR_` prefix as well? (I vote no, as it's prefix naming, and the suffix sounds weird).

Juuxel:
Yeah, that prefix is ugly and can be removed IMO.

liach:
Should remove `ATTR_` prefix. Don't think names will conflict after that prefix removal.

liach (approved):

dblsaiko (approved):

YanisBft (approved):

Earthcomputer (approved):

modmuss50 (approved):

PR #1394, opened 2020-05-24
Simplify living entity stack methods

ChloeDawn:


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -324,7 +324,7 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	METHOD method_5670 baseTick ()V
 	METHOD method_5671 getCommandSource ()Lnet/minecraft/class_2168;
 	METHOD method_5672 getHighSpeedSplashSound ()Lnet/minecraft/class_3414;
-	METHOD method_5673 equipStack (Lnet/minecraft/class_1304;Lnet/minecraft/class_1799;)V
+	METHOD method_5673 setStack (Lnet/minecraft/class_1304;Lnet/minecraft/class_1799;)V
```
Pyrofab, line 327: This method has some logic in it, it's not a simple setter. It's even a no-op in `Entity`, which is very counter-intuitive for a setter (setters that are not supported by a class would generally throw an `UnsupportedOperationException`).

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -317,7 +317,7 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 		ARG 5 z
 		ARG 7 particleEffects
 	METHOD method_6083 getLastAttackTime ()I
-	METHOD method_6084 hasStackEquipped (Lnet/minecraft/class_1304;)Z
+	METHOD method_6084 hasStack (Lnet/minecraft/class_1304;)Z
```
Pyrofab: `hasStack` would imply "including inventory" by default. Sure, you will quickly understand that it does not when you look at what `EquipmentSlot` is, but specifying `equipped` makes that immediately clear without being that verbose.

Pyrofab (changes requested):
I am not fond of those renames, they try too hard for brevity. Entities are (usually) not inventories nor some kind of map, so I think it is important to specify the context in which stack accesses are done.

mappings/net/minecraft/entity/LivingEntity.mapping
ChloeDawn: The arguments make it immediately clear, you don't need additional verbosity to make it clear. `entity.hasStack(EquipmentSlot.CHEST)` == "Has stack in chest slot"

mappings/net/minecraft/entity/Entity.mapping
ChloeDawn: The no-op method in Entity is just bad Mojang design from what I can tell, and shouldn't be considered when naming the methods. As for the "side-effects", all there is as far as that is concerned is a *callback* method. The callback is what has side-effects, not the setter itself. Your logic could also argue that `World#setBlockState` is not a setter because it also has side-effects.

Pyrofab: Sure, I would rather have a non-setter name for `setBlockState`. I don't have a suggestion though, so I don't complain about it. However `equipStack` is already a good name in my opinion, and I consider changing it to a nondescript setter a regression.

ChloeDawn: I would expect a method such as `equipStack` to return the previously equipped stack, but this method returns void like a typical setter.

mappings/net/minecraft/entity/LivingEntity.mapping
ChloeDawn: It's also worth noting that the "side-effects" only exist for 2 entity types, player and armor stand. For every other living entity, it is merely a standard setter.

mappings/net/minecraft/entity/Entity.mapping
ChloeDawn: It's also worth noting that the "side-effects" only exist for 2 entity types, player and armor stand. For every other living entity, it is merely a standard setter.

mappings/net/minecraft/entity/LivingEntity.mapping
ChloeDawn: A better option here may be `hasItem`, as technically there is always a stack in the slot, just that the stack can be empty.

Pyrofab: I don't think you should rely on arguments that much, especially when a call can look like `hasStack(var7)`.

ChloeDawn: Are we really mapping code based on decompiler generated garbage now?

Pyrofab: Considering what yarn is used for, I think it is important yes. Even discounting that, I think the code should stay clear when people shorten their `EquipmentSlot` variables to `slot`, although it could be argued that user code should not be taken into account.

mappings/net/minecraft/entity/Entity.mapping
Pyrofab: I was more thinking of the side effect in `AbstractSkeletonEntity`, which causes it to update its AI.

mappings/net/minecraft/entity/LivingEntity.mapping
ChloeDawn: Even if a variable were simplified to `slot`, `entity.getStack(slot)` reads perfectly fine and you are able to look at the type of `slot` if needed. This would make the method naming consistent with that of an inventory e.g `inventory.getStack(slot)` where `slot` is an `int`.

Pyrofab: I do think it is an issue to have the same method name for equipment and inventory methods. See `MinecartChestEntity`, which is both an entity and an inventory - it already has a `getStack(int)` method.

mappings/net/minecraft/entity/Entity.mapping
ChloeDawn: Again, that is merely a callback for when the stack changes as the AI needs to update how it handles the equipped item if it differs from the previously equipped item.

mappings/net/minecraft/entity/LivingEntity.mapping
ChloeDawn: I feel it is a non-issue when the argument of `EquipmentSlot` very clearly distinguishes it.

liach (approved):

Runemoro (approved):

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -185,7 +185,7 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 		ARG 1 attribute
 	METHOD method_5997 attackLivingEntity (Lnet/minecraft/class_1309;)V
 		ARG 1 target
-	METHOD method_5998 getStackInHand (Lnet/minecraft/class_1268;)Lnet/minecraft/class_1799;
+	METHOD method_5998 getStack (Lnet/minecraft/class_1268;)Lnet/minecraft/class_1799;
```
haykam821, line 188: The removal of `inHand` seems to imply that this method can get stacks from more than just the two hands. Even with the `hand` parameter, I think it's better to leave the additional context.

haykam821 (changes requested):

mappings/net/minecraft/entity/LivingEntity.mapping
ChloeDawn: How does it imply that? `Hand` is an enum of only two values, it's not like it is an ambiguous `int` index.

```java
ItemStack a = entity.getStack(Hand.MAIN_HAND);
ItemStack b = entity.getStack(Hand.OFF_HAND);
```

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -317,7 +317,7 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 		ARG 5 z
 		ARG 7 particleEffects
 	METHOD method_6083 getLastAttackTime ()I
-	METHOD method_6084 hasStackEquipped (Lnet/minecraft/class_1304;)Z
+	METHOD method_6084 hasItem (Lnet/minecraft/class_1304;)Z
```
Prospector, line 320: why `hasItem` but `getStack`? Shouldn't this be `hasStack`? This method just does `!this.getStack(slot).isEmpty()`

Prospector (changes requested):
Other than my comment, looks good.

mappings/net/minecraft/entity/LivingEntity.mapping
haykam821:
```suggestion
	METHOD method_6084 isStackEmpty (Lnet/minecraft/class_1304;)Z
```

liach: Seriously is this checking for an empty equpiment slot or a non-empty slot?

Pyrofab: Non-empty, so `isStackEmpty` is wrong.

ChloeDawn: I opted for `hasItem` over `hasStack`, as semantically speaking a slot will always have a stack. This method determines if the stack in the given slot is non-empty, i.e represents an item with a count of more than 0. If `hasItem(...)` returns true, it is guaranteed that `getStack(...).getCount()` will be more than 0 and `getStack(...).getItem()` will be a tangible item (not air).

haykam821: Oops, 
```suggestion
	METHOD method_6084 isStackFilled (Lnet/minecraft/class_1304;)Z
```

ChloeDawn: `isStackFilled` implies it would only return true if the stack was at its max count. Please think before making suggestions.

haykam821:
```suggestion
	METHOD method_6084 isStackNotEmpty (Lnet/minecraft/class_1304;)Z
```

Pyrofab: If anything, I think it should be `hasStackEquipped`, or maybe even `hasEquipmentIn`. `hasStack` and `hasItem` are both too generic in my opinion.

ChloeDawn: `isStackNotEmpty` is just an overly verbose way of saying `hasItem`

ChloeDawn: `hasEquipmentIn` implies the stack in the slot will always be a form of equipment, which is not the case. A pumpkin is not equipment, a dirt block is not equipment, etc.

ChloeDawn: `hasStackEquipped` introduces the same verbosity that this PR aims to remove. We either need to settle on `hasItem` or `hasStack` imo.

Pyrofab: I'd argue carved pumkin masks count as equipment, but I guess you're right about dirt.

Prospector: I see how hasItem is sorta more correct than hasStack, but I also think it could be confusing having one method refer to Stacks while the other refers to Items. I also don't really think it's unfair to say that if you have an empty stack then you don't really have a stack.

ChloeDawn: There is no such thing as a "carved pumpkin mask". It is just a carved pumpkin block that can be placed in the world. Essentially, any item can be set in the equipment slots, not just equipment, and that should be taken into account when naming imo.

ChloeDawn: If we agree that `hasStack` is more appropriate than `hasItem` then I'll make that change.

haykam821: I dislike `hasItem`

liach: Maybe `hasItemIn` `hasStackIn`

Runemoro (approved):

mappings/net/minecraft/entity/LivingEntity.mapping
Runemoro: I don't see what's wrong with `hasItem`. If you have a stack of items, then that implies you have an item...

haykam821 (commented):
I'm still not a fan of dropping the 'equipped' from method names, but removing 'inHand' is fine with me.

mappings/net/minecraft/entity/LivingEntity.mapping
Sollace: > 
> 
> There is no such thing as a "carved pumpkin mask". It is just a carved pumpkin block that can be placed in the world. Essentially, any item can be set in the equipment slots, not just equipment, and that should be taken into account when naming imo.

Taking that argument to the logical conclusion, it would mean renaming `EquipmentSlot` -> `ItemSlot`. It's not wrong, but I believe we should also take into account the _intended_ use of the function, and here I'd say it is to hold _equipment_.

Also since you have `equipStack` and `unequipStack`, taking equip as the performed action of placing an item into an equipment stack, I'd argue that the method to check if an item is in said equip stack (i.e. checking whether an item has been equipped) should be named `hasEquippedStack`, and anything shorter than that would be missing the necessary granularity of that name.

modmuss50:
Thanks for this PR, however it seems to be quite far behind and inactive for almost a year. If you wish to keep it going please open a new PR targetting 1.17. Thanks for understanding.

PR #1397, opened 2020-05-27
MaterialColor -> MapColor

dexman545:
Matches class name to its string.
Adds a tiny bit of javadoc.

This name is a bit more representative of what this class actually does.

It should be noted that the color names are bad/outdated in relation to the material name improvements, but this PR will not be touching those.


liach (approved):

YanisBft (approved):
You should just specify you talk about map items

dexman545:
But I do? Unless there's some other map in MC I missed.

liach:
People occasionally confuse it with `java.util.Map` `java.util.Optional.map` `java.util.stream.Stream.map` etc

YanisBft:
I was more thinking of world/level sometimes referenced as maps, but liach has a point too. To specify it's the item, you can just link "map" to the MapItem class

mappings/net/minecraft/block/MapColor.mapping
```diff
@@ -1,4 +1,5 @@
-CLASS net/minecraft/class_3620 net/minecraft/block/MaterialColor
+CLASS net/minecraft/class_3620 net/minecraft/block/MapColor
+	COMMENT The color that a block will have on a map. Selected by the Material the block uses.
```
liach:
```suggestion
	COMMENT The color that a block will have on a {@link net.minecraft.item.FilledMapItem map}. Selected by the Material the block uses.
```

mappings/net/minecraft/block/MapColor.mapping
```diff
@@ -1,4 +1,5 @@
-CLASS net/minecraft/class_3620 net/minecraft/block/MaterialColor
+CLASS net/minecraft/class_3620 net/minecraft/block/MapColor
+	COMMENT The color that a block will have on a map item. Selected by the Material the block uses.
```
liach:
```suggestion
	COMMENT The color that a block will have on a {@link net.minecraft.item.FilledMapItem map item}. Selected by the Material the block uses.
```

YanisBft (approved):

mappings/net/minecraft/block/MapColor.mapping
```diff
@@ -1,4 +1,5 @@
-CLASS net/minecraft/class_3620 net/minecraft/block/MaterialColor
+CLASS net/minecraft/class_3620 net/minecraft/block/MapColor
+	COMMENT The color that a block will have on a {@link net.minecraft.item.FilledMapItem map item}. Selected by the Material the block uses.
```
ChloeDawn:
```suggestion
	COMMENT Represents the surface color of a block when rendered from the {@link net.minecraft.client.gui.MapRenderer}.
```

ChloeDawn (changes requested):

shedaniel:
shouldn't this pr find reference to MaterialColor and fix them? i.e. DyeColor#getMaterialColor

mappings/net/minecraft/block/MapColor.mapping
ChloeDawn: "Selected by the Material the block uses." is redundant and confusing. That is a property of materials, not the color. And a map color can be specified for a block that is unrelated to the material.

dexman545:
Changed the method names - there is a fapi name that needs fixing as well in AbstractBlockSettingsAccessor

i509VCB:
Should be a quick fix since we extend AbstractBlock.Settings now in fapi. Should cause no issues updating.

shedaniel (approved):

liach (approved):

Juuxel (approved):

YanisBft (approved):

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -460,7 +460,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 			ARG 2 pos
 			ARG 3 direction
 		METHOD method_26204 getBlock ()Lnet/minecraft/class_2248;
-		METHOD method_26205 getTopMaterialColor (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_3620;
+		METHOD method_26205 getTopMapColor (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_3620;
```
ChloeDawn: Why is `Top` necessary here? As described in the documentation, map colors represent the surface color of the block. This is just a blockstate sensitive getter for the map color, no?

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -38,7 +38,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 	METHOD method_26160 asBlock ()Lnet/minecraft/class_2248;
 	METHOD method_26161 hasBlockEntity ()Z
 	METHOD method_26162 getLootTableId ()Lnet/minecraft/class_2960;
-	METHOD method_26403 getDefaultMaterialColor ()Lnet/minecraft/class_3620;
+	METHOD method_26403 getDefaultMapColor ()Lnet/minecraft/class_3620;
```
ChloeDawn, line 41: What is "default" about this? Does it not just return the blocks map color as defined by the settings?

ChloeDawn (changes requested):

Prospector (approved):
At first I wasn't a huge fan because MapColor sounds sorta like the color of a map when really it's the color of the block (often provided by the material), however the fact that it is only the color from above makes it pretty map-specific. I don't love MapColor but I can't think of anything better.

Just make sure it's updated everywhere the current name is used.

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -220,10 +220,10 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		FIELD field_23186 emissiveLightingPredicate Lnet/minecraft/class_4970$class_4973;
 		METHOD <init> (Lnet/minecraft/class_3614;Ljava/util/function/Function;)V
 			ARG 1 material
-			ARG 2 materialColorFactory
+			ARG 2 mapColorFactory
```
ChloeDawn: What makes this a factory? Is it constructing new colors? If not, it is a provider.

ChloeDawn (changes requested):

mappings/net/minecraft/block/AbstractBlock.mapping
i509VCB: Might have been matcher, since one of the constructors is a `Function<BlockState, MapColor>`

mappings/net/minecraft/block/AbstractBlock.mapping
dexman545: It pulls the default blockstate for the color

liach (approved):

Pyrofab (approved):

ChloeDawn (approved):

modmuss50:
This isnt ready for merge, its a very wide spread change, that will require yet another version of object builders.

modmuss50 (changes requested):
Im not a fan of this. Just becuase its only used by the map doesnt mean that can be its only use, almost seems yet another rename for the point of renaming things.

It contains more than just colors as well, as it has fields for `DIRT` for example. Im not even sure having color in the name is 100% correct.

liach:
Maybe like `MapTexture`, the texture of a block on a map?

shedaniel:
Don't think it is tied to the material, you can have blocks that has a color that is different than the material

shedaniel:
You can really just do MapColor or BlockColor imo

liach:
After looking at the field Imo we should call it something other than a `color`. `clay` is not a color.

dexman545:
As mentioned in the OP, the field names are bad and outdated. They should be ignored for the purpose of this PR.
This class is a wrapper around a list of color values for the purpose of maps. This list is a fixed size, so mods cannot add many more to it.
This PR will not handle the field renames. Will be more work to compile, and have more debate than the materials PR. I won't be doing that either, so someone else can.
MaterialColor makes no sense as the color is dependent on the blockstate.

liach:
Shall this pr refactor, it should just refactor fields as well. It is meaningless to split these into two prs that are both refactors. We can bikeshed one bikeshed at once than bikeshed bikes and the shed separately!

liach:
Given @dexman545 refuses to refactor the material color field names that he deem broken in this pr, I hereby close this pr. It is meaningless to split such a refactor into two debatable pull requests and introduce extra bikeshedding and migration difficulties.

shedaniel:
That shouldn't be a reason to close a PR, unless you've got another PR to fix this, I don't think we should close it.

i509VCB:
Since this is a smaller PR, it may make sense to bundle everything together.

However something larger like the past nbt PRs makes no sense to go hunting for every param name whenever it may be unmerged

liach:
@dexman545 just clarified that since material colors are just certain rgb colors and they are indeed very hard to name (like clay color doesn't really have a color name), so those field renames aren't made.

dexman545:
The constructor for MaterialColor takes in 2 ints: an index and an RGB value.
Its only usage is in relation to maps and for falling block particles (ie dripping lava).

Block(State)Color works, as this is the color of the blockstate. MapColor also works as it's primary purpose is for maps. MapTexture not so much as this is the color of each "pixel," not really a texture.

As for this being a wide-reaching change, I'm going to have to disagree. The color a mod's block will have on a map is, in most cases, an oversight by mod devs, as the material will provide a default template color for you, and this is the type of thing to get ignored. Also, **there are also only 12 available color slots left.** The values in this list are immutable, and its size is forced by the constructor (where the MapColor name comes from). Any mods that are adding colors to this are either a handful in number, or have changed the vanilla behavior.  Otherwise, we'd see tons of mod conflicts if this were common. As for mods that *make* a Material, they are probably using one of the pre-determined colors. In that case, migrateMappings will suffice.

This is what I meant when I said it's just a wrapper for a list: all it has are the RGB color values with named variables to easily pull them without nonsensical int values.

As for renaming the fields, I want that in a separate PR as it will be bikeshed to hell. Everyone sees color differently, and there are several very close colors in this list. The state of the current field names is rather poor. Some of them are named using the material that uses them (the old material names, or strange abbreviations of the material name), the color it actually is, or the block that has the materialcolor. 

Want to name them based on the color? No one will agree on the names then. Want to rename them based on the blocks that use them? That leads to issues with things like logs that have two different material colors for verticle and horizontal states. Hex values? Sure, we can all agree on each individual name, but as a whole, they then lose their functionality, especially considering you won't get a nifty color preview to know what it looks like.

There simply isn't a *good* way to rename them.

The only good field rename would be AIR -> CLEAR/INVISIBLE as it includes glass etc., and is not rendered on the map.

I can update the ones that were named after the material if you want, but that doesn't fix the others.

Here are all the color's hex values, and setup so idea will show you the color on the side: https://pastebin.com/nTXgt3Dr

As an aside: maybe fapi should have something to open up the colors, but again, as no one seems to be adding colors it's probably unneeded. Not to mention, the constructor is private anyways so I doubt anyone has gone through the trouble of opening it just to add a color.

Prospector (commented):
I personally am very neutral on this. I find MaterialColor fine, and MapColor is fine as well. I'd probably rather just not bother with this rename because it does mean some Fabric API methods will be wrongly named, and I don't think this PR is an improvement, really. The idea that a map can just display a material color isn't absurd or anything. The specificity seems unnecessary.

Edit: uh, is there no way to take back an approval? I kinda thought this would do that. I guess just pretend I don't approve of this anymore.

Edit 2: fixed lol

dexman545:
`MaterialColor` is a bad name for several reasons: 
 - just because Material has a field that uses this doesn't mean this should be named as such (we don't have MaterialPistonBehavior). Whenever this property is queried, it is always via the blockstate, as this is a property that varies based on blockstate, not to mention blocks have entire MaterialColors dedicated to them, and logs have a special helper method to give them two MaterialColors.
 - `MaterialColor` means nothing at all. What does that class do? Materials define basic properties for blocks, not their color in any meaningful sense. No other material property is so commonly overridden, or has so many predefined properties used for individual blocks, *not* materials.
 - why do blocks get not only a Material, but a MaterialColor? Perhaps because MaterialColor has no meaningful relation to Material
 - there are more MaterialColors than there are Materials, which is a pretty good indicator that they aren't related in the way the name implies.

This rename will actually clarify what this class is doing, and what the colors actually mean. 

As for object builders/fapi block settings, can't the pertinent methods just be deprecated, and properly named ones be added? Most people seemed fine with the rename aside from the fields.

Runemoro (approved):

YanisBft:
@dexman545 could you resolve the conflict please?

dexman545:
@Yanis48 fixed

YanisBft (approved):
Thanks :)

dexman545:
Can the outdated tag be removed? This is targeting the 1.17 snapshot.

modmuss50 (approved):

Issue #987, opened 2019-11-26
Drop 'is' prefixes for boolean fields


Runemoro:
A few boolean fields still use the "is" prefix, which doesn't respect the conventions.

natanfudge:
"Bug"

Runemoro:
> "Bug"

Hm? Not respecting the conventions is a bug the same way using a name using British spelling or making a typo is.

liach:
Most of the time the `is` prefix is extraneous (mainly for adjective phrases)

YanisBft:
If anyone wants to help, here's a very quick shell script I made to find all fields and parameters starting with "is": https://drive.google.com/open?id=1Ab3AVMFI9PaGeknCo-6C3gUKW8_xnpPY

PR #1403, opened 2020-05-29
Removed 'is' prefix from field/param names as per #987

dexman545:
Kept in some places where it doesn't really fit to remove them:
	World's isClient
	NativeImage's isSbtImage
	PageTurnWidget's isNextPageButton

I believe that's all of them.
In reference to #987

### DIFF
--- a/mappings/net/minecraft/block/AbstractBlock.mapping
+++ b/mappings/net/minecraft/block/AbstractBlock.mapping
@@ -304,7 +304,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		FIELD field_23166 shapeCache Lnet/minecraft/class_4970$class_4971$class_3752;
 		FIELD field_23167 luminance I
 		FIELD field_23168 hasSidedTransparency Z
-		FIELD field_23169 air Z
+		FIELD field_23169 isAir Z
 		FIELD field_23170 material Lnet/minecraft/class_3614;
 		FIELD field_23171 materialColor Lnet/minecraft/class_3620;
 		FIELD field_23172 hardness F
--- a/mappings/net/minecraft/block/AbstractBlock.mapping
+++ b/mappings/net/minecraft/block/AbstractBlock.mapping
@@ -211,7 +211,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		FIELD field_20721 opaque Z
 		FIELD field_21209 jumpVelocityMultiplier F
 		FIELD field_23179 velocityMultiplier F
-		FIELD field_23180 isAir Z
+		FIELD field_23180 air Z
 		FIELD field_23181 allowsSpawningPredicate Lnet/minecraft/class_4970$class_4972;
 		FIELD field_23182 solidBlockPredicate Lnet/minecraft/class_4970$class_4973;
 		FIELD field_23183 suffocationPredicate Lnet/minecraft/class_4970$class_4973;
@@ -304,7 +304,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		FIELD field_23166 shapeCache Lnet/minecraft/class_4970$class_4971$class_3752;
 		FIELD field_23167 luminance I
 		FIELD field_23168 hasSidedTransparency Z
-		FIELD field_23169 isAir Z
+		FIELD field_23169 air Z
 		FIELD field_23170 material Lnet/minecraft/class_3614;
 		FIELD field_23171 materialColor Lnet/minecraft/class_3620;
 		FIELD field_23172 hardness F
@@ -538,7 +538,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 			FIELD field_17651 exceedsCube Z
 			FIELD field_19360 collisionShape Lnet/minecraft/class_265;
 			FIELD field_19429 solidFullSquare [Z
-			FIELD field_20337 isFullCube Z
+			FIELD field_20337 FullCube Z
 			METHOD <init> (Lnet/minecraft/class_2680;)V
 				ARG 1 state
 	CLASS class_4972 TypedContextPredicate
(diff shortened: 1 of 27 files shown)


mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -534,7 +534,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 			FIELD field_17651 exceedsCube Z
 			FIELD field_19360 collisionShape Lnet/minecraft/class_265;
 			FIELD field_19429 solidFullSquare [Z
-			FIELD field_20337 isFullCube Z
+			FIELD field_20337 FullCube Z
```
i509VCB:
```suggestion
			FIELD field_20337 fullCube Z
```

mappings/net/minecraft/client/MinecraftClient.mapping
```diff
@@ -7,7 +7,7 @@ CLASS net/minecraft/class_310 net/minecraft/client/MinecraftClient
 	FIELD field_1689 framebuffer Lnet/minecraft/class_276;
 	FIELD field_1690 options Lnet/minecraft/class_315;
 	FIELD field_1692 targetedEntity Lnet/minecraft/class_1297;
-	FIELD field_1693 is64Bit Z
+	FIELD field_1693 64Bit Z
```
i509VCB: Is that a valid field name?

mappings/net/minecraft/client/MinecraftClient.mapping
```diff
@@ -17,7 +17,7 @@ CLASS net/minecraft/class_310 net/minecraft/client/MinecraftClient
 	FIELD field_1700 instance Lnet/minecraft/class_310;
 	FIELD field_1701 openProfilerSection Ljava/lang/String;
 	FIELD field_1702 toastManager Lnet/minecraft/class_374;
-	FIELD field_1703 IS_SYSTEM_MAC Z
+	FIELD field_1703 SYSTEM_MAC Z
```
i509VCB: Sounds kinda odd. Is there a better way to word this?

mappings/net/minecraft/entity/passive/HorseBaseEntity.mapping
```diff
@@ -1,7 +1,7 @@
 CLASS net/minecraft/class_1496 net/minecraft/entity/passive/HorseBaseEntity
 	FIELD field_18118 PARENT_HORSE_PREDICATE Lnet/minecraft/class_4051;
 	FIELD field_6955 temper I
-	FIELD field_6956 IS_BRED_HORSE Ljava/util/function/Predicate;
+	FIELD field_6956 BRED_HORSE Ljava/util/function/Predicate;
```
i509VCB: Hmm, what does this predicate do?

mappings/net/minecraft/world/gen/feature/ShipwreckFeatureConfig.mapping
```diff
@@ -1,4 +1,4 @@
 CLASS net/minecraft/class_3172 net/minecraft/world/gen/feature/ShipwreckFeatureConfig
-	FIELD field_13803 isBeached Z
+	FIELD field_13803 neached Z
```
i509VCB:
```suggestion
	FIELD field_13803 beached Z
```

i509VCB (changes requested):
I personally think the `isClient` field on the client should become `client`, but that deserves it's own consideration in a different pr.

mappings/net/minecraft/client/MinecraftClient.mapping
i509VCB: Not valid, failed to build:
`Caused by: java.lang.IllegalArgumentException: not a valid name: 64Bit`

mappings/net/minecraft/client/MinecraftClient.mapping
dexman545: CURRENT_SYSTEM_MAC?

i509VCB: yea that sounds better

mappings/net/minecraft/entity/passive/HorseBaseEntity.mapping
dexman545: Used to set the parent predicate, so that child stay around their parent

YanisBft:
The issues with this PR are:
- Non-adjective fields are confusing without "is" prefix
- Some fields use the "is" prefix because the related json strings use it (damage source and entity predicates)

dexman545:
Restored some old ones

mappings/net/minecraft/client/MinecraftClient.mapping
liach: Maybe `x64`?

Earthcomputer (changes requested):
I'm against some of these changes in cases where the removal of the `is` prefix causes the field name to be a noun phrase rather than an adjective phrase, which seems to be quite a few of them. Some examples:
- `RealmsWorldSlotButton.State.isCurrentlyActiveSlot` -> `currentlyActiveSlot`: makes it sound like the state "has" a currently active slot, rather than "is" one.
- `AbstractBlock.isAir` -> `air`: makes it sound like an `AbstractBlock` somehow "has" air, rather than "is" air.
- You get the pattern.

Some renames that are okay because they turn them into adjectives without losing clarity:
- `BipedEntityModel.isSneaking` -> `sneaking`
- ~~`PlayerEntity.isSubmergedInWater` -> `submergedInWater`~~ (edit: apparently this one is bad because there is already `protected Entity.submergedInWater`.)
- etc

mappings/net/minecraft/client/Mouse.mapping
```diff
@@ -3,7 +3,7 @@ CLASS net/minecraft/class_312 net/minecraft/client/Mouse
 	FIELD field_1780 activeButton I
 	FIELD field_1781 controlLeftTicks I
 	FIELD field_1782 cursorYSmoother Lnet/minecraft/class_3540;
-	FIELD field_1783 cursorLocked Z
+	FIELD field_1783 isCursorLocked Z
```
liach: No
```suggestion
	FIELD field_1783 cursorLocked Z
```

liach:
@dexman545 Please remove `Closes #987` from pr description. This one **does not** fix that issue as it misses many usages.

liach:
Mind fix the merge conflicts?

haykam821:
Fixes #987

YanisBft:
This PR is pretty much cursed, and has been badly merged with base branch. If you want to do it again, please make sure to take feedback here into account

PR #1404, opened 2020-05-29
New loot registries

YanisBft:


mappings/net/minecraft/loot/condition/LootCondition.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_5341
+CLASS net/minecraft/class_5341 net/minecraft/loot/condition/LootCondition
```
Juuxel, line 1: Was this not matched? :concern:

Juuxel (approved):

mappings/net/minecraft/loot/condition/LootCondition.mapping
YanisBft: Some mapped methods were moved in this class

Juuxel: Yes, but loot conditions were an existing thing already 🤔

YanisBft: Yes, but it's a new interface used by loot condition types, named after interface used by loot functions

Juuxel: Ah, turned it into an interface? Interesting.

modmuss50: In hindsight it might have been worth matching, however it would have broken anyway.

modmuss50 (approved):

Gegy (approved):

mappings/net/minecraft/loot/entry/LootPoolEntry.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_79 net/minecraft/loot/entry/LootEntry
+CLASS net/minecraft/class_79 net/minecraft/loot/entry/LootPoolEntry
```
liach, line 1: Why this rename? To be consistent with reference in `Registry`?

YanisBft: Yep

liach (approved):

PR #1409, opened 2020-05-29
remove an invalid mapping

TwilightFlower:
There's a mapping for a non-existent method `b` in `RedstoneWireBlock`. This happens to match up with an obfuscated name, and has been causing problems for me when I try to remap things from srg to intermediary and yarn.

This removes this 🅱️ad mapping.


TheGlitch76 (approved):
Seems simple enough. This is the only case of that happening?

TheGlitch76:
Also, this should probably be checked to make sure it didn't sneak into 1.16

YanisBft (approved):

liach (approved):

Earthcomputer (approved):

modmuss50:
Again 1.15.2 I will handle the merge for this

liach:
well this fix is valuable for 1.15.2, so i marked it as ready (note its release tag)

modmuss50:
I’ll try and sort the issue making it a slight inconvenience to merge PRs for older versions.

liach:
This broken entry doesn't appear to exist in 20w22a mappings

PR #1410, opened 2020-05-30
Couple of AbstractBlockState Mappings

dexman545:
AbstractBlock:
method_29291 -> toolRequired
field_25184 -> requiresTool

### DIFF
--- a/mappings/net/minecraft/block/AbstractBlock.mapping
+++ b/mappings/net/minecraft/block/AbstractBlock.mapping
@@ -312,6 +312,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		FIELD field_23176 blockVisionPredicate Lnet/minecraft/class_4970$class_4973;
 		FIELD field_23177 postProcessPredicate Lnet/minecraft/class_4970$class_4973;
 		FIELD field_23178 emissiveLightingPredicate Lnet/minecraft/class_4970$class_4973;
+		FIELD field_25184 requiresTool Z
 		METHOD <init> (Lnet/minecraft/class_2248;Lcom/google/common/collect/ImmutableMap;Lcom/mojang/serialization/MapCodec;)V
 			ARG 1 block
 			ARG 2 propertyMap
@@ -525,6 +526,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 			ARG 2 pos
 		METHOD method_27852 isOf (Lnet/minecraft/class_2248;)Z
 			ARG 1 block
+		METHOD method_29291 doesRequireTool ()Z
 		CLASS class_3752 ShapeCache
 			FIELD field_16555 lightSubtracted I
 			FIELD field_16556 translucent Z
--- a/mappings/net/minecraft/block/AbstractBlock.mapping
+++ b/mappings/net/minecraft/block/AbstractBlock.mapping
@@ -528,7 +528,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 			ARG 2 pos
 		METHOD method_27852 isOf (Lnet/minecraft/class_2248;)Z
 			ARG 1 block
-		METHOD method_29291 toolRequired ()Z
+		METHOD method_29291 isToolRequired ()Z
 		CLASS class_3752 ShapeCache
 			FIELD field_16555 lightSubtracted I
 			FIELD field_16556 translucent Z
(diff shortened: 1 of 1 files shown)


Prospector:
I'd name the method requiresTool as well, the `does` prefix is quite odd imo.

dexman545:
then the name would be the same as the setting's method that sets the block requiring a tool

kashike:
`toolRequired`

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -218,6 +218,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		FIELD field_23184 blockVisionPredicate Lnet/minecraft/class_4970$class_4973;
 		FIELD field_23185 postProcessPredicate Lnet/minecraft/class_4970$class_4973;
 		FIELD field_23186 emissiveLightingPredicate Lnet/minecraft/class_4970$class_4973;
+		FIELD field_25185 requiresTool Z
```
```suggestion
		FIELD field_25185 toolRequired Z
```

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -273,6 +274,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		METHOD method_26249 emissiveLighting (Lnet/minecraft/class_4970$class_4973;)Lnet/minecraft/class_4970$class_2251;
 			ARG 1 predicate
 		METHOD method_26250 air ()Lnet/minecraft/class_4970$class_2251;
+		METHOD method_29292 requiresTool ()V
```
liach, line 277:
```suggestion
		METHOD method_29292 toolRequired ()V
```

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -312,6 +314,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		FIELD field_23176 blockVisionPredicate Lnet/minecraft/class_4970$class_4973;
 		FIELD field_23177 postProcessPredicate Lnet/minecraft/class_4970$class_4973;
 		FIELD field_23178 emissiveLightingPredicate Lnet/minecraft/class_4970$class_4973;
+		FIELD field_25184 requiresTool Z
```
```suggestion
		FIELD field_25184 toolRequired Z
```

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -525,6 +528,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 			ARG 2 pos
 		METHOD method_27852 isOf (Lnet/minecraft/class_2248;)Z
 			ARG 1 block
+		METHOD method_29291 toolRequired ()Z
```
```suggestion
		METHOD method_29291 isToolRequired ()Z
```

liach (changes requested):

shedaniel:
I would prefer the old builder like names and requiresTool

liach:
Usually we don't use third person singular tense in method names, so imo `requiresTool` isn't a good method name, especially for a setter. `requireTool` may be better, like we would declare `eat` than `eats` in a sheep class.

Also `toolRequired` is an adjective phrase, which is usually a better choice for names of boolean fields/variables.

shedaniel:
then I would not use requireTool as the method, I can deal with toolRequired but please be consistent

liach:
Sure, good point, updated my suggestion

mappings/net/minecraft/block/AbstractBlock.mapping
dexman545: This change breaks with the other methods in setting's naming convention (being able to read the block settings in semi-natural English, ie. dropsNothing)

mappings/net/minecraft/block/AbstractBlock.mapping
```suggestion
		FIELD field_25184 toolRequired Z
```
This is **not** something in the builder. You argue for following the style in the builder, but this is not builder and you should use the regular naming convention.

liach (changes requested):

mappings/net/minecraft/block/AbstractBlock.mapping
liach: Mark this as resolved please.

YanisBft:
It's a bit confusing the AbstractBlock.Settings field/method are different from the AbstractBlock.AbstractBlockState ones

dexman545:
Woops, git messed up and had tried to push to the wrong place.
Field names should match now

YanisBft (approved):

dexman545:
To help clarify:
(requiresTool)
```
public AbstractBlock.Settings method_29292() {
         this.field_25185 = true;
         return this;
}
```

is in the block settings, setting the value to true as part of a build-like creator, while (toolRequired)
```
public boolean method_29291() {
         return this.field_25184;
}
```
Gets the value

liach (approved):

modmuss50 (approved):

Juuxel (approved):

PR #1414, opened 2020-05-31
Map ZOMBIE_TYPE

Earthcomputer:
I've been matching backwards; I matched back to 1.10 now and discovered what this field is from.

### DIFF
--- a/mappings/net/minecraft/entity/mob/ZombieEntity.mapping
+++ b/mappings/net/minecraft/entity/mob/ZombieEntity.mapping
@@ -4,7 +4,7 @@ CLASS net/minecraft/class_1642 net/minecraft/entity/mob/ZombieEntity
 	FIELD field_7425 CONVERTING_IN_WATER Lnet/minecraft/class_2940;
 	FIELD field_7426 inWaterTime I
 	FIELD field_7427 ZOMBIE_TYPE Lnet/minecraft/class_2940;
-		COMMENT Unused tracked data, left over from 1.10 when zombies, zombie villagers and husks were all the same entity type.
+		COMMENT Unused tracked data, left over from 1.10 when zombies, zombie villagers and husks were all the same type of entity.
 	FIELD field_7429 BABY_SPEED_ID Ljava/util/UUID;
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/entity/mob/ZombieEntity.mapping
```diff
@@ -3,6 +3,8 @@ CLASS net/minecraft/class_1642 net/minecraft/entity/mob/ZombieEntity
 	FIELD field_7424 ticksUntilWaterConversion I
 	FIELD field_7425 CONVERTING_IN_WATER Lnet/minecraft/class_2940;
 	FIELD field_7426 inWaterTime I
+	FIELD field_7427 ZOMBIE_TYPE Lnet/minecraft/class_2940;
+		COMMENT Unused tracked data, left over from 1.10 when zombies, zombie villagers and husks were all the same entity type.
```
i509VCB: Did they share the same entity type or were the all of them a `ZombieEntity` just with a flag

Earthcomputer: `EntityType` didn't exist back then in the code, there was only `Class<? extends Entity>`. They were all the same class with just this flag.

i509VCB: I'd probably say they were the same entity then in the comment

Earthcomputer: "same entity" would imply to me that they are the same entity instance

haykam821: Maybe "same type of entity"? That'd work around "entity type" being taken while still meaning the same time.

Juuxel (approved):

YanisBft (approved):

liach (approved):

i509VCB (approved):

haykam821 (approved):

modmuss50 (approved):
Great detective work.

PR #1416, opened 2020-06-01
Finish mapping campfires

haykam821:


mappings/net/minecraft/client/particle/CampfireSmokeParticle.mapping
```diff
@@ -7,6 +7,7 @@ CLASS net/minecraft/class_3937 net/minecraft/client/particle/CampfireSmokePartic
 		ARG 8 velocityX
 		ARG 10 velocityY
 		ARG 12 velocityZ
+		ARG 14 lastsLonger
```
Ueaj-Kerman: maybe "persistent" or somethng?

Ueaj-Kerman (approved):

mappings/net/minecraft/client/particle/CampfireSmokeParticle.mapping
hnOsmium0001: Persistent sounds like it would have something to do with IO to me
```suggestion
		ARG 14 extendedLifetime
```

liach:
```suggestion
		ARG 14 signalFire
```
Since the block state toggle flag is called `signal_fire`? Though this name is already good enough

liach (approved):

Juuxel (approved):

modmuss50 (approved):

PR #1431, opened 2020-06-05
String renderable (or visitable?)

liach:


Juuxel (changes requested):
I feel like `StringVisitable` is too abstract of a name. It takes the place of `Text` in text rendering methods, and it could possibly have a name related to that.

One idea I had was to map the new class as `Text` and rename `Text` to something else, but that feels wrong since most of the `Text` methods are still in `Text`.

Another option would be to call something like a `StyledString`, since it seems to be a string with a style and methods for working with that.

i509VCB:
`StylizedString` probably is a bit more grammatically correct over `StyledString`

liach:
This is definitely not a `Text`. This is not anything that is really string; it just can choose to feed whatever it wants to two types of specified visitors, one visiting raw strings and one visiting raw strings + a style

Juuxel:
After a bit of usage searching, I found out that `class_5348` is used 99% for text rendering (the rest is basically `Text` extending it), so it should just be called `RenderableString`. Calling it something like `StringVisitable` is far too generic and vague compared to what it's actually used for in the game. The visitor methods are just used for selecting parts of the text to render/hover/click.

liach:
It is used for rendering mostly, but what about the visitors? when you terminate a visit with the visitor, the string visitable would be only partially rendered. It can be used for many purposes like batch replacing special characters, not just rendering.

Juuxel:
It *can* be, but it's not. It's a specialised class that is only used for text rendering. There are other classes like this; `ScreenHandlerContext` is something that could be called "world visitable", but it's only used for screen handlers and it gets its name from that.

This class is similar: while it technically could be used for a lot of things, Mojang only uses it for a single purpose. We don't call something like `ChatMessages` "`ClientTexts`" either, even though it works with any `class_5348` instances. Classes should be named by their actual usages, not hypothetical ones.

liach:
Check out `TextHandler`. This thing is actually used for just removing surrogates or special characters from strings beyond rendering.

Juuxel:
All `TextHandler` methods are only used for -- you guessed it -- *rendering*. `getWidth`? Only used in `TextRenderer` and `BookEditScreen`. `trimToWidth`? Text rendering and book screens. `wrapLines`? Text rendering, screens and widgets. Even the `visit` methods in the new interface are mostly called on the client.

The interface represents a renderable string of characters, i.e. a `RenderableString`.

Juuxel (approved):
`StringRenderable` sounds better than `StringVisitable`, and we can always change it later if Mojang does yet another text refactor.

YanisBft (approved):

shedaniel (approved):

shedaniel:
StringRenderable sounds like something like Drawable, where there is a method to render it

liach:
Well juuz hates the `visitable` name, debatable

modmuss50 (approved):
Im happy with `StringRenderable` I think vistable also works fine.

liach:
Maybe we don't need to treat `render` strictly as drawing something on the client. dumping something that usually is a json tree as a string on console is probably "rendering" as well, and this class is used for that purpose aside from client rendering

PR #1438, opened 2020-06-07
Categorize entity packages a bit

i509VCB:
The entity packages are quite big imo. So I've categorized the entities by their type.

For an entity to get it's own package, it must meet the criteria of having multiple variants (such as skeletons), or be distinctly abstract (such as raiders and illagers).

This also moves `Raid` and `RaidManager` to it's own package since these are completely seperate constructs from just an entity


mappings/net/minecraft/raid/Raid.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_3765 net/minecraft/entity/raid/Raid
```
liach: Should raid really go to root package... it is such a small thing

liach (changes requested):
I personally disagree with splitting out fish, skeleton, and illager package. They can just go to their parent package.

mappings/net/minecraft/raid/Raid.mapping
i509VCB: maybe `world` or `server.world`

liach: maybe `server.raid` package? or village (no raid subpackage)?

mappings/net/minecraft/entity/golem/IronGolemEntity.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_1439 net/minecraft/entity/passive/IronGolemEntity
+CLASS net/minecraft/class_1439 net/minecraft/entity/golem/IronGolemEntity
```
liach: Let's just move all golem to the entity root package, no golem subpackage

mappings/net/minecraft/raid/Raid.mapping
i509VCB: Well villages are always the subject of a raid, so that makes sense

i509VCB:
Personally I'd keep the skeleton and illager packages. Skeletons because they are distinctly skeletons.

Illagers because they are an official category of mobs

liach:
Don't think grouping package by "official category" is a good idea. Java runtime library doesn't group collections into packages by what classes they extend.

i509VCB:
We aren't the java runtime library though

liach:
We don't have a reason to create super deep package trees. Super deep package trees with tiny packages are true nightmares

i509VCB:
How small is too small and how deep is too deep?

liach:
I think the current setup of packages is fairly fine. We just need to maybe split mob package into monster and neutral

i509VCB:
Yeah I've noticed that, since mob is meant to be generic to any entity with the capabilities to do basic pathfinding and move itself

Probably a later PR tbf

mappings/net/minecraft/entity/mob/hoglin/HoglinBrain.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_4835 net/minecraft/entity/mob/HoglinBrain
+CLASS net/minecraft/class_4835 net/minecraft/entity/mob/hoglin/HoglinBrain
```
YanisBft, line 1: Isn't there a package like `entity.ai.brain` or so for classes like these?

liach: it's where hoglin brain handling logic is, must be in same package as hoglin

modmuss50:
I think we should put this on hold along with other large scale refactors untill 1.17 snapshots start (hopefully only a few months at max).

modmuss50 (approved):
As this is only package renames it should be fine to go in now.

liach:
Per our discord discussion, this change will be on hold until 1.17 snapshots.

i509VCB:
Did you scroll down far enough?

liach:
Sure, but I still oppose this structural change. You split out way too many packages with too few classes.

i509VCB:
This will probably require another PR, will reopen in 1.17 snapshots

PR #1441, opened 2020-06-08
Four unmapped top level classes

liach:

With #1431 and #1440, all unmapped top level classes will be now mapped in 1.16-pre2..

### DIFF
--- a/mappings/net/minecraft/datafixer/fix/NonStretchingPackedArray.mapping
+++ b/mappings/net/minecraft/datafixer/fix/NonStretchingPackedArray.mapping
@@ -1,0 +1,24 @@
+CLASS net/minecraft/class_5298 net/minecraft/datafixer/fix/NonStretchingPackedArray
+	COMMENT A packed array of integers in which no value would stretch over more
+	COMMENT than one long values in the backing array. Introduced in 20w17a for the
+	COMMENT new block state storage format.
+	FIELD field_24641 array [J
+	FIELD field_24642 unitSize I
+	FIELD field_24643 maxValue J
+	FIELD field_24644 length I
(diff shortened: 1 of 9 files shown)


mappings/net/minecraft/datafixer/fix/NonStretchingPackedArray.mapping
```diff
@@ -0,0 +1,24 @@
+CLASS net/minecraft/class_5298 net/minecraft/datafixer/fix/NonStretchingPackedArray
```
i509VCB: Is this a util class or an actual data fix?

If it's a util class it should go in util package.

mappings/net/minecraft/datafixer/fix/NonStretchingPackedArray.mapping
```diff
@@ -0,0 +1,24 @@
...
+	COMMENT A packed array of integers in which no value would stretch over more
+	COMMENT than one long values in the backing array. Introduced in 20w17a for the
```
i509VCB:
```suggestion
	COMMENT than one long value in the backing array. Introduced in 20w17a for the
```

I assume you meant this?

mappings/net/minecraft/datafixer/fix/NonStretchingPackedArray.mapping
liach: A utility array that is only used by data fix. Hmm no other such helper object exist in fix package, sure

i509VCB (approved):
Looks fine to me otherwise

Gegy (approved):

modmuss50 (approved):

PR #1447, opened 2020-06-08
Did some mappings in BlockPos

Kroppeb:
The codec will be added in a later pr.

### DIFF
--- a/mappings/net/minecraft/util/math/BlockPos.mapping
+++ b/mappings/net/minecraft/util/math/BlockPos.mapping
@@ -80,9 +80,9 @@ CLASS net/minecraft/class_2338 net/minecraft/util/math/BlockPos
 	METHOD method_10093 offset (Lnet/minecraft/class_2350;)Lnet/minecraft/class_2338;
 		ARG 1 direction
 	METHOD method_10094 iterate (IIIIII)Ljava/lang/Iterable;
-		COMMENT Iterates over all block positions within a given area.
+		COMMENT Iterates over all block positions within a given box.
 		COMMENT
-		COMMENT <p>Note that the same object is returned each time.
+		COMMENT <p>Note that the same mutable block pos object is returned each time.
 		ARG 0 startX
 			COMMENT The lower x coordinate
 		ARG 1 startY
@@ -133,7 +133,7 @@ CLASS net/minecraft/class_2338 net/minecraft/util/math/BlockPos
 		COMMENT the one with a positive z offset is visited first before the one with a
 		COMMENT negative z offset.
 		COMMENT
-		COMMENT <p>Note that the same object is returned each time.
+		COMMENT <p>Note that the same mutable block pos object is returned each time.
 		ARG 0 center
 			COMMENT the center of iteration
 		ARG 1 xRange
@@ -156,7 +156,7 @@ CLASS net/minecraft/class_2338 net/minecraft/util/math/BlockPos
 		COMMENT Iterates random block positions in a given area.
 		COMMENT The same coordinates can ber returned multiple times.
 		COMMENT
-		COMMENT <p>Note that the same object is returned each time.
+		COMMENT <p>Note that the same mutable block pos object is returned each time.
 		ARG 0 random
 			COMMENT The {@code Random} object used to randomize coordinates
 		ARG 1 count
--- a/mappings/net/minecraft/util/math/BlockPos.mapping
+++ b/mappings/net/minecraft/util/math/BlockPos.mapping
@@ -154,7 +154,7 @@ CLASS net/minecraft/class_2338 net/minecraft/util/math/BlockPos
 		ARG 3 maxZ
 	METHOD method_27156 iterateRandom (Ljava/util/Random;IIIIIII)Ljava/lang/Iterable;
 		COMMENT Iterates random block positions in a given area.
-		COMMENT The same coordinates can ber returned multiple times.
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/util/math/BlockPos.mapping
```diff
@@ -80,12 +80,21 @@ CLASS net/minecraft/class_2338 net/minecraft/util/math/BlockPos
 	METHOD method_10093 offset (Lnet/minecraft/class_2350;)Lnet/minecraft/class_2338;
 		ARG 1 direction
 	METHOD method_10094 iterate (IIIIII)Ljava/lang/Iterable;
+		COMMENT Iterates over all block positions within a given area.
+		COMMENT
+		COMMENT <p>Note that the same object is returned each time.
```
liach:
```suggestion
		COMMENT <p>Note that the same mutable block pos object is returned each time.
```

mappings/net/minecraft/util/math/BlockPos.mapping
```diff
@@ -80,12 +80,21 @@ CLASS net/minecraft/class_2338 net/minecraft/util/math/BlockPos
 	METHOD method_10093 offset (Lnet/minecraft/class_2350;)Lnet/minecraft/class_2338;
 		ARG 1 direction
 	METHOD method_10094 iterate (IIIIII)Ljava/lang/Iterable;
+		COMMENT Iterates over all block positions within a given area.
```
liach:
```suggestion
		COMMENT Iterates over all block positions within a given box.
```

mappings/net/minecraft/util/math/BlockPos.mapping
```diff
@@ -123,6 +132,8 @@ CLASS net/minecraft/class_2338 net/minecraft/util/math/BlockPos
 		COMMENT positions with the same x and y offsets and the same manhattan distance,
 		COMMENT the one with a positive z offset is visited first before the one with a
 		COMMENT negative z offset.
+		COMMENT
+		COMMENT <p>Note that the same object is returned each time.
```
liach:
```suggestion
		COMMENT <p>Note that the same mutable block pos object is returned each time.
```

mappings/net/minecraft/util/math/BlockPos.mapping
```diff
@@ -141,6 +152,38 @@ CLASS net/minecraft/class_2338 net/minecraft/util/math/BlockPos
 		ARG 1 maxX
 		ARG 2 maxY
 		ARG 3 maxZ
+	METHOD method_27156 iterateRandom (Ljava/util/Random;IIIIIII)Ljava/lang/Iterable;
+		COMMENT Iterates random block positions in a given area.
+		COMMENT The same coordinates can ber returned multiple times.
+		COMMENT
+		COMMENT <p>Note that the same object is returned each time.
```
liach:
```suggestion
		COMMENT <p>Note that the same mutable block pos object is returned each time.
```

mappings/net/minecraft/util/math/BlockPos.mapping
```diff
@@ -151,6 +194,11 @@ CLASS net/minecraft/class_2338 net/minecraft/util/math/BlockPos
 		FIELD field_23101 limitY I
 		FIELD field_23102 dx I
 		FIELD field_23103 dy I
+		FIELD field_23375 centerZ I
+		FIELD field_23376 centerX I
+		FIELD field_23377 centerY I
+		FIELD field_23378 blockPos Lnet/minecraft/class_2338$class_2339;
```
liach, line 200: should be an opposite pos from the current calculated one

mappings/net/minecraft/util/math/BlockPos.mapping
```diff
@@ -141,6 +152,38 @@ CLASS net/minecraft/class_2338 net/minecraft/util/math/BlockPos
 		ARG 1 maxX
 		ARG 2 maxY
 		ARG 3 maxZ
+	METHOD method_27156 iterateRandom (Ljava/util/Random;IIIIIII)Ljava/lang/Iterable;
+		COMMENT Iterates random block positions in a given area.
+		COMMENT The same coordinates can ber returned multiple times.
```
dblsaiko: typo

mappings/net/minecraft/util/math/BlockPos.mapping
Kroppeb: what do you mean?

liach: if i recall right this should be an iterator that iterates outwards, isn't it?

mappings/net/minecraft/util/math/BlockPos.mapping
liach:
```suggestion
		COMMENT The same position may be returned multiple times in {@code next} calls.
```

mappings/net/minecraft/util/math/BlockPos.mapping
Kroppeb: Yes?

liach: The outward iterator is weird. It calculates a pair of positions at once, and holds one in a field so it would be returned in the next `next` call. if this is the block pos cached for the next `next` call imo `oppositeNext` is better; if this is the pos passed as the original center then `center` is better

PR #1452, opened 2020-06-09
Finish ProfilerTimer mapping

RDIL:
This is my first mapping that I've ever written, so please let me know if I did something wrong.

Also, this exists on latest, so figured I would add it.

Thanks in advance 😃

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,7 +12,7 @@ buildscript {
 		classpath "commons-io:commons-io:2.6"
 		classpath "com.google.guava:guava:28.0-jre"
 		classpath 'de.undercouch:gradle-download-task:3.4.3'
-		classpath 'net.fabricmc:tiny-remapper:+'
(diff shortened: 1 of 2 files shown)


liach:
I wonder if we still add to 1.14.4 mappings 🤔

liach (approved):

RDIL:
@liach modmuss said that if it targets latest it can be backported, and this change is on master but was never backported for some reason.

liach:
@modmuss50 decide then

modmuss50:
No reason why not, not sure why it failed to build.

liach:
The build failure may be caused by what's fixed in #1260?

RDIL:
I can cherry-pick that commit onto this branch if you want.

RDIL:
@liach that fixed it!

modmuss50 (approved):

PR #1457, opened 2020-06-10
DimensionArgumentType -> WorldArgumentType

i509VCB:
This is because the return type has now changed to return a ServerWorld.

Also includes some other relevant renames that this argument may call.


mappings/net/minecraft/client/network/ClientPlayNetworkHandler.mapping
```diff
@@ -2,7 +2,8 @@ CLASS net/minecraft/class_634 net/minecraft/client/network/ClientPlayNetworkHand
 	FIELD field_16771 sessionId Ljava/util/UUID;
 	FIELD field_19144 chunkLoadDistance I
 	FIELD field_24321 worldProperties Lnet/minecraft/class_638$class_5271;
-	FIELD field_25063 dimensionTracker Lnet/minecraft/class_5318;
+	FIELD field_25063 registryTracker Lnet/minecraft/class_5318;
+	FIELD field_25273 worldRegistryKeys Ljava/util/Set;
```
liach:
```suggestion
	FIELD field_25273 worldKeys Ljava/util/Set;
```

mappings/net/minecraft/client/network/ClientPlayNetworkHandler.mapping
```diff
@@ -57,3 +58,5 @@ CLASS net/minecraft/class_634 net/minecraft/client/network/ClientPlayNetworkHand
 	METHOD method_2888 validateResourcePackUrl (Ljava/lang/String;)Z
 		ARG 1 url
 	METHOD method_2890 getWorld ()Lnet/minecraft/class_638;
+	METHOD method_29091 getRegistryTracker ()Lnet/minecraft/class_5318;
+	METHOD method_29356 getWorldRegistryKeys ()Ljava/util/Set;
```
liach:
```suggestion
	METHOD method_29356 getWorldKeys ()Ljava/util/Set;
```

mappings/net/minecraft/command/arguments/WorldArgumentType.mapping
```diff
@@ -0,0 +1,14 @@
+CLASS net/minecraft/class_2181 net/minecraft/command/arguments/WorldArgumentType
+	FIELD field_9844 EXAMPLES Ljava/util/Collection;
+	FIELD field_9845 INVALID_DIMENSION_EXCEPTION Lcom/mojang/brigadier/exceptions/DynamicCommandExceptionType;
```
liach: Still?

mappings/net/minecraft/server/command/CommandSource.mapping
```diff
@@ -1,6 +1,7 @@
 CLASS net/minecraft/class_2172 net/minecraft/server/command/CommandSource
 	METHOD method_17771 getBlockPositionSuggestions ()Ljava/util/Collection;
 	METHOD method_17772 getPositionSuggestions ()Ljava/util/Collection;
+	METHOD method_29310 getWorldRegistryKeys ()Ljava/util/Set;
```
liach:
```suggestion
	METHOD method_29310 getWorldKeys ()Ljava/util/Set;
```

liach (commented):
World isn't in a registry, they are just using a class named `RegistryKey` as key, so Imo we should just call it `worldKey`

liach (approved):

liach:
I hereby block your attempt to change this name. This thing is called "dimension" still in a multitude of places, including translation keys and usages. The name change would add confusion than make things easier.

Edit: Juuz this is more like a joke comment. While I don't support it I am actually more like neutral on this.

Prospector (approved):

haykam821 (approved):

kvverti:
`World` now encapsulates both a world and the dimension that world is in. Code-wise they are the same object, so keeping `Dimension` is fine. 

As well, datapacks call them `dimension`

liach:
Maybe stay "dimension" if many exposed usages still call it dimension.

liach:
Should we make this a vote?

i509VCB:
So this is likely not going to be merged. Other parts of PR will be moved into a later pr

PR #1461, opened 2020-06-10
Map various, mostly private parts of client.texture classes

Sturmlilie:
This is mostly for documentation of the internal behavior of classes. There is one single mapping that I changed because it was a mistake (mipmap -> close in `NativeImage`), which becomes apparent when you compare the chained overload being called.  
Some more context on names:
* `atlasWidth/Height`: sprites need to know the final texture size of the atlas they're part of to calculate their position in [0,1] based coordinates.
* `maxLevel`: I introduced this term in my TextureUtil PR and keep on using it; it always refers to the highest miplevel index that is wanted (so 0 -> 1 miplevels).
* `NativeImage#write`: semantically the inverse operation to `read`
* `Sprite.Interpolation#apply` was a bit tricky to describe because it appears like it overwrites one of the frames it's interpolating, but I have a hunch the space it is writing to is reserved for just this interpolated image data.
* `SpriteInterpolation#mix` standard lerp, I chose the glsl naming (sans parameter names). bikeshed away =)
* `SpriteAtlasTexture#loadSprites`: based off of profiler scope
* many parameter names were given where a `this.x = x` correspondence with existing field existed in the constructor.


liach (approved):

mappings/net/minecraft/client/texture/Sprite.mapping
```diff
@@ -60,3 +68,17 @@ CLASS net/minecraft/class_1058 net/minecraft/client/texture/Sprite
 		FIELD field_21758 images [Lnet/minecraft/class_1011;
 		METHOD <init> (Lnet/minecraft/class_1058;Lnet/minecraft/class_1058$class_4727;I)V
 			ARG 3 mipmap
+		METHOD method_24128 apply ()V
+			COMMENT Linearly interpolate between the current and next frame on all miplevels
+			COMMENT based on the tick position within the current frame,
+			COMMENT and upload the results to the currently bound texture to the frame slot at position (0,0).
+		METHOD method_24129 mix (DII)I
+			ARG 1 alpha
```
Juuxel: `alpha`? We use `delta` for this in `lerp`.

mappings/net/minecraft/client/texture/Sprite.mapping
```diff
@@ -60,3 +68,17 @@ CLASS net/minecraft/class_1058 net/minecraft/client/texture/Sprite
 		FIELD field_21758 images [Lnet/minecraft/class_1011;
 		METHOD <init> (Lnet/minecraft/class_1058;Lnet/minecraft/class_1058$class_4727;I)V
 			ARG 3 mipmap
+		METHOD method_24128 apply ()V
+			COMMENT Linearly interpolate between the current and next frame on all miplevels
+			COMMENT based on the tick position within the current frame,
+			COMMENT and upload the results to the currently bound texture to the frame slot at position (0,0).
+		METHOD method_24129 mix (DII)I
```
Juuxel: Should be `lerp` to match `MathHelper.lerp` 😛

Juuxel (changes requested):

mappings/net/minecraft/client/texture/Sprite.mapping
Sturmlilie: As I said, this is based on glsl terminology (they use just `a` but it's clearly based on blending usecase, + this lerp is directly used  for blending).

liach: Think `alpha` for a blend of a background and a foreground color is fine

mappings/net/minecraft/client/texture/Sprite.mapping
```diff
@@ -60,3 +68,17 @@ CLASS net/minecraft/class_1058 net/minecraft/client/texture/Sprite
 		FIELD field_21758 images [Lnet/minecraft/class_1011;
 		METHOD <init> (Lnet/minecraft/class_1058;Lnet/minecraft/class_1058$class_4727;I)V
 			ARG 3 mipmap
+		METHOD method_24128 apply ()V
+			COMMENT Linearly interpolate between the current and next frame on all miplevels
+			COMMENT based on the tick position within the current frame,
+			COMMENT and upload the results to the currently bound texture to the frame slot at position (0,0).
+		METHOD method_24129 lerp (DII)I
```
liach, line 75: So this is lerping by each r/g/b portion? maybe mention that to clarify from regular lerp, like add javadoc or call this `lerpColor`

Sturmlilie: No, not at all, this is a purely value(int)-level lerp that doesn't know about colors or anything. The code that uses it though applies it to RGB components. Edit: Maybe I should mention that the javadoc above is part of `apply`, not `lerp`.

liach: Hmm, great. Think mathhelper only has int lerp. I saw that jd is for apply so i suggested "add javadoc" to this method, which now appears unnecessary.

Sturmlilie: Oh sorry I misread your comment earlier. I will add javadoc to lift the confusion then.

Sturmlilie: @liach if you could review whether the javadoc I added is adequate I'd be grateful =3

Juuxel (approved):

YanisBft (approved):

PR #1466, opened 2020-06-13
Map UUID in GameMessageS2CPacket

DreenDex:
Mojang, why?


i509VCB:
Maybe it's for blocking users messages on the client for the future?

liach (approved):
Hmm, thought someone already mapped them when it was in a few snapshots ago.

YanisBft (approved):

PR #1484, opened 2020-06-19
Glint layer mappings

Sollace:
Mapped all the new vertex consumer methods Mojang added to ItemRenderer (most likely in the process of fixing enchantment glints)

### DIFF
--- a/mappings/net/minecraft/client/render/RenderLayer.mapping
+++ b/mappings/net/minecraft/client/render/RenderLayer.mapping
@@ -115,8 +115,8 @@ CLASS net/minecraft/class_1921 net/minecraft/client/render/RenderLayer
 		ARG 0 texture
 	METHOD method_29380 getTranslucentMovingBlock ()Lnet/minecraft/class_1921;
 	METHOD method_29381 getItemPhaseData ()Lnet/minecraft/class_1921$class_4688;
-	METHOD method_29706 getGlintDirect ()Lnet/minecraft/class_1921;
-	METHOD method_29707 getEntityGlintDirect ()Lnet/minecraft/class_1921;
+	METHOD method_29706 getDirectGlint ()Lnet/minecraft/class_1921;
+	METHOD method_29707 getDirectEntityGlint ()Lnet/minecraft/class_1921;
 	METHOD method_29996 getTripwirePhaseData ()Lnet/minecraft/class_1921$class_4688;
 	METHOD method_29997 getTripwire ()Lnet/minecraft/class_1921;
 	CLASS class_4687 MultiPhase
--- a/mappings/net/minecraft/client/render/RenderLayer.mapping
+++ b/mappings/net/minecraft/client/render/RenderLayer.mapping
@@ -5,6 +5,8 @@ CLASS net/minecraft/class_1921 net/minecraft/client/render/RenderLayer
 	FIELD field_20975 hasCrumbling Z
 	FIELD field_21402 translucent Z
 	FIELD field_21850 optionalThis Ljava/util/Optional;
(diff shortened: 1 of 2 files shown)


liach (approved):

mappings/net/minecraft/client/render/item/ItemRenderer.mapping
```diff
@@ -42,7 +42,7 @@ CLASS net/minecraft/class_918 net/minecraft/client/render/item/ItemRenderer
 		ARG 4 stack
 		ARG 5 light
 		ARG 6 overlay
-	METHOD method_23181 getArmorVertexConsumer (Lnet/minecraft/class_4597;Lnet/minecraft/class_1921;ZZ)Lnet/minecraft/class_4588;
+	METHOD method_23181 getItemGlintConsumer (Lnet/minecraft/class_4597;Lnet/minecraft/class_1921;ZZ)Lnet/minecraft/class_4588;
 		ARG 0 vertexConsumers
```
YanisBft, line 46: Why is this param named `provider` in other methods?

YanisBft (approved):

mappings/net/minecraft/client/render/item/ItemRenderer.mapping
Sollace: Ah, sorry. It wasn't mapped in the other methods and I didn't remember what y'all like to call it.

It's a `VertexConsumerProvider` which gets shortened to `provider` in my own projects.

Sollace: Changed.

Though someone should maybe watershed these names, since the method to get a `VertexConsumer` is called `getBuffer`, it might be sensible to call it something like `VertexBuffer` or `LayerBuffer`

YanisBft (approved):

mappings/net/minecraft/client/render/RenderLayer.mapping
```diff
@@ -115,8 +115,8 @@ CLASS net/minecraft/class_1921 net/minecraft/client/render/RenderLayer
 		ARG 0 texture
 	METHOD method_29380 getTranslucentMovingBlock ()Lnet/minecraft/class_1921;
 	METHOD method_29381 getItemPhaseData ()Lnet/minecraft/class_1921$class_4688;
-	METHOD method_29706 getGlintDirect ()Lnet/minecraft/class_1921;
-	METHOD method_29707 getEntityGlintDirect ()Lnet/minecraft/class_1921;
+	METHOD method_29706 getDirectGlint ()Lnet/minecraft/class_1921;
+	METHOD method_29707 getDirectEntityGlint ()Lnet/minecraft/class_1921;
```
Juuxel, line 121: The getters should match the fields which are called `GLINT_DIRECT` and `ENTITY_GLINT_DIRECT`, which come from their IDs.

Juuxel (changes requested):

mappings/net/minecraft/client/render/RenderLayer.mapping
liach: Imo mojang used stupid naming there and we should not follow suit.

Sollace: What is the verdict? I changed it because the names were not fitting with our naming convention (they're stupid, as liach has said)

Juuxel: Hmm, I'd say renaming is fine if the fields are also renamed to match the methods.

Prospector: I agree. Rename field and method

Sollace:
Rebased and updated to 1.16.3

YanisBft (approved):

PR #1490, opened 2020-06-20
Rename the air material color to clear

### LINKED ISSUE #1451: Material-/MapColor Field Renames
For longterm discussion about how to best rename them. 

It should be noted that while bad, they are at least moderately functional (when you know they're for colors on a map).

The field names are currently named using three different systems:
- The color that it actually is
- The material that uses the color (outdated/abbreviated names)
- The block that uses the color

At the very least they should all follow a unified naming scheme, of which there are three practical ones:
- Naming color based on the color it actually is
    - Bikeshed time
    - Difficult as we have about 12 shades of brown.
- Based on the blocks that use it
    - More work to compile, maintain, and leads to names similar to the new Material one's, but with more mess as color is _per blockstate_, so this is likely impossible to do well.
- The hex value of the color
    -  Each rename is simple, but now each name is also useless unless you can interpret hex color codes.

The first option would be the most reasonable to use. We can also "cheat" a little and refer to MC textures, though that will require a Javadoc disclaimer (eg Emerald Green doesn't mean dark green, but MC's emerald color). Slightly complicated by MC's naming being off for some colors. Naming them in terms of *blockReference_baseColor* would make things more uniform and easier to understand, but would mean 16 more renames for the terracotta colors.

Here are all the color's hex values, and setup so Idea will show you the color on the side: https://pastebin.com/nTXgt3Dr

haykam821:
Partially fixes #1451

### DIFF
--- a/mappings/net/minecraft/block/MaterialColor.mapping
+++ b/mappings/net/minecraft/block/MaterialColor.mapping
@@ -31,7 +31,7 @@ CLASS net/minecraft/class_3620 net/minecraft/block/MaterialColor
 	FIELD field_16005 IRON Lnet/minecraft/class_3620;
 	FIELD field_16006 COLORS [Lnet/minecraft/class_3620;
 	FIELD field_16007 BLACK_TERRACOTTA Lnet/minecraft/class_3620;
-	FIELD field_16008 AIR Lnet/minecraft/class_3620;
+	FIELD field_16008 CLEAR Lnet/minecraft/class_3620;
 	FIELD field_16009 BLACK Lnet/minecraft/class_3620;
 	FIELD field_16010 YELLOW Lnet/minecraft/class_3620;
 	FIELD field_16011 color I


Prospector:
I sorta think TRANSPARENT would be better than CLEAR but we'll see what others say

liach (approved):

liach:
This should be `CLEAR` if it's used by like structure void (beyond 3 air), `AIR` if only used by void_air, cave_air, and air

dexman545 (approved):
Structure Void, Air, and Glass all use this, but the main point is that this color (pitch black) is not rendered on maps, so "clear" and it's various synonyms all work for this.

haykam821:
The `AIR`, `STRUCTURE_VOID`, `PORTAL`, `FIRE`, `SUPPORTED`, `REDSTONE_LAMP`, `GLASS`, `BARRIER`, and `CAKE` materials use this material color.

liach:
@dexman545 didn't you say that the map color isn't bound to material at all?

Prospector:
Actually, I think NONE would be the best, thinking about it more.

NONE > TRANSPARENT > CLEAR

dexman545:
@liach it isn't. Before 1.13, the Material of a block determined what color it would be on a map. This is no longer the case. 
All materials have a color, but not all blocks get their color from the material. MapColor is *blockstate* dependent. You should think of the value in Material as a default color for the block - one that can be overridden at any time, and by any block, which is often the case.

liach:
Yeah `NONE` may be good too

dexman545:
I'd go for `TRANSPARENT` over `NONE` as the `FilledMapItem` logic checks if the top color is AIR, then continues looking at lower y-levels for one that isn't AIR.

EDIT: NONE could also be confused with an empty collum (void), which this isn't used for.

liach:
Then I'd still support `CLEAR` over `TRANSPARENT`

haykam821:
Would `IGNORED` be good?

Prospector (approved):
tbh, clear is fine.

YanisBft (approved):

PR #1491, opened 2020-06-20
Map chunk and chunk statuses

i509VCB:


mappings/net/minecraft/block/Block.mapping
```diff
@@ -25,6 +25,12 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 	METHOD method_23350 getJumpVelocityMultiplier ()F
 	METHOD method_27839 is (Lnet/minecraft/class_2248;)Z
 		ARG 1 block
+	METHOD method_30094 replaceBlock (Lnet/minecraft/class_2680;Lnet/minecraft/class_2680;Lnet/minecraft/class_1936;Lnet/minecraft/class_2338;I)V
```
liach, line 28: `replaced`?

mappings/net/minecraft/world/chunk/ProtoChunk.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS net/minecraft/class_2839 net/minecraft/world/chunk/ProtoChunk
+	COMMENT
```
liach:
```suggestion
```

i509VCB: Hmm, I thought I purged all the orphaned comments

mappings/net/minecraft/block/Block.mapping
liach: If this is called when this block is replaced, I'd suggest
```suggestion
	METHOD method_30094 replaced (Lnet/minecraft/class_2680;Lnet/minecraft/class_2680;Lnet/minecraft/class_1936;Lnet/minecraft/class_2338;I)V
```

i509VCB: This is the exact same name as the other method that also passes the maximum depth. Unless we want to change that one also (not an issue tbf)

liach: Hmm, how often is this method overridden by mods? Need to weigh the benefit and the cost.

But replaceBlock sounds like this block is replacing other block than this block is getting replaced.

i509VCB:
I think I have to make sure this doesn't end up the same as #1493

mappings/net/minecraft/block/Block.mapping
i509VCB: This is a static method

mappings/net/minecraft/world/ChunkRegion.mapping
```diff
@@ -9,14 +9,14 @@ CLASS net/minecraft/class_3233 net/minecraft/world/ChunkRegion
 	FIELD field_14094 fluidTickScheduler Lnet/minecraft/class_1951;
 	FIELD field_14096 dimension Lnet/minecraft/class_2874;
 	FIELD field_14097 levelProperties Lnet/minecraft/class_5217;
-	FIELD field_14098 chunks Ljava/util/List;
+	FIELD field_14098 surroundingChunks Ljava/util/List;
```
Gegy, line 12: `surroundingChunks` in this context feels like it implies only the edge chunks. I think `chunks` might be descriptive enough?

i509VCB: maybe neighboring chunks?

Gegy: In the context of `ChunkStatus` that makes sense, but I think `ChunkRegion` is used more generically. It's just all the chunks contained within the area. If `chunks` doesn't work, maybe `containedChunks`/`includedChunks` or something along those lines?

i509VCB: On further thought, `chunks` should be fine

i509VCB:
Looks like I have to reopen this because of detached branch

liach:
Yeah just create a new fork and push this branch to the new fork's remote and open a new pr

i509VCB:
Reattached at #1511

PR #1502, opened 2020-06-24
Map TaskExecutor and MessageListener around actor model

Gegy:
The systems relevant here very closely represent an actor model. This PR updates the implementations and usages of `MessageListener` (`Actor`) and `TaskExecutor` (`ReschedulingActor`) to better reflect that model.
This is already referenced once in `ChunkTaskPrioritySystem`: https://github.com/FabricMC/yarn/blob/a8502591b0c69e46676f63245b2867f39f86ba5d/mappings/net/minecraft/server/world/ChunkTaskPrioritySystem.mapping#L4

`ReschedulingActor` is an implementation that runs its tasks by rescheduling itself to a java `Executor` whenever a new task is available. It specifically only handles `T extends Runnable`, so I'm conflicted on whether it should be named rather according to that (e.g. `TaskHandlingActor` or similar)


modmuss50 (approved):

YanisBft:
@Gegy any news about this?

Gegy:
I'm not super happy with these names- I may pick this up again in the future.

PR #1510, opened 2020-06-24
ItemUsageContext -> BlockUsageContext + related names

### LINKED ISSUE #1508: ItemUsageContext inconsistencies
The Vec3d position getter is called `getHitPos`, but the BlockPos one is just `getBlockPos` with no mention of `hit`. In my opinion, the `Hit` prefix in the double one should just be removed.

In addition, the class is more of a `BlockUsageContext` as it's only used when blocks are used with items. The same goes for `ItemPlacementContext`, which is really a `BlockPlacementContext` -- you can't place an item.

Juuxel:
Closes #1508.

The context is for using *blocks*, not items; it's only used when blocks are placed or blocks are used with items. It also doesn't make much sense to "place items" in Minecraft.

As this is a large rename, also specifically requesting a review from @modmuss50.

### DIFF
--- a/mappings/net/minecraft/block/AutomaticBlockPlacementContext.mapping
+++ b/mappings/net/minecraft/block/AutomaticBlockPlacementContext.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_2968 net/minecraft/item/AutomaticItemPlacementContext
+CLASS net/minecraft/class_2968 net/minecraft/block/AutomaticBlockPlacementContext
 	FIELD field_13362 facing Lnet/minecraft/class_2350;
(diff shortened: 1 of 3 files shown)


liach:
Is this called by item form of blocks?

Runemoro:
Placing a block is not the same as using a block. Using a block is right-clicking on a block that's already placed.

It should either be `BlockPlacementContext` or `BlockItemUsageContext`

Juuxel:
There are both. The one for using blocks with items (currently `ItemUsageContext`) is the *superclass* of the one used for block placement (currently `ItemPlacementContext`).

liach (approved):

Juuxel:
Can't be bothered to maintain this right now.

PR #1518, opened 2020-06-26
Amend prohibited mapping sources

i509VCB:
All of spigot's files related to their mappings have the following in their header: `(c) 2014-2020 SpigotMC Pty. Ltd.` This file header is effectively `All Rights Reserved` due to no legal language on the file or repository specifying any other licensing. 

Per the file header of spigot's mappings (albeit minimal mappings) is incompatible with yarn's license and as such cannot be used.

Further discussion about Paper, which uses the same mappings as spigot in their patches for the root of their project should be discussed. Paper's `OBFHELPER` commented method renames are pseudo dual licensed (LGPL or MIT) depending on the author of the patch (as specified by their license file: https://github.com/PaperMC/Paper/blob/master/LICENSE.md). Though, these may not be compatable with yarn's licensing with the LGPL or MIT licensed patches, or because they have been lifted from other projects (MCP, Spigot, or Mojang's mappings) that are incompatible with yarn's licensing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ To obtain a deobfuscated Minecraft jar, [`./gradlew mapNamedJar`](#mapNamedJar) 
 
 ## Contributing
 
-Please remember that copying and pasting mappings from alternate projects under more restrictive licenses (such as MCP or Mojang's obfuscation maps)
+Please remember that copying and pasting mappings from alternate projects under more restrictive licenses (such as MCP, Spigot's or Mojang's obfuscation maps)
 is **completely forbidden** without explicit permission from the owners of said mappings to distribute the names under the CC0 license.
 This includes using the names from those mappings for inspiration. Discussing the naming approaches used in said projects
 is also not welcome - you have been warned. However, it is a good idea to consult name changes with other people - use pull requests or our community spaces to ask questions!


Earthcomputer (approved):
btw im not a lawyer

i509VCB:
Neither am I lol.

asiekierka:
Not a lawyer, not legal advice.

IMO wholesale copying from Spigot/Paper mappings should be treated just like MCP or Mojang; given that we try to be public domain, we must impose no additional restrictions, not even light ones like MIT would.

zml2008:
i doubt this will affect the content of yarn -- spigot mappings are similar in value to just randomly mashing up minecrafty words with strings in files, have very limited coverage, and don't target snapshots in any publicly visible form.

paper mappings/obf helpers are probably more of a concern -- from what i gather many of those that aren't drawn from yarn or made up by authors are done with reference to Mojang's mappings.

liach:
You should also suggest sourcing like wiki.vg is even not allowed than just spigot.

i509VCB:
With the length of the prohibited stuff increasing, maybe a table describing why it isn't allow would be of benefit?

liach:
In my opinion, our end is to derive names based on evidence we can find when we look at the intermediary/official obf names in enigma. Looking at any other source is totally wrong, no matter what their license is.

Player3324 (approved):
It is a non-exhaustive list, mentioning the popular ones like this is fine

YanisBft (approved):

modmuss50 (approved):

PR #1521, opened 2020-06-27
Made it more explicit why one would want to run the yarn task

shartte:
Since I complained on Discord about initially not knowing why I should run the yarn task.
Here's what would have helped me.

(And yes, I fully admit, my attention span is .. uhm... what was I talking about?)


liach (approved):

YanisBft (approved):

LambdAurora (approved):

ramidzkh (approved):
So no profit? /s

Juuxel (approved):

PR #1524, opened 2020-06-27
Fix realms packages to be singular screen, map some other bits of it

i509VCB:


liach:
honestly we should move these into net/minecraft/client/gui/screen/realms instead, given now these extend vanilla screen directly than using some sort of handle.

i509VCB:
Yeah I could see the reasoning for truncating it to main screen package.

mappings/com/mojang/realmsclient/util.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_4429 com/mojang/realmsclient/gui/screens/UploadResult
+CLASS net/minecraft/class_4429 com/mojang/realmsclient/util
```
YanisBft, line 1: Bad rename here

mappings/com/mojang/realmsclient/gui/screen/RealmsBackupScreen.mapping
```diff
@@ -13,6 +13,8 @@ CLASS net/minecraft/class_4381 com/mojang/realmsclient/gui/screens/RealmsBackupS
 	FIELD field_19752 noBackups Ljava/lang/Boolean;
 	FIELD field_19753 serverData Lnet/minecraft/class_4877;
 	FIELD field_19754 titleLabel Lnet/minecraft/class_4903;
+	FIELD field_22686 PLUS_ICON_TEXTURE Lnet/minecraft/class_2960;
+	FIELD field_22687 RESTORE_ICON_TEXTURE Lnet/minecraft/class_2960;
```
YanisBft, line 17:
```suggestion
	FIELD field_22686 PLUS_ICON Lnet/minecraft/class_2960;
	FIELD field_22687 RESTORE_ICON Lnet/minecraft/class_2960;
```

mappings/com/mojang/realmsclient/util.mapping
liach: Ah, explains why java 11/14 build glitches out

liach:
marking wip as there is broken renames.

i509VCB:
I need to do this much more incrementally. This PR is HUGE. I'll open up a bunch of smaller PRs as time goes on

PR #1527, opened 2020-06-28
A couple of goal fixes

Earthcomputer:


liach (approved):

YanisBft (approved):

mappings/net/minecraft/entity/ai/goal/WanderNearTargetGoal.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_1369 net/minecraft/entity/ai/goal/GoToEntityTargetGoal
+CLASS net/minecraft/class_1369 net/minecraft/entity/ai/goal/WanderNearTargetGoal
```
liach, line 1: So this one wanders toward target than wanders around the target? If it's towards rune's suggestion is right

liach (approved):

mappings/net/minecraft/entity/ai/goal/WanderNearTargetGoal.mapping
Earthcomputer: `WanderNearTarget` can be taken 2 ways, both of which are correct. There is a hint of "towards" if the mob is far away, and there is a hint of "around" if the mob is close. This reflects exactly what the task does.

modmuss50 (approved):

PR #1528, opened 2020-06-29
Block.replaced/replaceBlock -> replace

Juuxel:
The two methods are really overloads of the same method, one just has a default parameter (`maxUpdateDepth` set to 512). I also added docs for the methods.

### DIFF
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -25,7 +25,7 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 	METHOD method_23350 getJumpVelocityMultiplier ()F
 	METHOD method_27839 is (Lnet/minecraft/class_2248;)Z
 		ARG 1 block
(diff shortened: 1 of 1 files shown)


liach:
Imo we should rename all `replaceBlock` to `replaced`. This is called when `this` block is replaced, not when `this` block replace some other block.

Juuxel:
There's no `this` since the method is static.

liach:
so how do the two methods (this method itself and the delegated method) look like?

Juuxel:
The one with fewer parameters (currently `replaced`) is just a call to the longer method with the last parameter defaulted to 512.

If `state != newState`, `replaceBlock` replaces the block at the position with `newState`. If `newState` is air, it will call `world.breakBlock`; otherwise, it will just call `world.setBlockState`.

liach:
then both methods should be called `replace`.

Juuxel:
@liach I changed them to `replace` and also added docs for the methods.

liach (approved):

YanisBft (approved):

Gegy (approved):

PR #1531, opened 2020-06-30
Drag and Drop Datapack Screen

ghost:
I added the mapping for the method responsible for opening the drag and drop confirmation screen when adding a datapack to the world creation screen.

### DIFF
--- a/mappings/net/minecraft/client/gui/screen/Screen.mapping
+++ b/mappings/net/minecraft/client/gui/screen/Screen.mapping
@@ -100,3 +100,5 @@ CLASS net/minecraft/class_437 net/minecraft/client/gui/screen/Screen
 	METHOD method_25441 hasControlDown ()Z
 	METHOD method_25442 hasShiftDown ()Z
 	METHOD method_25443 hasAltDown ()Z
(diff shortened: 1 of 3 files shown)


mappings/net/minecraft/client/gui/screen/Screen.mapping
```diff
@@ -100,3 +100,5 @@ CLASS net/minecraft/class_437 net/minecraft/client/gui/screen/Screen
 	METHOD method_25441 hasControlDown ()Z
 	METHOD method_25442 hasShiftDown ()Z
 	METHOD method_25443 hasAltDown ()Z
+	METHOD method_29638 dragPackConfirmation (Ljava/util/List;)V
```
liach, line 103: Are you sure this thing in `Screen` only handles pack drag and drop?

liach (changes requested):

modmuss50 (changes requested):
This seems to be an incorrect back port? 1.16.2 already has these names so ideally they should be the same here if wanted for 1.16.1. Im not sure if this PR was opened before 1.16.2 got the names.

The only name here thats not in 1.16.2 is copyPackToWorld, and im not sure that is correct either.

PR #1535, opened 2020-07-01
Map two methods in GeneratorType

YanisBft:


liach:
Hmm, what exactly is `first biome`?

YanisBft:
> Hmm, what exactly is `first biome`?

It gets the first biome in the biome list of biome source

liach (approved):

Gegy:
Hm, how is the first biome value used? Is it used in some more specific context, or is it totally arbitrary?

YanisBft:
> Hm, how is the first biome value used? Is it used in some more specific context, or is it totally arbitrary?

It's used when generating buffet worlds to get the only biome

Gegy (approved):

PR #1538, opened 2020-07-01
20w27a recipe book

liach:

### DIFF
--- a/mappings/net/minecraft/network/listener/ServerPlayPacketListener.mapping
+++ b/mappings/net/minecraft/network/listener/ServerPlayPacketListener.mapping
@@ -88,5 +88,5 @@ CLASS net/minecraft/class_2792 net/minecraft/network/listener/ServerPlayPacketLi
 		ARG 1 packet
 	METHOD method_27273 onJigsawGenerating (Lnet/minecraft/class_5194;)V
 		ARG 1 packet
-	METHOD method_30303 onRecipeBookOption (Lnet/minecraft/class_5427;)V
+	METHOD method_30303 onRecipeCategoryOption (Lnet/minecraft/class_5427;)V
 		ARG 1 packet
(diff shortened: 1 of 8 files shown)


mappings/net/minecraft/network/listener/ServerPlayPacketListener.mapping
```diff
@@ -88,3 +88,5 @@ CLASS net/minecraft/class_2792 net/minecraft/network/listener/ServerPlayPacketLi
 		ARG 1 packet
 	METHOD method_27273 onJigsawGenerating (Lnet/minecraft/class_5194;)V
 		ARG 1 packet
+	METHOD method_30303 onRecipeBookOption (Lnet/minecraft/class_5427;)V
```
Gegy: Should this maybe be like `onSetRecipeBookOptions`? (and the packet name too?)

liach: hmm, this packet sends the recipe book option for one recipe category (i.e. crafting, furnace, etc.)

imo the `set` part of the name you propose is extraneous.

Gegy: Okay right, I see. I think it might still make sense to have this be plural though, because there are still multiple options per category. In the case of the `set`, this is mainly motivated because `onRecipeBookOption` reads a bit weirdly with no verb

liach: So like `RecipeCategoryOption` then? it sends the option for like crafting, like two booleans of if the screen is open and if the filter craftable is on

Gegy: That’d be better, but I think it still would have the problems I mentioned before without the set and without the plural. Probably rather get a third opinion on this :)

haykam821: `onRecipeCategoryOption` seems good to me

haykam821 (approved):

mappings/net/minecraft/network/listener/ServerPlayPacketListener.mapping
Earthcomputer: I still think the plural is better here. There are two options being set, gui open and filtering craftable. 2 options in 1 category is "category options".

Earthcomputer (approved):

YanisBft (approved):

PR #1539, opened 2020-07-01
Map new tag logic.

i509VCB:
Much of the logic seems to have been seperated away from the `TagContainer` into a sort of `TagManager` now.

### DIFF
--- a/mappings/net/minecraft/tag/GlobalTagAccessor.mapping
+++ b/mappings/net/minecraft/tag/GlobalTagAccessor.mapping
@@ -3,11 +3,25 @@ CLASS net/minecraft/class_5120 net/minecraft/tag/GlobalTagAccessor
 	COMMENT tag changes through reloads/server tag sends.
 	FIELD field_23682 currentContainer Lnet/minecraft/class_5414;
 	FIELD field_23683 tags Ljava/util/List;
+	FIELD field_25740 managerGetter Ljava/util/function/Function;
+	METHOD <init> (Ljava/util/function/Function;)V
+		ARG 1 managerGetter
 	METHOD method_26770 getContainer ()Lnet/minecraft/class_5414;
 	METHOD method_26771 setContainer (Lnet/minecraft/class_5415;)V
+		ARG 1 manager
+	METHOD method_26772 (Lnet/minecraft/class_5414;Lnet/minecraft/class_5120$class_5121;)V
+		ARG 1 tag
 	METHOD method_26773 get (Ljava/lang/String;)Lnet/minecraft/class_3494$class_5123;
 		ARG 1 id
 	METHOD method_27061 markReady ()V
+	METHOD method_27062 (Lnet/minecraft/class_3494;Lnet/minecraft/class_5120$class_5121;)V
+		ARG 1 tag
+	METHOD method_27063 (Lnet/minecraft/class_3494;Lnet/minecraft/class_2960;)Lnet/minecraft/class_3494;
+		ARG 1 id
+	METHOD method_29224 getTagDifferences (Lnet/minecraft/class_5415;)Ljava/util/Set;
+		COMMENT Gets a set of identifiers which corresponds to tags not present in this tag accessor and another tag accessor.
+		ARG 1 manager
+	METHOD method_29902 getTags ()Ljava/util/List;
 	CLASS class_5121 CachedTag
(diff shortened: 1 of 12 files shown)


mappings/net/minecraft/tag/GlobalTagAccessor.mapping
```diff
@@ -3,11 +3,25 @@ CLASS net/minecraft/class_5120 net/minecraft/tag/GlobalTagAccessor
...
+	METHOD method_26772 (Lnet/minecraft/class_5414;Lnet/minecraft/class_5120$class_5121;)V
+		ARG 1 tag
 	METHOD method_26773 get (Ljava/lang/String;)Lnet/minecraft/class_3494$class_5123;
 		ARG 1 id
 	METHOD method_27061 markReady ()V
+	METHOD method_27062 (Lnet/minecraft/class_3494;Lnet/minecraft/class_5120$class_5121;)V
+		ARG 1 tag
+	METHOD method_27063 (Lnet/minecraft/class_3494;Lnet/minecraft/class_2960;)Lnet/minecraft/class_3494;
+		ARG 1 id
+	METHOD method_29224 getTagDifferences (Lnet/minecraft/class_5415;)Ljava/util/Set;
```
liach, line 21: This is really only reporting the elements that the required accessors has and the registry tag manager doesn't have. so this is like `getMissingTags`

liach (changes requested):
I really can't agree with this... what matcher retained as `TagContainer` is really the tag loader. What you mapped as `TagManager` is the actual tag container for each type of tag.

Also `GlobalTagAccessor` has some change that is big enough for a reevaluation as well. I'd suggest re-look into tags than basing off on previous works and matcher's matches.

mappings/net/minecraft/tag/GlobalTagAccessorRegistry.mapping
```diff
@@ -0,0 +1,18 @@
...
+	METHOD method_30199 (Lnet/minecraft/class_5415;Lnet/minecraft/class_5120;)V
+		ARG 1 accessor
+	METHOD method_30200 (Lcom/google/common/collect/Multimap;Lnet/minecraft/class_5415;Lnet/minecraft/class_2960;Lnet/minecraft/class_5120;)V
+		ARG 2 id
+		ARG 3 accessor
+	METHOD method_30201 registerAccessor (Lnet/minecraft/class_2960;Ljava/util/function/Function;)Lnet/minecraft/class_5120;
+		ARG 0 id
+		ARG 1 managerGetter
+	METHOD method_30202 validateRegistrations ()V
+	METHOD method_30203 getDifferences (Lnet/minecraft/class_5415;)Lcom/google/common/collect/Multimap;
```
Earthcomputer, line 17: Same as liach's comment here. It's `getMissingTags`. That's especially obvious if you look at the call hierarchy of this method.

mappings/net/minecraft/tag/RegistryTagManager.mapping
```diff
@@ -1,5 +1,16 @@
-CLASS net/minecraft/class_3505 net/minecraft/tag/RegistryTagManager
-	FIELD field_15612 blocks Lnet/minecraft/class_3503;
-	FIELD field_15613 items Lnet/minecraft/class_3503;
-	FIELD field_15614 fluids Lnet/minecraft/class_3503;
-	FIELD field_15615 entityTypes Lnet/minecraft/class_3503;
+CLASS net/minecraft/class_5415 net/minecraft/tag/RegistryTagManager
+	FIELD field_25744 EMPTY_TAG_MANAGER Lnet/minecraft/class_5415;
```
Earthcomputer, line 2:
```suggestion
	FIELD field_25744 EMPTY Lnet/minecraft/class_5415;
```

Earthcomputer (changes requested):
Several problems with this PR.
- You have named a class `RegistryTagManager` which contains multiple `TagManager`s. This is extremely confusing, the previous naming of `RegistryTagManager` containing `TagContainer`s was better than this, although I'm open to suggestions for better names.
- Does not rename most of the fields and methods in reference to the `TagContainer` -> `TagManager` rename.
- Leaves a class called `TagContainer` which now only has to do with tag deserializing.
- Does not rename anything in `SetTag`.

In addition, some other minor comments:

Earthcomputer:
Superseded by #1543

PR #1545, opened 2020-07-04
Adjustments to render layers

williewillus:
This is my first Yarn PR, so please point out any mistakes. 

This PR has adjusts some names related to render layers. I'll go through by class.

RenderLayers:
- Map the second bool parameter to numerous methods as `direct`. When true, the layer returned is the "direct" version, always rendering to the main framebuffer. When false, the newly-added off-screen framebuffers are returned when necessary.
- Name `getMovingBlockLayer`, used by moving piston and falling block renderers
- Confidence in these names: high

TexturedRenderLayers:
- map field `ITEM_ENTITY_TRANSLUCENT_CULL` and its getter according to their ID's
- Confidence: high

ModelTransformation:
- name isFirstPerson. straightforward.
- Confidence: very high

Now we have ItemRenderer:
1. Rename `method_23181` to `getGlintVertexConsumer`. The former name of `getArmorVertexConsumer` is a misnomer now as this is called any time enchantment glint is needed. Nothing to do with armor.
2. Map `method_23181`'s "direct" sibling `method_29711` to `getDirectGlintVertexConsumer`. The meaning of "direct" is as above.
3. Map `method_27952` as `getArmorVertexConsumer`, since this one is the one actually called by armor rendering
4. Map `method_30114` and its "direct" sibling `method_30115` as `getTransformingGlintVertexConsumer` (resp. `getTransformingDirectGlintVertexConsumer`). These are only used for enchanted compass rendering, but I decided to keep them general. Might be a bit long?

Confidence: high for items 1 and 2. medium-high for item 3 (only because the old name will silently change methods). medium for #4 (due to the name length).

Should apply cleanly to the snapshot but if not I'll forward port on merge.

### DIFF
--- a/mappings/net/minecraft/client/render/RenderLayers.mapping
+++ b/mappings/net/minecraft/client/render/RenderLayers.mapping
@@ -4,6 +4,7 @@ CLASS net/minecraft/class_4696 net/minecraft/client/render/RenderLayers
 	FIELD field_21472 fancyGraphicsOrBetter Z
 	METHOD method_23678 getItemLayer (Lnet/minecraft/class_1799;Z)Lnet/minecraft/class_1921;
 		ARG 0 stack
+		ARG 1 direct
 	METHOD method_23679 getBlockLayer (Lnet/minecraft/class_2680;)Lnet/minecraft/class_1921;
 		ARG 0 state
 	METHOD method_23680 getFluidLayer (Lnet/minecraft/class_3610;)Lnet/minecraft/class_1921;
@@ -12,3 +13,5 @@ CLASS net/minecraft/class_4696 net/minecraft/client/render/RenderLayers
 		ARG 0 fancyGraphicsOrBetter
 	METHOD method_23683 getEntityBlockLayer (Lnet/minecraft/class_2680;Z)Lnet/minecraft/class_1921;
 		ARG 0 state
+		ARG 1 direct
+	METHOD method_29359 getMovingBlockLayer (Lnet/minecraft/class_2680;)Lnet/minecraft/class_1921;
--- a/mappings/net/minecraft/client/render/item/ItemRenderer.mapping
+++ b/mappings/net/minecraft/client/render/item/ItemRenderer.mapping
@@ -42,7 +42,7 @@ CLASS net/minecraft/class_918 net/minecraft/client/render/item/ItemRenderer
 		ARG 4 stack
 		ARG 5 light
 		ARG 6 overlay
-	METHOD method_23181 getArmorVertexConsumer (Lnet/minecraft/class_4597;Lnet/minecraft/class_1921;ZZ)Lnet/minecraft/class_4588;
+	METHOD method_23181 getGlintVertexConsumer (Lnet/minecraft/class_4597;Lnet/minecraft/class_1921;ZZ)Lnet/minecraft/class_4588;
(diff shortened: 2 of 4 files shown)


mappings/net/minecraft/client/render/RenderLayers.mapping
```diff
@@ -4,6 +4,7 @@ CLASS net/minecraft/class_4696 net/minecraft/client/render/RenderLayers
 	FIELD field_21472 fancyGraphicsOrBetter Z
 	METHOD method_23678 getItemLayer (Lnet/minecraft/class_1799;Z)Lnet/minecraft/class_1921;
 		ARG 0 stack
+		ARG 1 direct
```
i509VCB, line 7: This might be a scope creep on the PR, but it would be nice if you could document the `direct` parameter and explain what it does like the PR description which was pretty clear.

You don't need to do it to every render layer as we can get those other ones in later PRs.

mappings/net/minecraft/client/render/item/ItemRenderer.mapping
```diff
@@ -62,11 +62,18 @@ CLASS net/minecraft/class_918 net/minecraft/client/render/item/ItemRenderer
...
 	METHOD method_27953 renderInGui (Lnet/minecraft/class_1799;II)V
 		COMMENT Renders an item in a GUI without an attached entity.
 		ARG 1 stack
 		ARG 2 x
 		ARG 3 y
+	METHOD method_29711 getDirectGlintVertexConsumer (Lnet/minecraft/class_4597;Lnet/minecraft/class_1921;ZZ)Lnet/minecraft/class_4588;
+		ARG 1 layer
+		ARG 3 glint
+	METHOD method_30114 getTransformingGlintVertexConsumer (Lnet/minecraft/class_4597;Lnet/minecraft/class_1921;Lnet/minecraft/class_4587$class_4665;)Lnet/minecraft/class_4588;
+	METHOD method_30115 getTransformingDirectGlintVertexConsumer (Lnet/minecraft/class_4597;Lnet/minecraft/class_1921;Lnet/minecraft/class_4587$class_4665;)Lnet/minecraft/class_4588;
```
i509VCB, line 76: If you could, it would be nice to describe how the `direct` one renders straight to the frame buffer.

i509VCB (approved):
Generally looks good. It would be nice to document those `direct` bits but it isn't required.

mappings/net/minecraft/client/render/RenderLayers.mapping
liach: I think this one looks good.

mappings/net/minecraft/client/render/item/ItemRenderer.mapping
liach: Imo that's extraneous.

liach (approved):

williewillus:
I'd like to document `direct` too, but that probably belongs in a higher-level (class-level) javadoc like on RenderLayers itself, not sure if it needs to go on the individual methods

i509VCB:
Yeah I could see `direct` being defined at a higher level. PR is pretty much good then

Gegy (approved):

YanisBft (approved):

modmuss50 (approved):

shartte (approved):
LGTM

PR #1549, opened 2020-07-06
Consoldiate all realms classes to net.minecraft.client.realms

i509VCB:
The client package was chosen since realms is only present in the client jar, hence all of it's classes are client only.


liach (commented):
I doubt if we should merge the net/minecraft/realms package with the com/mojang/realmsclient package.

i509VCB:
imo it makes more sense in net/minecraft.

Also realms is no longer a separate library, we can move it around a bit

liach (approved):
https://github.com/i509VCB/dmcyarn/tree/realmspkg/mappings/net/minecraft/client/realms
Looking good :+1:

Gegy (approved):

YanisBft (approved):

PR #1551, opened 2020-07-07
Mapped missing arg in Input.mapping.

andrew121410:
In ClientPlayerEntity it says this.input.tick(this.isHoldingSneakKey()); so that's how I got the name.

### DIFF
--- a/mappings/net/minecraft/client/input/Input.mapping
+++ b/mappings/net/minecraft/client/input/Input.mapping
@@ -10,3 +10,4 @@ CLASS net/minecraft/class_744 net/minecraft/client/input/Input
 	METHOD method_20622 hasForwardMovement ()Z
 	METHOD method_3128 getMovementInput ()Lnet/minecraft/class_241;
 	METHOD method_3129 tick (Z)V
+		ARG 1 isHoldingSneakKey


i509VCB (approved):

mappings/net/minecraft/client/input/Input.mapping
```diff
@@ -10,3 +10,4 @@ CLASS net/minecraft/class_744 net/minecraft/client/input/Input
 	METHOD method_20622 hasForwardMovement ()Z
 	METHOD method_3128 getMovementInput ()Lnet/minecraft/class_241;
 	METHOD method_3129 tick (Z)V
+		ARG 1 isHoldingSneakKey
```
liach, line 13:
```suggestion
		ARG 1 slowdown
```

liach (changes requested):
Turns out `isHoldingSneakyKey` is a wrong name.
```java
    public boolean isHoldingSneakKey() {
        return this.isInSneakingPose() || this.shouldLeaveSwimmingPose();
    }
```
That method should be `shouldSlowDown` instead.

andrew121410:
Well damn at least I tried 😅😥

liach:
Uh, you can fix that method and apply my suggestion in this pr and we can merge it in that case. feel free to open a new one that fixes this issue

PR #1552, opened 2020-07-07
Structure Placement

shartte:
I tried documenting how structure placement works in the Javadocs.
Debatable: StructureConfig seems like it's more like a StructurePlacementConfig to me, but I didn't rename it in this.

I've also started using the terms "placement grid" and "grid cell" because I think it describes pretty well how the algorithm works.

A bit of background info on the placement.

Given spacing=32, which means one structure placed per 32x32 chunks, the following shows where within that 32x32 chunks the structure could be placed (samples=1000):

image

Since the charts might need an explanation: I applied the spawn-position algorithm 1000 times for a single cell in the placement grid (32x32 chunks in this example), with various types of parameters. Each pixel in the chart indicates how many structures (out of 1000) ended up in that particular chunk. Darker = More, Brighter = Fewer.


mappings/net/minecraft/world/gen/chunk/StructureConfig.mapping
```diff
@@ -1,8 +1,29 @@
 CLASS net/minecraft/class_5314 net/minecraft/world/gen/chunk/StructureConfig
+	COMMENT Contains parameters for placement of a single type of {@link net.minecraft.world.gen.feature.StructureFeature} during chunk
+	COMMENT generation.
 	FIELD field_24917 CODEC Lcom/mojang/serialization/Codec;
 	FIELD field_24918 spacing I
+		COMMENT Defines how many chunks make up one cell in the grid used to place structures.
```
Gegy: Might be a bit confusing here- maybe rather:
```suggestion
		COMMENT Defines the width in chunks of one cell in the grid used to place structures.
```

shartte: It's also the height though :-P
Should probably say dimensions (in chunks) or something similar hmmm.

mappings/net/minecraft/world/gen/chunk/StructureConfig.mapping
```diff
@@ -1,8 +1,29 @@
...
+		COMMENT Each cell in the grid will receive one potential structure start.
+		COMMENT <p>
+		COMMENT Minimum is 1, which means the structure will potentially be placed in every chunk.
+		COMMENT 2 leads to one structure per 2x2 chunks, and so on.
 	FIELD field_24919 separation I
+		COMMENT Defines how varied the placement of a structure within the chunk grid (see spacing) is.
```
Gegy: It might help to define separation along the lines of being the minimum distance in chunks between structures

mappings/net/minecraft/world/gen/chunk/StructureConfig.mapping
Gegy: True :P 
`size` could potentially work here also

mappings/net/minecraft/world/gen/chunk/StructureConfig.mapping
shartte: I tried to improve this javadoc and also made a comment about it effectively being a margin that is subtracted from the width/height when actually placing the structure start.

mappings/net/minecraft/world/gen/chunk/StructureConfig.mapping
shartte: Changed it to "Defines the width and height of a cell in the structure placement grid in chunks."

liach (approved):

mappings/net/minecraft/world/gen/chunk/ChunkGenerator.mapping
```diff
@@ -4,6 +4,10 @@ CLASS net/minecraft/class_2794 net/minecraft/world/gen/chunk/ChunkGenerator
 	COMMENT Biome placement starts here, however all vanilla and most modded chunk generators delegate this to a {@linkplain net.minecraft.world.biome.source.BiomeSource biome source}.
 	FIELD field_12761 biomeSource Lnet/minecraft/class_1966;
 	FIELD field_16567 config Lnet/minecraft/class_5311;
+	FIELD field_24748 worldSeed_ J
```
shartte: Underscore sucks, but while it's private, Enigma won't let me name it the same as a field in the subclass.

liach:
```suggestion
	FIELD field_24748 worldSeed J
```
yeah you can do this hack here

liach (approved):
Dunno why but seems github action just broke again for the last commit

mappings/net/minecraft/structure/StructureStart.mapping
```diff
@@ -1,6 +1,16 @@
 CLASS net/minecraft/class_3449 net/minecraft/structure/StructureStart
+	COMMENT A structure start is created to describe a structure that will be generated by
+	COMMENT chunk generation. It contains a definition of its pieces and is associated
+	COMMENT with the chunk that the structure originates from.
 	FIELD field_15325 children Ljava/util/List;
 	FIELD field_15326 references I
+		COMMENT The number of chunks that inersect the structures bounding box,
```
Gegy:
```suggestion
		COMMENT The number of chunks that intersect the structures bounding box,
```

Player3324 (approved):

modmuss50 (approved):

PR #1560, opened 2020-07-08
data pack loaded registries util class

liach:
@modmuss50 requested at https://discordapp.com/channels/507304429255393322/521545796882006027/730484313065259078
Signed-off-by: liach <liach@users.noreply.github.com>

PR #1562, opened 2020-07-08
Builtin registries

YanisBft:
Maps the new dynamic registries
I know class_5458 is already covered by #1560 , it was just easier to name it before naming the other classes

### DIFF
--- a/mappings/net/minecraft/util/registry/DynamicRegistries.mapping
+++ b/mappings/net/minecraft/util/registry/DynamicRegistries.mapping
@@ -1,0 +1,9 @@
+CLASS net/minecraft/class_5458 net/minecraft/util/registry/DynamicRegistries
+	FIELD field_25925 LOGGER Lorg/apache/logging/log4j/Logger;
+	FIELD field_25927 CONFIGURED_SURFACE_BUILDER Lnet/minecraft/class_2378;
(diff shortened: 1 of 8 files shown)


Gegy (approved):

liach (approved):
Should be compatible with my pr after my rename

liach:
After another look by earthcomputer that class actually stores the hardcoded values for dynamic registries

i.e. mojang now keep dynamic registries at work in registry trackers, while this class is used for registry tracker to build dynamic registries and fill in builtin values

mappings/net/minecraft/util/registry/DynamicRegistries.mapping
```diff
@@ -0,0 +1,9 @@
+CLASS net/minecraft/class_5458 net/minecraft/util/registry/DynamicRegistries
```
liach: said to be `BuiltinRegistries` as there is a `missing built-in` message at where registry tracker uses  this class

liach (changes requested):

mappings/net/minecraft/util/registry/DynamicRegistries.mapping
YanisBft: renamed

Juuxel (approved):

liach (approved):
Thank you for fixing my typos :heart:

PR #1573, opened 2020-07-10
Some (hopefully) uncontroversial client.font mappings

Sturmlilie:
* `Font#method_27442` returns a set containing all the codepoints for which there are glyphs available in the Font. Alternatives: `getAvailableGlyphs`, `getContainedGlyphs`.
* `FontStorage#method_24290` just calls `close` on all fonts and then clears the list. Alternatives: `clearFonts`.
* Not sure whether `GlyphAtlasTexture#field_21691` should be `textSeeThroughLayer` (as in PR), which is consistent with `RenderLayer#getTextSeeThrough`, or `seeThroughTextLayer`.
* `GlyphRenderer` takes both of those layers in constructor so I named them the same here

The rest are mostly just parameters where the name was derived from the field they were assigned to, or obvious codepoint ints.

### DIFF
--- a/mappings/net/minecraft/client/font/Font.mapping
+++ b/mappings/net/minecraft/client/font/Font.mapping
@@ -1,2 +1,7 @@
 CLASS net/minecraft/class_390 net/minecraft/client/font/Font
 	METHOD method_2040 getGlyph (I)Lnet/minecraft/class_383;
+		ARG 1 codePoint
+	METHOD method_27442 getProvidedGlyphs ()Lit/unimi/dsi/fastutil/ints/IntSet;
+		COMMENT Returns the set of code points for which this font can provide glyphs.
+		COMMENT
+		COMMENT @return a set of integer code points.
--- a/mappings/net/minecraft/client/font/GlyphAtlasTexture.mapping
+++ b/mappings/net/minecraft/client/font/GlyphAtlasTexture.mapping
@@ -1,4 +1,6 @@
 CLASS net/minecraft/class_380 net/minecraft/client/font/GlyphAtlasTexture
+	FIELD field_21690 textLayer Lnet/minecraft/class_1921;
+	FIELD field_21691 seeThroughTextLayer Lnet/minecraft/class_1921;
 	FIELD field_2262 id Lnet/minecraft/class_2960;
 	FIELD field_2263 hasColor Z
 	FIELD field_2264 rootSlot Lnet/minecraft/class_380$class_381;
--- a/mappings/net/minecraft/client/font/GlyphRenderer.mapping
+++ b/mappings/net/minecraft/client/font/GlyphRenderer.mapping
@@ -1,4 +1,6 @@
 CLASS net/minecraft/class_382 net/minecraft/client/font/GlyphRenderer
+	FIELD field_21692 textLayer Lnet/minecraft/class_1921;
+	FIELD field_21693 seeThroughTextLayer Lnet/minecraft/class_1921;
 	FIELD field_2272 xMin F
 	FIELD field_2273 vMax F
 	FIELD field_2274 vMin F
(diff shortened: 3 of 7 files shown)


Sturmlilie:
Also, would it be possible to get this included in the 1.16.1 branch? Not sure if I want to hop onto the snapshots yet.

mappings/net/minecraft/client/font/GlyphAtlasTexture.mapping
```diff
@@ -1,4 +1,6 @@
 CLASS net/minecraft/class_380 net/minecraft/client/font/GlyphAtlasTexture
+	FIELD field_21690 textLayer Lnet/minecraft/class_1921;
+	FIELD field_21691 textSeeThroughLayer Lnet/minecraft/class_1921;
```
jaskarth: maybe `transparentTextLayer`?

mappings/net/minecraft/client/font/GlyphRenderer.mapping
```diff
@@ -1,4 +1,6 @@
 CLASS net/minecraft/class_382 net/minecraft/client/font/GlyphRenderer
+	FIELD field_21692 textLayer Lnet/minecraft/class_1921;
+	FIELD field_21693 textSeeThroughLayer Lnet/minecraft/class_1921;
```
jaskarth: ditto

mappings/net/minecraft/client/font/GlyphRenderer.mapping
```diff
@@ -23,6 +36,8 @@ CLASS net/minecraft/class_382 net/minecraft/client/font/GlyphRenderer
 		ARG 2 matrix
 		ARG 3 vertexConsumer
 		ARG 4 light
+	METHOD method_24045 getLayer (Z)Lnet/minecraft/class_1921;
+		ARG 1 seeThrough
```
jaskarth, line 40: `transparent`

jaskarth (changes requested):

jaskarth (approved):

mappings/net/minecraft/client/font/GlyphAtlasTexture.mapping
Sturmlilie: `seeThrough` is sourced from the layer name: https://github.com/FabricMC/yarn/blob/20w29a/mappings/net/minecraft/client/render/RenderLayer.mapping#L32

liach: maybe `seeThroughTextLayer`? Honestly we shouldn't fluctuate between prefix and suffix because mojang do

mappings/net/minecraft/client/font/GlyphRenderer.mapping
Sturmlilie: This would make sense if `seeThrough` had translucency enabled and the normal text layer didn't (in reality both do). However, the real difference is that `seeThrough` has depth test enabled, which implies that it's meant to be used in a 3D scene (as opposed to a 2D gui). The current name for this bool is purely naming related (ie. "give me the `see_through` layer instead of the normal one). If we want to redefine "seeThrough", it should in my opinion happen in a pure refactor PR that changes **all** instances of this naming scheme, not just in select places.

mappings/net/minecraft/client/font/GlyphAtlasTexture.mapping
Sturmlilie: @liach done.

liach (approved):
:heart:

mappings/net/minecraft/client/font/Font.mapping
```diff
@@ -1,2 +1,4 @@
 CLASS net/minecraft/class_390 net/minecraft/client/font/Font
 	METHOD method_2040 getGlyph (I)Lnet/minecraft/class_383;
+		ARG 1 codePoint
+	METHOD method_27442 getProvidedGlyphs ()Lit/unimi/dsi/fastutil/ints/IntSet;
```
shartte, line 4: What does this return? A set of the code points for which this font has glyphs?
Or a set of some integer related directly to the glyphs?

shartte: Ah forget about this comment. It's a nitpick and too much bikeshedding. getProvidedGlyphs should be fine

shartte (approved):

YanisBft (approved):

mappings/net/minecraft/client/font/Font.mapping
Sturmlilie: It's a fair point, because in the real world, glyphs and codepoints don't form a 1to1 relation. However in Minecraft they unfortunately do.

Sturmlilie:
This is "ready to be merged", but as shartte correctly noticed, the `getProvidedGlyphs` method isn't obvious at a glance, so I thought I'd push some yarn comments on top. Is that okay?

YanisBft:
Yes it is, please do :)

Sturmlilie:
I skimmed through the contribution guide again but couldn't find a section specifically on javadoc; hope what I wrote does more good than harm.

liach (approved):

Sturmlilie:
Will be force pushing to resolve.

Sturmlilie:
Should be functionally equal.

PR #1576, opened 2020-07-11
Corrected EyeOfEnderEntity mappings & comment

nskobelevs:
`EyeOfEnderEntity` has `velocityX`, `velocityY` and `velocityZ` properties in the mappings when these values are not velocities but the position to which the eye will fly to.

`lifespan` is also more appropriate instead of `useCount` as it's the number of ticks the entity has been alive for.

I'm still not sure about `initTargetPos` however it might be a bit better than `moveTowards` as the method itself doesn't actually do the moving.


nskobelevs:
I apologise, I branched of from 1.16.1 but the PR is into 20w28a

Pyrofab (approved):

Gegy (approved):

mappings/net/minecraft/entity/EyeOfEnderEntity.mapping
```diff
@@ -15,5 +15,12 @@ CLASS net/minecraft/class_1672 net/minecraft/entity/EyeOfEnderEntity
 	METHOD method_16934 (Lnet/minecraft/class_1799;)V
 		ARG 0 stack
 	METHOD method_16935 getTrackedItem ()Lnet/minecraft/class_1799;
-	METHOD method_7478 moveTowards (Lnet/minecraft/class_2338;)V
+	METHOD method_7478 initTargetPos (Lnet/minecraft/class_2338;)V
+		COMMENT Sets where the eye of ender will fly to.
+		COMMENT If within 12 horizontal blocks of the BlockPos, target is the block itself,
+		COMMENT     (Eye flying into the ground when close to the stronghold)
+		COMMENT otherwise, the target is 12 blocks in the horizontal direction of the BlockPos, and 8 blocks up.
+		COMMENT     (Eye flying up into the air)
```
YanisBft: Phrasing should be redone here

mappings/net/minecraft/entity/EyeOfEnderEntity.mapping
```diff
@@ -15,5 +15,12 @@ CLASS net/minecraft/class_1672 net/minecraft/entity/EyeOfEnderEntity
...
+		COMMENT
+		COMMENT @param pos The block the eye of ender is drawn towards
```
YanisBft:
```suggestion
		COMMENT @param pos the block the eye of ender is drawn towards
```

YanisBft (commented):
I don't get why already mapped members are being mapped here

nskobelevs:
> I don't get why already mapped members are being mapped here

The current mappings are incorrect / misleading. The fields are called `velocityX`, `velocityY` and `velocityZ` while they represent a position, not a velocity.
The `moveTowards` change is because the method itself doesn't actually do any moving of the entity, but only the logic of where it will fly to.

YanisBft:
> The current mappings are incorrect / misleading. The fields are called `velocityX`, `velocityY` and `velocityZ` while they represent a position, not a velocity.
> The `moveTowards` change is because the method itself doesn't actually do any moving of the entity, but only the logic of where it will fly to.

No, I mean the other classes like ProjectileEntity where the fields were already mapped with previous PRs

nskobelevs:
> No, I mean the other classes like ProjectileEntity where the fields were already mapped with previous PRs

Ah, yes that was my bad. I branched from 1.16.1 but the RP is into 20w28a. So that's where the extra commits are from. :/

jaskarth (approved):

YanisBft:
@nskobelevs mind update the branch and apply the suggestion please?

YanisBft (approved):

PR #1583, opened 2020-07-15
Map the angle argument type

haykam821:


mappings/net/minecraft/command/arguments/AngleArgument.mapping
```diff
@@ -0,0 +1,15 @@
+CLASS net/minecraft/class_5473 net/minecraft/command/arguments/AngleArgument
```
YanisBft: Should be `AngleArgumentType` to match the other arguments
Also make me realize `arguments` package should probably be singular like all other package names, but this would need a proper PR

haykam821: Oops, I don't know how I missed that. Fixed in c00551e64aaccbcda912c451a4db0004bff74f61.

YanisBft (approved):

shartte (approved):

PR #1587, opened 2020-07-17
ServerWorldAccess -> StructureWorldAccess

i509VCB:
This describes with more detail what this specific interface is.

### DIFF
--- a/mappings/net/minecraft/world/ServerWorldAccess.mapping
+++ b/mappings/net/minecraft/world/ServerWorldAccess.mapping
@@ -1,2 +1,0 @@
-CLASS net/minecraft/class_5281 net/minecraft/world/ServerWorldAccess
-	METHOD method_8412 getSeed ()J
--- a/mappings/net/minecraft/world/StructureWorldAccess.mapping
+++ b/mappings/net/minecraft/world/StructureWorldAccess.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_5281 net/minecraft/world/WorldAccessWithStructures
+CLASS net/minecraft/class_5281 net/minecraft/world/StructureWorldAccess
 	METHOD method_30275 getStructures (Lnet/minecraft/class_4076;Lnet/minecraft/class_3195;)Ljava/util/stream/Stream;
 		ARG 1 pos
 		ARG 2 feature
--- a/mappings/net/minecraft/world/WorldAccessWithStructures.mapping
+++ b/mappings/net/minecraft/world/WorldAccessWithStructures.mapping
@@ -1,0 +1,5 @@
+CLASS net/minecraft/class_5281 net/minecraft/world/WorldAccessWithStructures
+	METHOD method_30275 getStructures (Lnet/minecraft/class_4076;Lnet/minecraft/class_3195;)Ljava/util/stream/Stream;
+		ARG 1 pos
+		ARG 2 feature
+	METHOD method_8412 getSeed ()J


YanisBft:
What about `PostStructuresWorldAccess` or `StructureGeneratedWorldAccess`?

liach:
I agree with yanis. a naming like `XxxWithStructures` would sounds like `Strutures` while it is a world access.

Player3324:
The withXY name is supposed to declare the additional feature, Post* or StructureGenerated* can be misunderstood and implies some temporal relation that doesn't exist. Maybe something like *CapableWA or *ProvidingWA would work too.

It is also important to note that these are very special purpose interfaces, so more descriptive naming should be chosen over "nicer" names.

liach:
Then `StructurePopulatedWorld`, no need for `Access` part of name at least

Player3324:
That is wrong though, there are plenty of world-like classes that have structures populated without this interface.

liach:
Maybe `StructureAvailableWorld`? mojang is known for creating interfaces to restrict operations, like for text, so they don't need a state machine but just passing restricted interfaces for operational safety?

Player3324:
The structures are "available" (as in present) regardless of the interface - they are part of the world once generated. We need to talk about structure (meta)data I think.

liach:
> The structures are "available" (as in present) regardless of the interface

Well, this is exactly the same case in `Text`. Text implementations are regardless mutable. However, mutation operations are only exposed via the `MutableText` interface. Maybe they pass in this interface than a full `ServerWorld` to indicate that at where this interface is used, the operations this interface support is totally safe while other operations may not be guaranteed.

Player3324:
That's not what I mean, structures can be understood as just the blocks placed into the world. The client-side world has them, but not this interface. Your names imply otherwise with this fairly reasonable interpretation.

liach:
> structures can be understood as just the blocks placed into the world

no? structure has quite a few extra attached information afaik that is used by other world generation components. the thing called structure that is part of world generation is only safely accessible when this interface is provided.

Player3324:
That doesn't matter, some people will think of structures as I said, making the names insufficiently descriptive.

YanisBft (approved):
WorldAccessWithStructures is not that bad afterall (didn't want to start a bikeshedding lol, I'm fine with this name)

Runemoro:
`StructureView`

i509VCB:
🚲🏠

Rune I think it's been discussed before that WorldAccess and above interfaces aren't views in the typical sense. Also this is a very specific interface for a specific purpose, Player has said he wants more descriptive names for these very niche interfaces.

Prospector:
Is it really niche? Don't like loads of the features use this as a parameter? ServerWorldAccess seems good to me still.

Runemoro:
> WorldAccess and above interfaces aren't views in the typical sense

What's the difference between an access and a view? I thought we got rid of `Access` and renamed everything to `View`.

In any case, my problem is not with the word `Access`, but with the word `With`. It could be `StructureAccess` or `StructureWorldAccess` and mean the same thing with one word less.

i509VCB:
WorldAccess was added originally to get rid of the last yarn name that violated the naming standards, that name alone was a fucking nightmare to decide on. I'd be open to discussing the merits of keeping which names. I'd definitely say we are also somewhat restricted in naming because of the actual World classes (and no one wanting WorldWithXyz)

Runemoro:
> no one wanting WorldWithXyz

In every single case, `WorldWithXyz` should be `XyzWorld`. More generally, it's a common Java convention to use `XyzAbc` instead of `AbcWithXyz`.

liach:
> In every single case, `WorldWithXyz` should be `XyzWorld`. More generally, it's a common Java convention to use `XyzAbc` instead of `AbcWithXyz`.

This is literally the whole point of my argument. This thing would look like some type of "structures" on first sight while it is actually a world

jaskarth:
I would also vote for `StructureWorldAccess`, as `StructuresAccess` doesn't immediately convey that it's a world.

YanisBft:
+1 for `StructureWorldAccess`

i509VCB:
Now `StructureWorldAccess`

liach (approved):

YanisBft (approved):

Juuxel (approved):

Runemoro (approved):

Player3324 (approved):
better than before :D

modmuss50 (approved):

PR #1588, opened 2020-07-17
Move all realms screens to `realms.gui.screen`, move tasks to `task`package. And a few class renames.

i509VCB:


YanisBft (approved):

mappings/net/minecraft/client/realms/RealmsProxyConfig.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_4344 net/minecraft/client/realms/RealmsClientConfig
+CLASS net/minecraft/class_4344 net/minecraft/client/realms/RealmsProxyConfig
```
Runemoro: The `RealmsClientConfig` was based on the old name when the Realms jar was unobfuscated, so it seems that Mojang is using this class for storing all realms client config, but it just happens that the only config option right now is the proxy.

Runemoro: Having a separate class specifically for proxy config is weird, while having a config class makes more sense (and it's what it actually is). Matching the actual name will also make sure that it will stay accurate as Mojang possibly adds more things to it.

YanisBft:
Mind resolve the conflict and update branch @i509VCB ?

Runemoro (approved):

modmuss50 (approved):

natanfudge:
Very bad pr, so many conflicts with other prs and so many problems, trying to sneak in bad names but I hereby block your malicious attempt.

PR #1590, opened 2020-07-18
Redocument registry key

liach:
Since prospector said it's unclear, I'd figure to do it


liach:
Closing as these forks are borked. I will open another one from another fork

PR #1594, opened 2020-07-19
Fixes mapping issues in LevelInfo

### LINKED ISSUE #1593: LevelProperties#isHardcore returns whether the world has structures [20w29a]
image

shartte:
This is probably a matching artifact.
The "structures" flag has moved into the generator options.

This fixes #1593

### DIFF
--- a/mappings/net/minecraft/world/level/LevelInfo.mapping
+++ b/mappings/net/minecraft/world/level/LevelInfo.mapping
@@ -2,14 +2,30 @@ CLASS net/minecraft/class_1940 net/minecraft/world/level/LevelInfo
 	FIELD field_24105 name Ljava/lang/String;
 	FIELD field_24106 difficulty Lnet/minecraft/class_1267;
 	FIELD field_24107 gameRules Lnet/minecraft/class_1928;
+	FIELD field_25403 dataPackSettings Lnet/minecraft/class_5359;
 	FIELD field_9257 gameMode Lnet/minecraft/class_1934;
-	FIELD field_9261 hardcore Z
-	FIELD field_9262 structures Z
+	FIELD field_9261 allowCommands Z
+	FIELD field_9262 hardcore Z
 	METHOD <init> (Ljava/lang/String;Lnet/minecraft/class_1934;ZLnet/minecraft/class_1267;ZLnet/minecraft/class_1928;Lnet/minecraft/class_5359;)V
 		ARG 1 name
+		ARG 2 gameMode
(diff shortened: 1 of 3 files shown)


YanisBft (approved):

mappings/net/minecraft/world/level/LevelInfo.mapping
```diff
@@ -2,14 +2,30 @@ CLASS net/minecraft/class_1940 net/minecraft/world/level/LevelInfo
...
+		ARG 7 dataPackSettings
 	METHOD method_27339 getLevelName ()Ljava/lang/String;
 	METHOD method_27340 getDifficulty ()Lnet/minecraft/class_1267;
 	METHOD method_27341 getGameRules ()Lnet/minecraft/class_1928;
+	METHOD method_28381 withDifficulty (Lnet/minecraft/class_1267;)Lnet/minecraft/class_1940;
+		ARG 1 difficulty
+	METHOD method_28382 withGameMode (Lnet/minecraft/class_1934;)Lnet/minecraft/class_1940;
+		ARG 1 mode
+	METHOD method_28383 fromDynamic (Lcom/mojang/serialization/Dynamic;Lnet/minecraft/class_5359;)Lnet/minecraft/class_1940;
+	METHOD method_28385 withNewGameRules ()Lnet/minecraft/class_1940;
```
liach: Is each `LevelInfo` immutable? If yes `withDefaultGameRules` might be better.

shartte: It copies the current game rules attached to it though. I honestly didn't check further, but that only seemed to make sense if the LevelInfo was immutable, but the GameRules weren't. (I'll recheck though)

shartte: Yes after rechecking: GameRules seem to be mutable, while LevelInfo is not, so "withNewGameRules" might be used if one wants to have the same levelinfo, but with different rules

Runemoro (approved):

mappings/net/minecraft/world/level/LevelInfo.mapping
liach: How about `withCopiedGameRules`? this sounds like creating a new level info with vanilla's default game rules

shartte: Done

liach (approved):

modmuss50 (approved):

PR #1596, opened 2020-07-20
(Controversial) Rename client.font.TextureFont -> BitmapFont

Sturmlilie:
Two reasons for this:
1) There are no textures involved in this class,
   only CPU-side `NativeImage`s.
2) This name is consistent with the enum value in `FontType`.

Of course, naming it `ImageFont` or `NativeImageFont` would be the ultimate yarn-move, but in this case, the unobfuscated enum value is making this a clear-cut case in my view.

Edit: It has been pointed out to me that "bitmap font" is in fact an established term for exactly this type of font: https://en.wikipedia.org/wiki/Computer_font#Bitmap_fonts

### DIFF
--- a/mappings/net/minecraft/client/font/BitmapFont.mapping
+++ b/mappings/net/minecraft/client/font/BitmapFont.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_386 net/minecraft/client/font/TextureFont
+CLASS net/minecraft/class_386 net/minecraft/client/font/BitmapFont
 	FIELD field_2284 glyphs Lit/unimi/dsi/fastutil/ints/Int2ObjectMap;
 	FIELD field_2285 image Lnet/minecraft/class_1011;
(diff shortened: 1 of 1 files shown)


shartte:
I find the texture part unconvincing (the texture can be CPU side), but such fonts are indeed usually called bitmap fonts, so 👍

shartte (approved):

Runemoro (approved):
"Texture" sounds weird for something like a font.

liach (approved):

YanisBft (approved):

Sturmlilie:
> "Texture" sounds weird for something like a font.

Not at all; in fact many games upload glyph sheets directly to the GPU (unlike Minecraft which stitches them all into a giant atlas on-demand).

modmuss50 (approved):
Ill leave this another day becuase like you said it might be Controversial, but its not a wide spead change so I see no issue with it.

PR #1606, opened 2020-07-24
Did some mappings in RedstoneWireBlock

LostQuasar:


YanisBft (approved):

LostQuasar:
I feel like a better name could be used here but getReceivedPower makes the most sense, perhaps a change should be made to the method im world to be more clear about its purpose which is to find the highest power on all horizontal directions

mappings/net/minecraft/block/RedstoneWireBlock.mapping
```diff
@@ -28,6 +28,9 @@ CLASS net/minecraft/class_2457 net/minecraft/block/RedstoneWireBlock
 		ARG 1 world
 		ARG 2 state
 		ARG 3 pos
+	METHOD method_27842 geReceivedRedstonePower (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;)I
```
haykam821:
```suggestion
	METHOD method_27842 getReceivedRedstonePower (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;)I
```

haykam821 (changes requested):

liach (approved):

haykam821 (approved):
The rest of the methods can be fully mapped later.

liach:
Hmm, I doubt the rest of these methods may just be lambdas. Anyways look great, ready to go

PR #1607, opened 2020-07-24
(Almost) fully mapped Biome and BiomeEffects

YanisBft:
With this PR, BiomeEffects is fully mapped and Biome only has a couple of members unnamed left

### DIFF
--- a/mappings/net/minecraft/world/biome/Biome.mapping
+++ b/mappings/net/minecraft/world/biome/Biome.mapping
@@ -1,7 +1,11 @@
 CLASS net/minecraft/class_1959 net/minecraft/world/biome/Biome
 	FIELD field_20335 temperatureCache Ljava/lang/ThreadLocal;
 	FIELD field_22039 effects Lnet/minecraft/class_4763;
+	FIELD field_24677 REGISTRY_CODEC Lcom/mojang/serialization/Codec;
+	FIELD field_25819 CODEC Lcom/mojang/serialization/MapCodec;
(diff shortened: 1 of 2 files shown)


jaskarth (approved):

mappings/net/minecraft/world/biome/Biome.mapping
```diff
@@ -1,7 +1,11 @@
 CLASS net/minecraft/class_1959 net/minecraft/world/biome/Biome
 	FIELD field_20335 temperatureCache Ljava/lang/ThreadLocal;
 	FIELD field_22039 effects Lnet/minecraft/class_4763;
+	FIELD field_24677 REGISTRY_CODEC Lcom/mojang/serialization/Codec;
+	FIELD field_25819 CODEC Lcom/mojang/serialization/MapCodec;
```
liach, line 5: hmm, since this is a map codec, is there cases where like there is a json object which you can derive both a biome and some other stuff from that one json object? I assume this exists so the registry codec can either dispatch to an id field lookup or pass the object to this map codec?

liach (approved):

liach:
Removing refactor as you only rename private fields. Should be of minimal impact

mappings/net/minecraft/world/biome/Biome.mapping
YanisBft: I don't understand much about codecs, but the registry codec here is a registry element codec cast to a codec, hence the name

shartte (approved):

PR #1616, opened 2020-07-28
Random client mappings

### LINKED ISSUE #1597: MinecraftClient.openScreen -> setScreen
It first closes the current screen. `openScreen` makes it sound as if the previous screen stays open, and `setScreen(null)` makes more sense than `openScreen(null)`.

### LINKED ISSUE #1611: MinecraftClient.getEntityRenderManager -> getEntityRenderDispatcher
It returns the `entityRenderDispatcher` field.

YanisBft:
Closes #1611

### DIFF
--- a/mappings/net/minecraft/client/MinecraftClient.mapping
+++ b/mappings/net/minecraft/client/MinecraftClient.mapping
@@ -133,7 +133,7 @@ CLASS net/minecraft/class_310 net/minecraft/client/MinecraftClient
 	METHOD method_1506 handleGlErrorByDisableVsync (IJ)V
 		ARG 1 error
 		ARG 2 description
-	METHOD method_1507 setScreen (Lnet/minecraft/class_437;)V
+	METHOD method_1507 openScreen (Lnet/minecraft/class_437;)V
 		ARG 1 screen
 	METHOD method_1508 handleInputEvents ()V
 	METHOD method_1511 doItemPick ()V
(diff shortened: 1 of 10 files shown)


mappings/net/minecraft/client/MinecraftClient.mapping
```diff
@@ -130,7 +133,7 @@ CLASS net/minecraft/class_310 net/minecraft/client/MinecraftClient
 	METHOD method_1506 handleGlErrorByDisableVsync (IJ)V
 		ARG 1 error
 		ARG 2 description
-	METHOD method_1507 openScreen (Lnet/minecraft/class_437;)V
+	METHOD method_1507 setScreen (Lnet/minecraft/class_437;)V
```
shartte: Not sure about this one. It makes it sound as if it was a simple setter. I can understand that openScreen is bad because a null parameter just closes the current one. But given the significance of this one, I'd actually just leave it at openScreen (or maybe showScreen). **edit**: I just saw this is from the linked issue. So okay, approving anyway...

shartte (approved):

jaskarth (approved):

mappings/net/minecraft/client/MinecraftClient.mapping
liach: High impact rename, need to wait

YanisBft: I'm not sure of the change now. Didn't think at all it could be mislead with a simple setter, which is a fair point. Maybe we should keep openScreen actually, with docs explaining it, even if it's not accurate when you pass a null parameter

liach: Yeah maybe keep this rename in a separate pr

YanisBft:
Reverted the openScreen -> setScreen rename

liach (approved):

PR #1618, opened 2020-07-28
ServerWorldAccess is different now

i509VCB:
The interface below `StructureWorldAccess` is actually `ServerWorldAccess` as the interface has a `getWorld -> ServerWorld` method.

### DIFF
--- a/mappings/net/minecraft/world/ServerWorldAccess.mapping
+++ b/mappings/net/minecraft/world/ServerWorldAccess.mapping
@@ -3,4 +3,4 @@ CLASS net/minecraft/class_5425 net/minecraft/world/ServerWorldAccess
 	METHOD method_30771 spawnEntityAndPassengers (Lnet/minecraft/class_1297;)V
 		COMMENT Spawns an entity and all its passengers (recursively) into the world.
 		ARG 1 entity
-	METHOD method_8410 getServerWorld ()Lnet/minecraft/class_3218;
+	METHOD method_8410 toServerWorld ()Lnet/minecraft/class_3218;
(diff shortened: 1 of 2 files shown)


jaskarth (approved):

mappings/net/minecraft/world/ServerWorldAccess.mapping
```diff
@@ -0,0 +1,6 @@
+CLASS net/minecraft/class_5425 net/minecraft/world/ServerWorldAccess
+	COMMENT Represents access to a world on a logical Minecraft server.
+	METHOD method_30771 spawnEntityAndPassengers (Lnet/minecraft/class_1297;)V
+		COMMENT Spawns an entity and all it's passengers into the world.
```
liach:
```suggestion
		COMMENT Spawns an entity and all its passengers (recursively) into the world.
```

mappings/net/minecraft/world/ServerWorldAccess.mapping
```diff
@@ -0,0 +1,6 @@
...
+		ARG 1 entity
+	METHOD method_8410 getServerWorld ()Lnet/minecraft/class_3218;
```
liach:
```suggestion
	METHOD method_8410 toServerWorld ()Lnet/minecraft/class_3218;
```
Like how java's `CompletionStage` has `toCompletableFuture`

liach (approved):

mappings/net/minecraft/world/ServerWorldAccess.mapping
shartte: I can't check right now, but isn't this inconsistent with the other world access interfaces having "getWorld" instead?

i509VCB: There is no other world access interface with this method. This is the sole one

YanisBft (approved):

PR #1628, opened 2020-07-29
ChunkGeneratorType -> ChunkGeneratorSettings

YanisBft:
🚲 🏠 
~~This is the name of the datapack folder (`worldgen/noise_settings`), and it actually makes more sense than the old ChunkGeneratorType.~~
~~What a chance, there is not a single reference of the class in the code, so no param to rename :)~~

### DIFF
--- a/mappings/net/minecraft/util/registry/BuiltinRegistries.mapping
+++ b/mappings/net/minecraft/util/registry/BuiltinRegistries.mapping
@@ -12,7 +12,7 @@ CLASS net/minecraft/class_5458 net/minecraft/util/registry/BuiltinRegistries
 	FIELD field_25933 BIOME Lnet/minecraft/class_2378;
 	FIELD field_25934 DEFAULT_VALUE_SUPPLIERS Ljava/util/Map;
 	FIELD field_25935 ROOT Lnet/minecraft/class_2385;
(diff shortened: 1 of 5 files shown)


Player3324:
From discord:
- it does a lot more than noise and there is actually a field NoiseConfig
- NoiseConfig is also kind of wrong - height doesn't fit
- NoiseConfig defines the basic shape, ChunkGeneratorType the basic world (shape+how to translate the shape into blocks+some other fundamental settings)

jaskarth (approved):

shartte:
Also not a fan of `NoiseSettings`, they seem to be part of this class, but is it really 1:1?
ChunkGenerationSettings could be more suitable.

liach:
Yeah, makes no sense to have a mob spawning control boolean or bedrock y in a "noise settings"

Player3324 (approved):

liach (approved):

liach:
@haykam821 renamed, please take a look again

jaskarth:
Not a big fan of `GenerationShapeConfig` here- It seems a bit bulky to me. I would prefer `NoiseGenerationConfig`.

shartte (approved):

Runemoro (approved):

haykam821 (approved):

PR #1636, opened 2020-07-30
TransformingVertexConsumer -> OverlayVertexConsumer

JamiesWhiteShirt:
What is currently named TransformingVertexConsumer transforms input normals and ignores input texture coordinates. Instead, it computes texture coordinates based on the input vertex position and normal.

While it technically _is_ a transforming vertex consumer, it is not as general purpose as the name might imply. Minecraft uses its texture coordinate transformation for overlays on models such as enchantment glints and the cracks showing block breaking progress. It is not useful for much else.


Sturmlilie:
There are some uses of the term `transforming` left, eg. in `ItemRenderer.getTransforming(Direct)GlintVertexConsumer`, but other than that lgtm.

JamiesWhiteShirt:
I now removed the "Transforming" from the ItemRenderer method names as well.

liach (approved):

jaskarth (approved):

YanisBft (approved):

PR #1635, opened 2020-07-30
Text-related refactors & new text things

shedaniel:
may be very controversial

- Change StringRenderable to StringVisitable (it is no longer used for rendering, currently an intermediary until being turned to "OrderedText")
- OrderedText (I don't know about this but I can't think of a better one tbh, I'll throw in some more: CharacterVisitable, OrderedString, OrderedTextProvider)
- ReorderingUtils
- TextReorderingProcessor

Signed-off-by: shedaniel <daniel@shedaniel.me>

PR #1637, opened 2020-07-30
Miscellaneous Text and TranslatableText mappings

Sturmlilie:
* `Text.fromString(String)` returns a `LiteralText` instance constructed from the passed string (or the `EMPTY` constant in case of null). Alternatives: `asText`, `asLiteralText`. No known overrides.
* `TranslatableText.getArg(int)` returns an integer-indexed element of the `args` field (see: `#getArgs`), throwing an exception on out-of-bounds access with following message: `Invalid index %d requested for %s`.

This PR conflicts with #1635 , though we're mapping two different methods.

### DIFF
--- a/mappings/net/minecraft/text/Text.mapping
+++ b/mappings/net/minecraft/text/Text.mapping
@@ -42,6 +42,10 @@ CLASS net/minecraft/class_2561 net/minecraft/text/Text
 		COMMENT <p>A shallow copy is made for the siblings.</p>
 	METHOD method_27662 copy ()Lnet/minecraft/class_5250;
 		COMMENT Copies the text itself, excluding the styles or siblings.
+	METHOD method_30163 fromString (Ljava/lang/String;)Lnet/minecraft/class_2561;
+		COMMENT Creates a {@link Text} instance from a nullable string.
+		COMMENT @return the {@link Text} instance created from {@code string}
+		COMMENT @param string a nullable string.
 	CLASS class_2562 Serializer
 		COMMENT A JSON serializer for {@link Text}.
 		FIELD field_11752 JSON_READER_LINE_START Ljava/lang/reflect/Field;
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/text/Text.mapping
```diff
@@ -42,6 +42,10 @@ CLASS net/minecraft/class_2561 net/minecraft/text/Text
 		COMMENT <p>A shallow copy is made for the siblings.</p>
 	METHOD method_27662 copy ()Lnet/minecraft/class_5250;
 		COMMENT Copies the text itself, excluding the styles or siblings.
+	METHOD method_30163 fromString (Ljava/lang/String;)Lnet/minecraft/class_2561;
```
liach:
```suggestion
	METHOD method_30163 of (Ljava/lang/String;)Lnet/minecraft/class_2561;
```
or
```suggestion
	METHOD method_30163 literal (Ljava/lang/String;)Lnet/minecraft/class_2561;
```

mappings/net/minecraft/text/Text.mapping
```diff
@@ -42,6 +42,10 @@ CLASS net/minecraft/class_2561 net/minecraft/text/Text
...
+		COMMENT Creates a {@link Text} instance from a nullable string.
+		COMMENT @return the {@link Text} instance created from {@code string}
+		COMMENT @param string a nullable string.
```
liach:
```suggestion
		COMMENT Creates a literal text with the given string as content.
```

liach (changes requested):
If you document the param, please name the param "string" explicitly and document the param than adding param tag in the method body directly.

mappings/net/minecraft/text/Text.mapping
Sturmlilie: Do you describe interface methods based on their default behavior? If a subclass overrides this, the comment will be inaccurate.

mappings/net/minecraft/text/Text.mapping
Sturmlilie: I went with `of`, since `literal` would imply to me that null's aren't handled.

YanisBft (approved):

Juuxel (approved):

mappings/net/minecraft/text/Text.mapping
liach: Fyi this method is static. Procyon is known for displaying interface static as defaults.

Also side note (unrelated): this is client only

liach (approved):

Issue #1614, opened 2020-07-26
LivingEntity.canPickUp -> ?


Juuxel:
It seems to be more of a "can equip" check, where the implementations usually check whether the preferred slot for the item stack is empty.

(If this gets a new name, `EntityPredicates.CanPickup` should also be renamed to match.)

liach:
well, don't mobs with custom goals usually just pick up stuff they can make use of? like pick up helmets, armor, etc.

You should check their usage as well.

Juuxel:
This is used in places like the dispenser behavior that equips armor on entities. (edit: and possibly that entity pick up equipping, which is still equipping)

sylv256:
He's right. The entities equip the items. It would be a little easier to understand what "canEquip" does than "canPickUp" since the entity will equip the items it picks up. As far as I know, entities cannot "pick up" items without equipping them, so it only makes sense to rename it to "canEquip."

sylv256:
I don't think `canEquip` fits the mapping name since modded entities and piglins can pick up items (such as gold nuggets) without equipping them. `canTake` fits the name more because it's easier to type and makes more sense than `canEquip`.

sylv256:
Nevermind. Apparently piglins and other entities don't even use this method. I'll make sure to check how something is used before changing it. I'm re-opening the PR right now.

gbl:
It seems like #1640 breaks the build as there are a few more classes that use canEquip. See my comment in that PR.

liach:
#1643 should resolve that.

YanisBft:
Fixed by #1640

PR #1640, opened 2020-08-01
CanPickup/canPickUp -> Equipable/canEquip

sylv256:
# Changes
- LivingEntity.canPickUp -> canEquip
- EntityPredicates.CanPickup -> Equipable

#1614

### DIFF
--- a/mappings/net/minecraft/predicate/entity/EntityPredicates.mapping
+++ b/mappings/net/minecraft/predicate/entity/EntityPredicates.mapping
@@ -41,7 +41,7 @@ CLASS net/minecraft/class_1301 net/minecraft/predicate/entity/EntityPredicates
 		ARG 0 entity
 	METHOD method_5913 rides (Lnet/minecraft/class_1297;)Ljava/util/function/Predicate;
 		ARG 0 entity
-	CLASS class_1302 CanEquip
+	CLASS class_1302 Equipable
 		FIELD field_6158 stack Lnet/minecraft/class_1799;
 		METHOD <init> (Lnet/minecraft/class_1799;)V
 			ARG 1 stack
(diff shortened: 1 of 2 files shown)


Juuxel (approved):

sylv256:
According to people on the Fabricord server, modded entities and piglins can pick up items (like gold nuggets) without equipping them. I'm going to leave the PR open just in case I'm not entirely correct. However, I don't think canEquip is the right name.

sylv256:
I think `canTake` would be better since it's easier to type than canPickUp and makes more sense than canEquip. I'll close this PR and make a new one.

Juuxel:
`canPickUp` is only used for equipping checks. The method used for regular pick up checks is `canGather`, which is the one piglins, piglin brutes and villagers override to make gathering gold and food possible.

sylv256:
Piglins and other entities don't actually use this method for picking up items. It's only for equipping items on entities such as armor stands and also equipping items on entities via a dispenser.

Virtuoel (approved):

mappings/net/minecraft/predicate/entity/EntityPredicates.mapping
```diff
@@ -41,7 +41,7 @@ CLASS net/minecraft/class_1301 net/minecraft/predicate/entity/EntityPredicates
 		ARG 0 entity
 	METHOD method_5913 rides (Lnet/minecraft/class_1297;)Ljava/util/function/Predicate;
 		ARG 0 entity
-	CLASS class_1302 CanPickup
+	CLASS class_1302 CanEquip
```
liach: I'd suggest `Equipable` name instead

YanisBft (approved):

mappings/net/minecraft/predicate/entity/EntityPredicates.mapping
sylv256: Alright, I'll change the mappings real quick.

Mysterious-Dev (approved):

liach (approved):

PR #1652, opened 2020-08-03
Add Javadoc for WorldRenderer#processWorldEvent eventId

OroArmor:


liach (approved):
Hmm, I think these are fired in regular world class as well (sent from server)? Documenting there may be more helpful

OroArmor:
In `WorldAccess`?

liach (approved):
Yeah think this is it

PR #1657, opened 2020-08-04
Remove mapYarnJar task

Runemoro:
It does the same thing as `mapNamedJar` except without parameter names.


liach:
Hmm, when I added that task, I recall that task does not map stitch-inferred names.

Runemoro:
@liach Which task are you talking about?

I can confirm that `mapNamedJar` does map enum names. If you're talking about `mapYarnJar`, I don't see why we need a jar that's missing stitch names?

liach:
Yeah before I think I renamed old map named jar that does not include inferred names as this yarn jar for those that need such a jar. guess we can remove it

liach (approved):

modmuss50 (approved):
I see no use in it, will leave it a bit of time to make sure no one uses it for some reason.

YanisBft (approved):

modmuss50:
No one has said anything yet, so it can go.

PR #1668, opened 2020-08-08
Some item javadoc

i509VCB:


mappings/net/minecraft/item/Item.mapping
```diff
@@ -15,17 +15,34 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
...
+		COMMENT Checks if this item is food and therefore is edible.
 	METHOD method_19264 getFoodComponent ()Lnet/minecraft/class_4174;
 	METHOD method_21830 getEatSound ()Lnet/minecraft/class_3414;
 	METHOD method_21831 getDrinkSound ()Lnet/minecraft/class_3414;
 	METHOD method_24357 damage (Lnet/minecraft/class_1282;)Z
 		ARG 1 source
 	METHOD method_24358 isFireproof ()Z
 	METHOD method_7836 use (Lnet/minecraft/class_1937;Lnet/minecraft/class_1657;Lnet/minecraft/class_1268;)Lnet/minecraft/class_1271;
+		COMMENT Called when an item is used by a player.
+		COMMENT This typically is bound to using the right mouse button.
```
liach:
```suggestion
		COMMENT The use action, by default, is bound to the right mouse button.
```

mappings/net/minecraft/item/Item.mapping
```diff
@@ -15,17 +15,34 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
...
+		COMMENT
+		COMMENT <p>This method is called on both the logical client and logical server, so take caution when using this method.
```
liach:
```suggestion
		COMMENT <p>This method is called on both the logical client and logical server, so take caution when overriding this method.
```

mappings/net/minecraft/item/Item.mapping
```diff
@@ -15,17 +15,34 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
...
+		COMMENT The logical side can be checked using {@link net.minecraft.world.World#isClient() world.isClient()}.
+		COMMENT
+		COMMENT @return a typed action result that specifies if using the item was successful.
+		COMMENT The action result contains the new item stack that the player's hand will be set to.
```
liach: Make note that the use call goes from main hand to off hand and may perform hand animation based on the result

mappings/net/minecraft/item/Item.mapping
```diff
@@ -15,17 +15,34 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
...
 		ARG 1 world
+			COMMENT the world the item was used in
 		ARG 2 user
+			COMMENT the player which used the item
```
liach:
```suggestion
			COMMENT the player who used the item
```

mappings/net/minecraft/item/Item.mapping
```diff
@@ -101,7 +121,11 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
 	METHOD method_7875 byRawId (I)Lnet/minecraft/class_1792;
 		ARG 0 id
 	METHOD method_7876 getTranslationKey ()Ljava/lang/String;
+		COMMENT Gets the translation key of this item.
 	METHOD method_7877 isIn (Lnet/minecraft/class_1761;)Z
+		COMMENT Checks if this item is in a specified item group.
```
liach:
```suggestion
		COMMENT Checks if this item should appear in a specified item group.
```

mappings/net/minecraft/item/Item.mapping
```diff
@@ -118,7 +142,14 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
 		ARG 1 stack
 	METHOD method_7882 getMaxCount ()I
 	METHOD method_7884 useOnBlock (Lnet/minecraft/class_1838;)Lnet/minecraft/class_1269;
+		COMMENT Called when an item is used on a block.
+		COMMENT
+		COMMENT <p>This method is called on both the logical client and logical server, so take caution when using this method.
+		COMMENT The logical side can be checked using {@link net.minecraft.world.World#isClient() context.getWorld().isClient()}.
+		COMMENT
+		COMMENT @return an action result that specifies if using the item on a block was successful.
```
liach, line 150: Maybe make note about how the whole interaction process happens.

**This may be wrong!!!!** I recall that on client use on block is first checked and sent to server if possible. then check using item alone and send to server. Needa check and write in detail

mappings/net/minecraft/item/Item.mapping
```diff
@@ -127,6 +158,9 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
 	METHOD method_7886 hasGlint (Lnet/minecraft/class_1799;)Z
 		ARG 1 stack
 	METHOD method_7887 shouldSyncTagToClient ()Z
+		COMMENT Checks if an item should have additional data synced to the client.
+		COMMENT
+		COMMENT <p>If an item is damageable, this method is ignored.
```
liach:
```suggestion
		COMMENT <p>If an item is damageable, this method is ignored and data is always synced to client.
```
Correct me if i am wrong

mappings/net/minecraft/item/ItemConvertible.mapping
```diff
@@ -1,2 +1,4 @@
 CLASS net/minecraft/class_1935 net/minecraft/item/ItemConvertible
+	COMMENT Represents an object that has an item form.
 	METHOD method_8389 asItem ()Lnet/minecraft/class_1792;
+		COMMENT Gets this object in it's item form.
```
liach:
```suggestion
		COMMENT Gets this object in its item form.
```

mappings/net/minecraft/item/Item.mapping
i509VCB: Maybe add something like below to clarify part of the interaction process.
```
<p>If the action result is successful on a logical client, then the action will sent to the logical server for processing.
```

liach: Hmm, does the client only send the result when it succeed? needa check that minecraft client code

i509VCB: ~~yes~~ actually no, the packet is sent before `onUse` on the logical client

liach (approved):

mappings/net/minecraft/item/Item.mapping
```diff
@@ -15,17 +15,34 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
...
+		COMMENT The action result contains the new item stack that the player's hand will be set to.
 		ARG 1 world
+			COMMENT the world the item was used in
 		ARG 2 user
+			COMMENT the player who used the item
 		ARG 3 hand
+			COMMENT the hand used
 	METHOD method_7837 getEnchantability ()I
+		COMMENT Gets the enchantability of an item.
+		COMMENT This specifies the ability of an item to recieve enchantments when enchanted using an enchanting table.
```
YanisBft:
```suggestion
		COMMENT This specifies the ability of an item to receive enchantments when enchanted using an enchanting table.
```

mappings/net/minecraft/item/Item.mapping
```diff
@@ -15,17 +15,34 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
...
+		COMMENT As the value increases, the amount and level of enchantments applied increases.
```
YanisBft:
```suggestion
		COMMENT As the value increases, the amount and level of enchantments applied increase.
```

mappings/net/minecraft/item/Item.mapping
```diff
@@ -15,17 +15,34 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
...
+		COMMENT
+		COMMENT <p>If the value of this method is 0, the item cannot be enchanted using an enchantment table.
```
YanisBft:
```suggestion
		COMMENT <p>If the value of this method is 0, the item cannot be enchanted using an enchanting table.
```

mappings/net/minecraft/item/Item.mapping
```diff
@@ -68,7 +85,9 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
 	METHOD method_7856 isEffectiveOn (Lnet/minecraft/class_2680;)Z
 		ARG 1 state
 	METHOD method_7857 hasRecipeRemainder ()Z
+		COMMENT Checks if this item has a remainder item that should be left behind when crafting.
 	METHOD method_7858 getRecipeRemainder ()Lnet/minecraft/class_1792;
+		COMMENT Gets the remainder item that should be left behind when this item when crafting.
```
YanisBft:
```suggestion
		COMMENT Gets the remainder item that should be left behind when this item is crafted.
```

mappings/net/minecraft/item/Item.mapping
```diff
@@ -68,7 +85,9 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
 	METHOD method_7856 isEffectiveOn (Lnet/minecraft/class_2680;)Z
 		ARG 1 state
 	METHOD method_7857 hasRecipeRemainder ()Z
+		COMMENT Checks if this item has a remainder item that should be left behind when crafting.
```
YanisBft:
```suggestion
		COMMENT Checks if this item has a remainder item that should be left behind when crafted.
```

YanisBft (approved):

mappings/net/minecraft/item/Item.mapping
```diff
@@ -68,7 +85,9 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
 	METHOD method_7856 isEffectiveOn (Lnet/minecraft/class_2680;)Z
 		ARG 1 state
 	METHOD method_7857 hasRecipeRemainder ()Z
+		COMMENT Checks if this item has a remainder item that should be left behind when crafted.
```
liach:
```suggestion
		COMMENT Checks if this item has a remainder item that is left behind when used as a crafting ingredient.
```
Correct me if this is used for non-crafting like smelting as well.

mappings/net/minecraft/item/Item.mapping
```diff
@@ -68,7 +85,9 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
...
 	METHOD method_7858 getRecipeRemainder ()Lnet/minecraft/class_1792;
+		COMMENT Gets the remainder item that should be left behind when this item is crafted.
```
liach:
```suggestion
		COMMENT Gets the remainder item that should be left behind when this item is used as a crafting ingredient.
```

liach (changes requested):
this isn't when the item is crafted (i.e. got as output) but is when the item is consumed as input

mappings/net/minecraft/item/Item.mapping
i509VCB: This actually is called in the brewing stand weirdly enough. If there is a remainder in the brewing stand, then the item is just scattered into the world,

liach (approved):

mappings/net/minecraft/item/Item.mapping
```diff
@@ -127,6 +158,9 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
 	METHOD method_7886 hasGlint (Lnet/minecraft/class_1799;)Z
 		ARG 1 stack
 	METHOD method_7887 shouldSyncTagToClient ()Z
+		COMMENT Checks if an item should have additional data synced to the client.
```
shartte:
```suggestion
		COMMENT Checks if an item should have its NBT data stored in {@link #tag} sent to the client.
```

mappings/net/minecraft/item/Item.mapping
```diff
@@ -15,17 +15,34 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
...
 		ARG 1 source
 	METHOD method_24358 isFireproof ()Z
 	METHOD method_7836 use (Lnet/minecraft/class_1937;Lnet/minecraft/class_1657;Lnet/minecraft/class_1268;)Lnet/minecraft/class_1271;
+		COMMENT Called when an item is used by a player.
+		COMMENT The use action, by default, is bound to the right mouse button.
+		COMMENT
+		COMMENT <p>This method is called on both the logical client and logical server, so take caution when overriding this method.
+		COMMENT The logical side can be checked using {@link net.minecraft.world.World#isClient() world.isClient()}.
+		COMMENT
+		COMMENT @return a typed action result that specifies if using the item was successful.
```
haykam821:
```suggestion
		COMMENT @return a typed action result that specifies whether using the item was successful.
```

mappings/net/minecraft/item/Item.mapping
```diff
@@ -101,7 +121,11 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
 	METHOD method_7875 byRawId (I)Lnet/minecraft/class_1792;
 		ARG 0 id
 	METHOD method_7876 getTranslationKey ()Ljava/lang/String;
+		COMMENT Gets the translation key of this item.
 	METHOD method_7877 isIn (Lnet/minecraft/class_1761;)Z
+		COMMENT Checks if this item should appear in a specified item group.
```
haykam821:
```suggestion
		COMMENT Checks whether this item should appear in a specified item group.
```

mappings/net/minecraft/item/Item.mapping
```diff
@@ -101,7 +121,11 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
...
+		COMMENT
+		COMMENT @return true if the item is in the specified item group or the item group is {@link net.minecraft.item.ItemGroup#SEARCH}.
```
haykam821, line 128: Is there a special case for the `SEARCH` item group? I think this method just checks whether the item appears at all within the item group.

haykam821 (changes requested):

mappings/net/minecraft/item/Item.mapping
i509VCB: Yes, the search item group is a special case

liach: This is really `shouldBeAddedTo` or so as this is called when items are checked to be whether to added to groups, not when groups are ready and to query if an item is in a group

modmuss50 (approved):

PR #1671, opened 2020-08-08
Assorted mappings, mainly enchanting, some rendering params and generator type.

i509VCB:


mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
```diff
@@ -53,14 +58,25 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
...
 		ARG 0 stack
-	METHOD method_8209 forEachEnchantment (Lnet/minecraft/class_1890$class_1891;Ljava/lang/Iterable;)V
-		ARG 0 action
+	METHOD method_8207 (Lnet/minecraft/class_1309;Lnet/minecraft/class_1297;Lnet/minecraft/class_1887;I)V
+		ARG 2 enchantment
+		ARG 3 level
+	METHOD method_8208 (Lorg/apache/commons/lang3/mutable/MutableFloat;Lnet/minecraft/class_1310;Lnet/minecraft/class_1887;I)V
+		ARG 2 enchantment
+		ARG 3 level
+	METHOD method_8209 acceptEnchantments (Lnet/minecraft/class_1890$class_1891;Ljava/lang/Iterable;)V
```
liach: Hmm, debatable name

mappings/net/minecraft/enchantment/EnchantmentHelper.mapping
```diff
@@ -156,7 +175,10 @@ CLASS net/minecraft/class_1890 net/minecraft/enchantment/EnchantmentHelper
 			COMMENT whether treasure enchantments may appear
 	METHOD method_8234 getEfficiency (Lnet/minecraft/class_1309;)I
 		ARG 0 entity
-	CLASS class_1891 Consumer
+	CLASS class_1891 Visitor
```
liach: :-1: This is just a consumer.

mappings/net/minecraft/util/registry/Registry.mapping
```diff
@@ -148,3 +151,12 @@ CLASS net/minecraft/class_2378 net/minecraft/util/registry/Registry
 		ARG 1 value
 	METHOD method_29722 getEntries ()Ljava/util/Set;
 	METHOD method_30517 getKey ()Lnet/minecraft/class_5321;
+	METHOD method_31138 getLifecycle ()Lcom/mojang/serialization/Lifecycle;
+	METHOD method_31139 getEntryLifecycle (Ljava/lang/Object;)Lcom/mojang/serialization/Lifecycle;
+	METHOD method_31140 require (Lnet/minecraft/class_5321;)Ljava/lang/Object;
```
liach: Really need a better name for this.

liach (approved):

mappings/net/minecraft/util/registry/Registry.mapping
shartte: getRequired?

i509VCB: That makes me think of a method which gets a registry entry which is marked as required.

This is basically `get(RegistryKey<T>)` but non-null and throws if the value is not present

liach: Apparently cannot think of a better name than `require` unfortunately. I honestly would rename the optional one to `find` but that's out of scope. This is good.

YanisBft:
```suggestion
	METHOD method_31140 getRequired (Lnet/minecraft/class_5321;)Ljava/lang/Object;
```
I think `getRequired` is still better than `require`

Gegy: `getOrExcept`/`getOrThrow`?

i509VCB: getRequired makes me think it has to be a builtin biome to use the method. getOrThrow could work but is not as succinct as `require`

liach: `require` is better than `getRequired`

liach (approved):

YanisBft (approved):

Gegy (approved):

modmuss50 (approved):

PR #1677, opened 2020-08-11
Update ServerWorld mapping

MCMocoder:
Name method_27910.The new name of method_27910 is setWeather ,like method resetWeather.

### DIFF
--- a/mappings/net/minecraft/server/world/ServerWorld.mapping
+++ b/mappings/net/minecraft/server/world/ServerWorld.mapping
@@ -207,3 +207,9 @@ CLASS net/minecraft/class_3218 net/minecraft/server/world/ServerWorld
 	METHOD method_8554 setSpawnPos (Lnet/minecraft/class_2338;F)V
 		ARG 1 pos
 		ARG 2 angle
+        METHOD method_27910 setWeather (IIZZ)V
+	        ARG 1 clearWeatherTime
+		ARG 2 rainTime
+		ARG 3 thunderTime
+		ARG 4 isRaining
+		ARG 5 isThundering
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/server/world/ServerWorld.mapping
```diff
@@ -207,3 +207,9 @@ CLASS net/minecraft/class_3218 net/minecraft/server/world/ServerWorld
 	METHOD method_8554 setSpawnPos (Lnet/minecraft/class_2338;F)V
 		ARG 1 pos
 		ARG 2 angle
+        METHOD method_27910 setWeather (IIZZ)V
```
Juuxel: Enigma files should use tabs for indentation.

mappings/net/minecraft/server/world/ServerWorld.mapping
```diff
@@ -207,3 +207,9 @@ CLASS net/minecraft/class_3218 net/minecraft/server/world/ServerWorld
...
+	        ARG 1 clearWeatherTime
+		ARG 2 rainTime
+		ARG 3 thunderTime
+		ARG 4 isRaining
+		ARG 5 isThundering
```
Juuxel: There are only four parameters in this method.

Juuxel (changes requested):

mappings/net/minecraft/server/world/ServerWorld.mapping
MCMocoder: Sorry,I just made such a foolish mistake :(

Juuxel: You can edit the files with Enigma itself by running `./gradlew yarn` to prevent these types of problems (and it also makes mapping a lot easier).

MCMocoder: I fixed it.Thank you.

YanisBft:
This exact method was already mapped :) Please use Enigma next time to avoid this

PR #1679, opened 2020-08-12
1 outer class, 3 inner classes & some moves

### LINKED ISSUE #1666: Move AreaHelper and CharacterVisitor out of root package

YanisBft:
Closes #1666


mappings/net/minecraft/util/math/AreaHelper.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_2424 net/minecraft/AreaHelper
+CLASS net/minecraft/class_2424 net/minecraft/util/math/AreaHelper
```
liach: Not limited to math since it interacts with the world. It was originally an inner of like the portal block, so I'd expect this to be in like dimension package or similar.

liach (changes requested):

liach (approved):

Juuxel (approved):

PR #1688, opened 2020-08-13
Rename interface relating to registry at world class

ye-yu:
I use the name `RegistryProvider` for one of its methods to provide `DynamicRegistryManager` instance. However, most of the usage is to access `Registry.BIOME_KEY` and biome feature-related operation. Considering `DynamicRegistryManager` can be used for accessing other registry keys, I stick with `RegistryProvider` rather than `BiomeRegistryProvider` or so. 

~However, there are three redundant overrides to the extended interface method that has no relationship with registry. I have no idea why are they there and speculate that these methods may be properly overriden in the future.~ (It's hard to explain this part, pardon.) The overrides are to default them to the defaulted method of the right parent. e.g. Interface A and Interface B are exclusive but have the same method signature, but Interface A has it as a defaulted method, so an Interface C that extends A and B would like to use the defaulted method of Interface A. 

To add context, `class_5423` extends world-related interfaces: `EntityView`, `WorldView`, and `ModifiableTestableWorld` interfaces. All in all, I think this makes `class_5423` a world-specific interface.

The other change is made for the method_31081. It is pretty clear that the method is used to get the biome registry key from the given block position.

Also, I am new to contributing. I have read through contribution.md and pretty sure these mappings arent MCP/Mojang mapping. Do let me know what other stuffs that I am missing. Thanks.

### DIFF
--- a/mappings/net/minecraft/RegistryProvider.mapping
+++ b/mappings/net/minecraft/RegistryProvider.mapping
@@ -1,0 +1,4 @@
+CLASS net/minecraft/class_5423 net/minecraft/RegistryProvider
+	METHOD method_30349 getRegistryManager ()Lnet/minecraft/class_5455;
+	METHOD method_31081 getBiomeRegistryKey (Lnet/minecraft/class_2338;)Ljava/util/Optional;
+		ARG 1 atPosition
--- a/mappings/net/minecraft/world/RegistryProvider.mapping
+++ b/mappings/net/minecraft/world/RegistryProvider.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_5423 net/minecraft/RegistryProvider
+CLASS net/minecraft/class_5423 net/minecraft/world/RegistryProvider
 	METHOD method_30349 getRegistryManager ()Lnet/minecraft/class_5455;
 	METHOD method_31081 getBiomeRegistryKey (Lnet/minecraft/class_2338;)Ljava/util/Optional;
 		ARG 1 atPosition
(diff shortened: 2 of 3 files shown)


mappings/net/minecraft/RegistryProvider.mapping
```diff
@@ -0,0 +1,4 @@
+CLASS net/minecraft/class_5423 net/minecraft/RegistryProvider
```
liach: Package is a thing

liach (changes requested):

mappings/net/minecraft/RegistryProvider.mapping
ye-yu: sorry i dont seem to understand, the class is located at net/minecraft

williewillus: liach means that the class name needs to be assigned to a subpackage, right now you just have it in the root minecraft package `net.minecraft.RegistryProvider`

ye-yu: ah i think i get it, i thought the subpackages are provided by Mojang and the classes are already in their correct packages

mappings/net/minecraft/world/RegistryProvider.mapping
```diff
@@ -0,0 +1,4 @@
+CLASS net/minecraft/class_5423 net/minecraft/world/RegistryProvider
```
liach, line 1: Hmm, what's basis for putting this in world? should it be in `net/minecraft/util/registry`?

ye-yu: I put it in the world package solely because it extends interfaces that are related to world interfaces. Maybe the naming is not correct? The implementing classes are all world-related classes. Considering that and the overriden methods, what do you think of `RegistryfulWorld` instead? This is to show that the world has registry manager.

P/S: I have updated my 'analysis' in my first comment above.

YanisBft:
Superseded by #1894

PR #1697, opened 2020-08-15
Map tracked position stuff in Entity

Leo40Git:
Changes:
- `private Vec3d field_25750` -> `private Vec3d trackedPosition`
- `public void method_30228(Vec3d arg)` -> `public void updateTrackedPosition(Vec3d pos)` (based on its overload `public void updateTrackedPosition(double, double, double)`)
- `public Vec3d method_30227()` -> `public Vec3d getTrackedPositon()`

### DIFF
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -8,6 +8,7 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	FIELD field_22468 blockPos Lnet/minecraft/class_2338;
 	FIELD field_23807 inanimate Z
 	FIELD field_25154 chunkPosUpdateRequested Z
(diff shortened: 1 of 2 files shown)


liach (approved):

liach:
Just curious, what is this tracked position used for and where is it updated?

Leo40Git:
Possibly something to do with syncing entities from server to client, considering the getter is client-only.

liach:
interesting. I wonder what packet syncs it

YanisBft (approved):

Leo40Git:
Methods of interest in `EntityS2CPacket`:
- `public Vec3d method_30302(Vec3d arg)` - takes a `Vec3d`, returns a `Vec3d` offset by the `delta*` fields, possible name `calculateDeltaPosition`?
- `public double method_30301(long arg)` - seems to be the reverse of `encodePacketCoordinate`, possible name `decodePacketCoordinate`?

PR #1699, opened 2020-08-16
Various entity mappings

### LINKED ISSUE #804: MobEntity#isInDaylight is misleading
While it does check that the mob is exposed to the sun, it has a (high) random chance of returning false. Currently, the method is specifically used by undead mobs to start burning / damage their helmet, which should probably be emphasized.

### LINKED ISSUE #1662: PiglinBrain#onGuardedBlockBroken is wrongly named
This method is called when a guarded block is broken and is in the `guarded by piglins` tag *and* when barrels, chests, chest minecarts, enderchests or shulker boxes are opened.

YanisBft:
- `isInDaylight` -> `isAffectedByDaylight` as random is implied. Closes #804 
- `onGuardedBlockBroken` -> `onGuardedBlockInteraction` as it checks for open blocks too. Closes #1662 
- Some javadocs where the getters are not obvious
- `ANGER_TIME_RANGE` fields
- Random params

### DIFF
--- a/mappings/net/minecraft/entity/mob/EndermiteEntity.mapping
+++ b/mappings/net/minecraft/entity/mob/EndermiteEntity.mapping
@@ -11,3 +11,4 @@ CLASS net/minecraft/class_1559 net/minecraft/entity/mob/EndermiteEntity
 	METHOD method_7022 setPlayerSpawned (Z)V
 		ARG 1 playerSpawned
 	METHOD method_7023 isPlayerSpawned ()Z
+		COMMENT Returns whether this endermite was spawned from an enderpearl thrown by a player.
--- a/mappings/net/minecraft/entity/mob/EndermiteEntity.mapping
+++ b/mappings/net/minecraft/entity/mob/EndermiteEntity.mapping
@@ -11,4 +11,4 @@ CLASS net/minecraft/class_1559 net/minecraft/entity/mob/EndermiteEntity
 	METHOD method_7022 setPlayerSpawned (Z)V
 		ARG 1 playerSpawned
 	METHOD method_7023 isPlayerSpawned ()Z
-		COMMENT Returns whether this endermite was spawned from an enderpearl thrown by a player.
+		COMMENT Returns whether this endermite was spawned from an ender pearl thrown by a player.
--- a/mappings/net/minecraft/entity/mob/PiglinBrain.mapping
+++ b/mappings/net/minecraft/entity/mob/PiglinBrain.mapping
@@ -41,8 +41,9 @@ CLASS net/minecraft/class_4838 net/minecraft/entity/mob/PiglinBrain
 	METHOD method_24732 create (Lnet/minecraft/class_4836;Lnet/minecraft/class_4095;)Lnet/minecraft/class_4095;
 		ARG 0 piglin
 		ARG 1 brain
-	METHOD method_24733 onGuardedBlockBroken (Lnet/minecraft/class_1657;Z)V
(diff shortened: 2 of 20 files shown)


mappings/net/minecraft/entity/mob/EndermiteEntity.mapping
```diff
@@ -11,3 +11,4 @@ CLASS net/minecraft/class_1559 net/minecraft/entity/mob/EndermiteEntity
 	METHOD method_7022 setPlayerSpawned (Z)V
 		ARG 1 playerSpawned
 	METHOD method_7023 isPlayerSpawned ()Z
+		COMMENT Returns whether this endermite was spawned from an enderpearl thrown by a player.
```
liach:
```suggestion
		COMMENT Returns whether this endermite was spawned from an ender pearl thrown by a player.
```

liach:
For that day light method, it sounds like `tickInDayLight` or so

YanisBft:
`isAffectedByDaylight` is pretty accurate imo

liach:
I think `shouldApplyDaylightEffect` may be better as this is call every tick to check if daylight effect should be applied

YanisBft:
I'm not a big fan of the word "effect" here. Will wait for more input on this

liach (commented):
No problem besides that debatable tick-based sunlight affect method

mappings/net/minecraft/entity/mob/PiglinBrain.mapping
```diff
@@ -41,8 +41,9 @@ CLASS net/minecraft/class_4838 net/minecraft/entity/mob/PiglinBrain
 	METHOD method_24732 create (Lnet/minecraft/class_4836;Lnet/minecraft/class_4095;)Lnet/minecraft/class_4095;
 		ARG 0 piglin
 		ARG 1 brain
-	METHOD method_24733 onGuardedBlockBroken (Lnet/minecraft/class_1657;Z)V
+	METHOD method_24733 onGuardedBlockInteraction (Lnet/minecraft/class_1657;Z)V
```
Runemoro: Do we have a convention on how to name `on___` methods? Should they be nouns (`onBlockBreak`, `onBlockInteraction`) or verbs (`onBlockBroken`, `onBlockInteracted`)?

mappings/net/minecraft/entity/passive/BeeEntity.mapping
```diff
@@ -1,6 +1,6 @@
 CLASS net/minecraft/class_4466 net/minecraft/entity/passive/BeeEntity
-	FIELD field_20353 multipleByteTracker Lnet/minecraft/class_2940;
-	FIELD field_20354 anger Lnet/minecraft/class_2940;
+	FIELD field_20353 MULTIPLE_BYTE_TRACKER Lnet/minecraft/class_2940;
```
Runemoro: This is a really weird name... I think it should just be something like `STATUS_TRACKER`, rather than focusing on implementation details.

Runemoro (approved):

mappings/net/minecraft/entity/mob/PiglinBrain.mapping
YanisBft: `onBlockInteracted` is better, will change

Runemoro: This should probably also be added to conventions.md (assuming everyone agrees with verbs).

Gegy (approved):

liach (approved):

PR #1700, opened 2020-08-17
rayTrace -> raycast

Scotsguy:
Quoting from wikipedia,
> In 3D computer graphics, ray tracing is a rendering technique for generating an image by tracing the path of light as pixels in an image plane and simulating the effects of its encounters with virtual objects.

https://en.wikipedia.org/wiki/Ray_tracing_(graphics)
> Ray casting can refer to [...] a non-recursive ray tracing rendering algorithm that only casts primary rays

https://en.wikipedia.org/wiki/Ray_casting

For example, ProjectileUtil#rayCast is used by GameRenderer#updateTargetedEntity to get the first entity within reach distance. The ray is never bounced off walls.

Another example, Item#rayCast is used by LilyPadItem to find the water source to place the lilypad on.

There aren't any rayTraces left, because there's nothing that actually follows a ray bouncing on a surface.

### DIFF
--- a/mappings/net/minecraft/world/RayCastContext.mapping
+++ b/mappings/net/minecraft/world/RayCastContext.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_3959 net/minecraft/world/RayTraceContext
+CLASS net/minecraft/class_3959 net/minecraft/world/RayCastContext
 	FIELD field_17553 start Lnet/minecraft/class_243;
 	FIELD field_17554 end Lnet/minecraft/class_243;
 	FIELD field_17555 shapeType Lnet/minecraft/class_3959$class_3960;
--- a/mappings/net/minecraft/block/AbstractBlock.mapping
+++ b/mappings/net/minecraft/block/AbstractBlock.mapping
@@ -154,7 +154,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		ARG 1 state
 		ARG 2 world
 		ARG 3 pos
-	METHOD method_9584 getRayCastShape (Lnet/minecraft/class_2680;Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
+	METHOD method_9584 getRaycastShape (Lnet/minecraft/class_2680;Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_265;
 		ARG 1 state
 		ARG 2 world
 		ARG 3 pos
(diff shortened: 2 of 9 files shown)


Pyrofab (approved):

BoogieMonster1O1 (approved):

Earthcomputer (approved):

i509VCB (approved):
After this is merged, some javadoc would be nice for RayCastContext and other misc stuff

haykam821 (approved):

shartte (approved):

YanisBft (approved):

modmuss50 (approved):

liach (approved):
I wonder if we can call these "hit" if they call it "hit result"

dblsaiko (approved):

ramidzkh (approved):

LambdAurora (approved):

Mysterious-Dev (approved):

Player3324 (approved):

Runemoro (approved):

Runemoro:
I agree, but I think it should be `castRay` rather than `rayCast`.

Look at the Wikipedia page, for example, which uses "cast ray" rather than "ray cast" as the verb:
> **cast**s primary **ray**s

liach (changes requested):
Agreed with rune. Most "ray cast" should be replaced with "cast ray" except `getRayCastShape` `RayCastContext`

Earthcomputer:
Disagree. `castRay` doesn't sound as natural in English as `rayCast`. The Wikipedia page only makes sense because the "rays" is plural. Otherwise it needs an article, otherwise `rayCast` makes more sense. Also that part of the article is merely explaining how raycasting works, rather than using what is normally used as the verb.

https://docs.unity3d.com/ScriptReference/30_search.html?q=raycast

Scotsguy:
I agree with Earth, additionally rayCast is simply used more often, based on my GitHub search of "castRay" returning 60k results vs "rayCast" returning 6 million

Runemoro:
> `castRay` doesn't sound as natural in English as `rayCast`

Compare:
 1. I am going to cast a ray.
 2. I am going to raycast.
 3. I am going to ray cast.

The first two work. The third looks wrong and is a grammar error.

> rayCast is simply used more often, based on my GitHub search of "castRay" returning 60k results vs "rayCast" returning 6 million

GitHub ignores capitalization, but if you look at the first 10 pages of results for "rayCast", you won't find a single instance of `rayCast`. They're all `raycast`.

So the choice is between `raycast` and `castRay`. I'd prefer the second since Wikipedia calls it "ray casting" (verb form is cast rays), not "raycasting" (verb form is raycast). But I'm not strongly against `raycast`.

liach:
Yeah `raycast` works better. So `getRaycastShape` `RaycastContext` etc

Earthcomputer:
:+1: for changing to lowercase raycast

Scotsguy:
You didnt actually say that the capitalization was the issue, I agree
I'll change it to lowercase when I'm back at a pc

Runemoro (approved):

mappings/net/minecraft/world/RayCastContext.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_3959 net/minecraft/world/RayTraceContext
+CLASS net/minecraft/class_3959 net/minecraft/world/RayCastContext
```
liach: Capitalization here too

liach (changes requested):

liach (approved):

modmuss50 (approved):
Lets go with this, seems very well recieved.

PR #1706, opened 2020-08-19
Add a mapping for method_30843 in SignBlockEntity

enbrain:


mappings/net/minecraft/block/entity/SignBlockEntity.mapping
```diff
@@ -21,3 +21,5 @@ CLASS net/minecraft/class_2625 net/minecraft/block/entity/SignBlockEntity
 	METHOD method_16126 getTextColor ()Lnet/minecraft/class_1767;
 	METHOD method_16127 setTextColor (Lnet/minecraft/class_1767;)Z
 		ARG 1 value
+	METHOD method_30843 getTextOnRow (I)Lnet/minecraft/class_2561;
```
i509VCB, line 24: does this throw an exception if you use an invalid line number (for say -1) or is it just plain array access to the `Text` lines field on the block entity

enbrain: It's just a plain array access. `method_30843` doesn't validate the argument. If you use -1, you will get `ArrayIndexOutOfBoundsException`.

liach: I call this true bikeshedding if you ask to document this!

liach (approved):

YanisBft (approved):

Runemoro (approved):

PR #1714, opened 2020-08-22
Update BlockBox.mapping

enbrain:


mappings/net/minecraft/util/math/BlockBox.mapping
```diff
@@ -70,3 +70,5 @@ CLASS net/minecraft/class_3341 net/minecraft/util/math/BlockBox
 	METHOD method_22874 getCenter ()Lnet/minecraft/class_2382;
 		COMMENT @implNote Biased toward the minimum bound corner of the box.
 	METHOD method_27234 infinite ()Lnet/minecraft/class_3341;
+	METHOD method_29299 offset (Lnet/minecraft/class_2382;)V
+		ARG 1 dv
```
haykam821: What is `dv`? Please use a more descriptive name.

haykam821 (changes requested):

mappings/net/minecraft/util/math/BlockBox.mapping
enbrain: I named it `dv` for consistency with the arguments (`ARG 1 dx`, `ARG 2 dy` and `ARG 3 dz`) of `method_14661 offset (III)V`. Is `deltaVec` or something like that better?

haykam821: I would name it `vec3d` for consistency with [the corresponding method in `Box`](https://github.com/FabricMC/yarn/blob/37ee48f77ac187aeb6bbc72e5c3b97d54f6f2309/mappings/net/minecraft/util/math/Box.mapping#L98):

```suggestion
		ARG 1 vec3d
```

enbrain: Is it more important than the consistency with `dx`, `dy` and `dz`?

I don't think `vec3d` is descriptive because being vec3d is pretty obvious from the type information. It doesn't add any information to it.

On the other hand, `delta` gives you additional information, that is, the distance between the original box and the result box. (Even a letter `d` typically means delta or distance in math context.)

I'm still thinking what is the best name. (We might need to refactor other code later.)

haykam821: Those `dx`, `dy`, and `dz` names aren't right; they should just be `x`, `y`, and `z`. I think it's obvious in the context of the signature (`BlockBox.offset(Lnet/minecraft/util/math/Vec3i;)V`) that we're offsetting the `BlockBox` by a `Vec3i` and the name should reflect that. Specifying 'delta' is redundant and shortening it to 'd' simply makes it more confusing.

Also, just noticed, this actually takes in a `Vec3i`:

```suggestion
		ARG 1 vec3i
```

liach:
```suggestion
		ARG 1 delta
```

enbrain: I will keep the argument unmapped for now.

liach: what makes delta so bad? maybe offset if you want to name like builders

haykam821: It should be named like `Box.offset(III)` is, and `BlockBox.offset(III)` should as well.

liach: just `delta`. `dx` `dy` etc have their `d` stand for "delta", so `delta` is the fitting name.

haykam821: `dx`, `dy`, and `dz` should just be `x`, `y`, and `z` as well. It's much less redundant to say, 'I am offsetting this BlockBox by a Vec3i', rather than, 'I am offsetting this BlockBox by the delta Vec3i'.

liach: Would worth a debate in a separate issue.

liach (approved):

liach (changes requested):
Wait a sec... this offset name and the original offset names are bad.
The `offset` imo should be `move`, and `translated` should be `offset`. Similar to in mutable block pos, `offset` makes a copy object while `move` changes this object

liach (approved):

YanisBft (approved):

Gegy (approved):

haykam821 (approved):

modmuss50 (approved):

PR #1723, opened 2020-08-25
Clean up javadoc on Identifier(String) constructor

Leo40Git:
This was done without Enigma! Since this is, quite literally, changing 2 COMMENT lines.


YanisBft (approved):

liach (changes requested):
Should really move this to method docs

liach:
would be nice to note that this throws `InvalidIdentifierException` while `tryParse` returns null for invalid strings.

liach (approved):

Gegy (approved):

PR #1724, opened 2020-08-25
class_5423 -> LevelAccessor

Mysterious-Dev:
This PR complete class_5423 mappings.

### DIFF
--- a/mappings/net/minecraft/world/level/LevelAccessor.mapping
+++ b/mappings/net/minecraft/world/level/LevelAccessor.mapping
@@ -1,0 +1,3 @@
+CLASS net/minecraft/class_5423 net/minecraft/world/level/LevelAccessor
+	METHOD method_30349 getRegistryManager ()Lnet/minecraft/class_5455;
+	METHOD method_31081 getBiomeName (Lnet/minecraft/class_2338;)Ljava/util/Optional;
--- a/mappings/net/minecraft/class_5423.mapping
+++ b/mappings/net/minecraft/class_5423.mapping
@@ -1,2 +1,0 @@
-CLASS net/minecraft/class_5423
-	METHOD method_30349 getRegistryManager ()Lnet/minecraft/class_5455;


liach (changes requested):
The heck, this should really be registry accessible world. why does this have this anything to do with level
concern

liach:
uh no worries, you don't have to close... like you can wait for people to get a better name. Also see #1688

liach:
So for my concerns I will describe in detail:
1. There is nothing specific about level here. In yarn we only use level to refer to `level.dat` or related level properties content, unlike mojang who use it to refer to worlds
2. this class mainly offers access to dynamic registries, which is not conveyed in `LevelAccessor` name at all

mappings/net/minecraft/world/level/LevelAccessor.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS net/minecraft/class_5423 net/minecraft/world/level/LevelAccessor
+	METHOD method_30349 getRegistryManager ()Lnet/minecraft/class_5455;
+	METHOD method_31081 getBiomeName (Lnet/minecraft/class_2338;)Ljava/util/Optional;
```
shartte, line 3:
```suggestion
	METHOD method_31081 getBiomeKey (Lnet/minecraft/class_2338;)Ljava/util/Optional;
```

mappings/net/minecraft/world/level/LevelAccessor.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS net/minecraft/class_5423 net/minecraft/world/level/LevelAccessor
```
modmuss50, line 1: Yeah, this is just wrong, and sadly rasies my suspicions as to where you got this name from. For the sake of the yarn project I dont think I feel comfortable accepting this PR.

Thanks for understanding.

PR #1735, opened 2020-08-30
An extensions of  #1621

Suya1671:
yes


Suya1671:
(Also how is yarn so active? i started 2 hours ago and there is already 10 new commits)

modmuss50:
Sadly I don’t think we want this everywhere. One small thing is a fine but not this. Sorry.

Suya1671:
idc if this was too be pulled. Fun making it tho

PR #1743, opened 2020-09-03
Suggest using Block#asItem instead of Item.fromBlock

enbrain:
`Item.fromBlock` has `@Deprecated` annotation. `Block#asItem` calls `Item.fromBlock` internally but it also caches the result to improve performance. We should use `Block#asItem` instead of `Item.fromBlock`.


enbrain:
This PR was originally targeting 1.16.2, but it's still relevant in 1.16.3.

YanisBft:
Do we do that in yarn docs?

Gegy (approved):

YanisBft (approved):

PR #1756, opened 2020-09-10
Add some class javadocs

SoniEx2:
They're not very good because I'm making them up cleanroom from the mappings alone.

### DIFF
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -1,4 +1,5 @@
 CLASS net/minecraft/class_2248 net/minecraft/block/Block
+	COMMENT A block type.
 	FIELD field_10638 LOGGER Lorg/apache/logging/log4j/Logger;
 	FIELD field_10642 translationKey Ljava/lang/String;
 	FIELD field_10646 defaultState Lnet/minecraft/class_2680;
--- a/mappings/net/minecraft/item/BlockItem.mapping
+++ b/mappings/net/minecraft/item/BlockItem.mapping
@@ -1,4 +1,5 @@
 CLASS net/minecraft/class_1747 net/minecraft/item/BlockItem
+	COMMENT An item type for a block type.
 	FIELD field_7901 block Lnet/minecraft/class_2248;
 	METHOD <init> (Lnet/minecraft/class_2248;Lnet/minecraft/class_1792$class_1793;)V
 		ARG 1 block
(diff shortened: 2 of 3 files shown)


mappings/net/minecraft/block/Block.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS net/minecraft/class_2248 net/minecraft/block/Block
+	COMMENT A block type.
```
i509VCB, line 2: I'd argue we don't really need javadoc in this capacity since this is a pretty universally known concept.

mappings/net/minecraft/item/Item.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS net/minecraft/class_1792 net/minecraft/item/Item
+	COMMENT An item type.
```
i509VCB, line 2: Similar justification to `Block`

mappings/net/minecraft/item/BlockItem.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS net/minecraft/class_1747 net/minecraft/item/BlockItem
+	COMMENT An item type for a block type.
```
i509VCB, line 2: I would probably reword this to be a bit closer to the following:

```suggestion
	COMMENT Represents a block in item form.
	COMMENT This item handles the placement of a block in a world along with applying any block entity data present on the item stack.
```

i509VCB (changes requested):

SoniEx2:
We do believe javadocs for Block and Item are in order - Block could be an in-world location or subdivision (really, there's a free software voxel engine that does something like this), and Item could also be an in-world item or in-inventory item. And additionally, Java, Python, etc all have class docs for everything that's directly exposed to the API user.

If Block was some implementation detail, then sure, leave it without docs. but as far as we can tell it isn't.

frqnny:
But with minecraft this is really unnecessary. Blocks are blocks and it'll be confusing if you use "block" for something else they'll look at you weird.

SoniEx2:
Can't have yarnimpl without having these javadocs. :v

On a more serious note, we maintain our position.

mappings/net/minecraft/block/Block.mapping
Gegy: If these do have documentation, it's probably worth mentioning that they are singleton instances. Especially for items, it's a common mistake to attach state to an item implementation which becomes unintentionally global.

SoniEx2: That's not what a singleton is.

modmuss50:
The docs provided here dont provide any more help or context IMO. Ideally these classes would be fully documented but thats a large amount of work.

PR #1757, opened 2020-09-12
Some fixes

### LINKED ISSUE #1654: AbstractBlock.Settings.lightLevel -> luminance
It sets the `luminance` field and controls the value of `BlockState.getLuminance`.

### LINKED ISSUE #1740: ChestBlockEntity#countViewers parameter names are wrong
https://github.com/FabricMC/yarn/blob/097b138f6226efaf4c308e6334cf76da8a163839/mappings/net/minecraft/block/entity/ChestBlockEntity.mapping#L18-L20

Should be `x, y, z` not `ticksOpen, x, y`

`container` should also be changed to `inventory` ( although inventory is already used in generated locals and I'm not sure if that would cause issues. )

### LINKED ISSUE #1744: CommandSuggestor.slashRequired -> slashOptional
It does the opposite of the current name, i.e. makes the `/` optional when parsing the command.

### LINKED ISSUE #1745: Move CommandSource from net.minecraft.server.command to net.minecraft.command
It's implemented by `ClientCommandSource` which is client-only.

### LINKED ISSUE #1746: ChatListenerHud -> HudChatListener
It's a `(Client)ChatListener` that displays messages on the HUD, but not a HUD itself.

YanisBft:
Closes #1654 
Closes #1740 
Closes #1744 
Closes #1745 
Closes #1746

### DIFF
--- a/mappings/net/minecraft/block/AbstractBlock.mapping
+++ b/mappings/net/minecraft/block/AbstractBlock.mapping
@@ -316,6 +316,7 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		METHOD method_9630 copy (Lnet/minecraft/class_4970;)Lnet/minecraft/class_4970$class_2251;
 			ARG 0 block
 		METHOD method_9631 luminance (Ljava/util/function/ToIntFunction;)Lnet/minecraft/class_4970$class_2251;
+			COMMENT Specifies the light level emitted by a block.
 			ARG 1 luminance
 		METHOD method_9632 strength (F)Lnet/minecraft/class_4970$class_2251;
 			ARG 1 strength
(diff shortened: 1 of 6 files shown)


ghost (approved):

LambdAurora (changes requested):
I really disagree with the `luminance` rename, I think that `AbstractBlock$Settings#lightLevel` should be renamed to `luminance` instead.

`luminance` just describes the setting better when you look at the definition: "the intensity of light emitted from a surface per unit area in a given direction." in this case it's implicitly in "all direction". The definition just fits the variable very well and I prefer to use a word than to try to name a variable with multiple words.

I can understand people not being comfortable with `luminance` as it's a bit more uncommon than `lightLevel` but I think we should not lower the level of language used in yarn.

LambdAurora (approved):

Juuxel (approved):

Prospector (approved):

shartte (approved):

frqnny (approved):

BoogieMonster1O1 (approved):

modmuss50 (approved):

PR #1709, opened 2020-08-22
Mapped most of the Entity Model classes

halotroop2288:
Copied *most* of the changes from 1.16.1+build.21. This package should now be on-par with that build.
Some of the mappings are deductions of my own.

PR #1759, opened 2020-09-12
Entity model things

YanisBft:
These mappings are ported from the #1709 1.14.4 PR


frqnny (approved):
Lol, why did we name it Snowman instead of SnowGolem

haykam821:
The ID for snow golems was `minecraft:snowman` prior to the Flattening.

modmuss50 (approved):

PR #1764, opened 2020-09-20
Finish mapping the CreditsScreen class

haykam821:
This pull request maps the `field_24260` and `field_24261` fields in the `CreditsScreen` class to `COLORS` and `centeredLines` respectively.

### DIFF
--- a/mappings/net/minecraft/client/gui/screen/CreditsScreen.mapping
+++ b/mappings/net/minecraft/client/gui/screen/CreditsScreen.mapping
@@ -1,5 +1,5 @@
 CLASS net/minecraft/class_445 net/minecraft/client/gui/screen/CreditsScreen
-	FIELD field_24260 COLORS Ljava/lang/String;
+	FIELD field_24260 OBFUSCATION_PLACEHOLDER Ljava/lang/String;
 	FIELD field_24261 centeredLines Lit/unimi/dsi/fastutil/ints/IntSet;
 	FIELD field_2626 MINECRAFT_TITLE_TEXTURE Lnet/minecraft/class_2960;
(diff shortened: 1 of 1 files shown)


YanisBft (approved):

mappings/net/minecraft/client/gui/screen/CreditsScreen.mapping
```diff
@@ -1,4 +1,6 @@
 CLASS net/minecraft/class_445 net/minecraft/client/gui/screen/CreditsScreen
+	FIELD field_24260 COLORS Ljava/lang/String;
```
enbrain:
```suggestion
	FIELD field_24260 OBFUSCATE_PLACEHOLDER Ljava/lang/String;
```

Correct me if I'm wrong, but it's more like a placeholder rather than a mere collection of colors.

haykam821: It replaces the color with white obfuscation if it is not in the `COLORS` field.

enbrain: I can't find the code that checks if something is in the `COLORS` field. Only I found is `String#indexOf` taking the field **as an argument** and this means it searches for the field in the other string.

haykam821: It uses a string rather than a collection for the colors whitelist.

haykam821: Fixed in 1f1569f25302c22c08bcd8300e497b4e2e3d433e

enbrain (approved):

PR #1770, opened 2020-09-25
AbstractButtonWidget -> AbstractClickableWidget

### LINKED ISSUE #1752: AbstractButtonWidget is not only for buttons
Some implementations of `AbstractButtonWidget` are sliders. Since you can click on a slider to set the position of the slide, I'd say we could rename it to `(Abstract)ClickableWidget` since it handles the clicking of a mouse.

i509VCB:
Fixes #1752

Why?
`AbstractButtonWidget` is not used exclusively by buttons. This class is used by the text field and slider widgets. Whats common between those two widget types and button widgets is that you "click" the widget.

Also `renderBg` -> `renderBackground`

### DIFF
--- a/mappings/net/minecraft/client/gui/widget/AbstractPressableButtonWidget.mapping
+++ b/mappings/net/minecraft/client/gui/widget/AbstractPressableButtonWidget.mapping
@@ -1,2 +1,3 @@
 CLASS net/minecraft/class_4264 net/minecraft/client/gui/widget/AbstractPressableButtonWidget
+	COMMENT A simple widget which handles a clicking action and plays a click sound.
 	METHOD method_25306 onPress ()V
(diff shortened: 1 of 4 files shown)


YanisBft (approved):
The same could be done with `AbstractPressableButtonWidget` which is also used by checkboxes

shartte:
👎 This is more than a clickable widget.
It has focus behavior, positioning, size and narration too.

What about InteractiveWidget?

LoganDark:
GuiWidget?

i509VCB:
GuiWidget is far too vague

mappings/net/minecraft/client/gui/widget/AbstractPressableButtonWidget.mapping
```diff
@@ -1,2 +1,3 @@
 CLASS net/minecraft/class_4264 net/minecraft/client/gui/widget/AbstractPressableButtonWidget
```
Gegy, line 1: The distinction between being clickable and being pressable seems a bit unclear, might be worth renaming this?

Prospector:
If there's not a ClickableWidget already, you can drop the Abstract. Though I'm not convinced this name is ideal.

liach:
@i509VCB Mind update this to snapshots? I am willing to endorse if you do so.

liach:
@i509VCB mind resolve the merge conflicts?

liach:
If you don't want to maintain it, consider asking me to do so.

i509VCB:
Feel free to take it.

PR #1788, opened 2020-10-10
Stronghold, fortress, mineshaft, and general structure mapping

ShySkream:
First time contributing! I should have kept it smaller but I couldn't help myself.  Main changes include:

- all methods and most fields mapped (I wasn't sure on naming standards so I left some)
- removing all mention of jigsaw, since they aren't used in these old structures
- setting consistent names across files since the structures operate in a similar manner


Gegy (approved):

modmuss50 (approved):
Nice work 👍

PR #1789, opened 2020-10-10
Add Javadoc for property classes

enbrain:


Juuxel (approved):

mappings/net/minecraft/state/property/IntProperty.mapping
```diff
@@ -5,6 +5,14 @@ CLASS net/minecraft/class_2758 net/minecraft/state/property/IntProperty
 		ARG 2 min
 		ARG 3 max
 	METHOD method_11867 of (Ljava/lang/String;II)Lnet/minecraft/class_2758;
+		COMMENT Creates an integer property.
+		COMMENT
+		COMMENT <p>{@code min} must be non-negative and {@code max} must be greater than {@code min}.
+		COMMENT
+		COMMENT <p>Note that this method takes O({@code max} - {@code min}) time as it computes all possible values during instantiation.
```
YanisBft, line 12: What does `O({@code max} - {@code min}) time` mean?

YanisBft (approved):

mappings/net/minecraft/state/property/IntProperty.mapping
enbrain: It states that the time complexity of the method is up to linear in the distance between `min` and `max`.

https://en.wikipedia.org/wiki/Time_complexity
https://en.wikipedia.org/wiki/Big_O_notation

PR #1805, opened 2020-10-13
Add a simple `decompileCFR` task

modmuss50:
I have seen a number of people trying to use enigma's export source with mixed results. 

This provides a quick and simple way to export the source.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -679,10 +679,10 @@ task decompileCFR(type: JavaExec, dependsOn: "mapNamedJar") {
 	classpath = configurations.decompileClasspath
 	main = "org.benf.cfr.reader.Main"
 
-	args namedJar.getAbsolutePath(), "--outputdir", file("src").absolutePath
+	args namedJar.getAbsolutePath(), "--outputdir", file("namedSrc").absolutePath
 
 	doFirst {
-		file("src").deleteDir()
+		file("namedSrc").deleteDir()
 	}
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -37,6 +37,10 @@ repositories {
 		name "Fabric Repository"
 		url 'https://maven.fabricmc.net'
 	}
+	maven {
+		name "Mojang"
+		url 'https://libraries.minecraft.net/'
+	}
 }
 
 configurations {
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -667,6 +675,17 @@ task genFakeSource(dependsOn: ["buildYarnTiny", "mapNamedJar"]) {
 	}
 }
 
+task decompileCFR(type: JavaExec, dependsOn: "mapNamedJar") {
+	classpath = configurations.decompileClasspath
+	main = "org.benf.cfr.reader.Main"
+
+	args namedJar.getAbsolutePath(), "--outputdir", file("src").absolutePath
```
haykam821: An option should be exposed to change the output path from something other than `./src`.

build.gradle
```diff
@@ -667,6 +675,17 @@ task genFakeSource(dependsOn: ["buildYarnTiny", "mapNamedJar"]) {
...
+
+	doFirst {
+		file("src").deleteDir()
```
haykam821: Shouldn't this be `minecraftSrc` or similar to be in line with `remappedSrc`, and in certain scenarios, not overwrite an existing `src` folder?

haykam821 (changes requested):

build.gradle
modmuss50: I dont think thats needed, adds added compliexity for something that I dont think has a lot of gain.

build.gradle
modmuss50: there isnt an existing src, I did kinda think of dumping it in src/main/java and making it importable into an ide so you could explore the code a bit easier but thats prob not needed.

I think what we have here is fine? What do others think?

modmuss50: on 2nd throughts, remappedSrc does make more sense. I will change this

haykam821: Another solution would be to potentially halt the process if `src` already exists.

build.gradle
haykam821: If there's already an existing `src`, this is what users will have to use without an option:

```bash
mv src tempSrc
./gradlew decompileCFR
mv src minecraftSrc
mv tempSrc src
```

build.gradle
modmuss50: changed to `namedSrc` to follow the jar.

hYdos (approved):

YanisBft (approved):

PR #1807, opened 2020-10-15
Incorrect mapping in RealmsSlotOptionsScreen

ReeceXW:


enbrain (approved):

mappings/net/minecraft/client/realms/gui/screen/RealmsSlotOptionsScreen.mapping
```diff
@@ -48,9 +48,9 @@ CLASS net/minecraft/class_4423 net/minecraft/client/realms/gui/screen/RealmsSlot
...
 		FIELD field_22726 max D
 		METHOD <init> (Lnet/minecraft/class_4423;IIIIFF)V
-			ARG 2 id
-			ARG 3 x
-			ARG 4 y
-			ARG 5 width
+			ARG 2 x
+			ARG 3 y
+			ARG 4 width
+			ARG 5 height
```
YanisBft, line 54: According to where this method is called, this param seems to be the current value

ReeceXW: This inner class extends SliderWidget.  

Currently it calls super with
`super(id, x, y, 20, LiteralText.EMPTY, 0.0D);`

But the SliderWidget constructor is 
`SliderWidget(int x, int y, int width, int height, Text text, double value)`

You can see in the incorrect mapping that there is no height in that super call

YanisBft: I'm not sure I understood what you say. The height passed in super constructor is 20, not the param, so what you've called "height" is not the height, it's the value (look at the constructor calls)

PR #1829, opened 2020-11-04
Add entity mappings

LambdAurora:
`Entity#removed` is gone, now we have removal reason.


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -221,6 +222,7 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 		ARG 1 onGround
 	METHOD method_25936 getLandingBlockState ()Lnet/minecraft/class_2680;
 	METHOD method_27298 shouldSpawnSprintingParticles ()Z
+	METHOD method_29239 dismountVehicle ()V
```
Pyrofab, line 225: We should probably emphasize somehow that you are generally supposed to call `stopRiding` rather than this method

LambdAurora: Pushed some javadoc, hopefully it's enough.

modmuss50 (approved):

YanisBft (approved):

YTG1234 (approved):

PR #1836, opened 2020-11-05
BlockEntity tick and creation

AlexIIL:
Major changes:

* Named the new `Tickable` replacement class `BlockEntityTicker`
* Named the per-world chunk holder interface for `BlockEntityTicker` `BlockEntityTickInvoker`.
* Named the (package-private) functional interface for constructing BlockEntities (in `BlockEntityType`) `BlockEntityFactory`
* Named the side-agnostic various static BlockEntity tick methods `tick`, and the server-specific and client-specific tick methods `serverTick` and `clientTick` respectively.

### DIFF
--- a/mappings/net/minecraft/block/entity/BlockEntityType.mapping
+++ b/mappings/net/minecraft/block/entity/BlockEntityType.mapping
@@ -17,7 +17,12 @@ CLASS net/minecraft/class_2591 net/minecraft/block/entity/BlockEntityType
 		FIELD field_11915 supplier Lnet/minecraft/class_2591$class_5559;
 		FIELD field_19316 blocks Ljava/util/Set;
 		METHOD <init> (Lnet/minecraft/class_2591$class_5559;Ljava/util/Set;)V
+			ARG 1 creator
 			ARG 2 blocks
 		METHOD method_11034 build (Lcom/mojang/datafixers/types/Type;)Lnet/minecraft/class_2591;
 		METHOD method_20528 create (Lnet/minecraft/class_2591$class_5559;[Lnet/minecraft/class_2248;)Lnet/minecraft/class_2591$class_2592;
 			ARG 1 blocks
+	CLASS class_5559 BlockEntityCreator
+		METHOD create (Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;)Lnet/minecraft/class_2586;
+			ARG 1 pos
+			ARG 2 state
--- a/mappings/net/minecraft/block/entity/BlockEntityType.mapping
+++ b/mappings/net/minecraft/block/entity/BlockEntityType.mapping
@@ -1,23 +1,25 @@
 CLASS net/minecraft/class_2591 net/minecraft/block/entity/BlockEntityType
-	FIELD field_11892 supplier Lnet/minecraft/class_2591$class_5559;
+	FIELD field_11892 factory Lnet/minecraft/class_2591$class_5559;
 	FIELD field_11893 LOGGER Lorg/apache/logging/log4j/Logger;
 	FIELD field_11909 type Lcom/mojang/datafixers/types/Type;
 	FIELD field_19315 blocks Ljava/util/Set;
 	METHOD <init> (Lnet/minecraft/class_2591$class_5559;Ljava/util/Set;Lcom/mojang/datafixers/types/Type;)V
+		ARG 1 factory
 		ARG 2 blocks
 		ARG 3 type
 	METHOD method_11030 create (Ljava/lang/String;Lnet/minecraft/class_2591$class_2592;)Lnet/minecraft/class_2591;
 	METHOD method_11032 instantiate (Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;)Lnet/minecraft/class_2586;
 	METHOD method_11033 getId (Lnet/minecraft/class_2591;)Lnet/minecraft/class_2960;
 	METHOD method_20526 supports (Lnet/minecraft/class_2680;)Z
+		ARG 1 state
 	METHOD method_24182 get (Lnet/minecraft/class_1922;Lnet/minecraft/class_2338;)Lnet/minecraft/class_2586;
 		ARG 1 world
 		ARG 2 pos
 	CLASS class_2592 Builder
-		FIELD field_11915 supplier Lnet/minecraft/class_2591$class_5559;
+		FIELD field_11915 factory Lnet/minecraft/class_2591$class_5559;
 		FIELD field_19316 blocks Ljava/util/Set;
 		METHOD <init> (Lnet/minecraft/class_2591$class_5559;Ljava/util/Set;)V
-			ARG 1 creator
+			ARG 1 factory
 			ARG 2 blocks
 		METHOD method_11034 build (Lcom/mojang/datafixers/types/Type;)Lnet/minecraft/class_2591;
 		METHOD method_20528 create (Lnet/minecraft/class_2591$class_5559;[Lnet/minecraft/class_2248;)Lnet/minecraft/class_2591$class_2592;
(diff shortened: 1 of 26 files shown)


AlexIIL:
Oh, and it looks like this partially overlaps  #1838

kvverti:
I like my `TickAction` better but the rest is great and mostly-nonoverlapping

AlexIIL:
To be honest I'm not particularly fond of either name - `TickAction` sounds like it might apply to other things (although `Tickable` had that problem too in 1.16 - javadocs make this much better though), while `BlockEntityTicker` is a bit unwieldy and leads to long names like `ChunkBlockEntityTicker` or `BlockEntityTickerHolder`.

i509VCB:
Honestly I thought TickLogic originally but TickAction could also work

AlexIIL:
From what i can tell other interfaces that are related to `BlockEntity` seem to include `BlockEntity` in their name (`BlockEntityProvider`, `BlockEntityRenderer`) so I thought it would make sense to continue that.

Player3324:
To me *Ticker sounds more like something that handles ticking on a larger scale[1], while this just seems to be some wrapper class for a single thing-to-tick. This suggests names like BeTickWrapper etc.

Not sure why it exists, I don't see a reason other than pulling code even further apart. Any ideas?

[1] Think a thread with a loop to tick everything

AlexIIL:
As a sidenote, I'm currently naming the server/client tick methods as `tickServer` and `tickClient` - even when there's only one of them, but it's only used on one side. Is that a good idea?

AlexIIL:
Sadly there's a (mutable, package-private in `WorldChunk`) wrapper for this class anyway - so if `BlockEntityTicker/TickAction` were called `BeTickWrapper` then we'd end up with `BeTickWrapperWrapper`, which doesn't sound ideal.

Player3324:
It's not like it wasn't actually a wrapper of a wrapper :D

We need to establish the purpose of those classes a little more

`class_5558` is the new  tickable, BE specific this time
`class_5563` is the actual BE tick wrapper
`class_5562` is the interface abstracting the api of the BE tick wrapper, not sure if truly BE specific
`class_2818$1` is some no-op impl of the above interface
`class_5564` is a wrapper of the interface, no idea why - the wrapper of the wrapper in any case

AlexIIL:
`class_5564` seems to be used to allow `class_5558` to be swapped out without modifying the world's ticking list (maybe to ensure consistency? - this way the wrong `class_5558` is never called), and `class_2818$1` is used when the BlockEntity is removed (or the ticker is removed?)

As far as I can tell vanilla doesn't actually swap out the `class_5558` to a different one, ever - but mods might want to. *Edit: for example LibMultiPart might want to refresh it's ticker if a part is added that requires ticking, or remove it's ticker if all of the parts that tick are removed)*

AlexIIL:
*Sigh*. Turns out I've been confusing `class_5558` (the actual ticker) with `class_5562` (the named, positioned, tick wrapper) - thanks for listing them specifically :).

i509VCB:
Isn't the proper naming like serverTick in yarn standards?

AlexIIL:
`serverTick` makes more sense to me than `tickServer` too, I'll swap them over.

kvverti:
`class_5558` is _not_ the new `Tickable`. It is a functional interface and is implemented with method references to the static `tick*` methods. `TickAction` is to have similarity to the `Action` set of functional interfaces.

Player3324:
kvverti I am aware, but that it is no longer coupled to the BE instance directly doesn't fundamentally change what it is/does. Not sure which `Action` interfaces you mean, doesn't sound very fitting to me?

Wrt `class_5564` it seems to have its contents swapped around in `WorldChunk.method_31723`, but then why didn't they simply use `Map.Entry.setValue` or another `Map.put`? With the current use in the `Map<BlockPos, class_5564>` (`field_27222`) a name like *Slot for `class_5564` may fit the purpose better despite the "wrapper" `toString` mention?

AlexIIL:
Ah sorry I should have specified that replacing is primarily useful in the two `List<class_5562>`s in `World`. (Specifically `World.field_27082`, which is the one that is actually ticked).

Player3324:
Yes, however the chunk's map is where it actually gets used that way - for no benefit obvious to me. The swapping method is private, so only the chunk could use it as it is in the same compilation unit.

AlexIIL:
It looks like the World only ever gets instances of `class_5564` (passed in via `WorldChunk.method_31723`) which would make it useful? I assume the swapping method is private as it's only updated by `WorldChunk`, but it should affect the entry in the `World` list?

mappings/net/minecraft/block/entity/BlockEntityType.mapping
```diff
@@ -17,7 +17,12 @@ CLASS net/minecraft/class_2591 net/minecraft/block/entity/BlockEntityType
 		FIELD field_11915 supplier Lnet/minecraft/class_2591$class_5559;
 		FIELD field_19316 blocks Ljava/util/Set;
 		METHOD <init> (Lnet/minecraft/class_2591$class_5559;Ljava/util/Set;)V
+			ARG 1 creator
 			ARG 2 blocks
 		METHOD method_11034 build (Lcom/mojang/datafixers/types/Type;)Lnet/minecraft/class_2591;
 		METHOD method_20528 create (Lnet/minecraft/class_2591$class_5559;[Lnet/minecraft/class_2248;)Lnet/minecraft/class_2591$class_2592;
 			ARG 1 blocks
+	CLASS class_5559 BlockEntityCreator
```
Gegy: `BlockEntityFactory`?

AlexIIL: Yeah. that's much better. (And actually follows the conventions :p )

mappings/net/minecraft/block/entity/CampfireBlockEntity.mapping
```diff
@@ -12,3 +12,6 @@ CLASS net/minecraft/class_3924 net/minecraft/block/entity/CampfireBlockEntity
 	METHOD method_17507 saveInitialChunkData (Lnet/minecraft/class_2487;)Lnet/minecraft/class_2487;
 		ARG 1 tag
 	METHOD method_17510 updateListeners ()V
+	METHOD method_31666 tickLit (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Lnet/minecraft/class_3924;)V
+	METHOD method_31667 tickUnlit (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;Lnet/minecraft/class_3924;)V
```
i509VCB: `litTick` and `unlitTick`?

Also is this on logical server or logical client?

Player3324 (approved):

mappings/net/minecraft/block/entity/CampfireBlockEntity.mapping
AlexIIL: Logical server - there's another tick method for client only ticks. (Should I rename this to `litServerTick/unlitServerTick`?)

i509VCB: Hmm yeah that seems good.

mappings/net/minecraft/block/entity/BlockEntityType.mapping
```diff
@@ -17,7 +17,13 @@ CLASS net/minecraft/class_2591 net/minecraft/block/entity/BlockEntityType
 		FIELD field_11915 supplier Lnet/minecraft/class_2591$class_5559;
 		FIELD field_19316 blocks Ljava/util/Set;
 		METHOD <init> (Lnet/minecraft/class_2591$class_5559;Ljava/util/Set;)V
+			ARG 1 creator
```
jaskarth:
```suggestion
			ARG 1 factory
```
Seems like you missed a param 😅

AlexIIL: Woops, thanks. I've also updated the other `supplier` names to `factory` too.

modmuss50 (approved):

Gegy (approved):

PR #1837, opened 2020-11-05
Map NBT tag visitors.

kvverti:


mappings/net/minecraft/nbt/visitor/NbtOrderedStringFormatter.mapping
```diff
@@ -0,0 +1,30 @@
+CLASS net/minecraft/class_5625 net/minecraft/nbt/visitor/NbtOrderedStringFormatter
```
Gegy, line 1: Is this used specifically in pretty-printing like with the functions in `NbtHelper`? Would it make sense for the names to be similar?

kvverti: It's almost exactly like `NbtTextFormatter` except that it sorts the keys in a janky way.

haykam821 (approved):

YanisBft (approved):

PR #1839, opened 2020-11-05
Nomenclature surrounding how copper oxidizes in Minecraft

Jkmcameron:
This mapping differs from the one the others settled on because I don't agree with mapping this Interface to "Oxidizable" since there's no reason why the very same interface couldn't be used to create a plant that degrades over time. The name "Oxidizable" implies that it's usage is limited to metals, and "Corrodable" has the same issue, it is only "Degradable" that perfectly illustrates what this interface is responsible for without defining an implicit scope


haykam821 (changes requested):
I don't think that this nomenclature is any better than what was suggested in #1830. Degradation could imply that part of the block is broken; meanwhile, oxidization is the only behavior this interface is used for in vanilla.

OroArmor:
However, the features that the name `Degradation` implies is wider than `Oxidizable`. As well, (I haven't checked the code), copper still `oxidizes` even though it could be surrounded by blocks on all sides

i509VCB:
The name of this interface should relate to current uses imo. My guess is that only copper like blocks use the interface.

OroArmor:
Potentially, but for modders it shows a different feature aspect, something that these blocks can evolve over time. I do agree that things should be how they are in vanilla, but the use of this might make it a better choice to make it have a more abstract name.

Jkmcameron:
> The name of this interface should relate to current uses imo. My guess is that only copper like blocks use the interface.

> Degradation could imply that part of the block is broken; meanwhile, oxidization is the only behavior this interface is used for in vanilla.

What logic is this? Are we creating the mappings for Mojang or for the modding community? The name shouldn't be limited by its vanilla usage, it should be limited by its full range of potential usages. Sure, Mojang added this bit of code solely for the purpose of oxidizing copper, but a modder could just as easily use this very same code to create a plant that wilts over time.

jaskarth:
`Oxidizable` is the best choice here because only copper uses it currently. Instead, you should add javadoc to the class telling modders that it can have wider uses.

mappings/net/minecraft/block/CopperBlock.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS net/minecraft/class_5548 net/minecraft/block/Copper_Block
```
Gegy:
```suggestion
CLASS net/minecraft/class_5548 net/minecraft/block/CopperBlock
```

Jkmcameron:
> `Oxidizable` is the best choice here because only copper uses it currently. Instead, you should add javadoc to the class telling modders that it can have wider uses.

Following that logic we should name it "CopperOxidizable", wouldn't want anyone getting the wrong idea and thinking this interface was meant to be used by anything other than copper.

jaskarth:
Well, you can use the interface for anything that you want to, really. The names of classes should be used to describe the behavior by vanilla, as modders can use it for a wide range of behaviors. Likewise, we don't need to indicate to modders how to use a class as it's ultimately up to them.

YanisBft:
Closing this as people seem to be against this. Please open an issue if you still think your proposal is better

PR #1857, opened 2020-11-06
Use github actions to publish releases

modmuss50:
I dont see any issue with the current build number system, but now is a good chance to change that if there is a better solution.

This should be used from the start of the next snapshot.


coderbot16 (approved):
Seems pretty straightforward :+1:

YanisBft (approved):

jaskarth (approved):

haykam821:
Merged via b14dbfa228e5738d6f078f8a209b95f562dba1de

PR #1879, opened 2020-11-15
client.util.math.Vector3f -> util.math.Vec3f

### LINKED ISSUE #978: Vector3f is in client.util.math

### LINKED ISSUE #1716: Vec2f, Vec3d, Vector3f?
These names seem to be inconsistent. The fully named one is client only but that isn't a great naming standard. I feel in this case we need to consolidate these names in 1.17 snapshots

YanisBft:
Even though most of its methods are client only, Vector3f is not client only, so it has been moved to util.math. Also renamed to follow how the other vector classes are named there.
~~To be consistent with Vector3f, the other vector classes were renamed to Vector* instead of Vec*.~~

Closes #978 and closes #1716 

Also fixed invalid mappings and updates Enigma


Player3324:
It'd be more in line with the conventions to fix the abbreviations?

YanisBft:
wdym?

Juuxel:
Isn't `Vector3f` mutable unlike the `util.math.Vec*`s? That should be in its name IMO.

Player3324:
Yarn generally favors the long form. While we already have some uses of "vec", "vector" is used often enough and matches Matrix 

The (im)mutability is probably not worth mentioning. If you look at Vec3i, it's all inconsistent... Vec3i even has an `@Unmodifiable` annotation, but that's a lie for at least some sub-classes.. Currently only Vec2f and Vec3d are properly immutable, the other vectors and matrices are not.

YanisBft:
I agree, doesn't need to specify the immutability.

Regarding the abbreviation, I think "Vec" is clear enough, but if people prefer to use "Vector", I'll rename the 4 vector classes in the package

Player3324:
<https://github.com/FabricMC/yarn/search?q=vec> has a few locations where vector would fit better IMO, e.g getRotationVec (vs getRotVec / getRotationVector) or all the "vec" arg names. It is a little weird to go there, but not Matrix (Mat), Rotation (Rot), Direction (Dir), ...

The 2/3/4 i/d suffixes make it less bad, not sure what the community wants at large

Earthcomputer:
Disagree with lengthening the other class names. The abbreviations are so common that they are pretty standard. Also, would you be in favour of lengthening `BlockPos` to `BlockPosition`? I'd think not.
I agree that the method names you mentioned should be lengthened.

As for this PR, this vector class is clearly meant for client-side use so that needs to be preserved somewhere. It's unfortunate that the package needs to change, so maybe in the documentation? Agree with Juuxel that this could be renamed to `MutableVec3f`.

Runemoro:
I agree with Player about preferring not using abbreviations. It's easier to read text that isn't full of abbreviations, and likely not slower to type since your IDE will auto-complete it anyway.

YanisBft:
Renamed the Vec* classes to Vector*

Runemoro (approved):

i509VCB:
Could the issue name have that included so people see what is happening from a glance?

i509VCB (approved):

Gegy (approved):

Juuxel (changes requested):
Both `Vec` and `Vector` are fine, but the mutability should be in `class_1160`'s name. (Regarding `Vector3i`'s "immutability": even though it's internally mutable, it doesn't expose mutating methods publically)

YanisBft:
I don't get why mutability should be included in the name, at least when there is no immutable class it could be confused with. Adding `Mutable` as a prefix just makes the class both harder to find in class tree and via autocompletion

Juuxel:
There are already immutable/unmodifiable classes that follow the `Vec*` (or `Vector*`) pattern, which makes users expect that all `Vector*`s are unmodifiable, even though `Vector3f` is not.

Player3324:
Vector3i expresses no (im)mutability, it is internally mutable and its users can **not** expect it to stay the same, much like List. It isn't worth qualifying auxiliary properties if there's only one implemented variant. When a property like mutability changes within a type hierarchy, the base class needs to be the unqualified one.

Juuxel:
Sure enough. I would still add a small line of javadoc to each vector specifying whether it's publically mutable (`3f` and other client vectors), publically read-only but actually mutable (`3i`) or actually immutable.

Juuxel (approved):

jaskarth (approved):

dblsaiko (approved):

dexman545 (approved):
I agree that using abbreviations in class names are fine

Player3324 (approved):

modmuss50 (approved):

PR #1880, opened 2020-11-15
Scoreboard mappings

Juuxel:
- `PlayerListHud.method_27538` -> `applyGameModeFormatting`: makes the name italic if the player is in spectator mode
- `AbstractTeam.modifyText` and `Team.modifyText` -> `decorateName`: decorates an entity's name with the prefix, suffix and colour of a team
- `Team.field_24195` -> `nameStyle`: used in `Team.getFormattedName` to apply the name insertion and hover event to the name text
- `ScoreboardCriterion.OBJECTIVES` -> `CRITERIA`: it's a map of criteria, not objectives
- `ScoreboardCriterion.createStatCriterion` -> `getOrCreateStatCriterion`: the method checks the `CRITERIA` map for an existing criterion first
- `ScoreboardState`: changed all names to follow `from/toTag` format. The conventions say:
  > Use "read" for non-static methods that load data into the object. Use "write" for methods that save data to an existing object passed as a parameter.

  However, the superclass (`PersistentState`) uses `toTag` like many other classes for this purpose, so I used the same format to be consistent in the same class hierarchy.

### DIFF
--- a/mappings/net/minecraft/scoreboard/ScoreboardPlayerScore.mapping
+++ b/mappings/net/minecraft/scoreboard/ScoreboardPlayerScore.mapping
@@ -12,7 +12,7 @@ CLASS net/minecraft/class_267 net/minecraft/scoreboard/ScoreboardPlayerScore
 		ARG 3 playerName
 	METHOD method_1122 getScoreboard ()Lnet/minecraft/class_269;
 	METHOD method_1124 incrementScore (I)V
-		ARG 1 increment
+		ARG 1 amount
 	METHOD method_1125 setLocked (Z)V
 		ARG 1 locked
 	METHOD method_1126 getScore ()I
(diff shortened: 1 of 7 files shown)


mappings/net/minecraft/scoreboard/ScoreboardPlayerScore.mapping
```diff
@@ -12,6 +12,7 @@ CLASS net/minecraft/class_267 net/minecraft/scoreboard/ScoreboardPlayerScore
 		ARG 3 playerName
 	METHOD method_1122 getScoreboard ()Lnet/minecraft/class_269;
 	METHOD method_1124 incrementScore (I)V
+		ARG 1 increment
```
enbrain:
```suggestion
		ARG 1 amount
```

YanisBft (approved):

enbrain (approved):

haykam821 (approved):

PR #1865, opened 2020-11-11
Map classes and methods relating to powder snow and freezing

haykam821:

PR #1883, opened 2020-11-18
map powder snow & related stuff

pluiedev:
haha `FluidDrainable` and `FluidPlaceable`


pluiedev:
welp i guess i cant read, haykam already did the exact same thing in #1865

PR #1891, opened 2020-11-25
Dripstones

YanisBft:


mappings/net/minecraft/block/LandingBlock.mapping
```diff
@@ -1,3 +1,11 @@
-CLASS net/minecraft/class_5688
+CLASS net/minecraft/class_5688 net/minecraft/block/LandingBlock
```
Juuxel, line 1: This is a bit odd, but I can't think of a better name either...

Juuxel (approved):

haykam821 (approved):

PR #1912, opened 2020-12-09
TreeFeature.isDirtOrGrass -> canPlaceTreeOn

Juuxel:
It prevents the feature from placing a tree if it's false. The logic is pretty much `isSoil(block) || block == FARMLAND`.


YanisBft (approved):

Earthcomputer:
How about `canTreePlaceOn`?

Juuxel:
Hmm, that's actually more accurate (though `canPlaceTreeOn` sounds more natural so I'll go with that).

Earthcomputer (approved):

PR #1915, opened 2020-12-12
net.minecraft.world.level.entity package mappings

ghost:


Juuxel:
Most of this PR is copied straight from Mojang's official mappings, which are forbidden in Yarn (see CONTRIBUTING.md) since their license is incompatible.

PR #1934, opened 2020-12-23
Switch to mappingpoet 0.2.0

liach:
## Goal
 - Integrate mappingpoet 0.2.0
   - Now includes mapping information on javadoc generated. Would support type annotations if mojang add them.
 - Removed mappingpoet from build.gradle dependencies. Now only used when generating fake source/javadoc
 - Uses gradle 6.7 toolchain feature so running `./gradlew javadoc -Dorg.gradle.java.home="C:/Program Files/Java/jdk1.8.0_211"` (using java 8 to run gradle buildscript) would work even though poet now depends on java 11
 - Enables doclint and adds the 3 JDK command line taglets.
   - Fixed places in mappings where it does not comply with the doclint standards.
## Non Goals
 - Does not set up the buildscript for FabricMC/Enigma#321.
   - The commands need to be moved over from direct method calls to javaexec tasks individually.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,33 +22,6 @@ plugins {
 	id 'java' // for jd gen
 }
 
-def lowestLanguageVersion = JavaLanguageVersion.of(15) // Javadoc taglet requires it for css behavior
-def toolchainLanguageVersion = JavaLanguageVersion.of(JavaVersion.current().getMajorVersion())
-
-if (!toolchainLanguageVersion.canCompileOrRun(lowestLanguageVersion)) {
-	logger.warn ":The current Java Version ${toolchainLanguageVersion} is too old."
-	logger.warn ":Gradle may download newer Java online and the build may be slower."
-
-	toolchainLanguageVersion = lowestLanguageVersion
-}
-
-java.toolchain { // defined in JavaPluginExtension
-	languageVersion = toolchainLanguageVersion
-}
-
-// currently gradle does not yet propagate that version above, so
-tasks.withType(JavaExec) {
-	javaLauncher = javaToolchains.launcherFor {
-		languageVersion = toolchainLanguageVersion
-	}
-}
-
-tasks.withType(Javadoc) {
-	javadocTool = javaToolchains.javadocToolFor {
-		languageVersion = toolchainLanguageVersion
-	}
-}
-
 def minecraft_version = "20w51a"
 
 def ENV = System.getenv()
@@ -58,7 +31,6 @@ def build_number = ENV.BUILD_NUMBER ?: "local"
 def yarnVersion = "${minecraft_version}+build.$build_number"
 
 repositories {
-	mavenLocal()
 	mavenCentral()
 	maven {
 		name "Fabric Repository"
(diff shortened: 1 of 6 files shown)


modmuss50:
IDK if it would be better to just enfoce java 11 for yarn's build script. This does seem to increase the build time a lot on java 8 and 11.

liach:
I decide to emit a warning if the user is building with java 8 and use java 11 as the baseline now (LTS). Also made fake source gen use mergedv2 so it has access to obf information to put into javadoc. Uses 0.2.1 of mappingpoet now (thanks i509vcb)

build.gradle
```diff
@@ -23,6 +22,30 @@ plugins {
 	id 'java' // for jd gen
 }
 
+def toolchainLanguageVersion = JavaLanguageVersion.of(11)
+
+if (!JavaVersion.current().isCompatibleWith(JavaVersion.toVersion(toolchainLanguageVersion))) {
+	logger.warn ":The current Java Version ${JavaVersion.current()} is too old."
+	logger.warn ":Gradle may download newer Java online and the build may be slower."
```
modmuss50: I think this is good for now, we can look at enforcing java 11 in a while, possibly along with engima.

modmuss50 (approved):

build.gradle
```diff
@@ -23,6 +22,30 @@ plugins {
...
+	javaLauncher = javaToolchains.launcherFor {
+		languageVersion = toolchainLanguageVersion
+	}
+}
+
+tasks.withType(Javadoc) {
+	javadocTool = javaToolchains.javadocToolFor {
+		languageVersion = toolchainLanguageVersion
+	}
+}
```
modmuss50: This does use java 11 when using 15 for example, might be worth moving this to only apply when running on java versions bellow 11 (same time as the warning).

modmuss50 (changes requested):
Just noticed this is applying the toolchain stuff when using a newer java version than 11.

build.gradle
liach: I will just set the current java language version as the toolchain language version if it can compile or run the minimum version (this case 11). Done in last forced update.

modmuss50 (approved):

Mysterious-Dev (approved):

liach:
Mappingpoet is receiving an update. Will remove wip once the newer version is ready.

liach:
Here I encounter a problem with the current taglet setup:
The current taglet depends on javadoc's builtin stylesheet. However, some stylesheet's detected classes have been renamed in java 15, such as [`colFirst` to `col-first`](https://github.com/FabricMC/MappingPoet/commit/01e9d6a8dd5c10240f0b09cc5807b2fd37bfd621#diff-d99c522501c6e18f357c7a3abb246d38a80045ab150158103cfc23376ba14e44L69).

Currently, I resolve this by making java 15 the required min java toolchain version.

I wonder if this current approach is feasible, or if I should migrate to a standalone stylesheet for the render effects.

liach:
Hmm, a specific version indeed is better as some of the javadoc options used at least may not be forward compatible as they can be "changed without warning" per oracle.

liach:
@modmuss50 
This is ready.

This requires java 15 because of the css style of the mapping table depends on that of javadoc, which is version dependent.

liach:
I have removed enforced migration to higher java. Now, if someone is not running fake source gen or javadoc, they are free to use older java like 8 and won't spend time grabbing jdk 11.

modmuss50:
Nice, I think thats much better!

liach:
Also a pic of updated "fabric style" tables (usable in regular doc as well, just do `<div class="fabric"><table>...</table></div>`)
image

Juuxel (approved):

PR #1939, opened 2020-12-25
Complete RecipesProvider mapping

yyny:
Most of these are pretty trivial once you understand the code.

### DIFF
--- a/mappings/net/minecraft/data/server/RecipesProvider.mapping
+++ b/mappings/net/minecraft/data/server/RecipesProvider.mapping
@@ -129,6 +129,6 @@ CLASS net/minecraft/class_2446 net/minecraft/data/server/RecipesProvider
 	METHOD method_32814 offerSlabRecipe (Ljava/util/function/Consumer;Lnet/minecraft/class_1935;Lnet/minecraft/class_1935;)V
 		ARG 1 outputItem
 		ARG 2 inputItem
-	METHOD method_32815 offerStoneCutterRecipe (Ljava/util/function/Consumer;Lnet/minecraft/class_1935;Lnet/minecraft/class_1935;)V
+	METHOD method_32815 offerStonecutterRecipe (Ljava/util/function/Consumer;Lnet/minecraft/class_1935;Lnet/minecraft/class_1935;)V
 		ARG 1 outputItem
 		ARG 2 inputItem
--- a/mappings/net/minecraft/data/server/RecipesProvider.mapping
+++ b/mappings/net/minecraft/data/server/RecipesProvider.mapping
@@ -4,9 +4,131 @@ CLASS net/minecraft/class_2446 net/minecraft/data/server/RecipesProvider
 	FIELD field_11375 GSON Lcom/google/gson/Gson;
 	METHOD method_10419 generate (Ljava/util/function/Consumer;)V
 	METHOD method_10420 conditionsFromTag (Lnet/minecraft/class_3494;)Lnet/minecraft/class_2066$class_2068;
+		ARG 0 tag
 	METHOD method_10422 requireEnteringFluid (Lnet/minecraft/class_2248;)Lnet/minecraft/class_2037$class_2039;
+		ARG 0 block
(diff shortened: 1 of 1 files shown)


liach (approved):

mappings/net/minecraft/data/server/RecipesProvider.mapping
```diff
@@ -4,9 +4,131 @@ CLASS net/minecraft/class_2446 net/minecraft/data/server/RecipesProvider
...
+	METHOD method_32812 offerChiseledBlockRecipe (Ljava/util/function/Consumer;Lnet/minecraft/class_1935;Lnet/minecraft/class_1935;)V
+		ARG 1 outputItem
+		ARG 2 inputItem
+	METHOD method_32813 createPressurePlateRecipe (Ljava/util/function/Consumer;Lnet/minecraft/class_1935;Lnet/minecraft/class_1935;)V
+		ARG 1 outputItem
+		ARG 2 inputItem
+	METHOD method_32814 offerSlabRecipe (Ljava/util/function/Consumer;Lnet/minecraft/class_1935;Lnet/minecraft/class_1935;)V
+		ARG 1 outputItem
+		ARG 2 inputItem
+	METHOD method_32815 offerStoneCutterRecipe (Ljava/util/function/Consumer;Lnet/minecraft/class_1935;Lnet/minecraft/class_1935;)V
```
Juuxel:
```suggestion
	METHOD method_32815 offerStonecutterRecipe (Ljava/util/function/Consumer;Lnet/minecraft/class_1935;Lnet/minecraft/class_1935;)V
```

mappings/net/minecraft/data/server/RecipesProvider.mapping
```diff
@@ -4,9 +4,131 @@ CLASS net/minecraft/class_2446 net/minecraft/data/server/RecipesProvider
...
+		ARG 0 item
 	METHOD method_10427 saveRecipeAdvancement (Lnet/minecraft/class_2408;Lcom/google/gson/JsonObject;Ljava/nio/file/Path;)V
+		ARG 0 cache
+		ARG 1 json
+		ARG 2 path
 	METHOD method_17585 generateCookingRecipes (Ljava/util/function/Consumer;Ljava/lang/String;Lnet/minecraft/class_3957;I)V
+		ARG 1 cooker
+		ARG 2 serializer
+		ARG 3 cookingTime
+	METHOD method_24475 offerPlanksRecipe2 (Ljava/util/function/Consumer;Lnet/minecraft/class_1935;Lnet/minecraft/class_3494;)V
```
Juuxel, line 25: How does this differ from the other `offerPlanksRecipe`? (for example, is it only used for fungus planks or wooden planks)

Juuxel (changes requested):
Why do some of these use `offer` (as in `offerCandleDyeingRecipe` or `offerStoneCutterRecipe`), while others use `create` (as in `createSlabRecipe` or `createPressurePlateRecipe`)?

mappings/net/minecraft/data/server/RecipesProvider.mapping
yyny: > Why do some of these use `offer` (as in `offerCandleDyeingRecipe` or `offerStoneCutterRecipe`), while others use `create` (as in `createSlabRecipe` or `createPressurePlateRecipe`)?

`create` is the name used to create `net/minecraft/data/server/recipe/***RecipeJsonFactory`s, but there is a separate pull request #1940 for changing the names of those since they are very confusing so a name change might be in place here as well.

Similarly, `offer` is used because it was already mapped in the RecipeJsonFactory's, the difference between these kinds of functions is that the latter actually offers the recipe to the consumer via `createRecipe().offerTo(consumer);`, to reduce code duplication probably.

> How does this differ from the other offerPlanksRecipe? (for example, is it only used for fungus planks or wooden planks)

I think it's a bug or perhaps a typo, it's only used for Dark Oak and Acacia planks, it does the exact same thing but using a different "has_log" criteria (the other `offerPlanksRecipe` function has the "has_logs" criteria).

Juuxel (approved):

YanisBft (approved):

Issue #1938, opened 2020-12-25
Discussion and thoughts on naming conventions for `Builder`s (and `Factory`s)


yyny:
#### Context

I found some inconsistencies regarding `Builder`s and `Factory`s in some mappings I am working on and some more by doing a bit of `grep`ping, as well as having to spend more time than probably necessary figuring out what Builder methods were doing, but I'd like some discussion from the community before I start `rgrep`ping them all away, creating resistance by changing the status quo, and wasting everybody's time.

## Proposal Summary

Update `CONVENTIONS.md`:
 - [ ] Don't name `Builder`s as `Factory`s
 - [ ] Don't put `create` methods in `Builder`s
 - [ ] Don't use `Builder#property` for anything but simple setters.
 - [ ] Use `withProperty` for builders whenever possible, to increase readability.
 - [ ] Naming convention for array-like property setters in `Builder`s.

## `Factory`

There are quite a few classes named `Factory` that are actually `Builder`s.

I propose adding a section to [`CONVENTIONS.md`](https://github.com/FabricMC/yarn/blob/20w51a/CONVENTIONS.md) for this explicitly denouncing naming a class `Factory` when one of its methods returns a `Factory` (Thought: Would a `Factory` returning itself be a `FactoryFactory`?).

## `ThingBuilder#create`

The naming conventions contain a paragraph about naming constructing functions `create`.

This is fine, however, in the case of `Builder`s I feel like the name `build` is more appropriate. A `ThingBuilder#create` might be expected to be "creating a ThingBuilder", whereas a `ThingBuilder#build` pretty unambiguously "builds a Thing"

I propose adding an explicit convention for `Builder`s to use `build`, both to prevent people from adding `create` methods to builders in the future, and to standardize the naming convention for them.

Note that `GsonBuilder`, which is used by Minecraft, (unfortunately) uses `create`.

## `Builder` setters

The naming conventions contain a section recommending naming setter methods in Builders as the name of the property they are setting, without any prefix. This works great for builders of simple struct-like types, but quickly falls apart for more complicated `Thing`s.

As a solution, I propose using `Thing Thing#withProperty(Property property);`, which `CONVENTIONS.md` also recommends, to allow classes the freedom of adding more complicated setters without creating inconsistency (Where either some methods have no prefix and some do, or even worse, where all functions are unprefixed regardless of whether they are actually setters or not).

`withProperty` is a great convention since it makes it clear that the object is being returned again, unlike `setProperty`.
[...]

PR #1940, opened 2020-12-25
Apply builder pattern in a more reasonable manner

yyny:
Draft request as per #1938

The work in this commit was instrumental for getting yyny@a4d71af2812c95624eecec377a6c22b042578511 right

### DIFF
--- a/mappings/net/minecraft/advancement/Advancement.mapping
+++ b/mappings/net/minecraft/advancement/Advancement.mapping
@@ -28,7 +28,7 @@ CLASS net/minecraft/class_161 net/minecraft/advancement/Advancement
 	METHOD method_690 addChild (Lnet/minecraft/class_161;)V
 		ARG 1 child
 	METHOD method_691 getRewards ()Lnet/minecraft/class_170;
-	CLASS class_162 Task
+	CLASS class_162 Builder
 		FIELD field_1147 display Lnet/minecraft/class_185;
 		FIELD field_1148 criteria Ljava/util/Map;
 		FIELD field_1149 parentObj Lnet/minecraft/class_161;
@@ -42,7 +42,7 @@ CLASS net/minecraft/class_161 net/minecraft/advancement/Advancement
 			ARG 3 rewards
 			ARG 4 criteria
 			ARG 5 requirements
-		METHOD method_20416 display (Lnet/minecraft/class_1799;Lnet/minecraft/class_2561;Lnet/minecraft/class_2561;Lnet/minecraft/class_2960;Lnet/minecraft/class_189;ZZZ)Lnet/minecraft/class_161$class_162;
+		METHOD method_20416 withDisplay (Lnet/minecraft/class_1799;Lnet/minecraft/class_2561;Lnet/minecraft/class_2561;Lnet/minecraft/class_2960;Lnet/minecraft/class_189;ZZZ)Lnet/minecraft/class_161$class_162;
 			ARG 1 icon
 			ARG 2 title
 			ARG 3 description
@@ -54,14 +54,14 @@ CLASS net/minecraft/class_161 net/minecraft/advancement/Advancement
 		METHOD method_692 fromJson (Lcom/google/gson/JsonObject;Lnet/minecraft/class_5257;)Lnet/minecraft/class_161$class_162;
 			ARG 0 obj
 			ARG 1 predicateDeserializer
-		METHOD method_693 display (Lnet/minecraft/class_185;)Lnet/minecraft/class_161$class_162;
+		METHOD method_693 withDisplay (Lnet/minecraft/class_185;)Lnet/minecraft/class_161$class_162;
 			ARG 1 display
 		METHOD method_694 build (Ljava/util/function/Consumer;Ljava/lang/String;)Lnet/minecraft/class_161;
 		METHOD method_695 build (Lnet/minecraft/class_2960;)Lnet/minecraft/class_161;
 			ARG 1 id
 		METHOD method_696 fromPacket (Lnet/minecraft/class_2540;)Lnet/minecraft/class_161$class_162;
 			ARG 0 buf
-		METHOD method_697 display (Lnet/minecraft/class_1935;Lnet/minecraft/class_2561;Lnet/minecraft/class_2561;Lnet/minecraft/class_2960;Lnet/minecraft/class_189;ZZZ)Lnet/minecraft/class_161$class_162;
+		METHOD method_697 withDisplay (Lnet/minecraft/class_1935;Lnet/minecraft/class_2561;Lnet/minecraft/class_2561;Lnet/minecraft/class_2960;Lnet/minecraft/class_189;ZZZ)Lnet/minecraft/class_161$class_162;
 			ARG 1 icon
 			ARG 2 title
 			ARG 3 description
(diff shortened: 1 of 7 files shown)


yyny:
In it's current state, this pull request is only meant to show the changes I made to be able to understand the code in yyny@a4d71af2812c95624eecec377a6c22b042578511, and show some potential improvements for the future.

Some thoughts remaining:

 * I feel like `JsonRecipe` is more correct than `RecipeJson`
 * `createRecipe ` could probably be `beginRecipe` or even stay as `create` if we get `RecipeJsonFactory` -> `JsonRecipeBuilder`
 * `withAddedCriterion` is probably overly verbose.

Anyway, this commit has lots of issues but so do the old mappings, and I'd rather get a discussion going in #1938 first before fixing anything.

mappings/net/minecraft/data/server/recipe/SmithingRecipeJsonBuilder.mapping
```diff
@@ -9,11 +9,11 @@ CLASS net/minecraft/class_5377 net/minecraft/data/server/recipe/SmithingRecipeJs
 		ARG 2 base
 		ARG 3 addition
 		ARG 4 result
-	METHOD method_29729 create (Lnet/minecraft/class_1856;Lnet/minecraft/class_1856;Lnet/minecraft/class_1792;)Lnet/minecraft/class_5377;
+	METHOD method_29729 createRecipe (Lnet/minecraft/class_1856;Lnet/minecraft/class_1856;Lnet/minecraft/class_1792;)Lnet/minecraft/class_5377;
```
liach, line 12: Let's loot at this.

## The class
:+1:
First, this class is actually a builder, even though its fields are all final. So the rename looks good.

## The method
:confused:
This method returns a new instance of this class. Your current name `createRecipe` is really weird, as it sounds like creating a recipe than creating a builder of/for/from a recipe. Consider this code:
```java
SmithingRecipeJsonBuilder builder = SmithingRecipeJsonBuilder.createRecipe(/* */);
```
Am I creating a builder or a recipe?

I suggest a name like `forRecipe`, `ofRecipe`, `of`, or `create` instead. They are all less confusing.

mappings/net/minecraft/data/server/recipe/SmithingRecipeJsonBuilder.mapping
```diff
@@ -9,11 +9,11 @@ CLASS net/minecraft/class_5377 net/minecraft/data/server/recipe/SmithingRecipeJs
...
 		ARG 0 base
 		ARG 1 addition
 		ARG 2 result
-	METHOD method_29730 criterion (Ljava/lang/String;Lnet/minecraft/class_184;)Lnet/minecraft/class_5377;
+	METHOD method_29730 withAddedCriterion (Ljava/lang/String;Lnet/minecraft/class_184;)Lnet/minecraft/class_5377;
```
liach, line 16: This modifies itself. So `criterion` is good enough.

mappings/net/minecraft/data/server/recipe/SingleItemRecipeJsonBuilder.mapping
```diff
@@ -10,14 +10,14 @@ CLASS net/minecraft/class_3981 net/minecraft/data/server/recipe/SingleItemRecipe
...
+	METHOD method_17968 createRecipe (Lnet/minecraft/class_1856;Lnet/minecraft/class_1935;)Lnet/minecraft/class_3981;
 		ARG 0 input
 		ARG 1 output
-	METHOD method_17969 create (Lnet/minecraft/class_1856;Lnet/minecraft/class_1935;I)Lnet/minecraft/class_3981;
+	METHOD method_17969 createRecipe (Lnet/minecraft/class_1856;Lnet/minecraft/class_1935;I)Lnet/minecraft/class_3981;
 		ARG 0 input
 		ARG 1 output
 		ARG 2 outputCount
-	METHOD method_17970 create (Ljava/lang/String;Lnet/minecraft/class_184;)Lnet/minecraft/class_3981;
+	METHOD method_17970 withAddedCriterion (Ljava/lang/String;Lnet/minecraft/class_184;)Lnet/minecraft/class_3981;
```
liach, line 20: Yes this is a bug, should have been `criterion`

mappings/net/minecraft/advancement/Advancement.mapping
```diff
@@ -28,7 +28,7 @@ CLASS net/minecraft/class_161 net/minecraft/advancement/Advancement
 	METHOD method_690 addChild (Lnet/minecraft/class_161;)V
 		ARG 1 child
 	METHOD method_691 getRewards ()Lnet/minecraft/class_170;
-	CLASS class_162 Task
+	CLASS class_162 Builder
```
liach, line 31: Yeah good rename, I named it a bad name previously.

This is not only a builder but a mutable representation of advancement before their parent relationships are set up, though calling it builder is good enough.

liach:
One of yarn's main goal is to stay concise as long as it can convey meanings correctly and effectively. So for like `SmithingRecipeJsonBuilder`'s what you've named `createRecipe`, I think we can just call it `recipe` as it is not confusing in context and looks straightforward at where it is used in code.

liach:
Similarly, the `create` in the shapeless recipe json builder might be named like `output`, so you can have
```java
ShapelessRecipeJsonFactory.output(arg, 4).input(arg2).group("planks").criterion("has_log", RecipesProvider.conditionsFromTag(arg2)).offerTo(consumer);
```

mappings/net/minecraft/data/server/recipe/SmithingRecipeJsonBuilder.mapping
yyny: `createRecipe` is just something I quickly came up with that was less confusing than `SmithingRecipeJsonFactory::create`, it probably isn't a lot more descriptive. Thinking about it more, I think any kind of `create` is pretty confusing in `Builder`s to begin with, it might make sense to have a `static Thing ThingBuilder#create` in a `Builder` when it doubles as a `Factory`, but when it comes to `ThingBuilder ThingBuilder::create`, at the very least `ThingBuilder::begin`, but there is almost always a more descriptive name to come up with. I like the `output` suggestion further below. Perhaps `ThingBuilder::forOutput`, to distinguish the static methods from the rest.

yyny: I don't like `forRecipe`, in that case I prefer `recipeFor` since it's how you would write it in english.

mappings/net/minecraft/data/server/recipe/SmithingRecipeJsonBuilder.mapping
yyny: I will continue this particular discussion in #1938 to keep everything in one place.

mappings/net/minecraft/data/server/recipe/SmithingRecipeJsonBuilder.mapping
yyny: See also #1939 for how choosing a name here might affect mappings elsewhere.

YanisBft:
@yyny Would be nice to update this :)

modmuss50:
This PR is on the right track 👍 Is it still draft and do you plan to contiune with it? Or could someone else pick it up?

modmuss50:
I think this PR has been stale long enough, if someone else or the OP wishes to contiune with this pelase open a new PR. 👍

PR #1952, opened 2021-01-05
Constant Unpicking

modmuss50:
# Constant Unpicking

## Before

```java
private static void setLit(WorldAccess world, BlockState state, BlockPos pos, boolean lit) {
   world.setBlockState(pos, (BlockState)state.with(LIT, lit), 11);
}
```

## After

```java
private static void setLit(WorldAccess world, BlockState state, BlockPos pos, boolean lit) {
   world.setBlockState(pos, (BlockState)state.with(LIT, lit), SetBlockStateFlags.PROPAGATE_CHANGE | SetBlockStateFlags.NOTIFY_LISTENERS | SetBlockStateFlags.REDRAW_ON_MAIN_THREAD);
}
```

This works by using the [WeaveMC/unpick](https://github.com/WeaveMC/unpick) library created by @Daomephsta, unpick files are created to define how unpick will unline the constatnts. You can read about the [Unpick format here](https://github.com/WeaveMC/unpick/wiki/Unpick-Format-V2).

Compile time only classes are also provided that will be exposed to your mods. Only the generated sources in loom will contain the uninlined constants to ensure that all existing mixins that depend on the constants will contiune to work.

This PR currenly only contains a basic but common example (based off the work already done in [weft](https://github.com/WeaveMC/weft)) of the setBlockState method. There are a range of other constants that can be unpicked in the future with this.

I have added a `extras/definitions.unpick` to the v2 yarn mapping jars, I dont believe this will break any existing tools, and will make it easy for loom to pick them up and apply them if required. A -constants jar file along with sources is also published to maven.

Please let me know of any suggestions, this is still a WIP. I wont be merging this untill its had some time to be reviewed as well as the loom PR is complete. This currenly targets the 1.17 snapshots, it can easily be backported to 1.16 once complete.


src/main/java/net/fabricmc/yarn/constants/SetBlockStateFlags.java
```diff
@@ -0,0 +1,46 @@
...
+	 */
+	public static final int PROPAGATE_CHANGE = 1;
+
+	/**
+	 * Notifies listeners and clients who need to react when the block changes.
+	 */
+	public static final int NOTIFY_LISTENERS = 2;
+
+	/**
+	 * Used in conjunction with NOTIFY_LISTENERS to suppress the render pass on clients.
```
Juuxel:
```suggestion
	 * Used in conjunction with {@link NOTIFY_LISTENERS} to suppress the render pass on clients.
```

build.gradle
```diff
@@ -582,6 +582,154 @@ task mapNamedJar(dependsOn: ["mergeV2", mapIntermediaryJar]) {
...
+
+	@Override
+	ClassReader resolveClass(String internalName) throws IOException {
+		def resource = classLoader.getResourceAsStream(internalName + ".class")
+
+		if (resource != null) {
+			return new ClassReader(resource)
+		}
+
+		throw new RuntimeException("Failed to find ${internalName}.class")
```
Juuxel:
```suggestion
		throw new IOException("Failed to find ${internalName}.class")
```
Might be a bit more appropriate here, and it's a part of the method signature as well.

Juuxel (changes requested):
The suggestions are some minor details.

Also, should Enigma support the Unpick format? The files seem to be using Yarn names, so changing the names would invalidate the Unpick files. (`setBlockState` might be pretty vulnerable to this, considering Mojang moves world logic around in every update)

modmuss50:
Ah yes, I forgot to mention enigma should at the very least show it. I will have to check that the build fails with unknown names

build.gradle
```diff
@@ -582,6 +582,154 @@ task mapNamedJar(dependsOn: ["mergeV2", mapIntermediaryJar]) {
...
+			def definition = it.text
+			if (!definition.startsWith("v2")) {
+				throw new UnsupportedOperationException(it.name + " has an invalid unpick header")
+			}
+
+			// Remove the first, and append to the combined definition
+			combinedDefinitions += definition.split("\n").drop(1).join("\n")
+		}
+
+		output.text = combinedDefinitions
```
Daomephsta: If desired, syntax checking can be performed here using unpick-format-utils reader and writer classes. 
`UnpickV2Writer` doesn't visit the version, so stripping it isn't necessary. 
```suggestion
		def combinedDefinitions = new UnpickV2Writer()

		file("unpick-definitions").eachFile { file ->
			file.withInputStream { is -> 
				byte[] version = new byte [2];
				is.read(version);
				if (version != ['v', '2'])
					throw new UnsupportedOperationException(file.name + " has an invalid unpick header")
				// Reader does not need closing, because is will be closed at the closure end
				new UnpickV2Reader(is).accept(combinedDefinitions) 
			}
		}

		output.text = combinedDefinitions.output
```

build.gradle
```diff
@@ -675,11 +829,11 @@ task genFakeSource(dependsOn: ["buildYarnTiny", "mapNamedJar"]) {
 	}
 }
 
-task decompileCFR(type: JavaExec, dependsOn: "mapNamedJar") {
+task decompileCFR(type: JavaExec, dependsOn: [unpickNamedJar]) {
 	classpath = configurations.decompileClasspath
 	main = "org.benf.cfr.reader.Main"
 
-	args namedJar.getAbsolutePath(), "--outputdir", file("namedSrc").absolutePath
+	args unpickedJar.getAbsolutePath(), "--outputdir", file("namedSrc").absolutePath
```
Daomephsta: Won't this result in the JAR used in a development environment containing unpicked bytecode?

modmuss50: No this is just yarn and the basic decompile task. The loom changes are separate and still to come.

build.gradle
modmuss50: That sounds like a much better idea than what I have here. I’ll take a look 👍

build.gradle
Daomephsta: What's it decompiling & unpicking, if not the mapped MC JAR?

modmuss50: It is, but this is in yarn and not loom/dev environments

Daomephsta: Oh right, for enigma

build.gradle
modmuss50: Made the changes, its much better thanks.

src/main/java/net/fabricmc/yarn/constants/SetBlockStateFlags.java
```diff
@@ -0,0 +1,46 @@
+package net.fabricmc.yarn.constants;
```
liach: To reduce confusion I'd suggest moving this to another source set. Putting this in main is a bit confusing.

liach (commented):
We can consider moving unpick handling code away from buildscript into a separate source set and dependencies under a separate configuration (like what's going on with javadoc dependencies).

When we need the unpick code, we can execute a java exec task on the output of that source set and specify a main class, somewhat like what's done in #1934.

The advantage is that when someone runs a gradle task that does not touch unpick-related stuff, unpick won't be grabbed by gradle. Currently it is always grabbed when running any gradle task.

Earthcomputer:
Not sure if it's in the scope of this PR, but wouldn't it make sense to have a shortcut constant in SetBlockStateFlags for 3? I know that Mojang has a shortcut for 3 from snippets of original source code I've seen.

modmuss50:
Yes, that sounds reasonable, but is prob out of scope of this PR. This PR is more for the toolchain side of things, other PRs can come later to expand on that kind of thing.

build.gradle
```diff
@@ -578,10 +578,209 @@ task mapNamedJar(dependsOn: ["mergeV2", mapIntermediaryJar]) {
...
+import java.util.jar.JarOutputStream
+import org.objectweb.asm.tree.ClassNode
+import org.objectweb.asm.ClassWriter
+import net.fabricmc.mapping.tree.TinyMappingFactory
+
+class JarClassResolver implements IClassResolver, Closeable {
+	private final URLClassLoader classLoader;
+
+	JarClassResolver(Set<File> jars) {
+		this.classLoader = new URLClassLoader(
```
liach: Can't we use a java util's jar file instead? jar file afaik supports multi-release jars.

liach (commented):
Now that I think if we add these constants, the new source code files for storing these constants
can go with vanilla package infos so they can both be taken account by javadoc generation (if we want javadoc gen to take care of these constant holding classes).

src/main/java/net/fabricmc/yarn/constants/SetBlockStateFlags.java
modmuss50: Done.

modmuss50:
This should be ready to review, I have added a new `yarnUnpicked` task to run enigma with the unpicked names. Its not default right now as its a bit slower due to it needing to build and run a lot more. 

I think we should get this in and have a play with as we can now easily make changes as the unpick version is defined in via a json file now.

src/constants/java/net/fabricmc/yarn/constants/SetBlockStateFlags.java
```diff
@@ -0,0 +1,46 @@
+package net.fabricmc.yarn.constants;
```
liach, line 1: Should have used singular `constant` per our naming style

build.gradle
```diff
@@ -5,13 +5,17 @@ buildscript {
...
 		mavenCentral()
+		mavenLocal()
 	}
 	dependencies {
 		classpath "cuchaz:enigma-cli:${project.enigma_version}"
 		classpath "net.fabricmc:stitch:${project.stitch_version}"
 		classpath "commons-io:commons-io:2.6"
 		classpath 'de.undercouch:gradle-download-task:4.0.4'
 		classpath 'net.fabricmc:tiny-remapper:0.3.1.72'
+		classpath 'net.fabricmc:mappingpoet:0.1.0+build.2'
```
liach, line 16: Turns out... Will fix this in my next pr that fixes javadoc build

Drakesdrs:
thanks

PR #1959, opened 2021-01-15
Readd documentation for the ServerPlayerEntity.closeHandledScreen method

haykam821:
This should also be merged into the 1.17 snapshot branches because it was removed [here](https://github.com/FabricMC/yarn/commit/c91f48214181289cfe3485646ef0e254934684db#diff-8ce52bd08b24691684887c72b113c8da5e7901e5cd679ededc78c9914cea8926L144-L145) as well.


jaskarth (approved):

Gegy (approved):

YanisBft (commented):
As the method is now declared in PlayerEntity, the docs should go there

haykam821:
@Juuxel added this documentation to `ServerPlayerEntity` because the documented behavior is only present on the server. It was always a method from `PlayerEntity`.

YanisBft (approved):
My bad then

liach:
Hmm, does yarn support adding javadoc like this?

haykam821:
@liach It's not dropped as an invalid mapping and is recognized by Enigma, so I'd say it is.

liach (approved):

Juuxel (approved):

PR #1965, opened 2021-01-20
Map block families

haykam821:


i509VCB:
I have no damn clue why `Family` was the name chosen here

mappings/net/minecraft/data/server/RecipesProvider.mapping
```diff
@@ -21,84 +23,143 @@ CLASS net/minecraft/class_2446 net/minecraft/data/server/RecipesProvider
...
 		ARG 2 inputItem
 	METHOD method_32814 offerSlabRecipe (Ljava/util/function/Consumer;Lnet/minecraft/class_1935;Lnet/minecraft/class_1935;)V
+		ARG 0 exporter
 		ARG 1 outputItem
 		ARG 2 inputItem
+	METHOD method_33531 getWallRecipe (Lnet/minecraft/class_1935;Lnet/minecraft/class_1856;)Lnet/minecraft/class_5797;
+		ARG 0 outputItem
+		ARG 1 inputItem
+	METHOD method_33533 getVariantRecipeInput (Lnet/minecraft/class_5794;Lnet/minecraft/class_5794$class_5796;)Lnet/minecraft/class_2248;
+		COMMENT Gets the block used to craft a certain {@linkplain variant BlockFamily.Variant} of a base block.
```
liach: Shouldn't it be
```suggestion
		COMMENT Gets the block used to craft a certain {@linkplain BlockFamily.Variant variant} of a base block.
```

haykam821:
This is the method suggesting 'family':

```java
private static BlockFamily.Builder register(Block baseBlock) {
	BlockFamily.Builder builder = new BlockFamily.Builder(arg);

	BlockFamily oldFamily = BASE_BLOCKS_TO_FAMILIES.put(baseBlock, builder.build());
	if (oldFamily != null) {
		throw new IllegalStateException("Duplicate family definition for " + Registry.BLOCK.getId(baseBlock));
	}

	return builder;
}
```

mappings/net/minecraft/data/server/RecipesProvider.mapping
haykam821: Probably, I wasn't sure because Enigma simply inserted `{@linkplain variant}` when I was selecting `variant`.

liach:
Family is not bad. Worse choices would include class, type, kind, etc.

mappings/net/minecraft/RecipeJsonFactory.mapping
```diff
@@ -0,0 +1,8 @@
+CLASS net/minecraft/class_5797 net/minecraft/RecipeJsonFactory
```
Juuxel: This feels like it should be in `net.minecraft.data` or a subpackage.

haykam821: Oops, it should be in the same package as the two classes that implement it, `net.minecraft.data.server.recipe`. Fixed in 2d00a36969705a1edecc251736ffe4e0b0f40ff3.

mappings/net/minecraft/data/server/RecipesProvider.mapping
haykam821: Fixed in c9c561c58d9da83f0cbc0e4c47a973fd570e3d7c

YanisBft (approved):

liach (approved):

jaskarth (approved):

PR #1905, opened 2020-12-03
Map the world entity managers & surrounding classes

hedgehog1029:
Continuation & update of #1851.

- `ServerEntityManager`: server's entity management, tracks entities in each chunk
- `ClientEntityManager`: equivalent for client
- `EntityLike`: what it says on the tin
- `EntityTracker`: given to each entity as a way to call back in to the entity manager
- `EntityLoadingStage`: stage of loading entity from disk
- `TypeFilter`: downcasting utility
- `ChunkDataAccess`: interface for specific access to chunk data, future-based
- `ChunkDataList`: very generic list wrapper which includes a position (ChunkPos)
- `ChunkStatusChangeListener`: callback for when the chunk status changes, i guess
- `EntityChunkStatus`: encodes whether an entity should be ticked and/or be tracked by the manager
- `EntityIdMapSet`, `EntitySet`, etc.: it stores da entities by id efficiently
- `EntityTrackingGroup`: group of tracked entities treated as a blob in terms of `EntityChunkStatus` and such
- `SectionedEntityCache`: stores tracking groups keyed by subchunk position and does a Whole bunch of stuff with that

lotsa very generic classes here tbh

PR #1967, opened 2021-01-21
Entity stuff, see #1905

liach:
Supersedes #1905, mapping entity stuff


mappings/net/minecraft/world/entity/EntityTrackingGroup.mapping
```diff
@@ -0,0 +1,24 @@
+CLASS net/minecraft/class_5572 net/minecraft/world/entity/EntityTrackingGroup
```
Gegy: To carry over from: https://github.com/FabricMC/yarn/pull/1905#discussion_r539274537
"Group" is a very vague name to describe how this class is used. It is specifically used to represent a section within the entity manager. Especially given they reference chunk section state, it makes sense to name according to that.

Proposed names following that discussion were `EntityChunkSection` and `EntityTrackingSection`.

liach: group isn't vague. It is a tracking group where a few entities are tracked together because they are in the same section. This is in fact the original meaning of "group"

Gegy: Yea, it is an accurate description- but it doesn't give any indication as to why the entities are grouped. There is not really any reason to be less specific.

liach: Imo I would keep `Entity` and `Tracking` in the name. After that `Group` or `Section` doesn't really matter; I am currently working on something else. Will come back to this a bit later.

liach: Has been renamed @Gegy

mappings/net/minecraft/world/entity/EntityIdMapSet.mapping
```diff
@@ -0,0 +1,6 @@
+CLASS net/minecraft/class_5578 net/minecraft/world/entity/EntityIdMapSet
```
Gegy: `MapSet` is a very confusing name. Proposed `Sectioned` prefix as `SectionedEntityIdMap` in the last PR.

liach: Honestly all those idmap, idset, idmapset are poor names. I took them from the previous contributors without looking for alternatives. Btw if you don't mind I hope you can peek at https://github.com/FabricMC/Enigma/pull/338 real quick

liach: So now my proposal is like this:
`class_5578`: `EntityLookup` (old `EntityIdMap`)
`class_5570`: `EntityIndex` (old `EntitySet`)
`class_5573`: `EntityVoxels`, `EntitySections`, or something (old `SectionedEntityCache`)

liach: Decided to keep `SectionedEntityCache` but changed the other two. `Cache` might mean like storage in general, though in programming it often means expiring temporary storage

mappings/net/minecraft/world/entity/BasicEntityLookup.mapping
```diff
@@ -0,0 +1,10 @@
+CLASS net/minecraft/class_5578 net/minecraft/world/entity/BasicEntityLookup
```
Gegy:
```suggestion
CLASS net/minecraft/class_5578 net/minecraft/world/entity/SectionedEntityLookup
```

liach: Not sectioned. @Gegy This look up composes of two independent objects: an id index and a section cache. Both are independently maintained by the caller of the constructor than by this object. Sectioned is only a minor implementation detail of this.

Gegy (approved):

mappings/net/minecraft/client/world/ClientEntityManager.mapping
```diff
@@ -0,0 +1,37 @@
...
+	METHOD method_31871 (Lnet/minecraft/class_5572;)V
+		ARG 1 sections
+	METHOD method_31874 getEntityCount ()I
+	METHOD method_31875 stopTicking (Lnet/minecraft/class_1923;)V
+		ARG 1 pos
+	METHOD method_31877 (Lnet/minecraft/class_5572;)V
+		ARG 1 sections
+	METHOD method_31879 getDebugString ()Ljava/lang/String;
+	CLASS class_5583 Listener
+		FIELD field_27285 this$0 Lnet/minecraft/class_5582;
```
YanisBft:
```suggestion
```

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -208,7 +209,17 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 		ARG 1 x
 		ARG 3 y
 		ARG 5 z
-	METHOD method_24204 streamPassengersRecursively ()Ljava/util/stream/Stream;
+	METHOD method_24204 streamSelfAndPassengers ()Ljava/util/stream/Stream;
+		COMMENT Returns a stream consisting of this entity and its passengers recusively.
```
YanisBft:
```suggestion
		COMMENT Returns a stream consisting of this entity and its passengers recursively.
```

mappings/net/minecraft/server/world/ServerEntityManager.mapping
```diff
@@ -1,33 +1,132 @@
...
+	METHOD method_31855 getLoadedChunks ()Lit/unimi/dsi/fastutil/longs/LongSet;
+	METHOD method_31857 (Lnet/minecraft/class_5568;)V
 		ARG 1 entity
 	METHOD method_31863 (Lnet/minecraft/class_5568;)V
 		ARG 1 entity
 	METHOD method_31864 (Lnet/minecraft/class_5568;)V
 		ARG 1 entity
-	CLASS class_5580
+	CLASS class_5580 Listener
+		FIELD field_27271 this$0 Lnet/minecraft/class_5579;
```
YanisBft:
```suggestion
```
What is this?

mappings/net/minecraft/util/collection/TypeFilterableList.mapping
```diff
@@ -1,4 +1,12 @@
 CLASS net/minecraft/class_3509 net/minecraft/util/collection/TypeFilterableList
+	COMMENT A collection allowing getting all elements of a specific type. Backed
+	COMMENT by {@link java.util.ArrayList}s.
+	COMMENT
+	COMMENT <p>This implementation is not effcient for frequent modifications. You
```
YanisBft:
```suggestion
	COMMENT <p>This implementation is not efficient for frequent modifications. You
```

mappings/net/minecraft/util/collection/TypeFilterableList.mapping
```diff
@@ -7,8 +15,22 @@ CLASS net/minecraft/class_3509 net/minecraft/util/collection/TypeFilterableList
 	METHOD add (Ljava/lang/Object;)Z
 		ARG 1 e
 	METHOD contains (Ljava/lang/Object;)Z
+		COMMENT {@inheritDoc}
+		COMMENT
+		COMMENT @throws IllegalArgumentException if {@code o} is not an instance of
+		COMMENT {@link #elementType}
 		ARG 1 o
 	METHOD method_15216 getAllOfType (Ljava/lang/Class;)Ljava/util/Collection;
+		COMMENT Returns all elements in this collection's that are instances of {@code type}.
```
YanisBft:
```suggestion
		COMMENT Returns all elements in this collection that are instances of {@code type}.
```

mappings/net/minecraft/server/world/ServerEntityManager.mapping
liach: Think cfr isn't able to identify these as this$0 because the two listeners have generic fields that blows up cfr's guessing logic

YanisBft: Ah okay they don't even appear with Procyon. Wouldn't name them `entityManager` be better?

liach: I think you can actually see synthetic parameters in procyon as well (not outer class instance, but outer local var instance in some blockpos methods).
Also for naming `entityManager`, guess just `manager` (anyways, the field's synthetic flag is stored in its access modifier, so name can be freely chosen)
```suggestion
		FIELD field_27271 manager Lnet/minecraft/class_5579;
```

mappings/net/minecraft/client/world/ClientEntityManager.mapping
liach:
```suggestion
		FIELD field_27285 manager Lnet/minecraft/class_5582;
```

YanisBft (approved):

liach:
hmm, someone mind review this one, like point out the problems?

mappings/net/minecraft/world/entity/BasicEntityLookup.mapping
i509VCB: Maybe `SimpleEntityLookup`?

Or would this be a sort of `EntityLookupImpl`

liach: This is just a lookup composed of one id lookup and one collision lookup.

jaskarth: +1 for `SimpleEntityLookup`

liach: Done

jaskarth (approved):

PR #1977, opened 2021-01-22
Added some recipe things

mounderfod:
~~For some reason `Ingredient.test` was in some kind of limbo stage where it existed, but had no intermediary, but simultaneously had one in Enigma, but was not unique. This seemed like a massive problem as any file that used it called it `method_8093` as opposed to `test`. To fix it, I have renamed it altogether as this way it will be working again.~~ **This was me being a moron, ignore.**

I've also patched up some other missing method names that are related to recipes, if something is not right/not up to convention then do let me know.

### DIFF
--- a/mappings/net/minecraft/recipe/Ingredient.mapping
+++ b/mappings/net/minecraft/recipe/Ingredient.mapping
@@ -35,6 +35,7 @@ CLASS net/minecraft/class_1856 net/minecraft/recipe/Ingredient
 	METHOD method_8107 entryFromJson (Lcom/google/gson/JsonObject;)Lnet/minecraft/class_1856$class_1859;
 		ARG 0 json
 	METHOD test testStack (Ljava/lang/Object;)Z
+		ARG 1 itemStack
 	CLASS class_1857 StackEntry
 		FIELD field_9021 stack Lnet/minecraft/class_1799;
 		METHOD <init> (Lnet/minecraft/class_1799;)V
--- a/mappings/net/minecraft/recipe/Ingredient.mapping
+++ b/mappings/net/minecraft/recipe/Ingredient.mapping
@@ -18,7 +18,7 @@ CLASS net/minecraft/class_1856 net/minecraft/recipe/Ingredient
 		ARG 0 items
 	METHOD method_8092 ofEntries (Ljava/util/stream/Stream;)Lnet/minecraft/class_1856;
 		ARG 0 entries
-	METHOD method_8093 test (Ljava/lang/Object;)Z
+	METHOD method_8093 test (Lnet/minecraft/class_1799;)Z
 		ARG 1 stack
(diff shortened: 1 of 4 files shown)


mappings/net/minecraft/recipe/Ingredient.mapping
```diff
@@ -32,8 +34,8 @@ CLASS net/minecraft/class_1856 net/minecraft/recipe/Ingredient
 		ARG 0 tag
 	METHOD method_8107 entryFromJson (Lcom/google/gson/JsonObject;)Lnet/minecraft/class_1856$class_1859;
 		ARG 0 json
-	METHOD test (Ljava/lang/Object;)Z
-		ARG 1 stack
+	METHOD test testStack (Ljava/lang/Object;)Z
```
liach: No, you broke the method that properly overrides the method `boolean test(Object)` in `Predicate`.

mappings/net/minecraft/recipe/Ingredient.mapping
```diff
@@ -18,6 +18,8 @@ CLASS net/minecraft/class_1856 net/minecraft/recipe/Ingredient
 		ARG 0 items
 	METHOD method_8092 ofEntries (Ljava/util/stream/Stream;)Lnet/minecraft/class_1856;
 		ARG 0 entries
+	METHOD method_8093 test (Ljava/lang/Object;)Z
```
liach: This descriptor seems wrong. From my extracted class_1856 the descriptor of method_8093 according to javap is `public boolean method_8093(net.minecraft.class_1799);`, so this should be
```suggestion
	METHOD method_8093 test (Lnet/minecraft/class_1799;)Z
```
Dunno how your enigma get this wrong.

mappings/net/minecraft/recipe/Ingredient.mapping
```diff
@@ -32,8 +34,8 @@ CLASS net/minecraft/class_1856 net/minecraft/recipe/Ingredient
 		ARG 0 tag
 	METHOD method_8107 entryFromJson (Lcom/google/gson/JsonObject;)Lnet/minecraft/class_1856$class_1859;
 		ARG 0 json
-	METHOD test (Ljava/lang/Object;)Z
-		ARG 1 stack
+	METHOD test testStack (Ljava/lang/Object;)Z
+		ARG 1 itemStack
```
liach:
```suggestion
	METHOD test (Ljava/lang/Object;)Z
		ARG 1 stack
```
This method is actually synthetic as it is final and always calls 8093. This is a java 5 bridge to allow overriding classes to have concrete returns or return more specific types.

liach (changes requested):

liach (approved):

liach:
Hmm, if enigma renders method_8093 as having `(Ljava/lang/Object;)Z` descriptor then we need to check enigma for bugs

YanisBft (approved):

PR #180, opened 2018-10-31
More GUI stuff

maxpowa:
Highlights:

Fixed incorrect Gui.onMousePressed(game, mouseY, mouseButton) -> Gui.onScaleChange(game, windowWidth, windowHeight)
Added mapping for WidgetButton.onPressed / onReleased
Mapped most of GuiBackupLevel / GuiBackupPrompt

PR #1992, opened 2021-01-31
Some changes and enhancements

teddyxlandlee:
SplashScreen -> MojangLogoLoadingScreenOverlay
class_5493 -> util/MobUtils

### DIFF
--- a/mappings/net/minecraft/client/gui/screen/MojangLogoLoadingScreenOverlay.mapping
+++ b/mappings/net/minecraft/client/gui/screen/MojangLogoLoadingScreenOverlay.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_425 net/minecraft/client/gui/screen/SplashScreen
+CLASS net/minecraft/class_425 net/minecraft/client/gui/screen/MojangLogoLoadingScreenOverlay
 	FIELD field_17767 reloadMonitor Lnet/minecraft/class_4011;
 	FIELD field_17770 progress F
 	FIELD field_17771 applyCompleteTime J
--- a/mappings/net/minecraft/entity/EntityManager.mapping
+++ b/mappings/net/minecraft/entity/EntityManager.mapping
@@ -1,4 +1,5 @@
-CLASS net/minecraft/class_5569
+CLASS net/minecraft/class_5569 net/minecraft/entity/EntityManager
+	FIELD field_27243 EMPTY Lnet/minecraft/class_5569;
 	METHOD method_31749 updateEntityPosition ()V
 	METHOD method_31750 remove (Lnet/minecraft/class_1297$class_5529;)V
 		ARG 1 reason
--- a/mappings/net/minecraft/util/MobUtils.mapping
+++ b/mappings/net/minecraft/util/MobUtils.mapping
@@ -1,0 +1,2 @@
+CLASS net/minecraft/class_5493 net/minecraft/util/MobUtils
+	METHOD method_30955 withValidNavigation (Lnet/minecraft/class_1308;)Z
--- a/mappings/net/minecraft/class_5565.mapping
+++ b/mappings/net/minecraft/class_5565.mapping
@@ -1,4 +1,7 @@
 CLASS net/minecraft/class_5565
+	FIELD field_27231 dataFixer Lcom/mojang/datafixers/DataFixer;
+	FIELD field_27232 LOGGER Lorg/apache/logging/log4j/Logger;
+	FIELD field_27233 serverWorld Lnet/minecraft/class_3218;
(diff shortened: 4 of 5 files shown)


mappings/net/minecraft/client/gui/screen/MojangLogoLoadingScreenOverlay.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_425 net/minecraft/client/gui/screen/SplashScreen
+CLASS net/minecraft/class_425 net/minecraft/client/gui/screen/MojangLogoLoadingScreenOverlay
```
ramidzkh: Not a fan of such a verbose name. `LoadingScreenOverlay` is enough

ramidzkh (changes requested):

mappings/net/minecraft/client/gui/screen/MojangLogoLoadingScreenOverlay.mapping
teddyxlandlee: @ramidzkh What about the others? Do I need to reopen a PR?

liach: You can just go back to your local yarn, rename this class, save, and use git to commit and push to your branch

mappings/net/minecraft/util/MobUtils.mapping
```diff
@@ -0,0 +1,2 @@
+CLASS net/minecraft/class_5493 net/minecraft/util/MobUtils
```
liach, line 1: Hmm, we usually only put really generic, i.e. not-that-much game related stuff in util. This thing we usually put in the entity package. In addition, I guess we might come up with a more specific name?

teddyxlandlee: Hmm... I guess I have to put it in `net.minecraft.entity.mob` package and rename it. Let me take a century 😄   
Plus. What about the `LoadingScreenOverlay`? I have just pushed it on.

mappings/net/minecraft/entity/EntityManager.mapping
```diff
@@ -1,4 +1,5 @@
-CLASS net/minecraft/class_5569
+CLASS net/minecraft/class_5569 net/minecraft/entity/EntityManager
```
liach, line 1: Lol no, this is an entity listener/callback assigned by the manager to each entity.
See https://github.com/FabricMC/yarn/pull/1967/files#diff-ef57025da02812a1e3d26e02f5612f228e64800e7afc5a84ef00d55818833599R1

liach (changes requested):

liach:
Wait, does anyone know why the popup loading screen was named "splash"? I think that's because it pops up, right? Then SplashScreen is a better name. Fyi it's from #180

mappings/net/minecraft/client/gui/screen/MojangLogoLoadingScreenOverlay.mapping
pluiedev: +1 for `LoadingScreenOverlay` (that was called a `Screen` even though it doesn't inherit from it; it literally wasted half an hour of my life trying to find it :sadtater:)

liach:
Also, look at https://www.dictionary.com/browse/splash?s=t which defines splash as a patch of color or light and a striking impression, I think the name splash should be kept.

Juuxel:
@liach It's also a *splash screen* with its own definition (see [Wikipedia](https://en.wikipedia.org/wiki/Splash_screen) or [JDK docs](https://docs.oracle.com/en/java/javase/15/docs/api/java.desktop/java/awt/SplashScreen.html))

mappings/net/minecraft/client/gui/screen/MojangLogoLoadingScreenOverlay.mapping
liach: Should probably be like `SplashOverlay` then, see https://github.com/FabricMC/yarn/pull/1992#issuecomment-770912109

mappings/net/minecraft/util/MobUtils.mapping
liach: Hmm, I've named this class in #2030. I think my pr is more comprehensive as it covers the related code as a system.

liach:
For your 4 changes:
classes 5565 and 5569 and the fields you've mapped are already mapped in 47bdb3b9e45e939fb9adf618d062fe6268f66eca
class_5493 is part of #2030 which maps the whole path targeting logic more comprehensively
now this pr is really only for the rename of SplashScreen.

YanisBft:
Closing as this is outdated

PR #2001, opened 2021-01-31
codec utilities

liach:


mappings/net/minecraft/util/dynamic/Codecs.mapping
```diff
@@ -0,0 +1,35 @@
+CLASS net/minecraft/class_5699 net/minecraft/util/dynamic/Codecs
+	COMMENT A few static utilities that should have been in {@link Codec} or {@link DynamicOps}.
```
Juuxel: This feels kind of opinionated for a javadoc comment...

liach: Hmm, the double stream codec is very similar to the long stream or int stream codecs (public static final fields) in `Codec` interface. Same for the 2 public static methods; they are almost identical to the default methods of `getIntStream` `createIntList` (also for long variants) in `DynamicOps`. I actually don't know why mojang moved these to some independent classes than just include them in DFU.

Juuxel: Minecraft's code base could've been a test ground for these, and they just forgot to move them to DFU (I wouldn't be surprised with Mojang lol)

liach:
```suggestion
	COMMENT A few extensions for {@link Codec} or {@link DynamicOps}.
```
Sounds good? Like in the sense of kotlin extensions

Juuxel (approved):

YanisBft (approved):

modmuss50 (approved):

PR #2002, opened 2021-01-31
Document one interesting method

liach:
We should clean up these portal mess one day!

### DIFF
--- a/mappings/net/minecraft/class_5459.mapping
+++ b/mappings/net/minecraft/class_5459.mapping
@@ -1,6 +1,6 @@
 CLASS net/minecraft/class_5459
-	METHOD method_30576 getLargestRectangle ([I)Lcom/mojang/datafixers/util/Pair;
-		COMMENT Gets the largest rectangle within a histogram, where the vertical bars each have
+	METHOD method_30576 findLargestRectangle ([I)Lcom/mojang/datafixers/util/Pair;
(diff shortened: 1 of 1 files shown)


YanisBft (approved):

enbrain (approved):

Juuxel (approved):
Why are portals doing this :concern:

mappings/net/minecraft/class_5459.mapping
```diff
@@ -1,4 +1,24 @@
 CLASS net/minecraft/class_5459
+	METHOD method_30576 getLargestRectangle ([I)Lcom/mojang/datafixers/util/Pair;
+		COMMENT Gets the largest rectangle within a histogram, where the vertical bars each have
```
enbrain:
```suggestion
	METHOD method_30576 findLargestRectangle ([I)Lcom/mojang/datafixers/util/Pair;
		COMMENT Finds the largest rectangle within a histogram, where the vertical bars each have
```

`find` could be better since the largest rectangle is not obvious from a histogram.

haykam821 (approved):

mappings/net/minecraft/class_5459.mapping
liach: Lol, leetcode has bad names as well :laughing: Let me check if there is other associated renames

PR #2007, opened 2021-02-01
Element charTyped wrong parameter

skyecodes:
The parameter was named keyCode but the modifier keys are passed as parameter (you can check by looking at `Keyboard.onChar` and `GLFWCharModsCallbackI.invoke`'s third parameter).

Also this is my first Pull request here and I wasn't sure whether to target the latest release or the latest snapshot because the issue is present on both.


liach (approved):

liach:
You can open a pr later to cherry pick to 1.16.5 as well.

Juuxel (approved):

modmuss50:
Looks good, looks like a nice target for unpicking once ive finshed that 👍

PR #2009, opened 2021-02-03
Map moss

haykam821:
This pull request maps the moss carpets and moss blocks added in 21w05a.

### DIFF
--- a/mappings/net/minecraft/item/Items.mapping
+++ b/mappings/net/minecraft/item/Items.mapping
@@ -242,6 +242,7 @@ CLASS net/minecraft/class_1802 net/minecraft/item/Items
 	FIELD field_28101 SCULK_SENSOR Lnet/minecraft/class_1792;
 	FIELD field_28409 GLOW_LICHEN Lnet/minecraft/class_1792;
 	FIELD field_28653 MOSS_CARPET Lnet/minecraft/class_1792;
+	FIELD field_28654 MOSS_BLOCK Lnet/minecraft/class_1792;
 	FIELD field_8043 ORANGE_TERRACOTTA Lnet/minecraft/class_1792;
 	FIELD field_8047 JUNGLE_PRESSURE_PLATE Lnet/minecraft/class_1792;
 	FIELD field_8048 SPRUCE_BUTTON Lnet/minecraft/class_1792;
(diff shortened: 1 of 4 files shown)


OroArmor:
Shouldnt this also include the moss blocks in `Blocks.java`? I only see the items for the new moss blocks

haykam821:
That isn't necessary since they are [auto-mapped](https://github.com/FabricMC/yarn/blob/323e6f8e29e7064360fecb0037e076c5e073dc72/mappings/net/minecraft/block/Blocks.mapping) based on [the string passed to the register method](https://github.com/FabricMC/yarn/blob/323e6f8e29e7064360fecb0037e076c5e073dc72/mappings/net/minecraft/block/Blocks.mapping#L147).

OroArmor:
oh wait, yarn is way smarter than i thought

OroArmor (approved):

mappings/net/minecraft/item/Items.mapping
```diff
@@ -241,6 +241,8 @@ CLASS net/minecraft/class_1802 net/minecraft/item/Items
 	FIELD field_28043 DRIPSTONE_BLOCK Lnet/minecraft/class_1792;
 	FIELD field_28101 SCULK_SENSOR Lnet/minecraft/class_1792;
 	FIELD field_28409 GLOW_LICHEN Lnet/minecraft/class_1792;
+	FIELD field_28653 MOSS_CARPET Lnet/minecraft/class_1792;
+	FIELD field_28654 MOSS_BLOCK Lnet/minecraft/class_1792;
```
liach, line 245: Is the item id `moss_block` and the block it `moss`? If yes guess it's just another lilypad or cauldron

liach (approved):

mappings/net/minecraft/item/Items.mapping
haykam821: No, the block and item identifiers are the same. The sound events just use `moss`, though.

liach: Hmm, If they are both `moss_block` I'd suggest naming the block field `MOSS_BLOCK` as well, given how `GOLD_BLOCK` etc are named

haykam821: Hm? It is already auto-mapped as `Blocks#MOSS_BLOCK`; the `MOSS` in this pull request is a `BlockSoundGroup` containing sound events such as `block.moss.break`.

jaskarth (approved):

PR #2011, opened 2021-02-03
Bundle tutorial

liach:

### DIFF
--- a/mappings/net/minecraft/client/tutorial/BundleTutorial.mapping
+++ b/mappings/net/minecraft/client/tutorial/BundleTutorial.mapping
@@ -1,0 +1,13 @@
+CLASS net/minecraft/class_5829 net/minecraft/client/tutorial/BundleTutorial
+	FIELD field_28798 manager Lnet/minecraft/class_1156;
+	FIELD field_28799 options Lnet/minecraft/class_315;
+	FIELD field_28800 toast Lnet/minecraft/class_372;
+	METHOD <init> (Lnet/minecraft/class_1156;Lnet/minecraft/class_315;)V
+		ARG 1 manager
+		ARG 2 options
+	METHOD method_33701 start ()V
+	METHOD method_33702 onRegularSlotClick (Lnet/minecraft/class_1799;Lnet/minecraft/class_1799;Lnet/minecraft/class_5536;)V
+		ARG 1 cursorStack
+		ARG 2 slotStack
+		ARG 3 clickType
+	METHOD method_33703 end ()V
--- a/mappings/net/minecraft/client/tutorial/BundleTutorial.mapping
+++ b/mappings/net/minecraft/client/tutorial/BundleTutorial.mapping
@@ -6,7 +6,10 @@ CLASS net/minecraft/class_5829 net/minecraft/client/tutorial/BundleTutorial
 		ARG 1 manager
 		ARG 2 options
 	METHOD method_33701 start ()V
-	METHOD method_33702 onRegularSlotClick (Lnet/minecraft/class_1799;Lnet/minecraft/class_1799;Lnet/minecraft/class_5536;)V
+	METHOD method_33702 onPickupSlotClick (Lnet/minecraft/class_1799;Lnet/minecraft/class_1799;Lnet/minecraft/class_5536;)V
+		COMMENT A callback for starting the bundle tutorial.
+		COMMENT
+		COMMENT @see TutorialManager#onPickupSlotClick(ItemStack, ItemStack, ClickType)
 		ARG 1 cursorStack
 		ARG 2 slotStack
 		ARG 3 clickType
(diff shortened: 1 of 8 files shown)


mappings/net/minecraft/screen/ScreenHandler.mapping
```diff
@@ -37,7 +37,7 @@ CLASS net/minecraft/class_1703 net/minecraft/screen/ScreenHandler
 	METHOD method_17696 (Lnet/minecraft/class_2248;Lnet/minecraft/class_1657;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;)Ljava/lang/Boolean;
 		ARG 2 world
 		ARG 3 pos
-	METHOD method_30010 removeStack (IILnet/minecraft/class_1713;Lnet/minecraft/class_1657;)Lnet/minecraft/class_1799;
+	METHOD method_30010 onSlotClickUnsafe (IILnet/minecraft/class_1713;Lnet/minecraft/class_1657;)Lnet/minecraft/class_1799;
```
jaskarth: Why is this unsafe? Perhaps some javadoc explaining it could be useful here.

liach: This one is wrapped in a call from `onSlotClick` like how entity/block entity ticks are wrapped. Unfortunately private.

liach: Since this is a private method, I think the context is quite straightforward and doesn't need javadocs.

Juuxel: If it's just the internal implementation of `onSlotClick`, naming it `onSlotClickImpl` or something similar would be better. Unless there's something that actually makes this method unsafe?

mappings/net/minecraft/client/tutorial/BundleTutorial.mapping
```diff
@@ -0,0 +1,13 @@
+CLASS net/minecraft/class_5829 net/minecraft/client/tutorial/BundleTutorial
+	FIELD field_28798 manager Lnet/minecraft/class_1156;
+	FIELD field_28799 options Lnet/minecraft/class_315;
+	FIELD field_28800 toast Lnet/minecraft/class_372;
+	METHOD <init> (Lnet/minecraft/class_1156;Lnet/minecraft/class_315;)V
+		ARG 1 manager
+		ARG 2 options
+	METHOD method_33701 start ()V
+	METHOD method_33702 onRegularSlotClick (Lnet/minecraft/class_1799;Lnet/minecraft/class_1799;Lnet/minecraft/class_5536;)V
```
Gegy: It's a bit unclear how a "regular" slot click is different from just a slot click. Unqualified would be expected to be regular, so it is potentially confusing.

liach: I find no good term to summarize those non-quick craft non shift click etc clicks. It is just an adhoc created by mojang so they can call the bundle tutorial

liach:
Will update when I have time later today.

mappings/net/minecraft/screen/ScreenHandler.mapping
liach: Now I think should name it `internalOnSlotClick`

mappings/net/minecraft/client/tutorial/BundleTutorial.mapping
liach: Think it is actually pickup slot action type click. rename coming

mappings/net/minecraft/client/tutorial/TutorialManager.mapping
```diff
@@ -1,16 +1,44 @@
...
+		ARG 1 toast
+			COMMENT the tutorial toast
+		ARG 2 ticks
+			COMMENT the time the toast will last, in client ticks
+	METHOD method_31366 (Lnet/minecraft/class_372;Lnet/minecraft/class_1156$class_5524;)Z
+		ARG 1 entry
+	METHOD method_31367 (Ljava/lang/Object;)Z
+		ARG 0 entry
+	METHOD method_33704 onPickupSlotClick (Lnet/minecraft/class_1799;Lnet/minecraft/class_1799;Lnet/minecraft/class_5536;)V
+		COMMENT Called when a player performs a {@link net.minecraft.screen.slot.SlotActionType
```
liach:
```suggestion
		COMMENT Called when a player performs a {@link net.minecraft.screen.slot.SlotActionType#PICKUP
```

mappings/net/minecraft/entity/player/PlayerEntity.mapping
```diff
@@ -76,6 +76,17 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 	METHOD method_31548 getInventory ()Lnet/minecraft/class_1661;
 	METHOD method_31549 getAbilities ()Lnet/minecraft/class_1656;
 	METHOD method_31550 isUsingSpyglass ()Z
+	METHOD method_33592 onPickupSlotClick (Lnet/minecraft/class_1799;Lnet/minecraft/class_1799;Lnet/minecraft/class_5536;)V
+		COMMENT Called when a player performs a {@link net.minecraft.screen.slot.SlotActionType
```
liach:
```suggestion
		COMMENT Called when a player performs a {@link net.minecraft.screen.slot.SlotActionType#PICKUP
```

Gegy (approved):

YanisBft (approved):

jaskarth (approved):

PR #2012, opened 2021-02-04
Game options visitor

liach:
so they don't write a code twice for save and load

### DIFF
--- a/mappings/net/minecraft/client/option/GameOptions.mapping
+++ b/mappings/net/minecraft/client/option/GameOptions.mapping
@@ -122,6 +122,7 @@ CLASS net/minecraft/class_315 net/minecraft/client/option/GameOptions
 		ARG 1 category
 	METHOD method_1631 togglePlayerModelPart (Lnet/minecraft/class_1664;Z)V
 		ARG 1 part
+		ARG 2 enabled
 	METHOD method_1632 getCloudRenderMode ()Lnet/minecraft/class_4063;
 	METHOD method_1635 setPlayerModelPart (Lnet/minecraft/class_1664;Z)V
 		ARG 1 part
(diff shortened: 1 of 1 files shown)


jaskarth (approved):

mappings/net/minecraft/client/option/GameOptions.mapping
```diff
@@ -139,10 +140,65 @@ CLASS net/minecraft/class_315 net/minecraft/client/option/GameOptions
...
+		ARG 0 value
+	METHOD method_33675 loadArm (Ljava/lang/String;)Lnet/minecraft/class_1306;
+		ARG 0 arm
+	CLASS 2
+		METHOD method_33676 find (Ljava/lang/String;)Ljava/lang/String;
+			ARG 1 key
+	CLASS 3
+		METHOD method_33677 print (Ljava/lang/String;)V
+			ARG 1 key
+	CLASS class_5823 Visitor
```
i509VCB, line 179: Hmm I wonder, would this allow injection of custom options.

liach: Yes if you can bytecode hack `accept` to add an extra call to account for another option.
Unfortunately afaik, mixin still doesn't work nicely with these package classes, right? Kaboom time as this interface indeed is package private https://maven.fabricmc.net/docs/yarn-21w05a+build.1/net/minecraft/client/option/GameOptions.class_5823.html

Juuxel: You can inject to them just fine but mixin doesn't really help with extending/implementing/creating them.

liach: Well, mixin has trouble calling methods from this interface, though you most likely don't need new methods in this interface that handle extra data types

Juuxel (approved):

PR #2022, opened 2021-02-04
Add unnecessary Eclipse files to .gitignore

haykam821:


liach (approved):
Dunno why you run eclipse here but sure

modmuss50:
Should really be using a git include solves issues like this as someone is bound to come along with something else to ignore.

haykam821:
I have no idea where I got these files, but I accidentally committed them in 9aecefec711f879199fa0a37e374a5ae6ff04f43. They might as well be included if IntelliJ IDEA gets its own section.

modmuss50:
The change is fine, im just saying there is a better way to do the thing overall. (Prob out of scope for this PR)

modmuss50 (approved):

PR #2032, opened 2021-02-08
item slot

liach:


mappings/net/minecraft/inventory/ItemSlot.mapping
```diff
@@ -0,0 +1,26 @@
+CLASS net/minecraft/class_5630 net/minecraft/inventory/ItemSlot
```
Juuxel: This might get confused with the other item slot class, `Slot` which is used by menus/SHs.

liach: So `EntityItemSlot`? `ArgumentItemSlot`? `CommandItemSlot`?

liach: I will use `CommandItemSlot` as this represents an item slot for command access.

YanisBft (approved):

Juuxel (approved):

Mysterious-Dev (approved):

PR #2033, opened 2021-02-08
entity tracking listener

liach:

### DIFF
--- a/mappings/net/minecraft/server/world/EntityTrackingAudience.mapping
+++ b/mappings/net/minecraft/server/world/EntityTrackingAudience.mapping
@@ -1,5 +1,0 @@
-CLASS net/minecraft/class_5629 net/minecraft/server/world/EntityTrackingAudience
-	COMMENT An audience to entity tracking within threaded anvil chunk storage.
-	METHOD method_14364 sendPacket (Lnet/minecraft/class_2596;)V
-		ARG 1 packet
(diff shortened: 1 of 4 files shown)


mappings/net/minecraft/server/world/EntityTrackingAudience.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS net/minecraft/class_5629 net/minecraft/server/world/EntityTrackingAudience
```
i509VCB: So the mappings are going on an *adventure*?

liach: What? Don't understand

haykam821 (approved):

YanisBft (approved):

jaskarth (changes requested):
Why "audience" instead of "listener"? I'm not sure if I fully understand the meaning here.

liach:
Good point, i can rename to listener. Wait for my free time

i509VCB:
How I see it, an audience would refer to an object that can receive messages in very generic terms.

YanisBft (approved):

jaskarth (approved):

liach:
@modmuss50 mind review?

modmuss50 (approved):

PR #2058, opened 2021-02-13
Map Sprite sub classes

FlashyReese:
Just some animation stuff Mojang moved.


liach (approved):

liach:
nice, now just some client option nested classes and we will get 100% for classes

modmuss50 (approved):

PR #2064, opened 2021-02-13
Skeleton conversion

YanisBft:
Mapped the skeleton conversion behavior added in 1.17, and somee random mappings

### DIFF
--- a/mappings/net/minecraft/entity/projectile/ProjectileEntity.mapping
+++ b/mappings/net/minecraft/entity/projectile/ProjectileEntity.mapping
@@ -2,7 +2,7 @@ CLASS net/minecraft/class_1676 net/minecraft/entity/projectile/ProjectileEntity
 	FIELD field_22478 ownerUuid Ljava/util/UUID;
 	FIELD field_23739 ownerEntityId I
 	FIELD field_23740 leftOwner Z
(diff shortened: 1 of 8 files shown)


mappings/net/minecraft/entity/projectile/ProjectileEntity.mapping
```diff
@@ -2,6 +2,7 @@ CLASS net/minecraft/class_1676 net/minecraft/entity/projectile/ProjectileEntity
 	FIELD field_22478 ownerUuid Ljava/util/UUID;
 	FIELD field_23739 ownerEntityId I
 	FIELD field_23740 leftOwner Z
+	FIELD field_28646 hasBeenShot Z
```
liach: Maybe just
```suggestion
	FIELD field_28646 shot Z
```

liach (approved):

mappings/net/minecraft/entity/projectile/ProjectileEntity.mapping
YanisBft: It's from a string but yeah it's really the same

enbrain (approved):

modmuss50 (approved):

Issue #2057, opened 2021-02-13
NoiseChunkGenerator should be in net.minecraft.world.gen


OroArmor:
```
java.lang.IllegalAccessError: class net.minecraft.world.gen.chunk.NoiseChunkGenerator tried to access protected method 'double net.minecraft.world.gen.StructureWeightSampler.getWeight(int, int, int)' (net.minecraft.world.gen.chunk.NoiseChunkGenerator and net.minecraft.world.gen.StructureWeightSampler are in unnamed module of loader 'app')
```

While I think the package for `NoiseChunkGenerator` is currently better, running Minecraft in a development environment causes this crash.

liach:
Hmm, seems the package visibility index from enigma has stopped working for long

liach:
I doubt that the index works but the command line interface broke. Need to check.

modmuss50:
I did start looking at improving and fixing the mappings validator. Would be nice to try and get rid of these issues. But we cannot guarantee it

liach:
Turns out check mappings was never broke. It was this that broke:
https://github.com/FabricMC/yarn/blame/21w06a/build.gradle#L396
Should be using intermediary jar

And this check mappings never worked since the intermediary update 2 years ago!

liach:
<details>
<summary>
Look at the current results:
</summary>

```
ERROR: Must be in one package:
net/minecraft/client/gui/screen/ConnectScreen
net/minecraft/client/gui/screen/ConnectScreen$1
net/minecraft/client/gui/screen/CustomizeBuffetLevelScreen
net/minecraft/client/gui/screen/CustomizeBuffetLevelScreen$1
net/minecraft/client/gui/screen/CustomizeBuffetLevelScreen$BuffetBiomesListWidget
net/minecraft/client/gui/screen/CustomizeBuffetLevelScreen$BuffetBiomesListWidget$BuffetBiomeItem
net/minecraft/client/gui/screen/CustomizeFlatLevelScreen
net/minecraft/client/gui/screen/CustomizeFlatLevelScreen$1
net/minecraft/client/gui/screen/CustomizeFlatLevelScreen$SuperflatLayersListWidget
net/minecraft/client/gui/screen/CustomizeFlatLevelScreen$SuperflatLayersListWidget$SuperflatLayerItem
net/minecraft/client/gui/screen/DialogScreen
net/minecraft/client/gui/screen/DialogScreen$ChoiceButton
[...]

PR #2066, opened 2021-02-13
Make mapping check work again

liach:
It never worked since the intermediary update. So 2 years of "Error: Must be in one package"

Now made the task only print and swallows the illegal state it throws

See #2057. It's quite debatable if we should merge these packages or keep them split; I recall we had a similar debate when we were eliminating the `impl` packages for blocks, entities, etc.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -393,11 +393,15 @@ task checkMappings {
 		logger.lifecycle(":checking mappings")
 
 		String[] args = [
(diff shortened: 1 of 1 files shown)


liach:
Fyi see https://github.com/FabricMC/yarn/runs/1895221699#step:5:185

modmuss50 (approved):

PR #2068, opened 2021-02-14
Move registry lookup codec to dynamic

liach:
Solves the one package warning for registry lookup codec.


liach:
Keeping this as the rename somewhat makes sense. Many other registry impl codecs are in the dynamic package, and registry package is exclusively registry implementations.

YanisBft (approved):

jaskarth (approved):

PR #2070, opened 2021-02-14
Move chunk data access related stuff to storage

liach:
I added them when I mapped entity system. They fit storage well too

Resolves Entity chunk data access must be in same package as storage stuff


liach:
Keeping this as Imo this move is an improvement since chunk package is chunk logic than chunk save/load.

jaskarth (approved):

YanisBft (approved):

PR #2073, opened 2021-02-14
Move dilation to be with modelcuboid data that accesses its package

liach:
radiusXYZ fields


liach:
I decide to keep this pr as other users outside of the package still cannot really use these package-private fields before the move, and this class is dedicated to that single usage in model cuboid data.

modmuss50 (approved):

YanisBft (approved):

PR #2074, opened 2021-02-14
Move affine transformation and vector 4f to regular util math

liach:


liach:
Keeping this as mojang seems to randomly use some of these client math classes on server. Matrix4f rename was an example.

jaskarth (approved):

modmuss50 (approved):

PR #2078, opened 2021-02-15
Properly fix #2057 by asking tiny remapper to patch package access

### LINKED ISSUE #2057: NoiseChunkGenerator should be in net.minecraft.world.gen

liach:
Needs a loom pr as well. The loom pr can properly fix the problem
in dev env.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -387,6 +387,7 @@ task yarn(dependsOn: setupYarn) {
 }
 
 task checkMappings {
+	enabled = false
 	group = buildMappingGroup
 	inputs.dir mappingsDir
 	doLast {
@@ -397,6 +398,8 @@ task checkMappings {
 				mappingsDir.getAbsolutePath()
 		]
 
+		// todo: this checks package visibility which is already handled by tiny remapper
+		// need to add like check if two files map the same class, such as from multiple prs
 		try {
 			new CheckMappingsCommand().run(args)
 		} catch (IllegalStateException ignored) {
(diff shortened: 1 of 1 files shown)


liach:
@Chocohead Thank you for your suggestion from https://github.com/FabricMC/yarn/issues/2057#issuecomment-778848695

modmuss50:
No, this shouldn’t be used in loom as it will change it for mods compiling. I wouldn’t change it here either as there is no need to it will just cause added confusion between the two envs

liach:
then should yarn output like an access widener alongside mappings indicating members that need widening? or what approach do you suggest?

modmuss50:
I suggest we dont touch it. Mod dev envs should not have mapping specific access wideners as we would need to bake them into each mod.

Even we if fix all of the package access issues it will need to stay as it will break when an update comes along and adds a new class.

liach:
> I suggest we dont touch it. Mod dev envs should not have mapping specific access wideners as we would need to bake them into each mod.
> 
> Even we if fix all of the package access issues it will need to stay as it will break when an update comes along and adds a new class.

so you suggest keeping the status quo, which i agree. and we leave the handling to loader exclusively?

modmuss50:
Yes. 

I think mixin does complain a little in the log. I wouldn’t be against removing that log line though.

liach:
Hmm, should we add to tiny v2 metadata that some namespaces needs package access checks?
I mean, we can add extra metadata like `\tpackage-access-checks\tfalse\ttrue` etc. the values are tab split and each value indicates if each namespace may have package access issues

liach:
Closing in favor of https://github.com/FabricMC/tiny-remapper/issues/46

PR #2079, opened 2021-02-16
TextFieldWidget tweaks

shadowfacts:
The field focused clashes with superclass field, and is not accurate. It
actually controls whether the background is drawn. The getter and setter
methods are also not correct, as the field does not just control the
border:
image

The setSelected method does not set whether the field is selected, but
focused. It only exists to expose the protected AbstractButtonWidget.setFocused
method. It's named setTextFieldFocused not setFocused because that would
clash with the superclass name.

### DIFF
--- a/mappings/net/minecraft/client/gui/widget/TextFieldWidget.mapping
+++ b/mappings/net/minecraft/client/gui/widget/TextFieldWidget.mapping
@@ -3,7 +3,7 @@ CLASS net/minecraft/class_342 net/minecraft/client/gui/widget/TextFieldWidget
 	FIELD field_2088 changedListener Ljava/util/function/Consumer;
 	FIELD field_2092 text Ljava/lang/String;
 	FIELD field_2094 editable Z
-	FIELD field_2095 focused Z
+	FIELD field_2095 drawsBackground Z
 	FIELD field_2096 focusUnlocked Z
 	FIELD field_2098 uneditableColor I
 	FIELD field_2099 renderTextProvider Ljava/util/function/BiFunction;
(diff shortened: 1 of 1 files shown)


YanisBft (approved):

mappings/net/minecraft/client/gui/widget/TextFieldWidget.mapping
```diff
@@ -74,8 +74,8 @@ CLASS net/minecraft/class_342 net/minecraft/client/gui/widget/TextFieldWidget
 		ARG 1 newText
 	METHOD method_1875 setSelectionStart (I)V
 		ARG 1 cursor
-	METHOD method_1876 setSelected (Z)V
-		ARG 1 selected
+	METHOD method_1876 setTextFieldFocused (Z)V
```
liach, line 77: Hmm, having `TextField` in name doesn't sound quite good. Is this about whether the cursor for typing is in the field? If yes I'd say `setHasCursor` or alike

shadowfacts: I don't think `setHasCursor` is accurate, because it does more than that. The focused property in `AbstractButtonWidget` controls whether the text field controls both whether the cursor is shown and keyboad input is accepted. All this method does is call `super.setFocused`.

liach: hmm, what does this method do? is it an explicit override to return this? worth checking in javap, as i recall vec3i and blockpos has such up and down methods that override for a more specific return type

shadowfacts: It literally just calls the `AbstractButtonWidget.setFocused` because the superclass method is protected and the text field needs to make it public.

In `TextFieldWidget`:
image
In `AbstractButtonWidget`:
image

liach: After a look at the class file it appears they are indeed distinct methods and `method_1876` is not some sort of bridge. :+1:

liach (approved):

jaskarth (approved):

PR #2037, opened 2021-02-08
Refactor toTag/fromTag to follow current conventions.

LambdAurora:
This refactor is done following [the current conventions about serialization](https://github.com/FabricMC/yarn/blob/21w05b/CONVENTIONS.md#serialization).

The refactor has been assisted by a script: https://gist.github.com/LambdAurora/04608dac389e91cb5c41553641f6b8d2#file-tag_nbt_refactor-mjs

Everything has been checked afterwards and should be good.

Partially fixing #1989.

Issue #1989, opened 2021-01-28
Make NBT serialisation methods follow the conventions


Juuxel:
The NBT serialisation methods are currently a mess. Most of them were mapped before the conventions were a thing, and as a result they're pretty inconsistent.

For example, the conventions say that you should use `read`/`write` for non-static methods like the ones in `BlockEntity` - but `BlockEntity` uses `fromTag` and `toTag` instead. (The names should also probably use `-Nbt` instead of `-Tag`, as that seems to be the new standard.)

Should we make a bunch of PRs fixing this, or change the conventions to adapt to current naming? (I'd prefer the former for clean mappings, but the latter is better for modders.)

Technici4n:
I would be in favor of changing the conventions. That part was written 2 years ago anyway.

If the parameter is going to be a `Tag` or a `CompoundTag`, then keeping `fromTag` and `toTag` is better imo.

As for confusion between `nm.nbt.Tag` and `nm.tag.Tag`, I don't think you are deserializing from datapack tags anyway, and I don't think `toTag` returning `nm.tag.Tag` is supposed to happen either.

Earthcomputer:
> As for confusion between nm.nbt.Tag and nm.tag.Tag, I don't think you are deserializing from datapack tags anyway, and I don't think toTag returning nm.tag.Tag is supposed to happen either.

There have been multiple times where I have had to use both classes in the same class, and I am only able to import one. They should have different names.

Prospector:
> If the parameter is going to be a `Tag` or a `CompoundTag`, then keeping `fromTag` and `toTag` is better imo.

The format is NBT, though. When serializing to JSON, you wouldn't call that method `toObject`, you would call it `toJson` because the format is JSON, and that is what you care about, not the arbitrary name that the format chooses to call its objects/elements/tags/nodes/properties/values/entries.

> As for confusion between `nm.nbt.Tag` and `nm.tag.Tag`, I don't think you are deserializing from datapack tags anyway, and I don't think `toTag` returning `nm.tag.Tag` is supposed to happen either.

Until you work with items or entities where the phrase "get tag" could refer to either getting an NBT tag or a datapack tag. Really, the only time this isn't an issue is with blocks, because they don't store NBT.

[...]

PR #2090, opened 2021-02-17
Refactor the whole net.minecraft.nbt package.

LambdAurora:
The NBT package is quite a mess, with NBT `Tag` having the same name as <insert Registry> `Tag`s.

This PR drops entirely the `Tag` naming in the NBT package and is a follow up to #2037 and is related to #1989.

The naming scheme adopted by this PR is similar to what Gson has, or any other type libraries.

Some of the most controversial renames might be:
 - `EndTag` -> `NbtNull`, the reason of its rename is it's misleading as type `0` is also used as a `null` value in some cases and does not always represent the end of an object.
 - ~~`CompoundTag` -> `NbtObject`, similar to how JSON objects classes are named `JsonObject`, and like it's literally an object.~~ (`CompoundTag` -> `NbtCompound` now)
 
 The PR also includes a fair amount of javadocs.
 
 This PR won't touch anything outside of the nbt package, it can be delegated to another PR later.


mappings/net/minecraft/nbt/NbtInt.mapping
```diff
@@ -1,4 +1,7 @@
-CLASS net/minecraft/class_2497 net/minecraft/nbt/IntTag
+CLASS net/minecraft/class_2497 net/minecraft/nbt/NbtInt
+	COMMENT Represents a NBT 32-bits integer.
```
i509VCB:
```suggestion
	COMMENT Represents a NBT 32-bit integer.
```

mappings/net/minecraft/nbt/NbtLong.mapping
```diff
@@ -1,4 +1,7 @@
-CLASS net/minecraft/class_2503 net/minecraft/nbt/LongTag
+CLASS net/minecraft/class_2503 net/minecraft/nbt/NbtLong
+	COMMENT Represents a NBT 64-bits integer.
```
i509VCB:
```suggestion
	COMMENT Represents a NBT 64-bit integer.
```

mappings/net/minecraft/nbt/NbtLongArray.mapping
```diff
@@ -1,4 +1,7 @@
-CLASS net/minecraft/class_2501 net/minecraft/nbt/LongArrayTag
+CLASS net/minecraft/class_2501 net/minecraft/nbt/NbtLongArray
+	COMMENT Represents a NBT 64-bits integer array.
```
i509VCB:
```suggestion
	COMMENT Represents a NBT 64-bit integer array.
```

i509VCB (commented):
Typically we refer to a number like an integer as a `32-bit` number.

mappings/net/minecraft/nbt/NbtInt.mapping
LambdAurora: I always get it wrong, then learn about it and still manage to forget it. Sorry! Should be fixed now.

mappings/net/minecraft/nbt/NbtObject.mapping
```diff
@@ -1,9 +1,12 @@
-CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
-	FIELD field_11515 tags Ljava/util/Map;
+CLASS net/minecraft/class_2487 net/minecraft/nbt/NbtObject
+	COMMENT Represents an NBT object which can hold values associated to a key.
```
liach: Maybe mention this is unordered key-value pairs with distinct string keys, case sensitive

mappings/net/minecraft/nbt/AbstractNbtList.mapping
```diff
@@ -0,0 +1,19 @@
+CLASS net/minecraft/class_2483 net/minecraft/nbt/AbstractNbtList
```
liach, line 1: need javadocs, such as this list holds elements of the same type

mappings/net/minecraft/nbt/AbstractNbtNumber.mapping
```diff
@@ -0,0 +1,32 @@
+CLASS net/minecraft/class_2514 net/minecraft/nbt/AbstractNbtNumber
+	COMMENT Represents an NBT number.
+	COMMENT <p>
+	COMMENT The type value {@code 99} is an alias for elements representing numbers.
+	METHOD method_10696 getShort ()S
+		COMMENT Gets the value as a 16-bit integer.
+		COMMENT
+		COMMENT @return the value as a short
+	METHOD method_10697 getDouble ()D
+		COMMENT Gets the value as a double.
```
liach: No double, use 64-bit floating-point number/double-precision floating-point number

mappings/net/minecraft/nbt/AbstractNbtNumber.mapping
```diff
@@ -0,0 +1,32 @@
...
+		COMMENT
+		COMMENT @return the value as a double
+	METHOD method_10698 getByte ()B
+		COMMENT Gets the value as a byte.
```
liach: Mention this is 8 bit

mappings/net/minecraft/nbt/AbstractNbtNumber.mapping
```diff
@@ -0,0 +1,32 @@
...
+		COMMENT
+		COMMENT @return the value as a byte
+	METHOD method_10699 getLong ()J
+		COMMENT Gets the value as a 64-bit integer.
+		COMMENT
+		COMMENT @return the value as long
+	METHOD method_10700 getFloat ()F
+		COMMENT Gets the value as a float.
```
liach: Use 32-bit floating-point number/single-precision floating-point number

mappings/net/minecraft/nbt/NbtElementReader.mapping
```diff
@@ -1,4 +1,5 @@
-CLASS net/minecraft/class_4614 net/minecraft/nbt/TagReader
+CLASS net/minecraft/class_4614 net/minecraft/nbt/NbtElementReader
```
liach: More like Nbt Type

liach (commented):
need to fix byte/short/float/double to say xx-bit number if you do that for int/long.

mappings/net/minecraft/nbt/AbstractNbtList.mapping
```diff
@@ -0,0 +1,20 @@
...
+	COMMENT Represents an abstraction of a mutable NBT list which holds elements of the same type.
+	METHOD add (ILjava/lang/Object;)V
+		ARG 1 index
+		ARG 2 value
+	METHOD method_10533 addElement (ILnet/minecraft/class_2520;)Z
+		ARG 1 index
+		ARG 2 element
+	METHOD method_10535 setElement (ILnet/minecraft/class_2520;)Z
```
liach, line 9:
```suggestion
	METHOD method_10533 add (ILnet/minecraft/class_2520;)Z
		ARG 1 index
		ARG 2 element
	METHOD method_10535 set (ILnet/minecraft/class_2520;)Z
```

The synthetic `void add(int, Object)` and `set(int, Object)` generated by javac to override the superclass methods call these (because jvm calls by exact descriptor match), so per javac behavior they are named so

liach (commented):
Also I wonder, someone talked about renaming Nbt Reader before to like some type thing. Does it need a rename?

mappings/net/minecraft/nbt/AbstractNbtList.mapping
LambdAurora: ~~those are not similar to the superclass ones as it returns boolean value if the passed element has been successfully added (if it's the good type)~~

~~superclass set is `<T> T set(int index, T value)` and `method_10535` is `boolean set(int index, NbtElement element)` but I can agree with renaming those still but I feel like we lose some explicit naming.~~

I have said nothing, Enigma did not displayed the override annotations and forgot the actual signatures present in `AbstractList`.

LambdAurora:
> Also I wonder, someone talked about renaming Nbt Reader before to like some type thing. Does it need a rename?

I'm not sure of what you exactly mean? The only reader class I can think of is `TagReader` that has been renamed to `NbtTypeReader`.

Juuxel (changes requested):
Compounds should stay as compounds (I'd also prefer end as end since it's the established name for that, but that's not as common anyway). This type of naming is really widespread and used in many places like the [Minecraft wiki](https://minecraft.gamepedia.com/NBT_format) and even [unofficial NBT libraries](https://github.com/KyoriPowered/adventure/blob/1718f52c8ef065dcc6e42668621bd670bb96f72e/nbt/src/main/java/net/kyori/adventure/nbt/CompoundBinaryTag.java).

LambdAurora:
> Compounds should stay as compounds (I'd also prefer end as end since it's the established name for that, but that's not as common anyway). This type of naming is really widespread and used in many places like the [Minecraft wiki](https://minecraft.gamepedia.com/NBT_format) and even [unofficial NBT libraries](https://github.com/KyoriPowered/adventure/blob/1718f52c8ef065dcc6e42668621bd670bb96f72e/nbt/src/main/java/net/kyori/adventure/nbt/CompoundBinaryTag.java).

@Juuxel I honestly think that the "Compound" name doesn't make a lot of sense. Though I don't think I'll be able to push for `NbtObject` so I switched to `NbtCompound` but javadocs refer it as `compound object` as compound alone wouldn't make sense.

Also end can't stay as end, it's misleading and plain wrong when looking at its usage. Null makes much more sense.

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -109,14 +120,14 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
 	METHOD method_10583 getFloat (Ljava/lang/String;)F
 		ARG 1 key
 	METHOD method_25926 getUuid (Ljava/lang/String;)Ljava/util/UUID;
-		COMMENT Reads a {@link UUID} from its NBT representation in this {@code CompoundTag}.
+		COMMENT Reads a {@link UUID} from its NBT representation in this {@code NbtObject}.
```
haykam821:
```suggestion
		COMMENT Reads a {@link UUID} from its NBT representation in this {@code NbtCompound}.
```

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -109,14 +120,14 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
...
 		ARG 1 key
 	METHOD method_25927 putUuid (Ljava/lang/String;Ljava/util/UUID;)V
-		COMMENT Writes a {@link UUID} to its NBT representation in this {@code CompoundTag}.
+		COMMENT Writes a {@link UUID} to its NBT representation in this {@code NbtObject}.
```
haykam821:
```suggestion
		COMMENT Writes a {@link UUID} to its NBT representation in this {@code NbtCompound}.
```

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -109,14 +120,14 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
...
 		ARG 1 key
 		ARG 2 value
 	METHOD method_25928 containsUuid (Ljava/lang/String;)Z
-		COMMENT Returns {@code true} if this {@code CompoundTag} contains a valid UUID representation associated with the given key.
+		COMMENT Returns {@code true} if this {@code NbtObject} contains a valid UUID representation associated with the given key.
```
haykam821:
```suggestion
		COMMENT Returns {@code true} if this {@code NbtCompound} contains a valid UUID representation associated with the given key.
```

mappings/net/minecraft/nbt/visitor/NbtElementVisitor.mapping
```diff
@@ -0,0 +1,28 @@
+CLASS net/minecraft/class_5627 net/minecraft/nbt/visitor/NbtElementVisitor
+	COMMENT A visitor interface for NBT elements.
+	METHOD method_32290 visitNbtByteArray (Lnet/minecraft/class_2479;)V
+		ARG 1 element
+	METHOD method_32291 visiteNbtByte (Lnet/minecraft/class_2481;)V
+		ARG 1 element
+	METHOD method_32292 visitNbtObject (Lnet/minecraft/class_2487;)V
```
haykam821:
```suggestion
	METHOD method_32292 visitNbtCompound (Lnet/minecraft/class_2487;)V
```

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -57,9 +64,10 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
 		ARG 1 key
 	METHOD method_10559 createCrashReport (Ljava/lang/String;Lnet/minecraft/class_4614;Ljava/lang/ClassCastException;)Lnet/minecraft/class_128;
 		ARG 1 key
+		ARG 2 reader
 	METHOD method_10561 getIntArray (Ljava/lang/String;)[I
 		ARG 1 key
-	METHOD method_10562 getCompound (Ljava/lang/String;)Lnet/minecraft/class_2487;
+	METHOD method_10562 getObject (Ljava/lang/String;)Lnet/minecraft/class_2487;
```
haykam821:
```suggestion
	METHOD method_10562 getCompound (Ljava/lang/String;)Lnet/minecraft/class_2487;
```

haykam821 (changes requested):

LambdAurora:
Thank you @haykam821 I honestly have no idea how I missed those are they are quite obvious.

LemmaEOF:
[Compound is the official name in the original NBT spec](https://web.archive.org/web/20110723210920/http://www.minecraft.net/docs/NBT.txt). Whether we want to respect Notch or not is up to us.

liach:
Hmm, should we call compounds "compound objects" or "compound maps"? dfu calls them maps

Prospector:
I could go either way on NbtObject vs NbtCompound. I do kinda like the consistency with GSON's JsonObject and I think it would be pretty intuitive for users.

LambdAurora:
I find `NbtObject` more intuitive and natural than `NbtCompound` tbh.

LemmaEOF:
having it be called `NbtObject` or even maybe `NbtMap` would definitely be an accessility buff tbh

Earthcomputer (approved):
:ok_hand: finally! Just the other day I *again* had to use the fully qualified `net.minecraft.nbt.Tag` name due to that stupid name conflict, since the last time we spoke about this.

Earthcomputer:
`NbtObject` would be putting us at odds with the rest of the Minecraft community, and anyone who comes from more than a casual-Minecraft-gameplay background (and seen NBT before) would be confused by it. I would be willing to compromise on it but I still think `NbtCompound` is better. Far bigger issue is the name conflict with `Tag` which we all agree needs to be solved.

liach:
yeah, for relation to mc, compound is a better name. also we need to make sure people realize the difference between json and nbt, so just copying gson names straight isn't perfect (e.g. json list does not limit item types, nbt enforce same type)

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -86,6 +94,9 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
 		ARG 1 key
 		ARG 2 value
 	METHOD method_10573 contains (Ljava/lang/String;I)Z
+		COMMENT Returns whether the compound object contains an element of the specified type at the specified key.
+		COMMENT
+		COMMENT @return {@code true} if the key exist and the element's type is compatible with {@code type}, else {@code false}
```
YanisBft:
```suggestion
		COMMENT @return {@code true} if the key exists and the element's type is compatible with {@code type}, else {@code false}
```

YanisBft (approved):

Prospector (approved):

liach:
Note: this implements #241.

jaskarth (approved):

mappings/net/minecraft/nbt/AbstractNbtList.mapping
```diff
@@ -0,0 +1,20 @@
+CLASS net/minecraft/class_2483 net/minecraft/nbt/AbstractNbtList
+	COMMENT Represents an abstraction of a mutable NBT list which holds elements of the same type.
+	METHOD add (ILjava/lang/Object;)V
+		ARG 1 index
+		ARG 2 value
```
liach, line 5: Hmm, seems mappingpoet generates java code for these bridge methods and causes javadoc errors. I will fix that later

mappings/net/minecraft/nbt/AbstractNbtNumber.mapping
```diff
@@ -0,0 +1,32 @@
...
+		COMMENT
+		COMMENT @return the value as a float
+	METHOD method_10701 getInt ()I
+		COMMENT Gets the value as a 32-bit integer.
+		COMMENT
+		COMMENT @return the value as an int
+	METHOD method_10702 getNumber ()Ljava/lang/Number;
+		COMMENT Gets the value as a generic number.
+		COMMENT
+		COMMENT @return the value as {@link java.lang.Number}
```
liach:
```suggestion
		COMMENT @return the value as a {@link Number}
```

mappings/net/minecraft/nbt/NbtOps.mapping
```diff
@@ -18,4 +18,58 @@ CLASS net/minecraft/class_2509 net/minecraft/nbt/NbtOps
...
+		ARG 1 list
+		ARG 2 value
+	METHOD mergeToMap (Ljava/lang/Object;Lcom/mojang/serialization/MapLike;)Lcom/mojang/serialization/DataResult;
+		ARG 1 element
+	METHOD mergeToMap (Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;)Lcom/mojang/serialization/DataResult;
+		ARG 1 map
+		ARG 2 key
+		ARG 3 value
+	METHOD method_10666 (Ljava/lang/String;Ljava/lang/String;)Z
+		ARG 1 oKey
```
liach:
```suggestion
		ARG 1 k
```
Assuming this is a computeIfAbsent lambda

mappings/net/minecraft/nbt/NbtOps.mapping
```diff
@@ -18,4 +18,58 @@ CLASS net/minecraft/class_2509 net/minecraft/nbt/NbtOps
...
+	METHOD method_10670 (Lnet/minecraft/class_2487;Lnet/minecraft/class_2487;Ljava/lang/String;)V
+		ARG 2 oKey
```
liach:
```suggestion
		ARG 2 k
```
same assumption as above.

mappings/net/minecraft/nbt/visitor/NbtElementVisitor.mapping
```diff
@@ -0,0 +1,28 @@
+CLASS net/minecraft/class_5627 net/minecraft/nbt/visitor/NbtElementVisitor
+	COMMENT A visitor interface for NBT elements.
+	METHOD method_32290 visitNbtByteArray (Lnet/minecraft/class_2479;)V
```
liach: Maybe we should drop all these `Nbt` in name, i.e. just `visitByteArray`, etc.?

mappings/net/minecraft/nbt/AbstractNbtList.mapping
```diff
@@ -0,0 +1,20 @@
...
+	METHOD method_10533 add (ILnet/minecraft/class_2520;)Z
+		ARG 1 index
+		ARG 2 element
+	METHOD method_10535 set (ILnet/minecraft/class_2520;)Z
```
liach: Sorry, my bad, but 10533 and 10535 are not list methods but distinct methods for other purposes. Maybe restore their previous `addElement` `setElement` names or you can find better names

LambdAurora: I honestly don't have better names so I'm gonna just restore the previous names.

liach (approved):
Sorry for the add/setElement mess. It's indeed weird there are two versions of these methods.

Juuxel (approved):
Looks good. One minor thing is that `<p>` tags are generally placed on the same line as the first line of the paragraph they start.

Example:
```
abcdef

<p>ghijkl
```

(also `NbtElement` sounds a bit odd but it's consistent with JSON libraries such as Gson, so I guess it's okay)

TheGlitch76 (approved):

mappings/net/minecraft/nbt/AbstractNbtNumber.mapping
```diff
@@ -0,0 +1,32 @@
...
+		COMMENT
+		COMMENT @return the value as a double
+	METHOD method_10698 getByte ()B
+		COMMENT Gets the value as an 8-bit integer.
+		COMMENT
+		COMMENT @return the value as a byte
+	METHOD method_10699 getLong ()J
+		COMMENT Gets the value as a 64-bit integer.
+		COMMENT
+		COMMENT @return the value as long
```
haykam821:
```suggestion
		COMMENT @return the value as a long
```

mappings/net/minecraft/nbt/AbstractNbtList.mapping
```diff
@@ -0,0 +1,20 @@
...
+	METHOD method_10533 addElement (ILnet/minecraft/class_2520;)Z
+		ARG 1 index
+		ARG 2 element
+	METHOD method_10535 setElement (ILnet/minecraft/class_2520;)Z
+		ARG 1 index
+		ARG 2 element
+	METHOD method_10601 getElementType ()B
+		COMMENT Gets the type of elements held in this list.
+		COMMENT
+		COMMENT @return the type of elements
```
haykam821: This wording is a bit confusing and to me suggests that a list can hold different types, so I would rename it to `getType`:

```suggestion
	METHOD method_10601 getType ()B
		COMMENT Gets the type of element that this list holds.
		COMMENT
		COMMENT @return the type of element that this list holds
```

mappings/net/minecraft/nbt/NbtList.mapping
```diff
@@ -1,4 +1,11 @@
-CLASS net/minecraft/class_2499 net/minecraft/nbt/ListTag
+CLASS net/minecraft/class_2499 net/minecraft/nbt/NbtList
+	COMMENT Represents an NBT list.
+	COMMENT <p>
+	COMMENT Type value: {@code 9}.
```
haykam821: The punctuation here does not match the similar documentation on other classes. All of the 'type value' lines should have consistent punctuation with each other.

mappings/net/minecraft/nbt/NbtNull.mapping
```diff
@@ -0,0 +1,8 @@
+CLASS net/minecraft/class_2491 net/minecraft/nbt/NbtNull
+	COMMENT Represents the NBT null value.
+	COMMENT Delimits the end of a compound object,
+	COMMENT represents nonexistent values in a compound object and the type of a list if empty.
```
haykam821:
```suggestion
	COMMENT Defines the end of an NBT compound, represents nonexistent
	COMMENT values in an NBT compound, and is the type of empty NBT lists.
```

haykam821 (changes requested):

mappings/net/minecraft/nbt/NbtList.mapping
LambdAurora: Only `NbtIntArray` seems to be missing the punctuation, just to be sure, should all `Type value:` lines ends with a comma or not?

mappings/net/minecraft/nbt/AbstractNbtList.mapping
LambdAurora: `getType` is a big no as it's already used in `NbtElement` with the same signature.

mappings/net/minecraft/nbt/StringNbtReader.mapping
```diff
@@ -16,19 +16,19 @@ CLASS net/minecraft/class_2522 net/minecraft/nbt/StringNbtReader
...
 	METHOD method_10716 readComma ()Z
-	METHOD method_10717 parseTagArray ()Lnet/minecraft/class_2520;
+	METHOD method_10717 parseArray ()Lnet/minecraft/class_2520;
 	METHOD method_10718 parse (Ljava/lang/String;)Lnet/minecraft/class_2487;
 		ARG 0 string
 	METHOD method_10719 expect (C)V
-	METHOD method_10721 readCompoundTag ()Lnet/minecraft/class_2487;
-	METHOD method_10722 parseTagPrimitive ()Lnet/minecraft/class_2520;
-	METHOD method_10723 parseTag ()Lnet/minecraft/class_2520;
+	METHOD method_10721 readObject ()Lnet/minecraft/class_2487;
```
Gegy:
```suggestion
	METHOD method_10721 readCompound ()Lnet/minecraft/class_2487;
```

mappings/net/minecraft/nbt/NbtList.mapping
haykam821: I would change them all to something like `This class uses the {@linkplain NbtElement#getType NBT type} {@code 9}.`

mappings/net/minecraft/nbt/AbstractNbtList.mapping
haykam821: Would `getListType` work instead?

```suggestion
	METHOD method_10601 getListType ()B
		COMMENT Gets the {@linkplain NbtElement#getType type} of element that this list holds.
		COMMENT
		COMMENT @return the type of element that this list holds
```

mappings/net/minecraft/nbt/NbtList.mapping
LambdAurora: That's very verbose for something as simple as this.

haykam821: I think that it is better to use full sentences in documentation, even if it means adding a few additional words.

mappings/net/minecraft/nbt/AbstractNbtList.mapping
LambdAurora: yes

mappings/net/minecraft/nbt/NbtList.mapping
LambdAurora: Erf, in this case it's the type of the element, which I expect to be found and read quickly.

Like "damn what's the type id of NbtByteArray again? ah it's there!" will be much quicker if it's something short and separated.

haykam821: The most important part is the number, which is distinguishable easily from the rest because it's in monospaced font with a different background at the end of the line. Both versions are only a single line.

Gegy (approved):

Player3324 (approved):

modmuss50 (approved):
Im happy with this, I will put this into last call on discord to ensure that everyone has enough time to voice their input.

We also think its worth backporting this (and #2037) to 1.16.5 to allow for people to adopt these changes sooner rarther than later.

i509VCB (approved):

zml2008 (approved):

enbrain (approved):

Earthcomputer:
image

liach:
Really? Imo we should leave some time for voting. The current vote is close, like 9 for and 7 against. Notice a few people also don't explicitly support or oppose this change as well (such as myself). You should probably conduct a vote in discord as well.

LambdAurora:
> Really? Imo we should leave some time for voting. The current vote is close, like 9 for and 7 against. Notice a few people also don't explicitly support or oppose this change as well (such as myself). You should probably conduct a vote in discord as well.

How do you even define voting here?

Imo, if you want to oppose these changes you should leave a review and not just leave a reaction.
It's literally the entire purpose of reviews.

Technici4n:
I am pretty sure this was decided a long time ago and is being applied only now. You can debate some individual names, but I don't think the overall refactor is debatable.

liach:
Voting here I define as the reactions left on the issue body: :+1: and :heart: are counted as supports and :-1: are :confused: are counted as oppositions, with each person only counted once.

I think I've left reviews before indicating that it's not for or against a rename but just that the rename would pass the yarn criteria. Looking at the actions I guess you've reached a consensus on discord. I just want to make sure, as I see a lot of opposition reactions on this pull request (an interesting phenomenon indeed), even though the comments here are supportive.

So yeah, besides that reaction anomaly, everything looks great.

i509VCB:
There is no project policy atm the consider reactions on the issue body. The only real determinant in this case is approval by the yarn maintainers.

liach:
> The only real determinant in this case is approval by the yarn maintainers.

Concerning!

Anyways, modmuss said he put this in last call. So if people really oppose this change, they can speak out and it'd be good enough.

jamierocks:
> they can speak out and it'd be good enough.

I oppose this change.

mappings/net/minecraft/nbt/AbstractNbtNumber.mapping
```diff
@@ -0,0 +1,32 @@
+CLASS net/minecraft/class_2514 net/minecraft/nbt/AbstractNbtNumber
+	COMMENT Represents an NBT number.
+	COMMENT <p>
+	COMMENT The type value {@code 99} is an alias for elements representing numbers.
```
haykam821: I am a bit confused about this line. Is this just the NBT type returned by `AbstractNbtNumber`, or does it have some other functionality? Please clarify this.

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -15,6 +18,9 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
 		ARG 1 key
 		ARG 2 value
 	METHOD method_10540 getType (Ljava/lang/String;)B
+		COMMENT Gets the type of the element stored at the specified key.
```
haykam821:
```suggestion
		COMMENT Gets the {@linkplain NbtElement#getType NBT type} of the element stored at the specified key.
```

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -26,6 +32,7 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
 		ARG 1 key
 		ARG 2 value
 	METHOD method_10545 contains (Ljava/lang/String;)Z
+		COMMENT Returns whether the compound object contains the specified key.
```
haykam821: This should be consistent with other documentation:

```suggestion
		COMMENT Returns whether the NBT compound object contains the specified key.
```

Is there any particular reason why 'NBT compound object' is used and not 'NBT compound'?

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -86,6 +94,9 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
 		ARG 1 key
 		ARG 2 value
 	METHOD method_10573 contains (Ljava/lang/String;I)Z
+		COMMENT Returns whether the compound object contains an element of the specified type at the specified key.
+		COMMENT
+		COMMENT @return {@code true} if the key exists and the element's type is compatible with {@code type}, else {@code false}
```
haykam821: What is compatibility? The main documentation for this method describes it as 'of the specified type', implying that the type has to exactly match.

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -86,6 +94,9 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
 		ARG 1 key
 		ARG 2 value
 	METHOD method_10573 contains (Ljava/lang/String;I)Z
+		COMMENT Returns whether the compound object contains an element of the specified type at the specified key.
```
haykam821:
```suggestion
		COMMENT Returns whether the NBT compound object contains an element of the specified type at the specified key.
```

mappings/net/minecraft/nbt/NbtList.mapping
```diff
@@ -1,4 +1,11 @@
-CLASS net/minecraft/class_2499 net/minecraft/nbt/ListTag
+CLASS net/minecraft/class_2499 net/minecraft/nbt/NbtList
+	COMMENT Represents an NBT list.
+	COMMENT <p>
+	COMMENT An NBT list holds values of the same type.
```
haykam821:
```suggestion
	COMMENT An NBT list holds values of the same {@linkplain NbtElement#getType NBT type}.
```

mappings/net/minecraft/nbt/NbtList.mapping
```diff
@@ -1,4 +1,11 @@
...
+	COMMENT An empty NBT list doesn't have an associated type,
+	COMMENT the type will be determined when the first element is inserted.
```
haykam821:
```suggestion
	COMMENT The {@linkplain AbstractNbtList#getListType NBT type} of an NBT list is determined
	COMMENT once its first element is inserted; empty NBT lists do not have an associated type.
```

If it does not have an associated type, what would `AbstractNbtList#getListType` return? I think that would also be useful to place in this documentation.

modmuss50:
I think its quite clear that there is an issue here that needs solving. There seems to be good support being this PR (look at all the people who have approved it). If someone is unhappy with the changes I urge them to comment/request changes instead of being slient now and complaining when its too late.

Providing a suggestion of how you might go about changing it, is a lot more helpful than adding a 👎. (Its also unclear at what stage of the PR this recation was added, it could have been before a change was made.) At the end of the day everyone cannot have what they want so trying to find that middle ground is a lot easier if we try work togeather. It is going through last call to ensure everyone gets a chance to provide their input.

mappings/net/minecraft/nbt/NbtOps.mapping
```diff
@@ -18,4 +18,58 @@ CLASS net/minecraft/class_2509 net/minecraft/nbt/NbtOps
...
+		ARG 1 value
+	METHOD createDouble (D)Ljava/lang/Object;
+		ARG 1 value
+	METHOD createFloat (F)Ljava/lang/Object;
+		ARG 1 value
+	METHOD createInt (I)Ljava/lang/Object;
+		ARG 1 value
+	METHOD createLong (J)Ljava/lang/Object;
+		ARG 1 value
+	METHOD createNumeric (Ljava/lang/Number;)Ljava/lang/Object;
```
haykam821, line 35: Should this match up with `AbstractNbtNumber`'s `getNumber` method?

Prospector:
^ At this time, there are 13 approving reviews, and 0 requesting changes and 0 comments. That's overwhelming support. If you have an idea for a different way to solve the problem, make sure to submit your review.

mappings/net/minecraft/nbt/NbtElement.mapping
```diff
@@ -0,0 +1,37 @@
+CLASS net/minecraft/class_2520 net/minecraft/nbt/NbtElement
+	COMMENT Represents an NBT element.
+	COMMENT <p>
+	COMMENT NBT elements are identified by a type, here's a list of the possible type values:
```
haykam821:
```suggestion
	COMMENT Each type of NBT element has an associated integer {@linkplain NbtElement#getType NBT type}. The following is a list of possible types:
```

mappings/net/minecraft/nbt/NbtElement.mapping
```diff
@@ -0,0 +1,37 @@
...
+	COMMENT <li>{@code 4} - long</li>
+	COMMENT <li>{@code 5} - float</li>
+	COMMENT <li>{@code 6} - double</li>
+	COMMENT <li>{@code 7} - byte array</li>
+	COMMENT <li>{@code 8} - string</li>
+	COMMENT <li>{@code 9} - list</li>
+	COMMENT <li>{@code 10} - compound object</li>
+	COMMENT <li>{@code 11} - int array</li>
+	COMMENT <li>{@code 12} - long array</li>
+	COMMENT <li>{@code 99} - number</li>
```
haykam821: I think it would be beneficial to document exactly what the generic number type means here as well.

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -15,6 +18,9 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
 		ARG 1 key
 		ARG 2 value
 	METHOD method_10540 getType (Ljava/lang/String;)B
+		COMMENT Gets the type of the element stored at the specified key.
+		COMMENT
+		COMMENT @return {@code 0} if the key does not exist, else the type
```
haykam821:
```suggestion
		COMMENT @return the NBT type of the element stored at the specified key, or {@code 0} if the key does not exist
```

haykam821 (changes requested):

jamierocks:
> make sure to submit your review.

The issue is the conflicting name, which I don't think is exactly a major issue - but the change that is required needs only be that name to solve that issue.

The is a huge refactor and goes beyond resolving that issue. It changes the paradigm of suffix vs prefix, and other stuff that was discussed years back.

To come up with something quick `net.minecraft.nbt.AbstractTag`.

Prospector:
The literal conflict is the smallest of the problems this solves, the much bigger one is it being confusing. No one actually refers to NBT as "tags", you just call it NBT. "The player's NBT", "The ItemStack's NBT", "Load the structure NBT".

The reason for the prefix is that the format is an adjective that describes a type (standard english word order). Not every class hierarchy needs to be suffixed with its superclass when adding the suffix would not change the meaning of the name in the slightest. This is also consistent with GSON's naming which helps greatly for intuitively writing JSON and NBT serialization.

Tl;dr: This PR replaces an obtuse notch name that no one uses colloquially *and* conflicts with another concept with one that is consistent with other serialization formats and has no conflict.

haykam821:
@jamierocks This PR has been in the works for a long time considering the many issues, pull requests, discussions, and other troubles relating to it, such as #241, #1077, #1078, #1275, #1989, and #2037. Many thanks go out to @ChloeDawn, @Prospector, @Juuxel, @LambdAurora (who has had the most success making these changes in), and the various other people who have contributed to any of these discussions.

Most of the old names were unchanged from #2 where they were added. At the time of this pull request (August 12, 2016):

* The latest full stable release was 1.10 (released June 8, 2016)
* It was only one week into the 1.11 snapshot cycle
* Actual tags would not be added until over a year later in a 1.13 snapshot (17w49a, released December 6, 2017)

The latter `net.minecraft.tag.Tag`, as many people have come to know, would later cause conflicts with `net.minecraft.nbt.Tag`. Of course, this PR is over four years old right now; over this time, the various changes to Minecraft and the overwhelming amount of struggles various inexperienced and proficient modders have come to face simply due to Yarn continuing to use `net.minecraft.nbt.Tag` have finally caught up to it.

The majority of people consider `NbtElement` to be the next step in evolution for `net.minecraft.nbt.Tag`, even if they haven't directly participated in Yarn discussions, and there is very little that can be done to change that. Simply put, there is no place for `Tag` being used to refer to NBT anymore; even a compromise that would be less of a breaking change like `AbstractTag` is completely out of the question, since that still violates one of the few reasons why `net.minecraft.nbt.Tag` has been forced to evolve: it simply does not match what the already-evolved community uses.

In the more technical sense, the renaming of `net.minecraft.nbt.Tag` is still very important. Luckily, @LambdAurora, the other NBT rename contributors I mentioned before, and all the reviewers on this PR have worked hard since the first proposal in #241 in December 2018 to make sure the change will be as complete as possible before it is merged into the snapshot, and then backported. In addition, Loom supports mappings migrations which should mean projects will be able to easily update their mappings when they are ready. Even if a breaking change along the lines of this one will be very impactful, I think that it is approved enough among the community to last us several years at the very least before even considering making a change.

LemmaEOF:
> It changes the paradigm of suffix vs prefix

I don't think it does. The prefix vs suffix thing is about what's being used as an adjective. In `GrassBlock`, "Grass" is modifying "Block", so we have "Block" as the noun and "Grass" as the adjective. In the current mappings, we use the type as an adjective on top of the `Tag` noun, but this proposes that we use `NBT` as the adjective and the type as noun. While this might seemingly not mesh with the convention for naming blocks and items and such, it's a different part of the code. The suffix rule as we know it is almost exclusively for registered objects, but NBT tag types aren't registered. They're a data storage format, and so it's a lot more clear to the end user if we phrase them in the same way JSON does - the format then the type. Consistency for consistency's sake at the cost of readability is a bad move and gets in the way of Yarn's overall goal: to make a more friendly set of mappings. This PR should go through.

LemmaEOF (approved):

mappings/net/minecraft/nbt/NbtReaders.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_4615 net/minecraft/nbt/TagReaders
+CLASS net/minecraft/class_4615 net/minecraft/nbt/NbtReaders
```
ChloeDawn: These are types, not readers. The objects are used in equality comparisons.

mappings/net/minecraft/nbt/NbtTypeReader.mapping
```diff
@@ -1,4 +1,5 @@
-CLASS net/minecraft/class_4614 net/minecraft/nbt/TagReader
+CLASS net/minecraft/class_4614 net/minecraft/nbt/NbtTypeReader
```
ChloeDawn: This is a type, not a reader. They are used in equality comparisons.

ChloeDawn (changes requested):

mappings/net/minecraft/nbt/NbtList.mapping
LambdAurora: It would return `0` iirc, as it was one of the reason for `EndTag` -> `NbtNull`

mappings/net/minecraft/nbt/NbtOps.mapping
LambdAurora: `AbstractNbtNumber#getNumber` could be renamed to `AbstractNbtNumber#getNumeric`

but NbtOps cannot be renamed as it's an overriden method.

mappings/net/minecraft/nbt/AbstractNbtNumber.mapping
liach: it's used as a parameter for getting subtags i think so you can get a number from tag of any number type

LambdAurora: It's only used for checking the existence of an element of the given type in `NbtCompound`

liach:
another side note: one piece of existing javadoc will be broken by this rename. will prob fix that with other javadoc cleanups

LambdAurora:
Expanding on @liach's side note: this creates inconsistencies with stuff like the PacketByteBuf and command argument types, would it be worth it to fix some of these in this PR or should it be done separately?

liach:
probably separately, just like the docs. i think the names in packet byte buf is already compatible with this rename.

most likely we don't need extra renames since nbt number and number nbt ain't different as the thing is both nbt and number (maybe a bit more nbt), unlike server world that's definitely not a server

LambdAurora:
> probably separately, just like the docs. i think the names in packet byte buf is already compatible with this rename.
> 
> most likely we don't need extra renames since nbt number and number nbt ain't different as the thing is both nbt and number (maybe a bit more nbt), unlike server world that's definitely not a server

In PacketByteBuf there's writeCompoundTag and readCompoundTag which conflicts quite a lot with the new names :)

liach:
Yeah, can just simplify them to `read/writeCompound` in a subsequent pr (that fixes the javadoc errors as well)

LambdAurora:
> Yeah, can just simplify them to `read/writeCompound` in a subsequent pr (that fixes the javadoc errors as well)

Was thinking about readNbt and writeNbt directly as it's the name we already use for lot of serialization stuff. But that's a debate for that other PR anyway :)

ChloeDawn (changes requested):
The "raw type" would be better described as an "id", IMO

Earthcomputer:
Some of those magic numbers are going to be a bit obsolete anyway when we get weave up and running

mappings/net/minecraft/nbt/AbstractNbtNumber.mapping
```diff
@@ -0,0 +1,33 @@
+CLASS net/minecraft/class_2514 net/minecraft/nbt/AbstractNbtNumber
+	COMMENT Represents an NBT number.
+	COMMENT <p>
+	COMMENT The type value {@code 99} is an alias for
+	COMMENT {@linkplain NbtCompound#contains checking the existence of a number element in a NBT compound object}.
```
haykam821: If `numeric` is what will be used to refer to `99`-type NBT elements:

```suggestion
	COMMENT The type value {@code 99} can be used to
	COMMENT {@linkplain NbtCompound#contains check for the existence of any numeric element in a NBT compound object}.
```

mappings/net/minecraft/nbt/AbstractNbtNumber.mapping
```diff
@@ -0,0 +1,33 @@
...
+		COMMENT @return the value as a long
+	METHOD method_10700 getFloat ()F
+		COMMENT Gets the value as a 32-bit floating-point number.
+		COMMENT
+		COMMENT @return the value as a float
+	METHOD method_10701 getInt ()I
+		COMMENT Gets the value as a 32-bit integer.
+		COMMENT
+		COMMENT @return the value as an int
+	METHOD method_10702 getNumber ()Ljava/lang/Number;
```
haykam821: Is renaming this to `getNumeric` a good choice?

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -26,6 +32,7 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
 		ARG 1 key
 		ARG 2 value
 	METHOD method_10545 contains (Ljava/lang/String;)Z
+		COMMENT Returns whether the NBT compound object contains the specified key.
```
haykam821:
```suggestion
		COMMENT Determines whether the NBT compound object contains the specified key.
		COMMENT
		COMMENT @return {@code true} if the key exists, else {@code false}
```

mappings/net/minecraft/nbt/NbtElement.mapping
```diff
@@ -0,0 +1,37 @@
...
+	COMMENT <li>{@code 4} - long</li>
+	COMMENT <li>{@code 5} - float</li>
+	COMMENT <li>{@code 6} - double</li>
+	COMMENT <li>{@code 7} - byte array</li>
+	COMMENT <li>{@code 8} - string</li>
+	COMMENT <li>{@code 9} - list</li>
+	COMMENT <li>{@code 10} - compound object</li>
+	COMMENT <li>{@code 11} - int array</li>
+	COMMENT <li>{@code 12} - long array</li>
+	COMMENT <li>{@code 99} - only used for checking the existence of a number element in a {@linkplain NbtCompound NBT compound object}.</li>
```
haykam821: This should probably be moved out of the list if it is never returned by `NbtElement#getType`.

mappings/net/minecraft/nbt/NbtList.mapping
```diff
@@ -1,7 +1,14 @@
-CLASS net/minecraft/class_2499 net/minecraft/nbt/ListTag
+CLASS net/minecraft/class_2499 net/minecraft/nbt/NbtList
+	COMMENT Represents an NBT list.
+	COMMENT <p>
+	COMMENT An NBT list holds values of the same {@linkplain NbtElement#getType NBT type}.
+	COMMENT The {@linkplain AbstractNbtList#getListType NBT type} of an NBT list is determined
+	COMMENT once its first element is inserted; empty NBT lists return {@code 0} as their held {@linkplain AbstractNbtList#getListType NBT type}.
```
haykam821: Maybe `AbstractNbtList#getListType` should be renamed to `getHeldType`, since this seems to be a better name unless anyone disagrees.

haykam821 (changes requested):

mappings/net/minecraft/nbt/AbstractNbtNumber.mapping
LambdAurora: Feels a bit weird.

liach: Maybe `getValue` or just `get`? `number.getNumber()` feels a bit weird compared to `number.getValue()`. Also we have type-specific getInt/Double etc. and this shouldn't be too confusing

haykam821: What about `getGeneric`?

mappings/net/minecraft/nbt/NbtElement.mapping
haykam821: The line was removed completely in 223dd3a6791f9954a7fc92a16108cd787f5cdfb8, but I think that it would be better if it was just an additional note at the end.

mappings/net/minecraft/nbt/AbstractNbtNumber.mapping
ChloeDawn: `value`/`numberValue`, `doubleValue`, `intValue`, etc. Akin to JDK's `Number` class

Player3324: JDK's Number is immutable, NBT's isn't.

LambdAurora: Except NBT's is, all the number type definitions (or reader before) override a method which checks if the type is immutable and return `true` for the number types.

liach: we shouldn't expect immutability. it was added in sone snapshot, and they can always change nbt behavior as they see fit.

Player3324: You are correct, it even switched to constructor init only lately. Adopting the Number conventions may be reasonable then. `numberValue` above `Value` though, don't want to encourage using it.

Player3324: liach they were semi-immutable before with a package private deserializer, now they are fully so. Doesn't look like that'd change.

liach: Agreed. *If* we decide to use `xxValue` names, we should change other value getters from `getXxx` to `xxxValue`, including `getInt` to `intValue`, `getNumber` to `numberValue`, etc.

mappings/net/minecraft/nbt/NbtList.mapping
ChloeDawn: Describing it as an element type rather than a held type would be more appropriate for a list

haykam821: That's where we started. I don't think it makes sense to name it `getElementType` since everything is an `NbtElement`.

liach: I feel like we are falling into the rabbit hole where `Element` `Entry` `Content` can all be ambiguous and cause problems

mappings/net/minecraft/nbt/NbtElement.mapping
```diff
@@ -0,0 +1,36 @@
...
+	COMMENT <li>{@code 9} - list</li>
+	COMMENT <li>{@code 10} - compound object</li>
+	COMMENT <li>{@code 11} - int array</li>
+	COMMENT <li>{@code 12} - long array</li>
+	COMMENT </ul>
+	METHOD method_10707 copy ()Lnet/minecraft/class_2520;
+		COMMENT Copies this NBT element.
+		COMMENT
+		COMMENT @return the copied element
+	METHOD method_10711 getType ()B
```
ChloeDawn, line 24:
```suggestion
	METHOD method_10711 getId ()B
```

mappings/net/minecraft/nbt/NbtElement.mapping
```diff
@@ -0,0 +1,36 @@
...
+		COMMENT Gets the type of this NBT element.
+		COMMENT
+		COMMENT @return the type
+	METHOD method_10713 write (Ljava/io/DataOutput;)V
+		ARG 1 output
+	METHOD method_10714 asString ()Ljava/lang/String;
+	METHOD method_23258 getNbtType ()Lnet/minecraft/class_4614;
```
ChloeDawn, line 31:
```suggestion
	METHOD method_23258 getType ()Lnet/minecraft/class_4614;
```

mappings/net/minecraft/nbt/NbtFloat.mapping
```diff
@@ -0,0 +1,13 @@
+CLASS net/minecraft/class_2494 net/minecraft/nbt/NbtFloat
+	COMMENT Represents an NBT 32-bit floating-point number.
+	COMMENT <p>
+	COMMENT This class uses the {@linkplain NbtElement#getType NBT type} {@code 5}.
```
ChloeDawn, line 4:
```suggestion
	COMMENT This class uses the {@linkplain NbtElement#getId NBT id} {@code 5}.
```

mappings/net/minecraft/nbt/NbtInt.mapping
```diff
@@ -1,6 +1,9 @@
-CLASS net/minecraft/class_2497 net/minecraft/nbt/IntTag
+CLASS net/minecraft/class_2497 net/minecraft/nbt/NbtInt
+	COMMENT Represents an NBT 32-bit integer.
+	COMMENT <p>
+	COMMENT This class uses the {@linkplain NbtElement#getType NBT type} {@code 3}.
```
ChloeDawn, line 4:
```suggestion
	COMMENT This class uses the {@linkplain NbtElement#getId NBT id} {@code 3}.
```

mappings/net/minecraft/nbt/NbtIntArray.mapping
```diff
@@ -0,0 +1,17 @@
+CLASS net/minecraft/class_2495 net/minecraft/nbt/NbtIntArray
+	COMMENT Represents an NBT 32-bit integer array.
+	COMMENT <p>
+	COMMENT This class uses the {@linkplain NbtElement#getType NBT type} {@code 11}.
```
ChloeDawn, line 4:
```suggestion
	COMMENT This class uses the {@linkplain NbtElement#getId NBT id} {@code 11}.
```

mappings/net/minecraft/nbt/NbtList.mapping
```diff
@@ -1,7 +1,14 @@
-CLASS net/minecraft/class_2499 net/minecraft/nbt/ListTag
+CLASS net/minecraft/class_2499 net/minecraft/nbt/NbtList
+	COMMENT Represents an NBT list.
+	COMMENT <p>
+	COMMENT An NBT list holds values of the same {@linkplain NbtElement#getType NBT type}.
+	COMMENT The {@linkplain AbstractNbtList#getHeldType NBT type} of an NBT list is determined
+	COMMENT once its first element is inserted; empty NBT lists return {@code 0} as their held {@linkplain AbstractNbtList#getHeldType NBT type}.
+	COMMENT <p>
+	COMMENT This class uses the {@linkplain NbtElement#getType NBT type} {@code 9}.
```
ChloeDawn, line 8:
```suggestion
	COMMENT This class uses the {@linkplain NbtElement#getId NBT id} {@code 9}.
```

mappings/net/minecraft/nbt/NbtLong.mapping
```diff
@@ -1,6 +1,9 @@
-CLASS net/minecraft/class_2503 net/minecraft/nbt/LongTag
+CLASS net/minecraft/class_2503 net/minecraft/nbt/NbtLong
+	COMMENT Represents an NBT 64-bit integer.
+	COMMENT <p>
+	COMMENT This class uses the {@linkplain NbtElement#getType NBT type} {@code 4}.
```
ChloeDawn, line 4:
```suggestion
	COMMENT This class uses the {@linkplain NbtElement#getId NBT id} {@code 4}.
```

mappings/net/minecraft/nbt/NbtLongArray.mapping
```diff
@@ -1,6 +1,9 @@
-CLASS net/minecraft/class_2501 net/minecraft/nbt/LongArrayTag
+CLASS net/minecraft/class_2501 net/minecraft/nbt/NbtLongArray
+	COMMENT Represents an NBT 64-bit integer array.
+	COMMENT <p>
+	COMMENT This class uses the {@linkplain NbtElement#getType NBT type} {@code 12}.
```
ChloeDawn, line 4:
```suggestion
	COMMENT This class uses the {@linkplain NbtElement#getId NBT id} {@code 12}.
```

mappings/net/minecraft/nbt/NbtNull.mapping
```diff
@@ -0,0 +1,9 @@
+CLASS net/minecraft/class_2491 net/minecraft/nbt/NbtNull
+	COMMENT Represents the NBT null value.
+	COMMENT Defines the end of an NBT compound object,
+	COMMENT represents nonexistent values in an NBT compound object,
+	COMMENT and is the type of empty NBT lists.
+	COMMENT <p>
+	COMMENT This class uses the {@linkplain NbtElement#getType NBT type} {@code 0}.
```
ChloeDawn, line 7:
```suggestion
	COMMENT This class uses the {@linkplain NbtElement#getId NBT id} {@code 0}.
```

mappings/net/minecraft/nbt/NbtShort.mapping
```diff
@@ -1,6 +1,9 @@
-CLASS net/minecraft/class_2516 net/minecraft/nbt/ShortTag
+CLASS net/minecraft/class_2516 net/minecraft/nbt/NbtShort
+	COMMENT Represents an NBT 16-bit integer.
+	COMMENT <p>
+	COMMENT This class uses the {@linkplain NbtElement#getType NBT type} {@code 2}.
```
ChloeDawn, line 4:
```suggestion
	COMMENT This class uses the {@linkplain NbtElement#getId NBT id} {@code 2}.
```

mappings/net/minecraft/nbt/NbtString.mapping
```diff
@@ -1,6 +1,9 @@
-CLASS net/minecraft/class_2519 net/minecraft/nbt/StringTag
+CLASS net/minecraft/class_2519 net/minecraft/nbt/NbtString
+	COMMENT Represents an NBT string.
+	COMMENT <p>
+	COMMENT This class uses the {@linkplain NbtElement#getType NBT type} {@code 8}.
```
ChloeDawn, line 4:
```suggestion
	COMMENT This class uses the {@linkplain NbtElement#getId NBT id} {@code 8}.
```

mappings/net/minecraft/nbt/NbtType.mapping
```diff
@@ -8,3 +9,6 @@ CLASS net/minecraft/class_4614 net/minecraft/nbt/TagReader
 		ARG 2 depth
 		ARG 3 tracker
 	METHOD method_23263 isImmutable ()Z
+		COMMENT Returns whether the NBT type is immutable or not.
```
ChloeDawn:
```suggestion
		COMMENT Determines the immutability of this type.
```

mappings/net/minecraft/nbt/NbtType.mapping
```diff
@@ -8,3 +9,6 @@ CLASS net/minecraft/class_4614 net/minecraft/nbt/TagReader
...
+		COMMENT
+		COMMENT @return {@code true} if the NBT type is immutable, else {@code false}
```
ChloeDawn:
```suggestion
		COMMENT @return {@code true} if this type is immutable
```

mappings/net/minecraft/nbt/NbtTypes.mapping
```diff
@@ -0,0 +1,7 @@
+CLASS net/minecraft/class_4615 net/minecraft/nbt/NbtTypes
+	FIELD field_21048 VALUES [Lnet/minecraft/class_4614;
+	METHOD method_23265 of (I)Lnet/minecraft/class_4614;
```
ChloeDawn:
```suggestion
	METHOD method_23265 getById (I)Lnet/minecraft/class_4614;
```
```java
NbtTypes.getById(...)
```

mappings/net/minecraft/nbt/NbtTypes.mapping
```diff
@@ -0,0 +1,7 @@
...
+		COMMENT Returns the {@linkplain NbtType NBT type} object from the type {@code id}.
```
ChloeDawn:
```suggestion
		COMMENT Gets the associated {@linkplain NbtType NBT type} for a given {@code id}.
```

mappings/net/minecraft/nbt/NbtTypes.mapping
```diff
@@ -0,0 +1,7 @@
...
+		COMMENT
+		COMMENT @return the {linkplain NbtType NBT type}
```
ChloeDawn:
```suggestion
		COMMENT @return The NBT type, or {@linkplain NbtType#createInvalid an invalid type} if the id is unknown
```

mappings/net/minecraft/nbt/NbtDouble.mapping
```diff
@@ -0,0 +1,13 @@
+CLASS net/minecraft/class_2489 net/minecraft/nbt/NbtDouble
+	COMMENT Represents an NBT 64-bit floating-point number.
+	COMMENT <p>
+	COMMENT This class uses the {@linkplain NbtElement#getType NBT type} {@code 6}.
```
ChloeDawn, line 4:
```suggestion
	COMMENT This class uses the {@linkplain NbtElement#getId NBT id} {@code 6}.
```

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -86,6 +96,11 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
 		ARG 1 key
 		ARG 2 value
 	METHOD method_10573 contains (Ljava/lang/String;I)Z
+		COMMENT Returns whether the NBT compound object contains an element of the specified type at the specified key.
+		COMMENT <p>
+		COMMENT The specified type can also be the alias type {@code 99} which is used for numbers.
+		COMMENT
+		COMMENT @return {@code true} if the key exists and the element's type is compatible with {@code type}, else {@code false}
 		ARG 1 key
 		ARG 2 type
```
ChloeDawn, line 105:
```suggestion
		ARG 2 id
```

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -86,6 +96,11 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
 		ARG 1 key
 		ARG 2 value
 	METHOD method_10573 contains (Ljava/lang/String;I)Z
+		COMMENT Returns whether the NBT compound object contains an element of the specified type at the specified key.
+		COMMENT <p>
+		COMMENT The specified type can also be the alias type {@code 99} which is used for numbers.
```
ChloeDawn:
```suggestion
```
This is not something that should be documented by general argument implementations

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -86,6 +96,11 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
...
+		COMMENT
+		COMMENT @return {@code true} if the key exists and the element's type is compatible with {@code type}, else {@code false}
```
ChloeDawn:
```suggestion
		COMMENT @return {@code true} if the key exists and the element type is equivalent to the given {@code id}
```

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -15,6 +18,9 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
 		ARG 1 key
 		ARG 2 value
 	METHOD method_10540 getType (Ljava/lang/String;)B
+		COMMENT Gets the {@linkplain NbtElement#getType NBT type} of the element stored at the specified key.
+		COMMENT
+		COMMENT @return the NBT type of the element stored at the specified key, or {@code 0} if the key does not exist
```
ChloeDawn:
```suggestion
		COMMENT @return The element type, or {@linkplain NbtNull 0} if it does not exist
```

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -1,9 +1,12 @@
-CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
-	FIELD field_11515 tags Ljava/util/Map;
-	FIELD field_21029 READER Lnet/minecraft/class_4614;
+CLASS net/minecraft/class_2487 net/minecraft/nbt/NbtCompound
+	COMMENT Represents an NBT compound object which holds unordered key-value pairs with distinct case-sensitive string keys.
+	COMMENT <p>
+	COMMENT This class uses the {@linkplain NbtElement#getType NBT type} {@code 10}.
```
ChloeDawn, line 4:
```suggestion
	COMMENT This class uses the {@linkplain NbtElement#getId NBT id} {@code 10}.
```

mappings/net/minecraft/nbt/NbtByteArray.mapping
```diff
@@ -0,0 +1,17 @@
+CLASS net/minecraft/class_2479 net/minecraft/nbt/NbtByteArray
+	COMMENT Represents an NBT byte array.
+	COMMENT <p>
+	COMMENT This class uses the {@linkplain NbtElement#getType NBT type} {@code 7}.
```
ChloeDawn, line 4:
```suggestion
	COMMENT This class uses the {@linkplain NbtElement#getId NBT id} {@code 7}.
```

mappings/net/minecraft/nbt/NbtByte.mapping
```diff
@@ -1,6 +1,9 @@
-CLASS net/minecraft/class_2481 net/minecraft/nbt/ByteTag
+CLASS net/minecraft/class_2481 net/minecraft/nbt/NbtByte
+	COMMENT Represents an NBT byte.
+	COMMENT <p>
+	COMMENT This class uses the {@linkplain NbtElement#getType NBT type} {@code 1}.
```
ChloeDawn, line 4:
```suggestion
	COMMENT This class uses the {@linkplain NbtElement#getId NBT id} {@code 1}.
```

mappings/net/minecraft/nbt/AbstractNbtNumber.mapping
```diff
@@ -0,0 +1,33 @@
+CLASS net/minecraft/class_2514 net/minecraft/nbt/AbstractNbtNumber
+	COMMENT Represents an NBT number.
+	COMMENT <p>
+	COMMENT The type value {@code 99} can be used to
+	COMMENT {@linkplain NbtCompound#contains check for the existence of any numeric element in a NBT compound object}.
```
ChloeDawn, line 5:
```suggestion
```

mappings/net/minecraft/nbt/AbstractNbtNumber.mapping
```diff
@@ -0,0 +1,33 @@
+CLASS net/minecraft/class_2514 net/minecraft/nbt/AbstractNbtNumber
+	COMMENT Represents an NBT number.
+	COMMENT <p>
+	COMMENT The type value {@code 99} can be used to
```
ChloeDawn, line 4:
```suggestion
	COMMENT Can be represented as {@linkplain NbtElement#getId NBT id} {@code 99} to alias all number types
```

mappings/net/minecraft/nbt/AbstractNbtList.mapping
```diff
@@ -0,0 +1,20 @@
...
+	METHOD add (ILjava/lang/Object;)V
+		ARG 1 index
+		ARG 2 value
+	METHOD method_10533 addElement (ILnet/minecraft/class_2520;)Z
+		ARG 1 index
+		ARG 2 element
+	METHOD method_10535 setElement (ILnet/minecraft/class_2520;)Z
+		ARG 1 index
+		ARG 2 element
+	METHOD method_10601 getHeldType ()B
```
ChloeDawn, line 12:
```suggestion
	METHOD method_10601 getElementType ()B
```
The name `getElementType` reads better than `getElementId`. The javadoc describes the specific return type.

mappings/net/minecraft/nbt/AbstractNbtList.mapping
```diff
@@ -0,0 +1,20 @@
...
+		COMMENT Gets the {@linkplain NbtElement#getType type} of element that this list holds.
```
ChloeDawn, line 13:
```suggestion
		COMMENT Gets the {@linkplain NbtElement#getId id} for this list's elements
```

mappings/net/minecraft/nbt/AbstractNbtList.mapping
```diff
@@ -0,0 +1,20 @@
...
+		COMMENT
+		COMMENT @return the type of element that this list holds
```
ChloeDawn, line 15:
```suggestion
		COMMENT @return the element type id
```

ChloeDawn (changes requested):
I feel that the id descriptions in each class header could be omitted. That kind of information can be documented elsewhere (e.g. in a constants class).

mappings/net/minecraft/nbt/NbtTypes.mapping
ChloeDawn: This Javadoc should also explain the following
```
COMMENT This method does not support id aliases
```

LambdAurora:
> I feel that the id descriptions in each class header could be omitted. That kind of information can be documented elsewhere (e.g. in a constants class).

imo, I prefer the id descriptions to be kept in each class header, if I'm looking at the documentation of NbtInt I would like to be able to see the id of NbtInt directly than trying to search in which class the id list is.

mappings/net/minecraft/nbt/AbstractNbtNumber.mapping
haykam821: This leaves a sentence fragment

mappings/net/minecraft/nbt/AbstractNbtList.mapping
haykam821: Element type is indistinguishable from the actual element type, which is `9` for lists. There should be a clear difference between the list itself and its contents.

ChloeDawn: Given two methods `NbtList#getType` and `NbtList#getElementType` I feel it is clear which is which (especially given `getType` would return `NbtType` and `getElementType` would return `byte`). Documentation then aids further on distinguishing them.

FoundationGames:
Does this PR rename methods/fields in `ItemStack`, `BlockEntity`, or `PersistentState` related to NBT?
If not it might get confusing when NBT is named NBT while `ItemStack`s have methods like `putSubTag`

LambdAurora:
> Does this PR rename methods/fields in `ItemStack`, `BlockEntity`, or `PersistentState` related to NBT?
> If not it might get confusing when NBT is named NBT while `ItemStack`s have methods like `putSubTag`

BlockEntity and PresistentState should have been covered in a previous PR (or at least partially)

ItemStack is a special case as the tag stuff is in the `tag` NBT compound object, so there's no actual need to rename those
only the param names should be renamed so `tag -> nbt`

liach:
I quite disagree with chloedawn's suggestions. Id should be unique to each object as it's an identifier. That's how it works everywhere in registries, and this `getId` names are just wrong. It would be at least `getTypeId`; and promoting this Nbt reader thing as type is not helpful either as it has really few functionalities (the only clue on how it's a type is that it's used in equality comparison; it is mainly a reader)

mappings/net/minecraft/nbt/NbtType.mapping
haykam821: It would also be beneficial to define exactly what immutability means for NBT types, and if it's just the Java definition of immutability.

mappings/net/minecraft/nbt/NbtTypes.mapping
haykam821: If using `id`:

```suggestion
	METHOD method_23265 byId (I)Lnet/minecraft/class_4614;
```

Personally I don't think `id` should be used, but `byId` should be used over `getById`.

mappings/net/minecraft/nbt/NbtTypes.mapping
haykam821: The first character of return value documentation should be lowercase:

```suggestion
		COMMENT @return the NBT type, or {@linkplain NbtType#createInvalid an invalid type} if the id is unknown
```

RedstoneParadox (approved):

liach:
@LambdAurora Mind fix the merge conflict?

LambdAurora:
@liach fixed

LambdAurora:
Is there any other opinions on @ChloeDawn suggestions?

As for now 2 people expressed disagreement. Would like to know what I should do about these now.

Earthcomputer:
I also disagree with ChloeDawn's suggestions. But I am too lazy to thumbs down every single one xD. I think unless anyone else expresses support for them, we should go ahead without them.

mappings/net/minecraft/nbt/NbtCompound.mapping
```diff
@@ -86,6 +96,11 @@ CLASS net/minecraft/class_2487 net/minecraft/nbt/CompoundTag
 		ARG 1 key
 		ARG 2 value
 	METHOD method_10573 contains (Ljava/lang/String;I)Z
+		COMMENT Returns whether the NBT compound object contains an element of the specified type at the specified key.
+		COMMENT <p>
+		COMMENT The specified type can also be the alias type id {@code 99} which is used for numbers.
```
haykam821:
```suggestion
		COMMENT The type restriction can also be {@code 99}, which only allows any type of number.
```

mappings/net/minecraft/nbt/NbtTypes.mapping
```diff
@@ -0,0 +1,9 @@
+CLASS net/minecraft/class_4615 net/minecraft/nbt/NbtTypes
+	FIELD field_21048 VALUES [Lnet/minecraft/class_4614;
+	METHOD method_23265 byId (I)Lnet/minecraft/class_4614;
+		COMMENT Gets the associated {@linkplain NbtType NBT type} for a given {@code id}.
+		COMMENT <p>
+		COMMENT This method does not support id aliases.
+		COMMENT
+		COMMENT @return the NBT type, or {@linkplain NbtType#createInvalid an invalid type} if the id is unknown
```
haykam821:
```suggestion
		COMMENT @return the NBT type, or {@linkplain NbtType#createInvalid an invalid type} if there is no type with the given {@code id}
```

haykam821 (changes requested):

haykam821 (approved):

modmuss50 (approved):
Still looks good, I agree that type is the correct way to go here. I will merge this in this evening and backport to 1.16.5 in a few days.

modmuss50:
🚀

Earthcomputer:
The bikes have been put away in the shed, now time for the party!

PR #2094, opened 2021-02-22
Make named jar use jetbrain annotations

### LINKED ISSUE #16: Annotation replacement
https://github.com/FabricMC/fabric-loom/blob/57c9a8f3209121f22e84b8b01b64f1093256f1d3/src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftMappedProvider.java#L46-L50

Should be easy to do. Also avoids tricky bad javadoc problems from jsr305 annotations

liach:
and subsequent javadoc tweaks

Closes FabricMC/MappingPoet#16

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -64,7 +64,7 @@ dependencies {
 	enigmaRuntime "cuchaz:enigma-swing:${project.enigma_version}"
 	enigmaRuntime "net.fabricmc:stitch:${project.stitch_version}"
 	javadocClasspath "net.fabricmc:fabric-loader:${project.fabric_loader_version}"
-	javadocClasspath "com.google.code.findbugs:jsr305:3.0.2"
+	javadocClasspath "org.jetbrains:annotations:${project.jetbrains_annotations_version}"
 	decompileClasspath "org.benf:cfr:0.150"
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -817,16 +828,21 @@ publishing {
 	}
 }
 
-
 void mapJar(File output, File input, File mappings, File libraries, String from, String to) {
+	mapJar(output, input, mappings, libraries, from, to, { })
+}
+
+void mapJar(File output, File input, File mappings, File libraries, String from, String to, Action<TinyRemapper.Builder> action) {
```
modmuss50: I think you can just do:

```groovy
void mapJar(File output, File input, File mappings, File libraries, String from, String to, Action<TinyRemapper.Builder> action = { }) {
```

modmuss50 (approved):
Looks fine, just that minor nitpick

build.gradle
liach: hmm, groovy has default args? let me try

liach: It works perfectly! wrapped the line for the action arg because otherwise the line feels too long

modmuss50: Yeah, groovy has a lot of little tricks like this.

modmuss50 (approved):

PR #2109, opened 2021-02-28
RegistryKey -> TypedKey

### LINKED ISSUE #1674: RegistryKey -> TypedIdentifier
Though this key is pretty much designed around registry, much of the key now is used for generic type-identification purpose (like for worlds), and `TypedIdentifier<Registry<T>>` is easier to understand than `RegistryKey<Registry<T>>`, or `TypedIdentifier<DimensionType>` vs `RegistryKey<DimensionType>`

liach:
A high impact rename, but this does clarify some things. Fixes #1674


Juuxel:
`TypedKey` seems really generic to me. Keys in a `Map<K, V>` are typed - is `K` a typed key as well? This needs to have a more specific name.

Earthcomputer:
`TypedIdentifier` is better, we discussed this before and the issue you linked to even suggests `TypedIdentifier` specifically instead of `TypedKey`.

liach:
Just need a quick poll, how many people support renaming this to `TypedIdentifier`? https://www.strawpoll.me/42715091

liach:
So as of comment in the strawpoll it's TypedIdentifier 3 vs RegistryKey 1. Will do the rename when I get some time

jaskarth:
I'm in favor of leaving this as `RegistryKey` as it's clear what it is from the name, the use, and the javadoc, and this would impact many mods.

liach:
a major reason of rename is that registry key cannot distinguish if a key is from or identifying a registry. while we can say type vs typed identifier. @SuperCoder7979 any good solution for this confusion?

Earthcomputer:
The main reason for using `TypedIdentifier` over `RegistryKey` is their `RegistryKey<World>`. We do not have a World registry.

PR #2113, opened 2021-03-01
Improve github actions for updating the base of a PR

modmuss50:
Opend a PR to try and see if there is any more input on this, I have tested it and it works as expected.

Now runs on pull_request_target, should save having to loop all prs every 20 mins.
Now also atempts to merge the changes into the PR, will add a comment + the outdated label when this fails due to merge conflicts.

I might look at moving these scripts out into their own repo as its a little gross having JS in yml, and having typescript would aid a lot when writing them (as the test feedback is really slow).

Do note this wont work on older PRs where the existing target is something that doesnt have this workflow. I dont think its a major issue as it will be much better going forward.

### DIFF
--- a/.github/workflows/update-base.yml
+++ b/.github/workflows/update-base.yml
@@ -1,13 +1,12 @@
 name: Update Base Branch
 
-# Run every 20 mins or manually
 on:
-  schedule:
-    - cron: '*/20 * * * *'
-  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


enbrain (approved):

modmuss50:
Going to merge this in now, can always improve it later. 👍

PR #2116, opened 2021-03-04
21w08b worldgen mappings

jaskarth:
This PR maps float providers, carvers, and some other misc stuff. It also Refactors some names in biome layers and features to make them more correct, and remaps some noise cave and aquifer sampler names to make them more descriptive.

### DIFF
--- a/mappings/net/minecraft/structure/rule/RuleTest.mapping
+++ b/mappings/net/minecraft/structure/rule/RuleTest.mapping
@@ -1,5 +1,6 @@
 CLASS net/minecraft/class_3825 net/minecraft/structure/rule/RuleTest
 	COMMENT Rule tests are used in structure generation to check if a block state matches some condition.
+	FIELD field_25012 TYPE_CODEC Lcom/mojang/serialization/Codec;
 	METHOD method_16766 getType ()Lnet/minecraft/class_3827;
 	METHOD method_16768 test (Lnet/minecraft/class_2680;Ljava/util/Random;)Z
 		ARG 1 state
--- a/mappings/net/minecraft/util/math/ChunkPos.mapping
+++ b/mappings/net/minecraft/util/math/ChunkPos.mapping
@@ -23,6 +23,14 @@ CLASS net/minecraft/class_1923 net/minecraft/util/math/ChunkPos
 		ARG 1 pos2
 	METHOD method_24022 getChebyshevDistance (Lnet/minecraft/class_1923;)I
 		ARG 1 pos
+	METHOD method_33939 getOffsetX (I)I
+		ARG 1 offset
+	METHOD method_33940 getCenterX ()I
+	METHOD method_33941 getOffsetZ (I)I
+		ARG 1 offset
+	METHOD method_33942 getCenterZ ()I
+	METHOD method_33943 getCenterWithY (I)Lnet/minecraft/class_2338;
+		ARG 1 y
 	METHOD method_8323 getStartPos ()Lnet/minecraft/class_2338;
 	METHOD method_8324 toLong ()J
 	METHOD method_8325 getPackedX (J)I
--- a/mappings/net/minecraft/util/math/ChunkPos.mapping
+++ b/mappings/net/minecraft/util/math/ChunkPos.mapping
@@ -29,7 +29,7 @@ CLASS net/minecraft/class_1923 net/minecraft/util/math/ChunkPos
 	METHOD method_33941 getOffsetZ (I)I
 		ARG 1 offset
 	METHOD method_33942 getCenterZ ()I
-	METHOD method_33943 getCenterWithY (I)Lnet/minecraft/class_2338;
+	METHOD method_33943 getCenterAtY (I)Lnet/minecraft/class_2338;
 		ARG 1 y
 	METHOD method_8323 getStartPos ()Lnet/minecraft/class_2338;
 	METHOD method_8324 toLong ()J
--- a/mappings/net/minecraft/util/math/floatprovider/ClampedNormalFloatProvider.mapping
+++ b/mappings/net/minecraft/util/math/floatprovider/ClampedNormalFloatProvider.mapping
@@ -1,0 +1,24 @@
+CLASS net/minecraft/class_5861 net/minecraft/util/math/floatprovider/ClampedNormalFloatProvider
+	FIELD field_28998 CODEC Lcom/mojang/serialization/Codec;
+	FIELD field_28999 mean F
+	FIELD field_29000 deviation F
+	FIELD field_29001 min F
+	FIELD field_29002 max F
+	METHOD <init> (FFFF)V
+		ARG 1 mean
+		ARG 2 deviation
+		ARG 3 min
+		ARG 4 max
+	METHOD equals (Ljava/lang/Object;)Z
+		ARG 1 object
+	METHOD method_33900 create (FFFF)Lnet/minecraft/class_5861;
+		ARG 0 mean
+		ARG 1 deviation
+		ARG 2 min
+		ARG 3 max
+	METHOD method_33903 get (Ljava/util/Random;FFFF)F
+		ARG 0 random
+		ARG 1 mean
+		ARG 2 deviation
+		ARG 3 min
+		ARG 4 max
--- a/mappings/net/minecraft/util/math/floatprovider/ConstantFloatProvider.mapping
+++ b/mappings/net/minecraft/util/math/floatprovider/ConstantFloatProvider.mapping
@@ -1,5 +1,5 @@
 CLASS net/minecraft/class_5862 net/minecraft/util/math/floatprovider/ConstantFloatProvider
-	FIELD field_29003 DEFAULT Lnet/minecraft/class_5862;
+	FIELD field_29003 ZERO Lnet/minecraft/class_5862;
 	FIELD field_29004 CODEC Lcom/mojang/serialization/Codec;
 	FIELD field_29005 value F
 	METHOD <init> (F)V
--- a/mappings/net/minecraft/util/math/floatprovider/ConstantFloatProvider.mapping
+++ b/mappings/net/minecraft/util/math/floatprovider/ConstantFloatProvider.mapping
@@ -1,0 +1,11 @@
+CLASS net/minecraft/class_5862 net/minecraft/util/math/floatprovider/ConstantFloatProvider
+	FIELD field_29003 DEFAULT Lnet/minecraft/class_5862;
+	FIELD field_29004 CODEC Lcom/mojang/serialization/Codec;
+	FIELD field_29005 value F
+	METHOD <init> (F)V
+		ARG 1 value
+	METHOD equals (Ljava/lang/Object;)Z
+		ARG 1 object
+	METHOD method_33908 create (F)Lnet/minecraft/class_5862;
+		ARG 0 value
+	METHOD method_33914 getValue ()F
--- a/mappings/net/minecraft/util/math/floatprovider/UniformFloatProvider.mapping
+++ b/mappings/net/minecraft/util/math/floatprovider/UniformFloatProvider.mapping
@@ -1,0 +1,12 @@
+CLASS net/minecraft/class_5866 net/minecraft/util/math/floatprovider/UniformFloatProvider
+	FIELD field_29016 CODEC Lcom/mojang/serialization/Codec;
+	FIELD field_29017 base F
+	FIELD field_29018 spread F
+	METHOD <init> (FF)V
+		ARG 1 base
+		ARG 2 sprea
+	METHOD equals (Ljava/lang/Object;)Z
+		ARG 1 object
+	METHOD method_33934 create (FF)Lnet/minecraft/class_5866;
+		ARG 0 base
+		ARG 1 spread
--- a/mappings/net/minecraft/util/math/floatprovider/UniformFloatProvider.mapping
+++ b/mappings/net/minecraft/util/math/floatprovider/UniformFloatProvider.mapping
@@ -4,7 +4,7 @@ CLASS net/minecraft/class_5866 net/minecraft/util/math/floatprovider/UniformFloa
 	FIELD field_29018 spread F
 	METHOD <init> (FF)V
 		ARG 1 base
-		ARG 2 sprea
+		ARG 2 spread
 	METHOD equals (Ljava/lang/Object;)Z
 		ARG 1 object
 	METHOD method_33934 create (FF)Lnet/minecraft/class_5866;
--- a/mappings/net/minecraft/util/registry/Registry.mapping
+++ b/mappings/net/minecraft/util/registry/Registry.mapping
@@ -112,6 +112,7 @@ CLASS net/minecraft/class_2378 net/minecraft/util/registry/Registry
 	FIELD field_28264 GAME_EVENT Lnet/minecraft/class_2348;
 	FIELD field_28265 POSITION_SOURCE_TYPE Lnet/minecraft/class_2378;
 	FIELD field_28266 GAME_EVENT_KEY Lnet/minecraft/class_5321;
+	FIELD field_29076 FLOAT_PROVIDER Lnet/minecraft/class_2378;
 	METHOD <init> (Lnet/minecraft/class_5321;Lcom/mojang/serialization/Lifecycle;)V
 		ARG 1 key
 		ARG 2 lifecycle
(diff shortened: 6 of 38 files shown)


mappings/net/minecraft/util/registry/Registry.mapping
```diff
@@ -112,6 +112,7 @@ CLASS net/minecraft/class_2378 net/minecraft/util/registry/Registry
 	FIELD field_28264 GAME_EVENT Lnet/minecraft/class_2348;
 	FIELD field_28265 POSITION_SOURCE_TYPE Lnet/minecraft/class_2378;
 	FIELD field_28266 GAME_EVENT_KEY Lnet/minecraft/class_5321;
+	FIELD field_29076 FLOAT_PROVIDER Lnet/minecraft/class_2378;
```
YanisBft:
```suggestion
	FIELD field_29076 FLOAT_PROVIDER_TYPE Lnet/minecraft/class_2378;
```

mappings/net/minecraft/util/math/floatprovider/UniformFloatProvider.mapping
```diff
@@ -0,0 +1,12 @@
+CLASS net/minecraft/class_5866 net/minecraft/util/math/floatprovider/UniformFloatProvider
+	FIELD field_29016 CODEC Lcom/mojang/serialization/Codec;
+	FIELD field_29017 base F
+	FIELD field_29018 spread F
+	METHOD <init> (FF)V
+		ARG 1 base
+		ARG 2 sprea
```
YanisBft:
```suggestion
		ARG 2 spread
```

mappings/net/minecraft/util/math/floatprovider/ClampedNormalFloatProvider.mapping
```diff
@@ -0,0 +1,24 @@
+CLASS net/minecraft/class_5861 net/minecraft/util/math/floatprovider/ClampedNormalFloatProvider
```
FoundationGames, line 1: `floatprovider` seems awfully verbose as a package name (and also as a class name to a small extent, although I'm mostly okay with it), and judging by the content of these classes I'd prefer something along the lines of `distribution` for these

mappings/net/minecraft/world/gen/treedecorator/AlterGroundTreeDecorator.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_4658 net/minecraft/world/gen/tree/AlterGroundTreeDecorator
+CLASS net/minecraft/class_4658 net/minecraft/world/gen/treedecorator/AlterGroundTreeDecorator
```
FoundationGames, line 1: `treedecorator` also seems a bit verbose, what was wrong with `tree`?

jaskarth: This was the only thing in "tree" so it felt wrong calling it that when it was all tree decorators

mappings/net/minecraft/util/math/floatprovider/ClampedNormalFloatProvider.mapping
jaskarth: I named the package this way since it's similar tree decorators, foliage placers, and trunk placers. `distribution` may work but it's similar size to `floatprovider` and since it's only float providers, we might as well be more specific

mappings/net/minecraft/world/gen/treedecorator/AlterGroundTreeDecorator.mapping
YanisBft, line 1: Oof it hurts my eyes, but it's technically more consistent

YanisBft (commented):
Even though the float provider is a worldgen registry as specified by the deobfuscated name, I think it's good to put in `util/math`

mappings/net/minecraft/util/math/floatprovider/ClampedNormalFloatProvider.mapping
FoundationGames: I see your point.
I'd also like to bring up that "float provider" is a bit of a general term, and that "distribution" (which may not being the right term) seems a bit more specific and more accurately describes what these classes are used for.
However since it seems like these "float providers" can provide a float by any logic (with a Random as input) I get your reasons for naming it this way. My only issue is how people will interpret the name of this class. I think i'll need another opinion on this.

jaskarth: I named it "float provider" as that's what the registry string calls them, to avoid confusion. They could definitely have better names but I feel it's out of our hands :P

mappings/net/minecraft/world/gen/treedecorator/AlterGroundTreeDecorator.mapping
jaskarth: Yeah, it's not the most elegant solution but keeping them in the "tree" namespace was more confusing, at least to me

YanisBft (approved):

mappings/net/minecraft/structure/rule/RuleTest.mapping
```diff
@@ -1,5 +1,6 @@
 CLASS net/minecraft/class_3825 net/minecraft/structure/rule/RuleTest
 	COMMENT Rule tests are used in structure generation to check if a block state matches some condition.
+	FIELD field_25012 TYPE_CODEC Lcom/mojang/serialization/Codec;
```
liach, line 3: Is this actually like `Registry.RULE_TEST.dispatch(xx)`? This is a codec for a whole rule test, as that dispatch assigns the type's codec as decoder in decoding branch (i.e. read "type" field for the type, and get the right decoder for the contents), and has a method ref to `getType` to obtain the type's codec for encoding on the other hand.

Just curious, what is the origin of the `TYPE_CODEC` names? Meaning it's a codec for the type only?

mappings/net/minecraft/util/math/ChunkPos.mapping
```diff
@@ -23,6 +23,14 @@ CLASS net/minecraft/class_1923 net/minecraft/util/math/ChunkPos
 		ARG 1 pos2
 	METHOD method_24022 getChebyshevDistance (Lnet/minecraft/class_1923;)I
 		ARG 1 pos
+	METHOD method_33939 getOffsetX (I)I
```
liach, line 26: Is this converting a local x in a chunk to a global x? I suggest `offsetX()` if that's the case.

mappings/net/minecraft/util/math/ChunkPos.mapping
```diff
@@ -23,6 +23,14 @@ CLASS net/minecraft/class_1923 net/minecraft/util/math/ChunkPos
...
+		ARG 1 offset
+	METHOD method_33940 getCenterX ()I
+	METHOD method_33941 getOffsetZ (I)I
+		ARG 1 offset
+	METHOD method_33942 getCenterZ ()I
+	METHOD method_33943 getCenterWithY (I)Lnet/minecraft/class_2338;
```
liach:
```suggestion
	METHOD method_33943 getCenterAtY (I)Lnet/minecraft/class_2338;
```

mappings/net/minecraft/util/registry/Registry.mapping
```diff
@@ -112,6 +112,7 @@ CLASS net/minecraft/class_2378 net/minecraft/util/registry/Registry
 	FIELD field_28264 GAME_EVENT Lnet/minecraft/class_2348;
 	FIELD field_28265 POSITION_SOURCE_TYPE Lnet/minecraft/class_2378;
 	FIELD field_28266 GAME_EVENT_KEY Lnet/minecraft/class_5321;
+	FIELD field_29076 FLOAT_PROVIDER_TYPE Lnet/minecraft/class_2378;
```
liach, line 115: Imo if we have these float providers in registries, I suggest raising the floatprovider pacakge out of math to just util

mappings/net/minecraft/world/gen/treedecorator/AlterGroundTreeDecorator.mapping
liach: Is it possible to put it in like `decorator.tree` or is this actually unrelated to decorators

mappings/net/minecraft/util/math/floatprovider/ConstantFloatProvider.mapping
```diff
@@ -0,0 +1,11 @@
+CLASS net/minecraft/class_5862 net/minecraft/util/math/floatprovider/ConstantFloatProvider
+	FIELD field_29003 DEFAULT Lnet/minecraft/class_5862;
```
enbrain:
```suggestion
	FIELD field_29003 ZERO Lnet/minecraft/class_5862;
```
Any reason `DEFAULT`? This field seems to be used only for avoiding duplicate object creation.

mappings/net/minecraft/world/gen/feature/OreFeature.mapping
```diff
@@ -14,3 +14,13 @@ CLASS net/minecraft/class_3122 net/minecraft/world/gen/feature/OreFeature
 		ARG 18 z
 		ARG 19 horizontalSize
 		ARG 20 verticalSize
+	METHOD method_33983 shouldPlace (Lnet/minecraft/class_2680;Ljava/util/function/Function;Ljava/util/Random;Lnet/minecraft/class_3124;Lnet/minecraft/class_3124$class_5876;Lnet/minecraft/class_2338$class_2339;)Z
+		ARG 0 state
+		ARG 1 posToStateFunction
```
enbrain:
```suggestion
		ARG 1 posToState
```
I think `function` is redundant as it's obvious from their type.

mappings/net/minecraft/world/gen/feature/Feature.mapping
```diff
@@ -32,3 +32,10 @@ CLASS net/minecraft/class_3031 net/minecraft/world/gen/feature/Feature
 		ARG 0 feature
 	METHOD method_28629 (Lnet/minecraft/class_3037;)Lnet/minecraft/class_2975;
 		ARG 1 config
+	METHOD method_33981 isAir (Ljava/util/function/Function;Lnet/minecraft/class_2338;)Z
+		ARG 0 posToStateFunction
```
enbrain:
```suggestion
		ARG 0 posToState
```

mappings/net/minecraft/world/gen/feature/Feature.mapping
```diff
@@ -32,3 +32,10 @@ CLASS net/minecraft/class_3031 net/minecraft/world/gen/feature/Feature
...
+		ARG 1 pos
+	METHOD method_33982 testState (Ljava/util/function/Function;Lnet/minecraft/class_2338;Ljava/util/function/Predicate;)Z
+		ARG 0 posToStateFunction
```
enbrain:
```suggestion
		ARG 0 posToState
```

mappings/net/minecraft/structure/rule/RuleTest.mapping
jaskarth: Yeah, that is usually what it means. I've also co-opted it to mean the codec for the base type, so conflicts with subclasses are avoided

mappings/net/minecraft/world/gen/treedecorator/AlterGroundTreeDecorator.mapping
jaskarth: They are unrelated to normal decorators. Normal decorators get positions for features to generate, but tree decorators add vines/cocoa/podzol/extra stuff to trees

mappings/net/minecraft/util/registry/Registry.mapping
jaskarth: I put it here because it is used directly in math, but I can do that too. Would like a second opinion

mappings/net/minecraft/util/math/ChunkPos.mapping
jaskarth: It's adding a value on the x and z to the block coords of the chunk's start, everything else uses `get` so I think this should too

mappings/net/minecraft/structure/rule/RuleTest.mapping
liach: should we call it BASE_CODEC then?

mappings/net/minecraft/util/math/ChunkPos.mapping
liach: how about calling this get global x/z and treat inputs as local x/zs? imo global is clearer than offset

mappings/net/minecraft/world/gen/feature/Feature.mapping
```diff
@@ -32,3 +32,10 @@ CLASS net/minecraft/class_3031 net/minecraft/world/gen/feature/Feature
 		ARG 0 feature
 	METHOD method_28629 (Lnet/minecraft/class_3037;)Lnet/minecraft/class_2975;
 		ARG 1 config
+	METHOD method_33981 isAir (Ljava/util/function/Function;Lnet/minecraft/class_2338;)Z
```
enbrain: `isAir` could be misleading since this method doesn't check if the block at `pos` is air. It returns true if any of the adjacent blocks is air.

mappings/net/minecraft/world/gen/feature/Feature.mapping
```diff
@@ -32,3 +32,10 @@ CLASS net/minecraft/class_3031 net/minecraft/world/gen/feature/Feature
...
+		ARG 0 posToStateFunction
+		ARG 1 pos
+	METHOD method_33982 testState (Ljava/util/function/Function;Lnet/minecraft/class_2338;Ljava/util/function/Predicate;)Z
```
enbrain:
```suggestion
	METHOD method_33982 testAdjacentStates (Ljava/util/function/Function;Lnet/minecraft/class_2338;Ljava/util/function/Predicate;)Z
```

mappings/net/minecraft/world/gen/feature/Feature.mapping
liach: So like `isExposedToAir` or `isAdjacentToAir`.

jaskarth: `isExposedToAir` is better because that's the wording that Mojang used in their article, will do that

mappings/net/minecraft/util/math/floatprovider/ConstantFloatProvider.mapping
jaskarth: Whoops. Thanks!

mappings/net/minecraft/structure/rule/RuleTest.mapping
jaskarth: Probably a good idea, yeah. We should put codec guidelines into the javadoc conventions I think, to standardize this

liach: Guess we will do that in a later pr. This looks good enough.

mappings/net/minecraft/util/math/ChunkPos.mapping
liach: After looking at actual usages, this name looks good.

mappings/net/minecraft/world/gen/feature/Feature.mapping
liach: Everything looks great. Just have this one updated.

liach:
```suggestion
	METHOD method_33981 isExposedToAir (Ljava/util/function/Function;Lnet/minecraft/class_2338;)Z
```

mappings/net/minecraft/world/gen/treedecorator/AlterGroundTreeDecorator.mapping
FoundationGames: Should they even be called `decorator`s then? Since feature placement decorators and what this appears to be seem to be two entirely different things

liach: Think this name comes from mojang registry

jaskarth:
I think I fixed everything. Would like re-reviews please!

enbrain (approved):

liach (approved):

PR #2120, opened 2021-03-08
Add a mapping for nether vine growing

Legend-of-iPhoenix:
Title says it all, this was a fairly obvious mapping that didn't have a name yet. The function itself just computes a random length for the nether vine growing algorithm.

### DIFF
--- a/mappings/net/minecraft/block/VineLogic.mapping
+++ b/mappings/net/minecraft/block/VineLogic.mapping
@@ -2,3 +2,4 @@ CLASS net/minecraft/class_4949 net/minecraft/block/VineLogic
 	METHOD method_25961 isValidForWeepingStem (Lnet/minecraft/class_2680;)Z
 		ARG 0 state
 	METHOD method_26381 getGrowthLength (Ljava/util/Random;)I
+		ARG 0 random
--- a/mappings/net/minecraft/block/VineLogic.mapping
+++ b/mappings/net/minecraft/block/VineLogic.mapping
@@ -1,3 +1,4 @@
 CLASS net/minecraft/class_4949 net/minecraft/block/VineLogic
 	METHOD method_25961 isValidForWeepingStem (Lnet/minecraft/class_2680;)Z
 		ARG 0 state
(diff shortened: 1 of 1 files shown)


Legend-of-iPhoenix:
I'm not sure if it makes more sense here to specify that it's a *random* growth length, but that's kinda redundant seeing as it takes a `Random`.

mappings/net/minecraft/block/VineLogic.mapping
```diff
@@ -1,3 +1,4 @@
 CLASS net/minecraft/class_4949 net/minecraft/block/VineLogic
 	METHOD method_25961 isValidForWeepingStem (Lnet/minecraft/class_2680;)Z
 		ARG 0 state
+	METHOD method_26381 getGrowthLength (Ljava/util/Random;)I
```
haykam821, line 4: The name for the method itself is good, but make sure to map the parameter to `random` as well.

enbrain (approved):

haykam821 (changes requested):
I would also map `method_26376` to the same name as well since that just calls `method_26381`.

I think that `AbstractPlantStemBlock`'s methods will have to be looked into more in a later PR, but other than the additional method above, I think this PR's scope is good.

Legend-of-iPhoenix:
> The name for the method itself is good, but make sure to map the parameter to `random` as well.

Oh, thanks- didn't know you had to do that too in this case :) Fixed in that commit ^
> I would also map `method_26376` to the same name as well since that just calls `method_26381`.

Is that not mapped already, per 
https://github.com/FabricMC/yarn/blob/69c470b145d3c0ef4c1d6a2b327760d5586d7fcf/mappings/net/minecraft/block/AbstractPlantStemBlock.mapping#L12-L13?

haykam821:
Oh, it looks like it is mapped already, but only on 1.17 snapshot branches.

haykam821 (approved):

jaskarth (approved):

liach (approved):

YanisBft (approved):

PR #2126, opened 2021-03-10
Add missing methods for util math from 10a update

liach:

### DIFF
--- a/mappings/net/minecraft/util/math/BlockBox.mapping
+++ b/mappings/net/minecraft/util/math/BlockBox.mapping
@@ -1,10 +1,22 @@
 CLASS net/minecraft/class_3341 net/minecraft/util/math/BlockBox
+	COMMENT A mutable box with integer coordinates. The box is axis-aligned and the
+	COMMENT coordinates are inclusive.
+	COMMENT
+	COMMENT <p>This box, though mutable, has proper {@code hashCode} and {@code
+	COMMENT equals} implementation and can be used as map keys if user can ensure
+	COMMENT they are not modified.
+	COMMENT
+	COMMENT @see Box
 	FIELD field_14376 maxZ I
 	FIELD field_14377 maxY I
 	FIELD field_14378 maxX I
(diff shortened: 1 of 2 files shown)


liach:
Requesting a review. This readds the lost `BlockPos.add`, which should be quite useful.

YanisBft (approved):

mappings/net/minecraft/util/math/BlockBox.mapping
```diff
@@ -1,17 +1,34 @@
 CLASS net/minecraft/class_3341 net/minecraft/util/math/BlockBox
+	COMMENT A mutable box with integer coordinates. The box is axis-aligned and the
+	COMMENT coordinates are inclusive.
+	COMMENT
+	COMMENT <p>This box, though mutable, has proper {@code hashCode} and {@code
+	COMMENT equals} implementation and can be used as map keys if user can ensure
```
haykam821: This is probably an existing issue but I would still change it:

```suggestion
	COMMENT equals} implementations and can be used as map keys if user can ensure
```

haykam821 (approved):

mappings/net/minecraft/util/math/Vec3i.mapping
```diff
@@ -57,3 +59,11 @@ CLASS net/minecraft/class_2382 net/minecraft/util/math/Vec3i
 	METHOD method_33106 getSquaredDistance (Lnet/minecraft/class_2382;Z)D
 		ARG 1 vec
 		ARG 2 treatAsBlockPos
+	METHOD method_34592 add (III)Lnet/minecraft/class_2382;
+		COMMENT Returns another Vec3i whose coordinates have the parameter x, y, and z
+		COMMENT added to the coordinates of this vector.
+		COMMENT
+		COMMENT <p>This method always return an immutable object.
```
jaskarth:
```suggestion
		COMMENT <p>This method always returns an immutable object.
```

PR #2127, opened 2021-03-10
Map resource factory

liach:
It's used in creating class_5944 instances and GameRenderer.method_34521,
both of which only uses the DefaultResourcePack as the only resource
factory impl effectively.

Restored removed javadoc. Dunno why matcher removed it

Renamed a method in MinecraftClient as that is not just a downloader but a general resource pack provider that provides both the default and the server specific resource packs. the field is called `builtinPackProvider`, and imo that's ok so didn't rename it.


YanisBft (approved):

enbrain (approved):

liach:
@modmuss50 Just curious, what happened during matching? The parameter name, documentations etc are all lost during the matching; only the method name remained. I wonder if this is a matcher mapping bug or if you accidentally broke something.

PR #2128, opened 2021-03-10
Map vertical surfaces

haykam821:


YanisBft (approved):
Seems good, not sure about the package tho

haykam821:
I placed it in the same package as `Direction`, and many classes are part of `util` rather than `world.gen` already.

liach:
Hmm, is there a vertical surface class? We can create packages for them if needed, just like how game event types are put in their packages

haykam821:
No, this is just an enum containing `FLOOR` and `CEILING`. I can name it `VerticalSurface` if that's better.

liach:
Hmm, looking at `AxisDirection` `Axis` etc in `Direction` imo it's better to remove `Type` suffix.

liach (approved):
I doubt they may change it in the future. This name appears good.

Side note: Might need some manual merging with #2133

mappings/net/minecraft/class_5927.mapping
```diff
@@ -0,0 +1,4 @@
+CLASS net/minecraft/class_5927
```
liach: Mind move this to `VegetationPatchFeatureConfig` instead?

mappings/net/minecraft/class_5931.mapping
```diff
@@ -0,0 +1,4 @@
+CLASS net/minecraft/class_5931
```
liach: Same, now this is `CaveSurfaceDecoratorConfig`

mappings/net/minecraft/class_5927.mapping
haykam821: Fixed in e762aacaed6c3345f0e94f4365d8d9d63f306ae8

mappings/net/minecraft/class_5931.mapping
haykam821: Fixed in e762aacaed6c3345f0e94f4365d8d9d63f306ae8

PR #2133, opened 2021-03-11
World generation in 21w10a

mschae23:
I mapped the new world generation features and decorators in 21w10a.

This is my first contribution to Yarn, so please point out any mistakes I made.


liach (approved):

mappings/net/minecraft/world/gen/feature/util/CaveSurface.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_5721 net/minecraft/world/gen/feature/util/DripstoneColumn
+CLASS net/minecraft/class_5721 net/minecraft/world/gen/feature/util/CaveSurface
```
liach, line 1: Hmm, interesting

mschae23: This class is used for finding the ceiling or floor of a cave. Before 21w10a, this was only used by dripstone, so that's where the name is coming from. In 21w10a, the `cave_surface` decorator was added, which also uses this.

liach (approved):

YanisBft (approved):

jaskarth (approved):

PR #2135, opened 2021-03-12
ScreenHandler adjacent

emilyploszaj:
There were a good number of changes to screen handlers, so I mapped them, I also cleaned up the unmapped slot methods that have been around for a while. I also finally remapped `Item.canBeStoredInItems` as the previous name was straight up wrong.

`ScreenHandlerSyncHandler` is the only new class name, and I imagine it'll be contentious.

Something that should be discussed for changing the name of `CommandItemSlot` as it's now passed by screen handlers to `ItemStack.onStackClicked` and serves as a mutable reference to the cursor stack and seems to be pretty off base.

### DIFF
--- a/mappings/net/minecraft/client/gui/screen/ingame/HandledScreen.mapping
+++ b/mappings/net/minecraft/client/gui/screen/ingame/HandledScreen.mapping
@@ -31,6 +31,7 @@ CLASS net/minecraft/class_465 net/minecraft/client/gui/screen/ingame/HandledScre
 	FIELD field_2801 BACKGROUND_TEXTURE Lnet/minecraft/class_2960;
 	FIELD field_2802 touchDropOriginSlot Lnet/minecraft/class_1735;
 	FIELD field_2803 draggedStackRemainder I
+	FIELD field_29347 displayName Lnet/minecraft/class_2561;
 	METHOD <init> (Lnet/minecraft/class_1703;Lnet/minecraft/class_1661;Lnet/minecraft/class_2561;)V
 		ARG 1 handler
 		ARG 2 inventory
--- a/mappings/net/minecraft/item/Item.mapping
+++ b/mappings/net/minecraft/item/Item.mapping
@@ -33,10 +33,10 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
 		ARG 3 slot
 		ARG 4 clickType
 		ARG 5 player
-		ARG 6 commandItemSlot
+		ARG 6 cursorSlot
 	METHOD method_31567 isItemBarVisible (Lnet/minecraft/class_1799;)Z
 		ARG 1 stack
-	METHOD method_31568 hasStoredInventory ()Z
+	METHOD method_31568 canBeStoredInItems ()Z
 	METHOD method_31569 getItemBarStep (Lnet/minecraft/class_1799;)I
 		ARG 1 stack
 	METHOD method_31570 getEquipSound ()Lnet/minecraft/class_3414;
--- a/mappings/net/minecraft/item/Item.mapping
+++ b/mappings/net/minecraft/item/Item.mapping
@@ -36,7 +36,8 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
 		ARG 6 cursorSlot
 	METHOD method_31567 isItemBarVisible (Lnet/minecraft/class_1799;)Z
 		ARG 1 stack
-	METHOD method_31568 canBeStoredInItems ()Z
+	METHOD method_31568 canBeNested ()Z
+		COMMENT @return true if the item can be placed inside of shulker boxes or bundles.
 	METHOD method_31569 getItemBarStep (Lnet/minecraft/class_1799;)I
 		ARG 1 stack
 	METHOD method_31570 getEquipSound ()Lnet/minecraft/class_3414;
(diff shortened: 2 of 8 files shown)


mappings/net/minecraft/item/Item.mapping
```diff
@@ -33,10 +33,10 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
 		ARG 3 slot
 		ARG 4 clickType
 		ARG 5 player
-		ARG 6 commandItemSlot
+		ARG 6 cursorSlot
 	METHOD method_31567 isItemBarVisible (Lnet/minecraft/class_1799;)Z
 		ARG 1 stack
-	METHOD method_31568 hasStoredInventory ()Z
+	METHOD method_31568 canBeStoredInItems ()Z
```
haykam821: This method specifically is used for determining whether an item can be placed into a bundle (an item) or shulker box slot (not an item!). As such, I would probably rename it to something more generic and add documentation listing those two usages.

haykam821 (changes requested):

mappings/net/minecraft/item/Item.mapping
```diff
@@ -33,10 +33,10 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
 		ARG 3 slot
 		ARG 4 clickType
 		ARG 5 player
-		ARG 6 commandItemSlot
+		ARG 6 cursorSlot
```
liach, line 36: Hmm, should the command item slot be renamed if its usage has expanded?

mappings/net/minecraft/item/Item.mapping
emilyploszaj: I think it's reasonable to call a shulker box an item, as its primarily a storage block that can be picked up, but I am open to name suggestions for this one. When I originally brought this up in #yarn this was the name people seemed most happy with.

mappings/net/minecraft/screen/ScreenHandler.mapping
```diff
@@ -44,6 +49,31 @@ CLASS net/minecraft/class_1703 net/minecraft/screen/ScreenHandler
...
+		ARG 1 slot
+		ARG 2 stack
+		ARG 3 copySupplier
+	METHOD method_34254 setCursorStack (Lnet/minecraft/class_1799;)V
+		ARG 1 stack
+	METHOD method_34255 getCursorStack ()Lnet/minecraft/class_1799;
+	METHOD method_34256 disableSyncing ()V
+	METHOD method_34257 enableSyncing ()V
+	METHOD method_34258 updateCursorStack ()V
+	METHOD method_34259 getCursorCommandItemSlot ()Lnet/minecraft/class_5630;
```
liach, line 76:
```suggestion
	METHOD method_34259 getCursorSlot ()Lnet/minecraft/class_5630;
```

mappings/net/minecraft/screen/ScreenHandlerSyncHandler.mapping
```diff
@@ -0,0 +1,17 @@
+CLASS net/minecraft/class_5916 net/minecraft/screen/ScreenHandlerSyncHandler
```
liach, line 1: Maybe just `ScreenHandlerSync`, two handlers in name sounds awkward

mappings/net/minecraft/item/Item.mapping
emilyploszaj: I touched on this in the description of the PR, it felt wrong to change it without talking about it though, so for now I just wanted t o bring it up

mappings/net/minecraft/screen/ScreenHandler.mapping
emilyploszaj: I think this name is misleading, since `CommandItemSlot`s (or whatever we end up naming them) aren't actually related to normal slots

mappings/net/minecraft/item/Item.mapping
liach: Maybe `canBeNested`? Something like java's nested class concepts

emilyploszaj: I think that's okay, but I find it a bit vague, I'll go with it if we can't find something better though

haykam821: I think that `canBeNested` would work as long as it is supplemented by documentation.

mappings/net/minecraft/screen/ScreenHandlerSyncHandler.mapping
emilyploszaj: I think ending with `Sync` feels more like an event rather than a continuous handler, I originally got suggested `ScreenHandlerSyncer` but found `Syncer` to not feel like a real word, is that potentially better?

CoolMineman (approved):

liach (approved):

mappings/net/minecraft/screen/ScreenHandlerSyncHandler.mapping
Earthcomputer: I wish I hadn't seen the mojmap name for this, or I would be suggesting better names for this. I can't get behind it, it's so awkward :/ But I suppose it's at least better than `class_5916`

liach: we can just use this name for now and update later

jaskarth (approved):

YanisBft (approved):

mappings/net/minecraft/client/gui/screen/ingame/HandledScreen.mapping
```diff
@@ -31,6 +31,7 @@ CLASS net/minecraft/class_465 net/minecraft/client/gui/screen/ingame/HandledScre
 	FIELD field_2801 BACKGROUND_TEXTURE Lnet/minecraft/class_2960;
 	FIELD field_2802 touchDropOriginSlot Lnet/minecraft/class_1735;
 	FIELD field_2803 draggedStackRemainder I
+	FIELD field_29347 displayName Lnet/minecraft/class_2561;
```
Juuxel, line 34: Hmm, is this different from the title?

Juuxel (approved):

PR #2138, opened 2021-03-12
Mapped world gen noise interpolator

mschae23:
Mapped `class_5917`, which is used to interpolate noise in the chunk generator.

### DIFF
--- a/mappings/net/minecraft/util/math/MathHelper.mapping
+++ b/mappings/net/minecraft/util/math/MathHelper.mapping
@@ -155,13 +155,13 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
 			COMMENT The x-coordinate on the unit square
 		ARG 2 deltaY
 			COMMENT The y-coordinate on the unit square
-		ARG 4 val00
+		ARG 4 x0y0
 			COMMENT The output if {@code deltaX} is 0 and {@code deltaY} is 0
-		ARG 6 val10
+		ARG 6 x1y0
 			COMMENT The output if {@code deltaX} is 1 and {@code deltaY} is 0
-		ARG 8 val01
+		ARG 8 x0y1
 			COMMENT The output if {@code deltaX} is 0 and {@code deltaY} is 1
-		ARG 10 val11
+		ARG 10 x1y1
 			COMMENT The output if {@code deltaX} is 1 and {@code deltaY} is 1
 	METHOD method_16438 lerp3 (DDDDDDDDDDD)D
 		COMMENT A three-dimensional lerp between values on the 8 corners of the unit cube. Arbitrary values are specified for the corners and the output is interpolated between them.
@@ -171,21 +171,21 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
 			COMMENT The y-coordinate on the unit cube
 		ARG 4 deltaZ
 			COMMENT The z-coordinate on the unit cube
-		ARG 6 val000
+		ARG 6 x0y0z0
 			COMMENT The output if {@code deltaX} is 0, {@code deltaY} is 0 and {@code deltaZ} is 0
-		ARG 8 val100
+		ARG 8 x1y0z0
 			COMMENT The output if {@code deltaX} is 1, {@code deltaY} is 0 and {@code deltaZ} is 0
-		ARG 10 val010
+		ARG 10 x0y1z0
 			COMMENT The output if {@code deltaX} is 0, {@code deltaY} is 1 and {@code deltaZ} is 0
-		ARG 12 val110
+		ARG 12 x1y1z0
 			COMMENT The output if {@code deltaX} is 1, {@code deltaY} is 1 and {@code deltaZ} is 0
-		ARG 14 val001
+		ARG 14 x0y0z1
 			COMMENT The output if {@code deltaX} is 0, {@code deltaY} is 0 and {@code deltaZ} is 1
-		ARG 16 val101
+		ARG 16 x1y0z1
 			COMMENT The output if {@code deltaX} is 1, {@code deltaY} is 0 and {@code deltaZ} is 1
-		ARG 18 val011
+		ARG 18 x0y1z1
 			COMMENT The output if {@code deltaX} is 0, {@code deltaY} is 1 and {@code deltaZ} is 1
-		ARG 20 val111
+		ARG 20 x1y1z1
 			COMMENT The output if {@code deltaX} is 1, {@code deltaY} is 1 and {@code deltaZ} is 1
 	METHOD method_16439 lerp (FFF)F
 		ARG 0 delta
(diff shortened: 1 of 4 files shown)


liach (approved):

YanisBft (approved):

YanisBft:
Would be interesting to document this class (not necessarily in this PR)

mschae23:
> Would be interesting to document this class (not necessarily in this PR)

That's a good idea, I can try doing that in another PR. Should all the `noisexxx` fields be documented too, or just the methods?

YanisBft:
> That's a good idea, I can try doing that in another PR. Should all the `noisexxx` fields be documented too, or just the methods?

As you feel it, if they need specific documentation it can be interesting

liach:
Yeah if an element's name is straightforward, you don't need to document. However, if a method has some weird behavior like under what condition it may throw some unchecked exception, if it modifies itself or returns a new copy, etc., these weird behaviors are usually worthy to document.

mappings/net/minecraft/util/math/NoiseInterpolator.mapping
```diff
@@ -0,0 +1,57 @@
+CLASS net/minecraft/class_5917 net/minecraft/util/math/NoiseInterpolator
+	FIELD field_29227 startNoiseBuffer [[D
+	FIELD field_29228 endNoiseBuffer [[D
+	FIELD field_29229 sizeY I
+	FIELD field_29230 sizeZ I
+	FIELD field_29231 minY I
+	FIELD field_29232 columnSampler Lnet/minecraft/class_5917$class_5918;
+	FIELD field_29233 noise000 D
```
jaskarth: Personally, I'm much more partial to the name `x0z0y0` instead of `noise000` for example. It's easier to see what each of the 3 numbers mean. Interested to hear what you all have to say

mappings/net/minecraft/util/math/NoiseInterpolator.mapping
```diff
@@ -0,0 +1,57 @@
...
+	FIELD field_29234 noise001 D
+	FIELD field_29235 noise100 D
+	FIELD field_29236 noise101 D
+	FIELD field_29237 noise010 D
+	FIELD field_29238 noise011 D
+	FIELD field_29239 noise110 D
+	FIELD field_29240 noise111 D
+	FIELD field_29241 noise0Y0 D
```
jaskarth: Same here. I'd prefer `x0z0`, as the y value is interpolated out.

mappings/net/minecraft/util/math/NoiseInterpolator.mapping
```diff
@@ -0,0 +1,57 @@
...
+	FIELD field_29242 noise1Y0 D
+	FIELD field_29243 noise0Y1 D
+	FIELD field_29244 noise1Y1 D
+	FIELD field_29245 noiseXY0 D
```
jaskarth: Perhaps `z0` as the x value is interpolated out.

jaskarth (commented):
I have some rather controversial opinions on what the noise sample fields should be named (I have mapped and documented version of the noise interpolation [here](https://github.com/SuperCoder7979/ecotones/blob/master/src/main/java/supercoder79/ecotones/world/gen/BaseEcotonesChunkGenerator.java#L396)) and would like to start a discussion on the best way to name them.

mappings/net/minecraft/util/math/NoiseInterpolator.mapping
liach: Just for reference, in a similar situation in matrix4f, we currently name fields like a00, a01, etc. Not saying that naming is the best, but it'd help if we can find a consistent reasoning for naming fields in different styles than adding 15 competing standards.

jaskarth: That's a good point. I'm mostly interested in naming it this way as the values get interpolated in the 3 axes separately, so it's easier to see how it occurs, e.g. when the cube gets interpolated down to a square, the y value is "lost" as it's interpolated into the x and z values, so `x0z0y0` becomes `x0z0` for example. The same happens when the square is interpolated into a line, where it becomes `z0` from `x0z0` as the x value is "lost" and interpolated into the z values.

liach: Yeah, we can choose to use `xz00` names, which has more specific prefixes (about the type of lerping they are doing) than the generic "noise", already implied by the class name "noise interpolator".

Imo we should pull out `xz` `xyz` as a whole to the prefix, like `xyz000` than `x0y0z0` so in code it's be more obvious which types of lerping arguments you are using.

jaskarth: I originally went with the numbers interlayed with the coord as it was originally how the noise was accessed out of the 3d array, so it was:
```java
double x0z0y0 = noiseData[0][noiseZ][noiseY];
double x0z1y0 = noiseData[0][noiseZ + 1][noiseY];
double x1z0y0 = noiseData[1][noiseZ][noiseY];
double x1z1y0 = noiseData[1][noiseZ + 1][noiseY];
```
so it was x first, z second, and y third. Since these are local variables it becomes a bit hairier, but I still think we should keep the interlayed coord as it's easier to me to see which axis it is interpolating. Doing the prefix would lose the meaning of `x0z0y0` meaning "x0, z0, y0" in my opinion.

liach: Agreed. If `x0z0y0` is better than `xzy000` for understanding code, I'm all for it. The purpose of yarn is to allow us understand minecraft code better (say, than if we look at mojang map)

mschae23: I named the fields that way because of `MathHelper.lerp3`. But you're right, `noise000` is meaningless on its own.

Earthcomputer: Who was that idiot who named the parameters in `lerp3`? Oh, darn, it was me. Tbh I couldn't think of any better parameter names at the time.
Anyway, the naming inside mat4f is pretty standard specifically for matrices, but it may be confusing outside of that context.

mschae23: > Who was that idiot who named the parameters in lerp3? Oh, darn, it was me. Tbh I couldn't think of any better parameter names at the time.

They were at least better than the names I originally came up for this.

liach (approved):

mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -155,13 +155,13 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
 			COMMENT The x-coordinate on the unit square
 		ARG 2 deltaY
 			COMMENT The y-coordinate on the unit square
-		ARG 4 val00
```
jaskarth, line 158: Personally, I think these should be left as-is because they're used in other places that don't need to follow this convention

mschae23: This would make the parameters clearer there too though. And they mean basically the same thing.

jaskarth: Took another look at it and I get what you're saying

jaskarth (approved):

mappings/net/minecraft/util/math/MathHelper.mapping
liach: i approce these param changes as well.

PR #2140, opened 2021-03-13
Improve ScreenHandlerContext mappings

Juuxel:
- Both getter methods renamed to `get` instead of `run`.
  - `run` was confusing as it's used more for getting a world value than running a function.
  - Side note: having two `run`s just taking a lambda also confused kotlinc's SAM interface conversions; this should make it nicer for Kotlin modders if JetBrains hasn't fixed it yet
- Improved documentation for `get` and `run`.
- Renamed `get`'s `function` parameter to `getter`.

### DIFF
--- a/mappings/net/minecraft/screen/ScreenHandlerContext.mapping
+++ b/mappings/net/minecraft/screen/ScreenHandlerContext.mapping
@@ -8,9 +8,24 @@ CLASS net/minecraft/class_3914 net/minecraft/screen/ScreenHandlerContext
 		ARG 0 world
 		ARG 1 pos
 	METHOD method_17393 run (Ljava/util/function/BiConsumer;)V
+		COMMENT Runs a {@link BiConsumer} with this context's world and position
+		COMMENT if this context is active.
 		ARG 1 function
-	METHOD method_17395 run (Ljava/util/function/BiFunction;)Ljava/util/Optional;
-		ARG 1 function
-	METHOD method_17396 run (Ljava/util/function/BiFunction;Ljava/lang/Object;)Ljava/lang/Object;
-		ARG 1 function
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/screen/ScreenHandlerContext.mapping
```diff
@@ -8,9 +8,24 @@ CLASS net/minecraft/class_3914 net/minecraft/screen/ScreenHandlerContext
...
+			COMMENT a function that gets a non-null value from this context's world and position
+	METHOD method_17396 get (Ljava/util/function/BiFunction;Ljava/lang/Object;)Ljava/lang/Object;
+		COMMENT Gets a value from this context's world and position
+		COMMENT with a {@link BiFunction} getter.
+		COMMENT
+		COMMENT @return the getter's return value if this context is active,
+		COMMENT         the default value otherwise
+		ARG 1 getter
+			COMMENT a function that gets a non-null value from this context's world and position
 		ARG 2 defaultValue
```
liach, line 30: I'm inclined to name all these default values `fallback`

liach (approved):

mappings/net/minecraft/screen/ScreenHandlerContext.mapping
Juuxel: Hmm, that might be worth a note in `CONVENTIONS.md` since it's a relatively common pattern in the code base.

liach: Yeah, blame java for reserving `default`. `defaultValue` is too long compared to `fallback`.

enbrain (approved):

YanisBft (approved):

PR #2144, opened 2021-03-14
Map some parts of the OpenGL state managing mechanisms

kb-1000:
This also changes some plural names to singular where it makes sense (as the LWJGL method handles creating the buffers, so they only ever deal with single OpenGL objects).

### DIFF
--- a/mappings/com/mojang/blaze3d/platform/GlStateManager.mapping
+++ b/mappings/com/mojang/blaze3d/platform/GlStateManager.mapping
@@ -210,24 +210,24 @@ CLASS net/minecraft/class_4493 com/mojang/blaze3d/platform/GlStateManager
 		ARG 0 program
 		ARG 1 maxLength
 	METHOD method_22053 disableBlend ()V
-	METHOD method_22054 deleteBuffers (I)V
+	METHOD method_22054 deleteBuffer (I)V
 		ARG 0 buffer
 	METHOD method_22056 enableBlend ()V
 	METHOD method_22058 polygonMode (II)V
 		ARG 0 face
 		ARG 1 mode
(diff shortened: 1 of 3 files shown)


mappings/com/mojang/blaze3d/platform/GlStateManager.mapping
```diff
@@ -286,6 +287,24 @@ CLASS net/minecraft/class_4493 com/mojang/blaze3d/platform/GlStateManager
...
 	METHOD method_31947 unmapBuffer (I)V
 		ARG 0 target
+	METHOD method_34404 bindAttribLocation (IILjava/lang/CharSequence;)V
+		ARG 0 program
+		ARG 1 index
+		ARG 2 name
+	METHOD method_34405 shaderSource (ILjava/util/List;)V
+		ARG 0 shader
+		ARG 1 strings
+	METHOD method_34406 vertexAttribIPointer (IIIIJ)V
```
liach, line 297: Is this supposed to have the `I` between `Attrib` and `Pointer`

kb-1000: I'm not sure, but it's not a typo by me at least. That `I` comes from the LWJGL/OpenGL method name (`GL30.glVertexAttribIPointer`).

Juuxel: It's an [OpenGL API function](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glVertexAttribPointer.xhtml), so not a LWJGL typo either. (Curiously, there's also a variant without the `I`)

liach (approved):

Juuxel (approved):

PR #2147, opened 2021-03-15
Some lambda parameters in packets

DreenDex:


liach:
imo you should give them names like b than buf1 for these params (effectively local vars), like what you would do in writing map computeifabsent

liach (approved):

YanisBft (approved):

enbrain (approved):

PR #2148, opened 2021-03-15
Two methods used on client player respawn

DreenDex:


mappings/net/minecraft/client/network/ClientPlayerEntity.mapping
```diff
@@ -98,3 +98,4 @@ CLASS net/minecraft/class_746 net/minecraft/client/network/ClientPlayerEntity
 		ARG 2 dz
 	METHOD method_3149 isAutoJumpEnabled ()Z
 	METHOD method_3151 getMountJumpStrength ()F
+	METHOD method_33689 onRespawnOrDimensionChange ()V
```
liach: This is more like `init` or `setup` as it sets up the position, velcity, and pose of the player. Also it initializes health and sets deathTime to 0 which I think prevents the player from rendering as "dying".
```suggestion
	METHOD method_33689 init ()V
```

liach (changes requested):

liach (approved):

enbrain (approved):

modmuss50 (approved):

PR #2170, opened 2021-03-20
Add misc javadoc

enbrain:


mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -278,6 +278,11 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_6024 getLeaningPitch (F)F
 		ARG 1 tickDelta
 	METHOD method_6025 heal (F)V
+		COMMENT Heals this entity by the given {@code amount} of hearts.
```
Juuxel: This actual seems to be HP (half-hearts) instead of hearts.

Juuxel (changes requested):

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -278,6 +278,11 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_6024 getLeaningPitch (F)F
 		ARG 1 tickDelta
 	METHOD method_6025 heal (F)V
+		COMMENT Heals this entity by the given {@code amount} of half-hearts.
+		COMMENT
+		COMMENT <p>You can't heal a dead entity.
```
jaskarth:
```suggestion
		COMMENT <p>A dead entity cannot be healed.
```
I'm not sure if we try to avoid "You" in javadoc but passive voice works a bit better here I think

jaskarth (commented):
Looks good, apart from this one nitpick

liach (approved):

jaskarth (approved):

modmuss50 (approved):

PR #2184, opened 2021-03-24
Document ItemEntity and improve a few names

### LINKED ISSUE #2110: ItemEntity#age shadows Entity#age
I think the name needs to be changed to something else, though I'm not sure exactly what. Calling it `age` is problematic because it shadows Entity#age and makes it private, making it difficult to reference the field from the parent class without a cast.

Also ItemEntity#age appears to function differently between client and server, meaning code that makes use of that field will cause weird desyncs and jumping of item entities, wheras Entity#age does not exhibit this behaviour.

ItemEntity#age also doesn't seem to properly reflect the way it's used. i.e. From implementing a similar entity for my own mod, I've been able to identify that it's used for two things:

1. It controls whether an item despawns, is merged, and can be picked up
2. It controls the rotation and vertical bobbing of the item on the client.

// ItemEntity.java
```
   @Environment(EnvType.CLIENT)
   public float method_27314/*getRotation*/(float tickDelta) {
      return ((float)this.getAge() + tickDelta) / 20F + this.hoverHeight;
   }
```
and
// ItemEntityRenderer.jsva
```
float verticalOffset = MathHelper.sin(((float)itemEntity.getAge() + tickDelta) / 10F + itemEntity.hoverHeight) * 0.1F + 0.1F;
```

I also suggest the following name changes:
`Item#method_27314` -> `Item#getRotation`
`Item#hoverHeight` -> `Item#uniqueOffset`

Maybe calling it `itemAge` would do as a possible workaround, since Mojang appears to have called it Age in the nbt? I would also add a javadoc to the field so people are aware of the above noted desync issue. I've seen it occur white a few times in the past, both in vanilla and with mods, where items jump sporadically so it may be a more common problem that anyone seems to have noticed.

Also before anyone asks: The reason i noticed this as an issue was because:
A: Way back in the forge days I was trying to make items float on clouds, and they would jitter and ghost through before jumping back.
B: More recently I was implementing logic to make feathers float to the ground realistically, and saw somewhat similar behaviour. It would jump occassionally when using ItemEntity#age but would move smoothly when using Entity#age, which is why I recommend the change.

YanisBft:
Fixes #2110

### DIFF
--- a/mappings/net/minecraft/entity/ItemEntity.mapping
+++ b/mappings/net/minecraft/entity/ItemEntity.mapping
@@ -5,7 +5,10 @@ CLASS net/minecraft/class_1542 net/minecraft/entity/ItemEntity
 	FIELD field_7202 pickupDelay I
 	FIELD field_7203 uniqueOffset F
 	FIELD field_7204 itemAge I
-		COMMENT The number of ticks since the item has been created. It is a short value with key {@code Age} in the NBT structure.
+		COMMENT The number of ticks since this item entity has been created.
+		COMMENT It is a short value with key {@code Age} in the NBT structure.
+		COMMENT
+		COMMENT <p>It differs from {@code Entity#age}.
 	FIELD field_7205 owner Ljava/util/UUID;
 	METHOD <init> (Lnet/minecraft/class_1542;)V
 		ARG 1 entity
@@ -79,13 +82,16 @@ CLASS net/minecraft/class_1542 net/minecraft/entity/ItemEntity
 		COMMENT <p>Increases every tick. When it equals to 6000 ticks (5 minutes),
 		COMMENT this item entity disappears.
 		COMMENT
-		COMMENT <p>Unlike {@linkplain Entity#age}, it is not synchronized between
-		COMMENT the client and the server.
+		COMMENT <p>Unlike {@linkplain Entity#age}, it is persistent and not synchronized
+		COMMENT between the client and the server.
 		COMMENT
 		COMMENT @see #tick()
 	METHOD method_6986 getOwner ()Ljava/util/UUID;
-		COMMENT Returns the UUID of the entity which owns this item entity,
+		COMMENT Returns the UUID of the entity to which belongs this item entity,
 		COMMENT or {@code null} if there is not.
+		COMMENT
+		COMMENT <p>If there is one, the owner is the only entity which can pick
+		COMMENT up this item entity.
 	METHOD method_6987 setDespawnImmediately ()V
 	METHOD method_6988 setToDefaultPickupDelay ()V
 		COMMENT Sets the number of ticks before this item entity can be picked up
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/entity/ItemEntity.mapping
```diff
@@ -3,8 +3,8 @@ CLASS net/minecraft/class_1542 net/minecraft/entity/ItemEntity
 	FIELD field_7200 thrower Ljava/util/UUID;
 	FIELD field_7201 health I
 	FIELD field_7202 pickupDelay I
-	FIELD field_7203 hoverHeight F
-	FIELD field_7204 age I
+	FIELD field_7203 uniqueOffset F
+	FIELD field_7204 itemAge I
```
liach, line 7: mind document this is `Age` in nbt and note the name discrepancy
```suggestion
	FIELD field_7204 itemAge I
		COMMENT The number of ticks since the item has been created. It is a short value with key {@code Age} in the NBT structure.
```

mappings/net/minecraft/entity/ItemEntity.mapping
```diff
@@ -37,26 +37,58 @@ CLASS net/minecraft/class_1542 net/minecraft/entity/ItemEntity
...
+		COMMENT
+		COMMENT <p>Increases every tick. When it equals to 6000 ticks (5 minutes),
+		COMMENT this item entity disappears.
+		COMMENT
+		COMMENT <p>Unlike {@linkplain Entity#age}, it is not synchronized between
+		COMMENT the client and the server.
+		COMMENT
+		COMMENT @see #tick()
 	METHOD method_6986 getOwner ()Ljava/util/UUID;
+		COMMENT Returns the UUID of the entity which owns this item entity,
```
liach: Prob note that the owner is the only one that can pick up the item if it exists

mappings/net/minecraft/entity/ItemEntity.mapping
```diff
@@ -37,26 +37,58 @@ CLASS net/minecraft/class_1542 net/minecraft/entity/ItemEntity
...
 		ARG 1 uuid
-	METHOD method_6985 getAge ()I
+	METHOD method_6985 getItemAge ()I
+		COMMENT Returns the number of ticks since this item entity has been created.
+		COMMENT
+		COMMENT <p>Increases every tick. When it equals to 6000 ticks (5 minutes),
+		COMMENT this item entity disappears.
+		COMMENT
+		COMMENT <p>Unlike {@linkplain Entity#age}, it is not synchronized between
+		COMMENT the client and the server.
```
liach: Maybe note this is persistent.

mappings/net/minecraft/entity/ItemEntity.mapping
```diff
@@ -5,7 +5,10 @@ CLASS net/minecraft/class_1542 net/minecraft/entity/ItemEntity
 	FIELD field_7202 pickupDelay I
 	FIELD field_7203 uniqueOffset F
 	FIELD field_7204 itemAge I
-		COMMENT The number of ticks since the item has been created. It is a short value with key {@code Age} in the NBT structure.
+		COMMENT The number of ticks since this item entity has been created.
+		COMMENT It is a short value with key {@code Age} in the NBT structure.
+		COMMENT
+		COMMENT <p>It differs from {@code Entity#age}.
```
liach:
```suggestion
		COMMENT <p>It differs from {@link Entity#age}.
```

liach (approved):

enbrain (approved):

IotaBread (approved):

jaskarth (approved):

PR #2185, opened 2021-03-24
More set block state flags

Earthcomputer:
Flags! :black_flag: :white_flag: :rainbow_flag:


unpick-definitions/set_block_state_flags.unpick
```diff
@@ -1,5 +1,6 @@
 v2
 
+#flag set_block_state_flags net/fabricmc/yarn/constants/SetBlockStateFlags DEFAULT
```
Earthcomputer: Unpick doesn't seem to handle this very well yet, so I commented this out for now.

modmuss50: Humm, yeah.. Might need to take a look at unpick to handle this.

liach (approved):

liach:
Just curious, is there any tool to modify unpick format easily, or is it usually modified by hand like access wideners?

Earthcomputer:
I did it by hand, hopefully there will be tools in Enigma in the near future.

liach:
Just a side note: I will put new/refactor on unpick prs based on how they change unpick definitions. In fact, I wonder how bad it is for mapping users if a constant holding utility class is changed. Does it cause more incompatibility than regular mapping refactors?

Earthcomputer:
The unpick issue affecting `SetBlockStateFlags.DEFAULT` is fixed by https://github.com/FabricMC/unpick/pull/3, going to re-enable the DEFAULT flag in this PR so that it works once unpick is fixed.

YanisBft (approved):
Works well

liach:
@Earthcomputer Feel free to revert #2189 in your changes. I think your patch fixed the spamming problem of unpick where it logs an info message for every method it visits, but it might need some testing to see if the log spam is actually gone.

modmuss50 (approved):

Earthcomputer:
@liach I didn't remove that log spam

liach:
Turns out the muting of unpick is still needed, sadly. Yeah I tested locally as well and the log spam is still there, but otherwise things still appear good :+1:

Issue #2195, opened 2021-03-27
Compatibility of Constant-holding classes


liach:
Now that we've added unpick, we have constant holding classes.

I want to talk about their binary compatibility across renames. Unlike yarn names, the constant-holding classes do not have stable intermediary names; they are more like classes in the api.

We need to determine our policy on these classes to minimize the negative effect of name changes or moves of these classes on the users.

liach:
Hmm, should we generate intermediary for the constant classes? Just in case we rename them and binary compatibility may break.

modmuss50:
They are binary compatible as they are compile only. They can be renamed and changed just like any yarn name without worry.

liach:
What if people try reflection on the field values?

modmuss50:
Nothing we can do about that, it will break in dev so seems fine to me. There are loads of things people can do to cause issues, this seems like a none issue to me.

liach:
Sure, I guess we will just enforce `compileOnly` configuration for the constant jar. And I've added a warning in the package javadoc as well.

On a side note, will this [`NbtType` from base api](https://github.com/FabricMC/fabric/blob/1.16/fabric-api-base/src/main/java/net/fabricmc/fabric/api/util/NbtType.java) be deprecated and scheduled for removal once we get #2201 in?

liach:
So we now have a general grasp on the compatibility of these constant classes. If compatibility problems arise in the future, we will address in separate issues.

PR #2200, opened 2021-03-28
Minor javadoc updates around unpick

liach:
A few things:
1. Added licenser and cc0 message to constant classes
2. Add constant classes to jd gen, wrote a package info that partially addressed #2195 
3. Fixed one javadoc error in current yarn
4. Changed set block state method doc to use the doc of the constant holding class instead

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,4 +12,4 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: gradle/wrapper-validation-action@v1
-      - run: ./gradlew build javadocJar checkMappings mapNamedJar --stacktrace
+      - run: ./gradlew checkLicenses build javadocJar checkMappings --stacktrace
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,4 +12,4 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: gradle/wrapper-validation-action@v1
-      - run: ./gradlew checkLicenses build javadocJar checkMappings --stacktrace
+      - run: ./gradlew build javadocJar checkMappings --stacktrace
--- a/mappings/net/minecraft/block/entity/AbstractFurnaceBlockEntity.mapping
+++ b/mappings/net/minecraft/block/entity/AbstractFurnaceBlockEntity.mapping
@@ -50,7 +50,8 @@ CLASS net/minecraft/class_2609 net/minecraft/block/entity/AbstractFurnaceBlockEn
 	METHOD method_17763 dropExperienceForRecipesUsed (Lnet/minecraft/class_3222;)V
 		ARG 1 player
 	METHOD method_26395 isNonFlammableWood (Lnet/minecraft/class_1792;)Z
-		COMMENT Returns whether the provided {@code item} is in the {@linkplain ItemTags#NON_FLAMMABLE_WOOD} tag.
+		COMMENT Returns whether the provided {@code item} is in the {@linkplain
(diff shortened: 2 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -12,4 +12,4 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: gradle/wrapper-validation-action@v1
-      - run: ./gradlew build javadocJar checkMappings mapNamedJar --stacktrace
+      - run: ./gradlew checkLicenses build javadocJar checkMappings --stacktrace
```
modmuss50: No, why remove `mapNamedJar` ?

modmuss50 (changes requested):

.github/workflows/build.yml
liach: It is a dependency of `genFakeSource`, which leads to `javadoc` and `javadocJar`

liach: The proof that this is ran: https://github.com/FabricMC/yarn/runs/2212467447#step:5:181

modmuss50: Ok cool, all good then 👍

modmuss50 (approved):

Earthcomputer (approved):

.github/workflows/build.yml
Juuxel: Hmm isn't `checkLicenses` a dependency of `check` (and thus `build`)?

mappings/net/minecraft/block/entity/AbstractFurnaceBlockEntity.mapping
```diff
@@ -50,7 +50,8 @@ CLASS net/minecraft/class_2609 net/minecraft/block/entity/AbstractFurnaceBlockEn
 	METHOD method_17763 dropExperienceForRecipesUsed (Lnet/minecraft/class_3222;)V
 		ARG 1 player
 	METHOD method_26395 isNonFlammableWood (Lnet/minecraft/class_1792;)Z
-		COMMENT Returns whether the provided {@code item} is in the {@linkplain ItemTags#NON_FLAMMABLE_WOOD} tag.
+		COMMENT Returns whether the provided {@code item} is in the {@linkplain
+		COMMENT net.minecraft.tag.ItemTags#NON_FLAMMABLE_WOOD non_flammable_wood} tag.
```
Juuxel:
```suggestion
		COMMENT Returns whether the provided {@code item} is in the {@link
		COMMENT net.minecraft.tag.ItemTags#NON_FLAMMABLE_WOOD non_flammable_wood} tag.
```
I'd say a non-plaintext link is better here.

Juuxel (approved):

.github/workflows/build.yml
liach:
```suggestion
      - run: ./gradlew build javadocJar checkMappings --stacktrace
```
It indeed is according to gradle build scan. Optimizing.

mappings/net/minecraft/block/entity/AbstractFurnaceBlockEntity.mapping
liach: Hmm, seems link is like wrapping stuff in `<code>`, so this actually works better I guess?

YanisBft (approved):

mappings/net/minecraft/block/entity/AbstractFurnaceBlockEntity.mapping
Juuxel: Yeah `{@link}` makes it use the code font which is better for a tag.

liach:
Additionally, after merging some unpick prs, we need to run `gradle licenseFormat` in order to add the cc0 license to our constant classes (arr bad)

PR #2205, opened 2021-03-29
Add package javadoc support

### LINKED ISSUE #2035: Add package javadoc
Since yarn javadocs is now published (https://maven.fabricmc.net/docs/yarn-21w05b+build.11/), it would be cool to have package info Javadocs.

liach:
Closes #2035

This is surprisingly easy given modmuss's work on constants. So here we go!

For now, the package info files aren't compiled; they are purely used to generate javadoc. In addition, these javadocs aren't exported to mappings in any way.

(in fact, the compilePackageDocsJava task is totally unused, but you can run it; I guess I added it to explicitly state it is java 11 as opposed to java 8 for constants)

I added comments on their treatments if there is plan in the future to build binary for these package-infos (possible if we want annotations).

To demonstrate the new package docs, I added package info for `net.minecraft` and `net.minecraft.resource` packages. Imo this description of resource package tells you what you should check out, doesn't it?
image

Another tiny change is to make `checkMappings` depend on `mapIntermediaryJar`, since mapping check requires the intermediary jar to be present.

### DIFF
--- a/src/packageDocs/java/net/minecraft/package-info.java
+++ b/src/packageDocs/java/net/minecraft/package-info.java
@@ -1,0 +1,17 @@
+/*
+ * This file is free for everyone to use under the Creative Commons Zero license.
+ */
+
+/**
+ * The base package for all Minecraft classes. All Minecraft classes belong to this package
+ * in their intermediary names.
+ *
+ * <p>Unmapped classes goes into this package by default. This package additionally contains
+ * {@link Bootstrap}, {@link SharedConstants}, and {@link MinecraftVersion} classes.
+ *
+ * <p>While it's known that some obfuscated Minecraft classes are under other packages like
+ * {@code com.mojang.*}, yarn keeps all mapped classes under {@code net.minecraft.*} since
+ * there is no convincing evidence those classes are independent from Minecraft.
+ */
+
+package net.minecraft;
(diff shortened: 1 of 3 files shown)


liach:
CC @LambdAurora for a review

LambdAurora (approved):

src/packageDocs/java/net/minecraft/package-info.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * This file is free for everyone to use under the Creative Commons Zero license.
+ */
+
+/**
+ * The base package for all Minecraft classes. All Minecraft classes belong to this package
+ * in their intermediary names.
+ *
+ * <p>Unmapped classes goes into this package by default. This package additionally contains
```
Earthcomputer:
```suggestion
 * <p>Unmapped classes go into this package by default. This package additionally contains
```

src/packageDocs/java/net/minecraft/resource/package-info.java
```diff
@@ -0,0 +1,71 @@
...
+ * <tr>
+ *     <th><b>Class</b></th><th><b>Usage</b></th>
+ * </tr>
+ * <tr>
+ *     <td>{@link Resource}</td>
+ *     <td>Accesses binary data.</td>
+ * </tr>
+ * <tr>
+ *     <td>{@link ResourceFactory}</td>
+ *     <td>Provides resource given an {@link net.minecraft.util.Identifier}.</td>
```
Juuxel:
```suggestion
 *     <td>Provides a resource given an {@link net.minecraft.util.Identifier}.</td>

```

Juuxel (changes requested):

modmuss50 (approved):

PR #2211, opened 2021-03-31
class_5945 -> net/minecraft/client/util/ParticleUtil

Sollace:
Just one class.

This contains various static methods that are called to spawn a particle.


liach (approved):
After looking, I think the util package is indeed better than the particle package for this one.

YanisBft (approved):

Shnupbups (approved):

PR #2216, opened 2021-03-31
Map WorldEventConstants

apple502j:
RIP unpick 21w11a-21w11a
All names copied from unpick mapping


liach (changes requested):
Where are my javadocs and upgraded unpick notations

YanisBft:
Do we agree on suffixing the constant classes with "Constants"? Because honestly WorldEvents would suffice

apple502j:
Maybe `WorldEvents` works, but unsure if the package name is not confusing - unpick is under `net.fabricmc.yarn.constants`, but I did not introduce "constants" package yet, maybe should I?

OroArmor:
Also shouldnt this pr fix the unpick file to use the new constant file rather than the old manual one?

YanisBft:
idk if a new package is needed, imo constant classes should be put under the same package of their use (the use we guessed)

liach:
If you find it confusing, `WorldEventIds` is an okay alternative

apple502j:
Someone can redo it later with javadocs and unpick, i'm gonna close

PR #2222, opened 2021-03-31
Entity model part constants

YanisBft:
Maps the class containing model part constants, and adds the unpick definition file using them.
~~I didn't get the unpick to work though, probably because there are strings?~~


modmuss50:
Constants in 13a need looking at, in theory all of the constants are now in the mc jar. Unpick will need a small tweak to allow using them.

mappings/net/minecraft/client/render/entity/model/EntityModelPartConstants.mapping
```diff
@@ -0,0 +1,58 @@
+CLASS net/minecraft/class_6230 net/minecraft/client/render/entity/model/EntityModelPartConstants
```
liach: is `EntityModelPartNames` more accurate?

mappings/net/minecraft/client/util/ParticleUtil.mapping
```diff
@@ -4,12 +4,12 @@ CLASS net/minecraft/class_5945 net/minecraft/client/util/ParticleUtil
...
 		ARG 2 direction
 		ARG 3 effect
-	METHOD method_34682 spawnParticle (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2394;Lnet/minecraft/class_4801;)V
+	METHOD method_34682 spawnParticle (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2394;Lnet/minecraft/class_6019;)V
 		ARG 0 world
 		ARG 1 pos
 		ARG 2 effect
 		ARG 3 range
-	METHOD method_34683 spawnParticle (Lnet/minecraft/class_2350$class_2351;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;DLnet/minecraft/class_2394;Lnet/minecraft/class_4801;)V
+	METHOD method_34683 spawnParticle (Lnet/minecraft/class_2350$class_2351;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;DLnet/minecraft/class_2394;Lnet/minecraft/class_6019;)V
```
liach, line 12: is this right?

YanisBft: Yes the methods changed in 21w13a but the PR was created in 21w11a

liach (approved):

OroArmor (approved):

unpick-definitions/entity_model_parts.unpick
```diff
@@ -0,0 +1,74 @@
...
+target_method net/minecraft/client/model/ModelPartBuilder cuboid (Ljava/lang/String;FFFIIIII)Lnet/minecraft/client/model/ModelPartBuilder;
+	param 1 entity_model_parts
+target_method net/minecraft/client/model/ModelPartBuilder cuboid (Ljava/lang/String;FFFFFF)Lnet/minecraft/client/model/ModelPartBuilder;
+	param 1 entity_model_parts
+target_method net/minecraft/client/model/ModelPartBuilder cuboid (Ljava/lang/String;FFFFFFLnet/minecraft/client/model/Dilation;)Lnet/minecraft/client/model/ModelPartBuilder;
+	param 1 entity_model_parts
+target_method net/minecraft/client/model/ModelPartData getChild (Ljava/lang/String;)Lnet/minecraft/client/model/ModelPartData;
+	param 1 entity_model_parts
+target_method net/minecraft/client/model/ModelPartData addChild (Ljava/lang/String;Lnet/minecraft/client/model/ModelPartBuilder;Lnet/minecraft/client/model/ModelTransform;)Lnet/minecraft/client/model/ModelPartData;
+	param 1 entity_model_parts
```
liach: Wait, I think these are supposed to be `param 0` in unpick format, even though these are instance methods

liach (changes requested):

unpick-definitions/entity_model_parts.unpick
liach: (At least that's how it worked when I last tested in #2246)

YanisBft: fml I tested everything but that. It's confusing to have different param indexing

liach: Indeed! Unpick is quite confusing. Even its wiki incorrectly described its format: when you need to specify constant value and desc, say value `255` and type `I`, you should have `255 I` than `I 255` as the wiki suggests. So yeah, this is indeed painstaking!

liach (approved):
:heart:

modmuss50:
Did you get unpick working?

YanisBft:
Yes it does for all the targeted methods

modmuss50:
Great 👍

modmuss50 (approved):

YanisBft:
I will open another PR to map and unpick other model part constants, declared in entity model classes

unpick-definitions/entity_model_parts.unpick
Daomephsta: > Wait, I think these are supposed to be `param 0` in unpick format, even though these are instance methods

They're parameter indices, not local variable indices. So there's no difference between static and instance methods.

>  Even its wiki incorrectly described its format: when you need to specify constant value and desc, say value 255 and type I, you should have 255 I than I 255 as the wiki suggests.

If you find an issue with the docs, please report it. Though you can't do that right now, since I archived WeaveMC/unpick last night.

PR #2238, opened 2021-04-01
Fixes javadoc gen for 21w13a

liach:
Closes #2217

Fixes some leftover minor stuff from unpick pr
Named `FLOAT_PROVIDER_TYPE_KEY` since the name inferer just infers `FLOAT_PROVIDER_TYPE` name, and it conflicts with the next registry field's name. Like a bug in stitch?

PR #2241, opened 2021-04-01
Two annotations. They write cursed annotations

liach:
Depends on #2238 for proper javadocs.

The `Debug` name is presumptive and most likely is something else. Feel free to suggest.

### DIFF
--- a/mappings/net/minecraft/obfuscate/DontObfuscate.mapping
+++ b/mappings/net/minecraft/obfuscate/DontObfuscate.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/obfuscate/DontObfuscate net/minecraft/obfuscate/DontObfuscate
+CLASS net/minecraft/obfuscate/DontObfuscate
 	COMMENT This annotation applies to some unobfuscated elements within the Minecraft
 	COMMENT source code.
 	COMMENT
--- a/mappings/net/minecraft/obfuscate/DontObfuscate.mapping
+++ b/mappings/net/minecraft/obfuscate/DontObfuscate.mapping
@@ -5,21 +5,19 @@ CLASS net/minecraft/obfuscate/DontObfuscate net/minecraft/obfuscate/DontObfuscat
 	COMMENT <p>Its behavior appears as follows:
 	COMMENT <ul>
 	COMMENT <li>The annotation itself is not obfuscated.</li>
-	COMMENT <li>If a class is annotated, it is not obfuscated.</li>
+	COMMENT <li>If a class is annotated, it is not obfuscated. It's not yet clear if its
+	COMMENT members will always become deobfuscated as well.</li>
 	COMMENT <li>If a member is annotated, it and its containing class is not obfuscated,
 	COMMENT but other members in the same class may stay obfuscated.</li>
 	COMMENT </ul>
 	COMMENT
+	COMMENT <p>Visit the use page for the usage of this annotation.
+	COMMENT
 	COMMENT <p>In addition, single-abstract-method interfaces used as lambda expressions
-	COMMENT have their single abstract method unobfuscated without this annotation.
+	COMMENT always have their single abstract method unobfuscated per proguard behavior.
 	COMMENT
 	COMMENT <p>This annotation is not {@link java.lang.annotation.Documented}, and hence
 	COMMENT will not appear in the generated javadoc for annotated elements.
 	COMMENT
-	COMMENT @see net.minecraft.client.Main#main(String[])
-	COMMENT @see net.minecraft.data.Main#main(String[])
-	COMMENT @see net.minecraft.server.Main#main(String[])
-	COMMENT @see net.minecraft.client.ClientBrandRetriever#getClientModName()
-	COMMENT @see net.minecraft.server.MinecraftServer#getServerModName()
 	COMMENT @see net.minecraft.util.CubicSampler.RgbFetcher
-	COMMENT @see net.minecraft.world.level.ColorResolver#getColor()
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/obfuscate/DontObfuscate.mapping
```diff
@@ -0,0 +1,25 @@
...
+	COMMENT <p>This annotation is not {@link java.lang.annotation.Documented}, and hence
+	COMMENT will not appear in the generated javadoc for annotated elements.
+	COMMENT
+	COMMENT @see net.minecraft.client.Main#main(String[])
+	COMMENT @see net.minecraft.data.Main#main(String[])
+	COMMENT @see net.minecraft.server.Main#main(String[])
+	COMMENT @see net.minecraft.client.ClientBrandRetriever#getClientModName()
+	COMMENT @see net.minecraft.server.MinecraftServer#getServerModName()
+	COMMENT @see net.minecraft.util.CubicSampler.RgbFetcher
+	COMMENT @see net.minecraft.world.level.ColorResolver#getColor()
```
Juuxel: Ah, that's why `ColorResolver` was unobfuscated... I wonder *why* they did this.

Juuxel (approved):

mappings/net/minecraft/obfuscate/DontObfuscate.mapping
liach: maybe they just had the anno by accident. i mean, that debug anno is always on methods except for one bee goal class

mappings/net/minecraft/obfuscate/DontObfuscate.mapping
```diff
@@ -0,0 +1,25 @@
...
+	COMMENT <p>Its behavior appears as follows:
+	COMMENT <ul>
+	COMMENT <li>The annotation itself is not obfuscated.</li>
+	COMMENT <li>If a class is annotated, it is not obfuscated.</li>
+	COMMENT <li>If a member is annotated, it and its containing class is not obfuscated,
+	COMMENT but other members in the same class may stay obfuscated.</li>
+	COMMENT </ul>
+	COMMENT
+	COMMENT <p>In addition, single-abstract-method interfaces used as lambda expressions
+	COMMENT have their single abstract method unobfuscated without this annotation.
```
Earthcomputer: I think this is a bug in proguard more than anything

mappings/net/minecraft/obfuscate/DontObfuscate.mapping
Earthcomputer: This is way too many usages in the docs, if I want to find all usages I have intellij. Probably better to just list one or two of them.

mappings/net/minecraft/obfuscate/DontObfuscate.mapping
```diff
@@ -0,0 +1,25 @@
+CLASS net/minecraft/obfuscate/DontObfuscate net/minecraft/obfuscate/DontObfuscate
+	COMMENT This annotation applies to some unobfuscated elements within the Minecraft
+	COMMENT source code.
+	COMMENT
+	COMMENT <p>Its behavior appears as follows:
+	COMMENT <ul>
+	COMMENT <li>The annotation itself is not obfuscated.</li>
+	COMMENT <li>If a class is annotated, it is not obfuscated.</li>
```
Earthcomputer: Does it make all members not obfuscated? Should probably clarify that.

liach: I don't know. In the only case it's used in a class, that class is used a functional interface whose member won't be obfuscated in any case.

mappings/net/minecraft/obfuscate/DontObfuscate.mapping
liach: Is it? Can't find anything at https://github.com/Guardsquare/proguard/issues

mappings/net/minecraft/obfuscate/DontObfuscate.mapping
liach: Hmm, this becomes a list when rendered, so it shouldn't be too bad? Also when browsing pure javadoc on web, you can't have access to the code unfortunately. I should test how the javadoc use page works with this. If they are shown in the use page, I will definitely cut it. Otherwise, keeping this long list isn't really problematic.

mappings/net/minecraft/obfuscate/DontObfuscate.mapping
```diff
@@ -0,0 +1,25 @@
+CLASS net/minecraft/obfuscate/DontObfuscate net/minecraft/obfuscate/DontObfuscate
```
enbrain:
```suggestion
CLASS net/minecraft/obfuscate/DontObfuscate
```

mappings/net/minecraft/obfuscate/DontObfuscate.mapping
liach: Have cut it already to only include the rgb fetcher and color resolver.

mappings/net/minecraft/obfuscate/DontObfuscate.mapping
liach: Edited to say this is proguard's behavior than anno related

mappings/net/minecraft/obfuscate/DontObfuscate.mapping
liach: Just added that it's not clear; but it won't obfuscate more stuff

YanisBft (approved):

enbrain (approved):

PR #2245, opened 2021-04-01
Generate use page for javadoc

### LINKED ISSUE #2093: Including use pages in javadoc
I recently explored that we can include the use pages for javadoc in the generated yarn javadoc.

Brief information about use pages:
Use pages details where a class occurs in code, but it only considers their presence in signatures, such as usage in fields, method return type/parameter types (including generics), subclasses. 

For instance, class with utility methods (`net.minecraft.util.Util`, `net.minecraft.util.dynamic.Codecs`) will not appear on use page as the class itself is not part of any signature (see [the use page of `java.util.Collections`](https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/util/class-use/Collections.html)). Meanwhile, `net.minecraft.util.BlockPos`, `net.minecraft.util.BlockState` would have larger use pages as they occur very frequently in method signatures and fields.

Drawbacks of use pages:
The built javadoc jar size grows in size by about 40% after generating use pages. So it's a deliberation on whether they are worth it.

Testing use pages:
You can start generating use pages by adding a line 
```gradle
		it.use()
```
at here: https://github.com/FabricMC/yarn/blob/0025fbf845a27efa50d44ba1e2bac7c028eea57d/build.gradle#L729

Then run `gradle javadocJar` to see if the output javadoc/Jar is desirable.

liach:
Closes #2093

I have decided to act now since the use page is a good way to look at annotations (see https://github.com/FabricMC/yarn/pull/2241#discussion_r605530446), which are now shipped by mojang. Also it's good for learning the usage of types as any part of signatures.

Notice that

> The built javadoc jar size grows in size by about 40% after generating use pages.

so this may need some deliberation.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -929,6 +929,7 @@ javadoc {
 		)
 		taglets "net.fabricmc.mappingpoet.jd.MappingTaglet"
 		// taglet path, header, extra stylesheet settings deferred
+		it.use()
 
 		addBooleanOption "-allow-script-in-comments", true
 		links(


modmuss50:
40% bigger is qutie a lot considering it was almost 30MB compressed already

liach:
> 
> 
> 40% bigger is qutie a lot considering it was almost 30MB compressed already

Yes. The current built javadocJar from this branch is 40MB.

modmuss50 (approved):
Sure, if it becomes an issue its easy enough to turn off. Storage isnt an issue atm anyway.

PR #2246, opened 2021-04-01
Some network stuff, and minor atomic stack/affine transformation revisit

liach:


Juuxel (approved):

YanisBft (approved):

modmuss50 (approved):

modmuss50:
Good to merge once conflict fixed.

liach:
Found another field that can be mapped. ALso I think I should try unpicking

liach:
Turns out unpick is weird. In my first run of `yarnU`, it unpicks the constant for max text length; in the second run, it unpicks the constant for the max var long size, but not that for the max var int size. I don't know how unpick works, but apparently it is working in an inconsistent and unreliable way. So much it... We are good with javadocs and `{@value}` links I guess

liach:
Hmm, so yeah, I shouldn't have used unpick. Those constants except the string and text length ones are private.

So is the goal of unpick to make constants exposed in public APIs more understandable? And it doesn't worry too much about impl detail constants, I assume?

PR #2247, opened 2021-04-01
Yeet the unobfuscated methods from GlStateManager

Sollace:
Yeeet

Also touched up some parameter names that were missing in there and in RenderSystem


Juuxel (approved):

liach (approved):

liach:
Honestly I think our names are better than mojang ones. But yes, we don't usually map mojang names again. We can always restore the names if we need in the future.

YanisBft (approved):
It hurts my eyes but if we don't the choice

Sollace:
> 
> 
> Honestly I think our names are better than mojang ones. But yes, we don't usually map mojang names again. We can always restore the names if we need in the future.

Same, but oh well. :/

modmuss50 (approved):

PR #2249, opened 2021-04-01
Add a task to generate mappings for the package-info classes.

modmuss50:
If you have a better idea of where to put these or what to call them let me know. I have spot checked them and they all seem correctly identified. Might not be a bad idea to put a comment on them? 

This is also a good exeriment with using a build-logic sub project to provide well, the build logic for yarn. The long term goal will be to cleanup the yarn build script and move the logic into well defined tasks.


BeetMacol (approved):

build-logic/src/main/groovy/net/fabricmc/yarn/buildlogic/task/GeneratePackageInfoMappingsTask.groovy
```diff
@@ -0,0 +1,79 @@
...
+import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.OutputDirectory
+import org.gradle.api.tasks.TaskAction
+import org.objectweb.asm.ClassReader
+import org.objectweb.asm.Opcodes
+import org.objectweb.asm.tree.ClassNode
+
+import java.util.zip.ZipFile
+
+@CompileStatic
```
Juuxel, line 17: Just out of curiosity, what does this annotation do?

build.gradle
```diff
@@ -985,6 +986,13 @@ task javadocJar(type: Jar, dependsOn: ["javadoc"]) {
 	archiveClassifier = 'javadoc'
 }
 
+generatePackageInfoMappings {
+	dependsOn mapIntermediaryJar
+	inputJar = intermediaryJar
+	packageName = "net/minecraft/util/packageinfo/"
+	outputDir = file("mappings/net/minecraft/util/packageinfo")
```
Juuxel: Not really sure if `util` is the best place for these. That's a package for actual utility classes, not a dump for unused things.

Juuxel (commented):
Naming these something like `PackageInfo1234` would be nicer IMO, but this works too.

build-logic/src/main/groovy/net/fabricmc/yarn/buildlogic/task/GeneratePackageInfoMappingsTask.groovy
kb-1000: This annotation makes Groovy less dynamic, and makes it generate more Java-like bytecode (as far as I know)

modmuss50: "This will let the Groovy compiler use compile time checks in the style of Java then perform static compilation" basically makes groovy a bit more type safe instead of it blowing up at runtime. Should really be put onto all groovy classes in this case.

build.gradle
modmuss50: Yeah... this was up in the air, do you have any suggestions?

Juuxel: Maybe even `net.minecraft.unused.packageinfo`? These classes shouldn't really be seeing any use (unless Yarn switches to follow Mojang's package structure :yeefuckinhaw:).

modmuss50: That makes more sense yes, its clear they wont be used, I will move them 👍

liach: imo util can just be unused.

liach (approved):

modmuss50:
They are now named like: `net/minecraft/unused/packageinfo/PackageInfo5984`

BeetMacol (approved):

Juuxel (approved):

PR #2263, opened 2021-04-02
Shorten set_block_state_flags

Earthcomputer:
Renamed `PROPAGATE_CHANGE` to `NOTIFY_NEIGHBORS`, which allows for `NOTIFY_ALL` to encompass both that and `NOTIFY_LISTENERS`.
Given that `NOTIFY_ALL` is really commonly used, it's important that it shouldn't be as long as it was.

### DIFF
--- a/mappings/net/minecraft/world/ModifiableWorld.mapping
+++ b/mappings/net/minecraft/world/ModifiableWorld.mapping
@@ -11,9 +11,7 @@ CLASS net/minecraft/class_1945 net/minecraft/world/ModifiableWorld
 		COMMENT change persisted across loads. It's recommended to check whether this world is client before
 		COMMENT interacting with the world in this way.</p>
 		COMMENT
-		COMMENT <p>For the accepted values of the flags, see {@link net.fabricmc.yarn.constants.SetBlockStateFlags}.
-		COMMENT
-		COMMENT @see net.fabricmc.yarn.constants.SetBlockStateFlags
+		COMMENT <p>See {@link #setBlockState(BlockPos, BlockState, int)} for a list of accepted flags.
 		ARG 1 pos
 			COMMENT the target position
 		ARG 2 state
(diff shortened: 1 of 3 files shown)


liach (approved):

liach:
On a side note, we might need a minor cleanup on this since the fields are no longer in the context of a constant holding class. We need to document for each field that it's used by setBlockState's flag. Also there is a few broken jd links in `ModifiableWorld`.

mappings/net/minecraft/world/ModifiableWorld.mapping
```diff
@@ -44,10 +42,20 @@ CLASS net/minecraft/class_1945 net/minecraft/world/ModifiableWorld
...
+		COMMENT <ul>
+		COMMENT <li>{@link net.minecraft.block.Block#NOTIFY_ALL Block.NOTIFY_ALL}</li>
+		COMMENT <li>{@link net.minecraft.block.Block#NOTIFY_NEIGHBORS Block.NOTIFY_NEIGHBORS}</li>
+		COMMENT <li>{@link net.minecraft.block.Block#NOTIFY_LISTENERS Block.NOTIFY_LISTENERS}</li>
+		COMMENT <li>{@link net.minecraft.block.Block#NO_REDRAW Block.NO_REDRAW}</li>
+		COMMENT <li>{@link net.minecraft.block.Block#REDRAW_ON_MAIN_THREAD Block.REDRAW_ON_MAIN_THREAD}</li>
+		COMMENT <li>{@link net.minecraft.block.Block#FORCE_STATE Block.FORCE_STATE}</li>
+		COMMENT <li>{@link net.minecraft.block.Block#SKIP_DROPS Block.SKIP_DROPS}</li>
+		COMMENT <li>{@link net.minecraft.block.Block#MOVED Block.MOVED}</li>
+		COMMENT <li>{@link net.minecraft.block.Block#SKIP_LIGHTING_UPDATES Block.SKIP_LIGHTING_UPDATES}</li>
```
liach, line 55: Consider using `{@value net.minecraft.block.Block#NOTIFY_ALL} - "the thing they do"`. In javadoc, it renders with the actual value of that flag, like 1, 2, 4, 8, etc.

Earthcomputer: No, I would rather encourage people to use the named constants than the literal values

liach (approved):

YanisBft (approved):

modmuss50 (approved):

PR #2265, opened 2021-04-03
Bee mappings

Juuxel:
:bee: :bee: :bee:

There are a lot more constants to map, but they sometimes have the same value so I just picked the easiest ones :smile:


liach:
imo for nbt keys, _KEY suffix is enough, no need of _NBT_KEY suffix

Juuxel:
Sure, probably deserves a mention in conventions.md if this pattern is used in other entities (didn't check since I only looked at bees lol)

Juuxel:
Okay, I renamed them to just `*_KEY` now. It also looks like bees are the exception here: most entity classes don't declare them in fields. (~~Another reason why they're the best mob~~)

liach (approved):

YanisBft (approved):

PR #2270, opened 2021-04-04
Convert unpick to intermediary

### LINKED ISSUE #2208: Use intermediary names in unpick configs
Why?
- There should be one authoritative source for what the yarn names are, and those are the files under the `mappings` directory. All other references to classes, fields and methods should be in intermediary format.
- Whenever a yarn name is in an unpick config, and that yarn name is renamed, you have to change it both in `mappings` and in the unpick config.
- Avoids the need for remapping unpick config in yarn's buildscript.
- Once Enigma has support for unpick config, it shouldn't be too much of a pain to edit or write unpick config in intermediary.

liach:
Done largely with https://gist.github.com/4100d54f8a8329e5484e715e15b7d710 and some style fixup.

The constants are still present and shows up in the decompiled CFR code.

Closes #2208


build.gradle
```diff
@@ -666,66 +666,7 @@ task combineUnpickDefinitions(type : FileOutput) {
 	}
 }
 
-task remapUnpickDefinitionsIntermediary(type : FileOutputInput, dependsOn: [combineUnpickDefinitions, buildYarnTiny]) {
```
modmuss50, line 669: The output in the jar is now going to be wrong, this needs to be changed to remap to named to be included in the exported jars.

modmuss50 (changes requested):

liach:
Side note: If you want to copy the notation for unpick, You can for now go to the javadocs of yarn, copy its mixin notation, and remove the `L` in front of the class name, replace `;` with a ` `, and finally add a space in front of the `(` in the desc. This is for methods. 

For fields, you can do largely the same of removing `L` and replacing `;` with ` `; in addition, you remove everything behind the `:` as unpick doesn't care about field type.

Shnupbups:
but why though
just makes the unpick files more annoying to write and harder to read

build.gradle
liach: What? Unpick is intermediary and applied to intermediary (unpicked intermediary). named jar has always been unpicked and derived from the unpicked intermediary.

I honestly don't know what you are talking about, as I am looking at the cfr decompiled result and it has no problems for remapping (except those ones where I explicitly override the desc)

liach:
> 
> 
> but why though
> just makes the unpick files more annoying to write and harder to read

Should have debated #2208. This presumably makes it more change resistant? Imo not really, we have javadoc using named and it works pretty well, doesn't it. So yeah this pr is just a platform for debate

build.gradle
modmuss50: The published jar files that loom use need to contain named unpick definitions, this is no longer the case.

liach: Done.

modmuss50 (approved):
LGTM.

Thanks 👍

Earthcomputer:
Can we wait until we have something in Enigma?

YanisBft:
I agree with Earth, it will be a pain to do manually

liach:
hmm, did you just type the whole desc out by hand? fyi, i always generated javadoc and started with the mixin notations from there. it's quite convenient.

Earthcomputer:
> hmm, did you just type the whole desc out by hand? fyi, i always generated javadoc and started with the mixin notations from there. it's quite convenient.

I've been using mcdev's copy mixin target reference tool, which gets you the yarn name and descriptor.

liach:
There is no significant advantage on leaving this open as the unpick definitions will change quickly and the script for migration is already available in the description. Closing.

PR #2279, opened 2021-04-05
Add MONOCHROME_LOGO to Option

Darxoon:
21w13a added the "Monochrome Logo" toggle in Accessibility settings, which makes the background in the mojang logo turn black instead of red, to reduce strain on the eyes. This change will give the Option for that the name MONOCHROME_LOGO

### DIFF
--- a/mappings/net/minecraft/client/option/GameOptions.mapping
+++ b/mappings/net/minecraft/client/option/GameOptions.mapping
@@ -109,7 +109,7 @@ CLASS net/minecraft/class_315 net/minecraft/client/option/GameOptions
 	FIELD field_26845 keySocialInteractions Lnet/minecraft/class_304;
 	FIELD field_26926 hideMatchedNames Z
 	FIELD field_28777 hideBundleTutorial Z
-	FIELD field_32156 monochromeBackground Z
+	FIELD field_32156 monochromeLogo Z
 	METHOD <init> (Lnet/minecraft/class_310;Ljava/io/File;)V
 		ARG 1 client
 		ARG 2 optionsFile
--- a/mappings/net/minecraft/client/option/GameOptions.mapping
+++ b/mappings/net/minecraft/client/option/GameOptions.mapping
@@ -109,6 +109,7 @@ CLASS net/minecraft/class_315 net/minecraft/client/option/GameOptions
 	FIELD field_26845 keySocialInteractions Lnet/minecraft/class_304;
 	FIELD field_26926 hideMatchedNames Z
 	FIELD field_28777 hideBundleTutorial Z
+	FIELD field_32156 monochromeBackground Z
 	METHOD <init> (Lnet/minecraft/class_310;Ljava/io/File;)V
 		ARG 1 client
 		ARG 2 optionsFile
--- a/mappings/net/minecraft/client/option/Option.mapping
+++ b/mappings/net/minecraft/client/option/Option.mapping
@@ -59,6 +59,7 @@ CLASS net/minecraft/class_316 net/minecraft/client/option/Option
 	FIELD field_27956 TOGGLE_TEXT Lnet/minecraft/class_2561;
 	FIELD field_27957 HOLD_TEXT Lnet/minecraft/class_2561;
 	FIELD field_32146 MONOCHROME_LOGO Lnet/minecraft/class_4064;
+	FIELD field_32148 MONOCHROME_LOGO_TOOLTIP Lnet/minecraft/class_2561;
 	METHOD <init> (Ljava/lang/String;)V
 		ARG 1 key
 	METHOD method_18513 (Lnet/minecraft/class_315;)Ljava/lang/Double;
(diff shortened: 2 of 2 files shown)


YanisBft (approved):
Wait, Mojang actually added kind of a dark mode at game start \*-\*

BeetMacol (approved):

haykam821 (approved):

haykam821:
Thanks for the pull request! :+1:

Darxoon:
Oh sorry, I didn't see that this was approved when I added those two commits

mappings/net/minecraft/client/option/GameOptions.mapping
```diff
@@ -109,6 +109,7 @@ CLASS net/minecraft/class_315 net/minecraft/client/option/GameOptions
 	FIELD field_26845 keySocialInteractions Lnet/minecraft/class_304;
 	FIELD field_26926 hideMatchedNames Z
 	FIELD field_28777 hideBundleTutorial Z
+	FIELD field_32156 monochromeBackground Z
```
haykam821:
```suggestion
	FIELD field_32156 monochromeLogo Z
```

Something that might be useful in a later pull request is to add Javadoc showing how these options are saved. For example, this one is `darkMojangStudiosBackground`.

mappings/net/minecraft/client/option/Option.mapping
```diff
@@ -58,6 +58,8 @@ CLASS net/minecraft/class_316 net/minecraft/client/option/Option
 	FIELD field_26925 HIDE_MATCHED_NAMES_TOOLTIP Lnet/minecraft/class_2561;
 	FIELD field_27956 TOGGLE_TEXT Lnet/minecraft/class_2561;
 	FIELD field_27957 HOLD_TEXT Lnet/minecraft/class_2561;
+	FIELD field_32146 MONOCHROME_LOGO Lnet/minecraft/class_4064;
```
haykam821, line 61: Can you map the lambda method parameters here?

haykam821 (changes requested):

mappings/net/minecraft/client/option/Option.mapping
```diff
@@ -58,6 +58,8 @@ CLASS net/minecraft/class_316 net/minecraft/client/option/Option
...
+	FIELD field_32148 MONOCHROME_LOGO_TOOLTIP Lnet/minecraft/class_2561;
```
BeetMacol, line 62: Wouldn't `MONOCHROME_BACKGROUND` be better here? I think it is called Monochrome Background, not logo in-game.

haykam821: It is called 'Monochrome Logo' in the Accessibility Settings screen:

Monochrome Logo option

Darxoon: No, actually it's called Monochrome Logo

mappings/net/minecraft/client/option/Option.mapping
Darxoon: Do you mean that? I'm not sure but for me, it is already mapped

Darxoon: image

haykam821: All four of the lambda parameters on the line are unmapped. In the Enigma editor launched through `./gradlew yarn`, they will be red.

liach (approved):

mappings/net/minecraft/client/option/Option.mapping
Darxoon: Oh okay. This is the first time I'm doing something with mappings, I don't know how to do that

liach: You just open a command line interface like how you open one for mod dev (at the root directory for yarn), and run `./gradlew yarnU` to open yarn with unpicked constants. Then you go to that class, and some lambda parameters will be mappable. You can rename there.

On a side note, I don't think it's worth requesting a change for lambda params, as they aren't really used outside of enigma; mapped source doesn't care, javadoc can't render them, and they are highly susceptible to code changes (lost across updates very easily)

enbrain (approved):

BeetMacol (approved):

mappings/net/minecraft/client/option/Option.mapping
```diff
@@ -440,3 +442,9 @@ CLASS net/minecraft/class_316 net/minecraft/client/option/Option
 		ARG 0 gameOptions
 	METHOD method_32593 (Lnet/minecraft/class_315;)Ljava/lang/Integer;
 		ARG 0 gameOptions
+	METHOD method_35708 (Lnet/minecraft/class_315;)Ljava/lang/Boolean;
+		ARG 0 gameOptions
+	METHOD method_35709 (Lnet/minecraft/class_315;Lnet/minecraft/class_316;Ljava/lang/Boolean;)V
+		ARG 0 gameOptions
+		ARG 1 option
+		ARG 2 value
```
haykam821: For consistency with the other lambda parameters in this class, this should be named like the `GameOptions` field:

```suggestion
		ARG 2 monochromeLogo
```

haykam821 (changes requested):

liach (approved):

haykam821 (approved):

enbrain (approved):

PR #2280, opened 2021-04-05
field_32657 -> MAX_PARTICLE_COUNT

Sollace:
This is a constant used to set the size of the queue containing particles on the client.

I don't know how to do unpick stuff, but this probably should be unpicked too. :/


liach (approved):

YanisBft (approved):

liach:
Fyi this one isn't worthy to unpick. Unpick is really designed for values exposed in public api (like int enumerations, int flags, special chars, etc) than for implementation details. If you want to unpick here, you would target all usages of evicting queue constructor, which imo is quite dumb.

Sollace:
@liach I thought unpick let you specify a call in a specific method?

liach:
> 
> 
> @liach I thought unpick let you specify a call in a specific method?

Nah. This is how it works:
You can specify constant groups, meaning a type of constant.
Then you can declare constants, which are related to each field.
Unpick allow you to associate constant groups with method params/returns, so it will scan how the methods are called/returned/return values used etc. and replace stuff correspondingly. And yes, if it interacts with an instance field, you cannot tell unpick to handle that (yet)

PR #2288, opened 2021-04-06
Map some constants found in the world classes

Sollace:


mappings/net/minecraft/server/world/ServerWorld.mapping
```diff
@@ -22,6 +22,8 @@ CLASS net/minecraft/class_3218 net/minecraft/server/world/ServerWorld
 	FIELD field_26934 entityList Lnet/minecraft/class_5574;
 	FIELD field_26935 entityManager Lnet/minecraft/class_5579;
 	FIELD field_28859 sleepManager Lnet/minecraft/class_5838;
+	FIELD field_29768 ENTITY_TICK_FREQUENCY I
+		COMMENT The number of ticks the world will wait between entity updates.
```
liach: Hmm, so this is not `1`? Interesting, I thought entities are always ticked

mappings/net/minecraft/world/World.mapping
```diff
@@ -19,7 +23,7 @@ CLASS net/minecraft/class_1937 net/minecraft/world/World
 	FIELD field_9234 thunderGradient F
 	FIELD field_9235 rainGradient F
 	FIELD field_9236 isClient Z
-	FIELD field_9238 unusedIncrement I
+	FIELD field_9238 lcgBlockSeedIncrement I
```
liach, line 26: Just curious, what is `lcg`?

mappings/net/minecraft/server/world/ServerWorld.mapping
liach:
```suggestion
		COMMENT The number of ticks ({@value}) the world will wait between entity updates.
```
Javadoc will render the value on the web page with this little tag.

mappings/net/minecraft/world/World.mapping
enbrain: It's linear congruential generator. Might be worth documenting it if this abbreviation is not so common.

Sollace: The abbreviation was used in existing names. I just copied the convention. If you want it documented we can do that in a separate PR.

mappings/net/minecraft/server/world/ServerWorld.mapping
Sollace: @liach I thought so too, but evidently not. At least, I think I should change the description a little. Looks like it only applies when the server is idle.

```
        boolean bl2 = bl4 = !this.players.isEmpty() || !this.getForcedChunks().isEmpty();
        if (bl4) {
            this.resetIdleTimeout();
        }
        if (bl4 || this.idleTimeout++ < 300 /* ENTITY_TICK_FREQUENCY */) {
            lv.push("entities");
```

liach (approved):

enbrain (approved):

YanisBft (approved):

PR #2289, opened 2021-04-06
Javadoc linting task

Juuxel:
Adds a `javadocLint` task that ensures the current javadoc comments follow Yarn's guidelines for javadocs, namely:

> Write quick descriptions for parameter javadocs as well as `@return` tags, with no uppercase or period. Add parameter docs to the parameter itself instead of using the `@param` tag.

(The `@return` tags are not currently checked, nor are some other guidelines defined in `CONVENTIONS.md`. The parameter style seems to be broken most often.)

All javadoc comments are also fixed here to follow the conventions.

<details><summary>Example output of <code>javadocLint</code> when there are errors:</summary>
<pre>
lint: net/minecraft/world/gen/chunk/ChunkGenerator.locateStructure.radius: parameter javadoc ends with '.'
lint: net/minecraft/world/gen/chunk/ChunkGenerator.locateStructure.radius: parameter javadoc starts with uppercase word 'The'
lint: net/minecraft/entity/player/PlayerEntity.dropItem.throwRandomly: parameter javadoc starts with uppercase word 'If'
lint: net/minecraft/entity/EyeOfEnderEntity.initTargetPos: method contains parameter docs, which should be on the parameter itself
lint: net/minecraft/screen/ScreenHandler.onSlotClick.actionType: parameter javadoc starts with uppercase word 'The'
</pre>
</details>

### DIFF
--- a/build-logic/src/main/groovy/net/fabricmc/yarn/buildlogic/task/JavadocLintTask.groovy
+++ b/build-logic/src/main/groovy/net/fabricmc/yarn/buildlogic/task/JavadocLintTask.groovy
@@ -57,7 +57,7 @@ class JavadocLintTask extends DefaultTask {
                     if (javadoc.contains(PARAMETER_TAG_START)) {
                         int paramIndex = javadoc.indexOf(PARAMETER_TAG_START) + PARAMETER_TAG_START.length()
 
-                        if (javadoc.length() > paramIndex || javadoc.charAt(paramIndex) != '<' as char) {
+                        if (javadoc.length() <= paramIndex || javadoc.charAt(paramIndex) != '<' as char) {
                             localErrors.add("method contains parameter docs, which should be on the parameter itself")
                         }
                     }
--- a/build-logic/src/main/groovy/net/fabricmc/yarn/buildlogic/task/JavadocLintTask.groovy
+++ b/build-logic/src/main/groovy/net/fabricmc/yarn/buildlogic/task/JavadocLintTask.groovy
@@ -13,14 +13,14 @@ import groovy.transform.CompileStatic
 import org.gradle.api.DefaultTask
 import org.gradle.api.GradleException
 import org.gradle.api.file.DirectoryProperty
-import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.InputDirectory
 import org.gradle.api.tasks.TaskAction
 
 @CompileStatic
 class JavadocLintTask extends DefaultTask {
     private static final String PARAMETER_TAG_START = "@param "
 
-    @InputFiles
+    @InputDirectory
     DirectoryProperty mappingDirectory = project.objects.directoryProperty()
 
     @TaskAction
--- a/build-logic/src/main/groovy/net/fabricmc/yarn/buildlogic/task/JavadocLintTask.groovy
+++ b/build-logic/src/main/groovy/net/fabricmc/yarn/buildlogic/task/JavadocLintTask.groovy
@@ -23,6 +23,11 @@ class JavadocLintTask extends DefaultTask {
     @InputDirectory
     DirectoryProperty mappingDirectory = project.objects.directoryProperty()
 
+    JavadocLintTask() {
+        // Only run when inputs have changed
+        outputs.upToDateWhen { true }
+    }
+
     @TaskAction
     def run() {
         def saveParameters = new MappingSaveParameters(MappingFileNameFormat.BY_DEOBF)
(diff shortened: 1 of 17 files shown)


liach:
Can you just reuse existing frameworks' javadoc lint tasks and run them on the generated fake source? Imo your current work would be prone to be out of date.

build-logic/src/main/groovy/net/fabricmc/yarn/buildlogic/task/JavadocLintTask.groovy
```diff
@@ -0,0 +1,103 @@
...
+                        // ignore single-letter "words" (like X or Z)
+                        if (word.length() > 1) {
+                            localErrors.add("parameter javadoc starts with uppercase word '${word}'")
+                        }
+                    }
+                }
+
+                if (entry instanceof MethodEntry) {
+                    // check for non-type parameters declared as @param
+                    if (javadoc.contains(PARAMETER_TAG_START)) {
```
liach, line 62: Lolno, what about type parameters

build-logic/src/main/groovy/net/fabricmc/yarn/buildlogic/task/JavadocLintTask.groovy
```diff
@@ -0,0 +1,103 @@
...
+                        int paramIndex = javadoc.indexOf(PARAMETER_TAG_START) + PARAMETER_TAG_START.length()
+
+                        if (javadoc.length() <= paramIndex || javadoc.charAt(paramIndex) != '<' as char) {
+                            localErrors.add("method contains parameter docs, which should be on the parameter itself")
```
liach, line 66: Rip my bad, you have considered this case, good

liach (commented):
Imo we should keep what's written on our own minimal still, so only that `@param` presence one. otherwise, for what we can check at source level, I still recommend source-based lints.

Juuxel:
Hmm, I think it might be possible to run something like Checkstyle on the generated sources instead? One caveat is that tools processing them can't detect `@param` declarations on methods because parameter docs are converted to that in the source.

build-logic/src/main/groovy/net/fabricmc/yarn/buildlogic/task/JavadocLintTask.groovy
liach: Hmm, there should be no need for `== paramIndex` case, maybe just use `javadoc.length() < paramIndex`

Juuxel: Wouldn't that technically be prone to throwing an index out of bounds exception when the doc ends with `@param `? (Very unlikely, but still possible)

build-logic/src/main/groovy/net/fabricmc/yarn/buildlogic/task/JavadocLintTask.groovy
```diff
@@ -0,0 +1,103 @@
...
+@CompileStatic
+class JavadocLintTask extends DefaultTask {
+    private static final String PARAMETER_TAG_START = "@param "
+
+    @InputDirectory
+    DirectoryProperty mappingDirectory = project.objects.directoryProperty()
+
+    JavadocLintTask() {
+        // Only run when inputs have changed
+        outputs.upToDateWhen { true }
```
liach, line 28: No? I think this will make groovy always treat this task as up-to-date and cause it to not run? If you depend on input you should just remove this up to date spec

Juuxel: No, I've tested that this works properly. The default up-to-date checking does not work because this task has no outputs, but the task still gets run if it hasn't been run with these inputs before.

Juuxel: If you're curious, the up-to-date checks are done like this (from [the gradle user guide](https://docs.gradle.org/6.8.3/userguide/more_about_tasks.html#sec:task_outcomes):

gradle up-to-date checks

Since there are no outputs, this fits into the second category. Note that up-to-date checks check both the outputs (that I configure manually) *and* the inputs.

Juuxel:
@liach I made a WIP (non-commited) Checkstyle check for the parameter checks, but I noticed one major issue: With Checkstyle, you can't see *what* causes the problem, only the file and the line/column numbers.

That's most likely true for other source-based tools as well (ktlint for example also does it like this iirc), so a custom mapping-based library is better IMO.

Juuxel:
Side note: using Groovy for the Gradle task was a bit of a pain as it can't handle the wildcard type `Entry<?>` that well. I had to add `as Entry<?>` casts to get the static compilation working (the dynamic compilation/typing would probably have produced more readable code).

modmuss50:
I’m not convinced groovy is the way forward for this build logic stuff. There are some other issues I came across with the whole buildscript plugin thing that aren’t a fault of groovy.

Juuxel:
Superseded by FabricMC/fabric-filament#2.

PR #2294, opened 2021-04-07
Ensure that releases are not ran in parallel

modmuss50:
This resolves builds beign pushed to maven in the wrong order: see the final few 13a builds here: https://meta.fabricmc.net/v2/versions/yarn/

This happens when PRs are merged all at once and its a race to release.


modmuss50:
Seems to have built fine, going to merge this, let me know if there are any issues.

modmuss50:
This seems to work well, I am going to bump the timeout on it however as 15 mins was pushing it when merging 6 prs.

PR #2295, opened 2021-04-07
Remove unnecessary paragraph end tags from documentation

haykam821:


liach (approved):

mappings/net/minecraft/item/Item.mapping
```diff
@@ -115,7 +115,7 @@ CLASS net/minecraft/class_1792 net/minecraft/item/Item
 		COMMENT Note that this is not the <b>only</b> way to achieve "effectiveness" when mining.
 		COMMENT Other items, such as shears on string, may use their own logic
 		COMMENT and calls to this method might not return a value consistent to this rule for those items.
-		COMMENT </p>
+		COMMENT
```
enbrain:
```suggestion
```

enbrain (changes requested):

YanisBft (approved):

enbrain (approved):
Approving because the above comment doesn't affect the final output. I can make a follow-up PR if you want me to do so.

PR #2297, opened 2021-04-08
Map Matrix3f and Matrix4f

enbrain:


Gegy (approved):

mappings/net/minecraft/util/math/Matrix4f.mapping
```diff
@@ -48,13 +54,36 @@ CLASS net/minecraft/class_1159 net/minecraft/util/math/Matrix4f
...
+	METHOD method_35434 load (Lnet/minecraft/class_1159;)V
+		ARG 1 source
+	METHOD method_35435 read (Ljava/nio/FloatBuffer;)V
+		ARG 1 buf
+	METHOD method_35436 read (Ljava/nio/FloatBuffer;Z)V
+		ARG 1 buf
+		ARG 2 transposed
+	METHOD method_35437 isInteger (F)Z
+		ARG 0 value
+	METHOD method_35438 readTransposed (Ljava/nio/FloatBuffer;)V
```
liach: Imo we should use row first/column first than transpose or not for the reading order.

enbrain: Just wondering, does unpick support boolean parameters? like `MatrixFormat.ROW_FIRST` / `MatrixFormat.COLUMN_FIRST`.

liach: Not yet. If it does, it will be in the `LiteralType` class of unpick.

liach (approved):
:heart:

YanisBft (approved):

PR #2309, opened 2021-04-10
Two new exception classes and a bunch of legacy names

liach:
Done in a way to avoid/minimize conflict with #2300

### DIFF
--- a/mappings/net/minecraft/client/render/model/json/JsonUnbakedModel.mapping
+++ b/mappings/net/minecraft/client/render/model/json/JsonUnbakedModel.mapping
@@ -1,5 +1,8 @@
 CLASS net/minecraft/class_793 net/minecraft/client/render/model/json/JsonUnbakedModel
 	FIELD field_21857 guiLight Lnet/minecraft/class_793$class_4751;
+	FIELD field_32793 TEXTURE_REFERENCE_INITIAL C
+		COMMENT The initial character ({@value}) of a texture reference in JSON; used to
+		COMMENT distinguish texture references from other references.
 	FIELD field_4244 ambientOcclusion Z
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/client/render/model/json/JsonUnbakedModel.mapping
```diff
@@ -56,25 +59,28 @@ CLASS net/minecraft/class_793 net/minecraft/client/render/model/json/JsonUnbaked
 		ARG 3 side
 		ARG 4 settings
 		ARG 5 id
+	METHOD method_35789 needsResolution ()Z
 	CLASS class_795 Deserializer
+		FIELD field_32794 DEFAULT_AMBIENT_OCCULSION Z
+			COMMENT The default value for ambient occulsion if unspecified in JSON; is {@value}.
```
YanisBft:
```suggestion
		FIELD field_32794 DEFAULT_AMBIENT_OCCLUSION Z
			COMMENT The default value for ambient occlusion if unspecified in JSON; is {@value}.
```

YanisBft (approved):

Juuxel (approved):

mappings/net/minecraft/client/render/model/json/JsonUnbakedModel.mapping
liach: Ah, have been always calling it incorrectly and wondering why it's named in such a "weird" way :laughing:

liach:
heck, my mobile is too trashy to fix conflicts. it's just reordering some entries to be in alphabetical order. hope modmuss can do it as i go to bed.

enbrain (approved):

PR #2317, opened 2021-04-10
Gradle 7 and JDK 16 on GHA

modmuss50:
Was easy enough.

Should be good to merge assuming the checks pass.


modmuss50:
Seems fine.

Issue #2217, opened 2021-03-31
Notice: Javadoc unavailable


liach:
Due to FabricMC/MappingPoet#21, Javadocs are currently unavailable for any 21w13a builds. Just a piece of notice.

PR #2318, opened 2021-04-10
Add javadoc linting task using Filament

Juuxel:
Basically #2289 but using Filament. Filament's `javadocLint` is also better and supports incremental running so it'll run faster and skip unmodified files :^)

Adds a `javadocLint` task that ensures the current javadoc comments follow Yarn's guidelines for javadocs, namely:

> Write quick descriptions for parameter javadocs as well as `@return` tags, with no uppercase or period. Add parameter docs to the parameter itself instead of using the `@param` tag.

(The `@return` tags are not currently checked, nor are some other guidelines defined in `CONVENTIONS.md`. The parameter style seems to be broken most often.)

All javadoc comments are also fixed here to follow the conventions.

<details><summary>Example output of <code>javadocLint</code> when there are errors:</summary>
<pre>
lint: net/minecraft/world/gen/chunk/ChunkGenerator.locateStructure.radius: parameter javadoc ends with '.'
lint: net/minecraft/world/gen/chunk/ChunkGenerator.locateStructure.radius: parameter javadoc starts with uppercase word 'The'
lint: net/minecraft/entity/player/PlayerEntity.dropItem.throwRandomly: parameter javadoc starts with uppercase word 'If'
lint: net/minecraft/entity/EyeOfEnderEntity.initTargetPos: method contains parameter docs, which should be on the parameter itself
lint: net/minecraft/screen/ScreenHandler.onSlotClick.actionType: parameter javadoc starts with uppercase word 'The'
</pre>
</details>

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,7 @@ plugins {
 	id 'maven-publish'
 	id 'java' // for jd gen
 	id 'org.cadixdev.licenser' version '0.5.1'
-	id 'net.fabricmc.filament' version '0.1.0'
+	id 'net.fabricmc.filament' version '0.2.0'
 }
 
 def minecraft_version = "21w14a"
@@ -990,6 +990,13 @@ generatePackageInfoMappings {
 	packageName = "net/minecraft/unused/packageinfo/"
 	outputDir = file("mappings/net/minecraft/unused/packageinfo")
 }
+
+javadocLint {
+	group = "javadoc generation"
+	mappingDirectory = file("mappings")
+}
+
+check.dependsOn javadocLint
 
 publishing {
 	publications {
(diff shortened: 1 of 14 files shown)


enbrain (approved):

modmuss50 (approved):

liach (approved):

liach:
On a side note, should we fail the build if the lint doesn't pass? Currently even if the javadoc task errors, the build passes.

Juuxel:
That's what the `check.dependsOn javadocLint` line does. When you or the CI runs `gradlew build`, it also runs the `check` task and therefore the `javadocLint` task. If I revert 2c90c8f locally and try to do `gradlew build`, the build fails.

YanisBft (approved):

liach:
Should we make regular javadoc just fail gradle build if it detects any format error as well?

Juuxel:
Hmm, I'm not sure about that. How aggressive is it with the format errors? (Like, are missing parameter docs or `@return`s considered format errors?)

liach:
> 
> 
> Hmm, I'm not sure about that. How aggressive is it with the format errors? (Like, are missing parameter docs or `@return`s considered format errors?)

Currently it's aggressive against lack of title for table, broken javadoc links, html format errors, etc.

In fact, the current check is too lenient; there can be errors that only break one page or break the whole gen. We currently cannot figure out a way to fail when the whole generation would fail (see #2217, javadoc was totally broken but doesn't fail because we have `failOnError = false`; if we have it `true` then it fails for the smallest error, and there is no choice in between apparently)

modmuss50 (approved):

PR #2319, opened 2021-04-10
A few more constant classes

liach:

### DIFF
--- a/mappings/net/minecraft/entity/player/HungerConstants.mapping
+++ b/mappings/net/minecraft/entity/player/HungerConstants.mapping
@@ -1,0 +1,24 @@
+CLASS net/minecraft/class_6075 net/minecraft/entity/player/HungerConstants
+	COMMENT Some constants on hunger values.
+	COMMENT
+	COMMENT @see HungerManager
+	FIELD field_30704 FULL_FOOD_LEVEL I
+		COMMENT The maximum food level ({@value}) allowed in a hunger manager.
+	FIELD field_30706 INITIAL_SATURATION_LEVEL F
+		COMMENT The initial saturation level ({@value}) for a newly created hunger manager.
+	FIELD field_30708 EXHAUSION_UNIT F
+		COMMENT A value {@value} that when the exhausion reaches, the exhausion minuses itself
+		COMMENT by to reduce the saturation or food level.
+	FIELD field_30709 SLOW_HEALING_STARVING_INTERVAL I
+		COMMENT When the food tick is a multiple of {@value}, the hunger manager may perform
+		COMMENT slow healing or starving logic.
+	FIELD field_30710 FAST_HEALING_INTERVAL I
+		COMMENT When the food tick is a multiple of {@value}, the hunger manager may perform
+		COMMENT fast healing logic.
+	FIELD field_30711 SLOW_HEALING_FOOD_LEVEL I
+		COMMENT The minimum food level ({@value}) required for the slow-healing mechanism.
+	FIELD field_30712 EXHAUSION_PER_HITPOINT I
+		COMMENT The exhausion from healing each hitpoint ({@value}), used for both fast and
+		COMMENT slow healing mechanisms.
+	FIELD field_30713 STARVING_FOOD_LEVEL I
+		COMMENT The maximum food level ({@value}) permitted for the starving mechanism to run.
--- a/mappings/net/minecraft/unused/UnknownWorldConstants6148.mapping
+++ b/mappings/net/minecraft/unused/UnknownWorldConstants6148.mapping
@@ -1,0 +1,7 @@
+CLASS net/minecraft/class_6148 net/minecraft/unused/UnknownWorldConstants6148
+	COMMENT A class holding unknown constants.
+	COMMENT
+	COMMENT @apiNote From the package structure, this class appears to be world-related,
+	COMMENT but it seems to be the exclusive class in a package. Its neighboring packages
+	COMMENT are about portal teleporting and persistent state, yet these constants don't
+	COMMENT seem to relate to those.
--- a/mappings/net/minecraft/class_6148.mapping
+++ b/mappings/net/minecraft/class_6148.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_6148 net/minecraft/unused/UnknownWorldConstants6148
+CLASS net/minecraft/class_6148
 	COMMENT A class holding unknown constants.
 	COMMENT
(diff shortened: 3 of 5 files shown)


mappings/net/minecraft/entity/player/HungerConstants.mapping
```diff
@@ -0,0 +1,24 @@
...
+		COMMENT A value {@value} that when the exhausion reaches, the exhausion minuses itself
+		COMMENT by to reduce the saturation or food level.
+	FIELD field_30709 SLOW_HEALING_STARVING_INTERVAL I
+		COMMENT When the food tick is a multiple of {@value}, the hunger manager may perform
+		COMMENT slow healing or starving logic.
+	FIELD field_30710 FAST_HEALING_INTERVAL I
+		COMMENT When the food tick is a multiple of {@value}, the hunger manager may perform
+		COMMENT fast healing logic.
+	FIELD field_30711 SLOW_HEALING_FOOD_LEVEL I
+		COMMENT The minimum food level ({@value}) required for the slow-healing mechanism.
```
haykam821, line 19: In a previous comment you did not use a hyphen with 'slow healing'.

mappings/net/minecraft/entity/player/HungerConstants.mapping
```diff
@@ -0,0 +1,24 @@
...
+	FIELD field_30712 EXHAUSION_PER_HITPOINT I
+		COMMENT The exhausion from healing each hitpoint ({@value}), used for both fast and
```
haykam821, line 21:
```suggestion
	FIELD field_30712 EXHAUSTION_PER_HITPOINT I
		COMMENT The exhaustion from healing each hitpoint ({@value}), used for both fast and
```

mappings/net/minecraft/entity/player/HungerConstants.mapping
```diff
@@ -0,0 +1,24 @@
...
+	COMMENT Some constants on hunger values.
+	COMMENT
+	COMMENT @see HungerManager
+	FIELD field_30704 FULL_FOOD_LEVEL I
+		COMMENT The maximum food level ({@value}) allowed in a hunger manager.
+	FIELD field_30706 INITIAL_SATURATION_LEVEL F
+		COMMENT The initial saturation level ({@value}) for a newly created hunger manager.
+	FIELD field_30708 EXHAUSION_UNIT F
+		COMMENT A value {@value} that when the exhausion reaches, the exhausion minuses itself
+		COMMENT by to reduce the saturation or food level.
```
haykam821, line 11:
```suggestion
	FIELD field_30708 MAX_EXHAUSTION
		COMMENT The value {@value} at which exhaustion resets
		COMMENT causing saturation or hunger to decrease.
```

haykam821 (changes requested):

mappings/net/minecraft/entity/player/HungerConstants.mapping
liach: It doesn't reset. I chose your wording initially but decided to use the current one because the exhausion may be 6 and it will decrease by 4 and become 2 than reset to 0.
```suggestion
	FIELD field_30708 EXHAUSTION_UNIT F
		COMMENT A value {@value} that when the exhaustion reaches, the exhaustion minuses itself
		COMMENT by, causing reduction in saturation or food level.
```

mappings/net/minecraft/entity/player/HungerConstants.mapping
liach:
```suggestion
		COMMENT The minimum food level ({@value}) required for the slow healing mechanism.
```

OroArmor (approved):

Mysterious-Dev (approved):

mappings/net/minecraft/unused/UnknownWorldConstants6148.mapping
```diff
@@ -0,0 +1,7 @@
+CLASS net/minecraft/class_6148 net/minecraft/unused/UnknownWorldConstants6148
```
YanisBft: While it's nice to reach 100% mapped classes, I think it makes no sense to name a class like that. Better to live unmapped

YanisBft (changes requested):

mappings/net/minecraft/unused/UnknownWorldConstants6148.mapping
liach: No? According to the reasoning above, all package info should be unmapped. This mapping at least have some information about what this class is; more info than an unmapped class. At least we know this is some world-related constant from this name; it's definitely more informative than `class_6148`.

Once before, we had a "sortme" package https://github.com/FabricMC/yarn/tree/18w50a/mappings/net/minecraft/sortme. This mapping is generally the same idea; we should make use of what information we currently have.

YanisBft: Because you state that this is world-related based on the package structure, and we can't be sure of that. The constant values are way too broad to be able to guess anything.
I remember someone saying on discord that it could also be related to redstone-related block power, but these values probably exist in entity classes, item classes or any other game part.

What I mean is naming this class like you did, with the docs you wrote, will make people think that this is 100% sure world-related.

YanisBft: And as well, naming a class for the sake of deobfuscating a class doesn't make sense. It makes sense for package info classes because we don't need them. But for this class, putting them in a package with "unknown" name will just burry it and we won't look at it anymore.
This is not a "sortme" or "temp" package, this is "unused". This is wrong. These constants are used by Mojang, like other constant classes. We just can't see it.

liach:
Hmm, why didn't the branch updater not run...

mappings/net/minecraft/unused/UnknownWorldConstants6148.mapping
liach: Done. Now this class is unnamed while its javadocs are kept.

YanisBft (approved):

mappings/net/minecraft/unused/UnknownWorldConstants6148.mapping
andreypfau: in official mappings this constants used for redstone signal:
image

Juuxel: @andreypfau Yarn naming cannot use Mojang's official mapping data for legal reasons (see CONTRIBUTING.md). Mojmap names can't be shared in Yarn PRs/issues either since they risk "tainting" the contributors with Mojmap names.

PR #2321, opened 2021-04-11
Last few top level classes (annotations)

liach:

### DIFF
--- a/mappings/net/minecraft/util/annotation/ClientMethodsReturnNonnullByDefault.mapping
+++ b/mappings/net/minecraft/util/annotation/ClientMethodsReturnNonnullByDefault.mapping
@@ -1,9 +1,11 @@
-CLASS net/minecraft/class_6328 net/minecraft/util/annotation/MethodsReturnNonnullByDefault
+CLASS net/minecraft/class_6179 net/minecraft/util/annotation/ClientMethodsReturnNonnullByDefault
 	COMMENT Specifies that all methods in the annotated package/class return nonnull
 	COMMENT values unless nullability is specified with a separate annotation.
 	COMMENT
 	COMMENT <p>While this annotation is meant to be a package/class annotation, it can
 	COMMENT be applied to anything as it does not restrict its targets.
 	COMMENT
-	COMMENT @see ClientMethodsReturnNonnullByDefault
+	COMMENT <p>This annotation is used on the client exclusively.
+	COMMENT
+	COMMENT @see MethodsReturnNonnullByDefault
 	COMMENT @see MathMethodsReturnNonnullByDefault
--- a/mappings/net/minecraft/util/annotation/DeobfuscateClass.mapping
+++ b/mappings/net/minecraft/util/annotation/DeobfuscateClass.mapping
@@ -1,0 +1,5 @@
+CLASS net/minecraft/class_6177 net/minecraft/util/annotation/DeobfuscateClass
+	COMMENT An annotation on classes. When a class is annotated, the class itself and its
+	COMMENT fields and methods are not obfuscated. Mainly used by blaze3d.
+	COMMENT
+	COMMENT @see net.minecraft.obfuscate.DontObfuscate
--- a/mappings/net/minecraft/util/annotation/FieldsAreNonnullByDefault2.mapping
+++ b/mappings/net/minecraft/util/annotation/FieldsAreNonnullByDefault2.mapping
@@ -1,12 +1,12 @@
-CLASS net/minecraft/class_6319 net/minecraft/util/annotation/FieldsAreNonnullByDefault
+CLASS net/minecraft/class_6037 net/minecraft/util/annotation/FieldsAreNonnullByDefault2
 	COMMENT Specifies that all fields in the annotated package/class are nonnull
 	COMMENT unless nullability is specified with a separate annotation.
 	COMMENT
 	COMMENT <p>While this annotation is meant to be a package/class annotation, it can
 	COMMENT be applied to anything as it does not restrict its targets.
 	COMMENT
-	COMMENT <p>A similar {@link FieldsAreNonnullByDefault2} exists; they don't coexist
-	COMMENT on any package info class.
+	COMMENT <p>This is pretty much identical to {@link FieldsAreNonnullByDefault} that they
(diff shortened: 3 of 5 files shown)


mappings/net/minecraft/util/annotation/FieldsAreNonnullByDefault2.mapping
```diff
@@ -0,0 +1,12 @@
+CLASS net/minecraft/class_6037 net/minecraft/util/annotation/FieldsAreNonnullByDefault2
```
OroArmor, line 1: Is there any better naming for this class? Or do we not see the usages and therefore don't know where it's used

liach: You just look it up. There is no recognizable pattern on how this is used than `FieldsAreNonnullByDefault`, hence I go with this. If you have a name, please suggest it than commenting "please get a better name". Thanks!

liach: On a side note, the math method return nonnull one originally I named `2` suffix as well, but it is only used in that b3d math package, so I named it `Math...`. This annotation unfortunately has no such comparable trait and hence I chose such a bad name. @Juuxel might have some better ideas?

mappings/net/minecraft/util/annotation/ClientMethodsReturnNonnullByDefault.mapping
```diff
@@ -0,0 +1,11 @@
+CLASS net/minecraft/class_6179 net/minecraft/util/annotation/ClientMethodsReturnNonnullByDefault
+	COMMENT Specifies that all methods in the annotated package/class return nonnull
+	COMMENT values unless nullability is specified with a separate annotation.
+	COMMENT
+	COMMENT <p>While this annotation is meant to be a package/class annotation, it can
```
Sollace, line 5: Why do we need this disclaimer? If it's not restructed, what evidence do you have to assert what it was "meant to be"?

Juuxel: Because as a `@TypeQualifierDefault` it has no effect on anything that's not a package or class.

mappings/net/minecraft/util/annotation/FieldsAreNonnullByDefault2.mapping
Juuxel: In mojang code they might be `com.mojang.blaze3d.something.FieldsAreNonnullByDefault` etc if they're only used there, but I'm not sure whether we should do that in yarn.

Sollace:
Something I noticed with these annotations, actually, is that they're client-only. Specifically class_6177, class_6178, and class_6179 so I wonder whether we need to have a `net/minecraft/client/util/annotation/` package?

mappings/net/minecraft/util/annotation/FieldsAreNonnullByDefault2.mapping
OroArmor: weird, it doesnt seem like it has any discernable use. i guess this is fine

OroArmor (approved):

mappings/net/minecraft/util/annotation/FieldsAreNonnullByDefault2.mapping
liach: This one's usage is scattered with the other non null by default, like in item and tag packages at least.

kb-1000:
> Something I noticed with these annotations, actually, is that they're client-only. Specifically class_6177, class_6178, and class_6179 so I wonder whether we need to have a `net/minecraft/client/util/annotation/` package?

Is it possible that all new code went client only?
If that's the case, I'd argue that the client package should be used for client specific code, not just for code that's stripped out from the server... although that might be confusing from a modder perspective, so idk if that's a good idea

liach:
Well, the 2 client package annotations are definitely client-only; previously client only methods (due to proguard, like getters in packets, a few mathhelper methods) are no longer client-only after 13a update.

Sollace (approved):
Well I think the annotation names in this one are _fiiine_. I would like to look at them again once https://github.com/FabricMC/Enigma/issues/371 is implemented.

liach:
Sure. Anyways, mods most likely shouldn't use these annotations since most of their setup stuff could have been in a separate annotation processor or such.

mappings/net/minecraft/util/annotation/DeobfuscateClass.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS net/minecraft/class_6177 net/minecraft/util/annotation/DeobfuscateClass
+	COMMENT An annotation on classes. When a class is annotated, the class itself and its
+	COMMENT fields and methods are not obfuscated. Mainly used by blaze3d.
```
Juuxel, line 3: This is probably a blaze3d variant of `DontObfuscate` in Mojang's names, but we can call it this IMO.

Juuxel:
@liach Most of these don't really need an annotation processor, since IDEA supports `@TypeQualifierDefault` annotations by default so using them is fine if you have javax annotations (but I'm pretty sure Loom breaks these by remapping the meta-annos)

mappings/net/minecraft/util/annotation/DeobfuscateClass.mapping
liach: Not sure; this one is actually always used on classes while the obf one is on methods.

liach:
> 
> 
> @liach Most of these don't really need an annotation processor, since IDEA supports `@TypeQualifierDefault` annotations by default so using them is fine if you have javax annotations (but I'm pretty sure Loom breaks these by remapping the meta-annos)

I guess there is a reason they have two different fields are non null by default annotations... and that previously mapped "debug" annotation

Juuxel (approved):
These names are fine for now. I'm pretty sure some of these duplicate annotations are specific to some internal structures of the game (think Blaze3D for example).

modmuss50 (approved):

modmuss50:
The B3D idea makes some sense, this is fine for now.

PR #2323, opened 2021-04-11
Yet more obscure stuff

Sollace:
I'd appreciate feedback/ideas on a better name for the new class. Took me a good few hours to figure out what it was doing.


Earthcomputer:
At first glance, this class looks like it's supposed to contain values from `BlockPos.toLong()`. The `MathHelper.log2(60000000)` is very suspicious.

Sollace:
@Earthcomputer I believe it's related to LevelPropagator, as that's the only place I can find which uses this type's parent class.

```
            this.pendingIdUpdatesByLevel[l] = new LongLinkedOpenHashSet(expectedLevelSize, 0.5f){
                @Override
                protected void rehash(int newN) {
                    if (newN > j) {
                        super.rehash(newN);
                    }
                }
            };
```
Likely implemented for use here, but left out for some unknown reason.

Earthcomputer:
Yes, I've taken a closer look at the class and I think you're right. I've taken a closer look and noticed the following:
- `method_35483` returns the same value for block positions in the same 2x2x2 area
- `method_35486` returns an int depending on which block position we are looking at in that 2x2x2
- So we're extending from `Long2LongLinkedOpenHashMap`. The keys are longs representing the cubes, and the values are bitflags of which blocks are present in each cube.

So not only are values grouped together, values that are geometrically close when converted to a BlockPos are grouped together.

mappings/net/minecraft/util/collection/ExpandingLongLinkedOpenHashSet.mapping
```diff
@@ -0,0 +1,55 @@
...
+	CLASS class_6137 Buffer
+		FIELD field_31716 STARTING_OFFSET I
+		FIELD field_31717 MIDDLE_OFFSET I
+		FIELD field_31718 BIT_OFFSET_DIFFERENCE I
+		FIELD field_31719 MIDDLE_BIT_OFFSET I
+		FIELD field_31720 LOWER_BIT_OFFSET I
+		FIELD field_31721 HIGHER_BIT_OFFSET I
+		FIELD field_31722 TOTAL_BIT_MASK J
+		FIELD field_31723 lasWrittenIndex I
+		FIELD field_31724 lasWrittenKey J
```
liach:
```suggestion
		FIELD field_31723 lastIndex I
		FIELD field_31724 lastKey J
```
methink those two are updated on reads, aren't they? Also use `last` than `las` prefix

Sollace: Keys and indexes are different. The parent class has an array of keys and an array of values, and here `lastWrittenKey = this.keys[index]` and `value = this.values[lastWrittenIndex]`

Sollace: > methink those two are updated on reads, aren't they?

No. They're updated when adding/removing. In fact there aren't any methods that read without a corresponding write.

The only operations we have here are adds and removes, with one of them performing a pop (read and remove).

liach: Why is `Written` needed here? It isn't crucial to the nature of the field and just makes the name longer. I think just `last` prefix suffices.

mappings/net/minecraft/util/collection/LinkedBlockPosHashSet.mapping
```diff
@@ -0,0 +1,92 @@
...
+		ARG 1 posLong
+	METHOD isEmpty ()Z
+		COMMENT Checks whether there are any block positions that have been "set".
+		COMMENT
+		COMMENT @return {@code true} is this collection is empty.
+	METHOD rem (J)Z
+		COMMENT Marks a block position as "not set". Effectively removing it from this collection.
+		ARG 1 posLong
+	METHOD removeFirstLong ()J
+		COMMENT  Pops first block position off of this set.
```
YanisBft:
```suggestion
		COMMENT Pops first block position off of this set.
```

mappings/net/minecraft/util/collection/LinkedBlockPosHashSet.mapping
```diff
@@ -0,0 +1,92 @@
...
+	METHOD size ()I
+		COMMENT @throws UnsupportedOperationException
+	CLASS class_6137 Storage
+		COMMENT Represents a three-dimensional mapping from a block position to a bitset
+		COMMENT of values set at that position.
+		FIELD field_31716 STARTING_OFFSET I
+		FIELD field_31717 HOR_COLUMN_BIT_SEPARATION I
```
YanisBft: `Hor`?

mappings/net/minecraft/util/collection/LinkedBlockPosHashSet.mapping
```diff
@@ -0,0 +1,92 @@
...
+		FIELD field_31725 expectedSize I
+		METHOD <init> (IF)V
+			ARG 1 expectedSize
+			ARG 2 loadFactor
+		METHOD method_35481 removeFirstLong ()J
+		METHOD method_35482 setBits (IJ)Z
+			COMMENT Sets flags within a specific 2x cubed region represented by {@code index}.
+			COMMENT
+			COMMENT @return {@code true} if the map already contained set bits for the indicated mask
+			COMMENT            (i.e. an overlap occured)
```
YanisBft:
```suggestion
			COMMENT            (i.e. an overlap occurred)
```

mappings/net/minecraft/util/collection/LinkedBlockPosHashSet.mapping
```diff
@@ -0,0 +1,92 @@
...
+			COMMENT Converts a individual position into a key
+			COMMENT representing the 2x cube region containing that position.
+			ARG 0 posLong
+		METHOD method_35484 getBlockPosLong (JI)J
+			ARG 0 key
+			ARG 2 valueLength
+		METHOD method_35485 unsetBitsAt (IJ)Z
+			COMMENT Unsets flags within a specific 2x cubed region contained within this map.
+			COMMENT
+			COMMENT @return {@code true} if the the collection was changed as a result of this call
```
YanisBft:
```suggestion
			COMMENT @return {@code true} if the collection was changed as a result of this call
```

mappings/net/minecraft/util/collection/LinkedBlockPosHashSet.mapping
```diff
@@ -0,0 +1,92 @@
...
+			ARG 1 posLong
+		METHOD method_35488 rem (J)Z
+			COMMENT Removes a block position from this map.
+			ARG 1 posLong
+		METHOD method_35489 unsetBits (J)Z
+			COMMENT Unsets flags within the last 2x cubed region contained within this map.
+			COMMENT <p>
+			COMMENT This is equivalent to the call {@code unsetBitsAt(this.n, mask)}
+			COMMENT
+			COMMENT @return {@code true} if the the collection was changed as a result of this call
```
YanisBft:
```suggestion
			COMMENT @return {@code true} if the collection was changed as a result of this call
```

mappings/net/minecraft/util/collection/LinkedBlockPosHashSet.mapping
Sollace: Short for HORIZONTAL.

Sollace: Basically, it's half the width of the middle field.

First field is `start --> (MID - HOR)`
Second field is `(MID - HOR) ---> (MID + HOR)`
Third field is `(MID + HOR) ---> end`

At least as best I can remember.

liach: Should just use `HORIZONTAL_COLUMN_BIT_SEPARATION` instead.

Sollace: It'll be too long tho >.>

Sollace:
```suggestion
		FIELD field_31717 HORIZONTAL_COLUMN_BIT_SEPARATION I
```

Sollace: Done

YanisBft (approved):

liach (approved):

PR #2326, opened 2021-04-14
Map marker entities

haykam821:


Shnupbups:
mfw i get sniped
very unpog

YanisBft (approved):

Shnupbups (approved):

Mysterious-Dev (approved):

mappings/net/minecraft/client/render/entity/EmptyEntityRenderer.mapping
```diff
@@ -0,0 +1,2 @@
+CLASS net/minecraft/class_6344 net/minecraft/client/render/entity/EmptyEntityRenderer
+	COMMENT A renderer used for markers and area effect clouds that does not render anything.
```
liach:
```suggestion
	COMMENT A renderer that does not render anything. Used for markers and area effect clouds.
```
Reordering since the `does not render anything` is more conclusive of this renderer. Also the entity types that use it may change in the future, so I put them a bit back.

liach (approved):

mappings/net/minecraft/client/render/entity/EmptyEntityRenderer.mapping
haykam821: Fixed in 268fe285b7edb31cfd92202e1fe77a98e373360d

PR #2327, opened 2021-04-14
Map raw metal blocks

haykam821:
This pull request maps the item fields of raw metal blocks, a recipe exporter for the compacting and decompacting recipes, and a map color (`#D8AF93`) used for raw iron blocks.

### DIFF
--- a/mappings/net/minecraft/block/MapColor.mapping
+++ b/mappings/net/minecraft/block/MapColor.mapping
@@ -64,7 +64,7 @@ CLASS net/minecraft/class_3620 net/minecraft/block/MapColor
 	FIELD field_25706 DARK_AQUA Lnet/minecraft/class_3620;
 	FIELD field_25707 DARK_DULL_PINK Lnet/minecraft/class_3620;
 	FIELD field_25708 BRIGHT_TEAL Lnet/minecraft/class_3620;
-	FIELD field_33533 IRON Lnet/minecraft/class_3620;
+	FIELD field_33533 RAW_IRON_PINK Lnet/minecraft/class_3620;
 	METHOD <init> (II)V
 		ARG 1 id
 		ARG 2 color
--- a/mappings/net/minecraft/block/MapColor.mapping
+++ b/mappings/net/minecraft/block/MapColor.mapping
@@ -64,6 +64,7 @@ CLASS net/minecraft/class_3620 net/minecraft/block/MapColor
 	FIELD field_25706 DARK_AQUA Lnet/minecraft/class_3620;
 	FIELD field_25707 DARK_DULL_PINK Lnet/minecraft/class_3620;
 	FIELD field_25708 BRIGHT_TEAL Lnet/minecraft/class_3620;
+	FIELD field_33533 IRON Lnet/minecraft/class_3620;
 	METHOD <init> (II)V
 		ARG 1 id
 		ARG 2 color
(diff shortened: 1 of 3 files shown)


mappings/net/minecraft/block/MapColor.mapping
```diff
@@ -64,6 +64,7 @@ CLASS net/minecraft/class_3620 net/minecraft/block/MapColor
 	FIELD field_25706 DARK_AQUA Lnet/minecraft/class_3620;
 	FIELD field_25707 DARK_DULL_PINK Lnet/minecraft/class_3620;
 	FIELD field_25708 BRIGHT_TEAL Lnet/minecraft/class_3620;
+	FIELD field_33533 IRON Lnet/minecraft/class_3620;
```
YanisBft: Hum, maybe a more precise name would be better, to make a difference with IRON_GRAY (used by regular iron blocks)

haykam821: I chose this name in order to be consistent with `MapColor.GOLD`.

dexman545: Gold is a color, iron is not.

haykam821: I think iron is fine as a color name.

dexman545: perhaps `RAW_IRON_ORANGE` or `PALE_BEIGE`.
Iron is not descriptive at all, it cannot be a color name.

dexman545: For reference, this is the color:
image
image

haykam821: It is definitely not orange and too pinkish to be beige. In my opinion, this color is the distinctive iron color.

dexman545: Iron in an MC context is gray.
This is more the color of the speckles on iron ore, but it is not the color of iron.

mappings/net/minecraft/data/server/RecipesProvider.mapping
```diff
@@ -232,3 +232,11 @@ CLASS net/minecraft/class_2446 net/minecraft/data/server/RecipesProvider
 		ARG 2 output
 		ARG 3 experience
 		ARG 4 cookingTime
+	METHOD method_36325 offerReversibleCompactingRecipes (Ljava/util/function/Consumer;Lnet/minecraft/class_1935;Lnet/minecraft/class_1935;)V
```
liach, line 235: Is there only 9 compact? I think this may be confused with 4 compact some stuff have, like copper?

mappings/net/minecraft/block/MapColor.mapping
liach: Why not `RAW_IRON`?

dexman545: At the very least `RAW_IRON_PINK` so it follows the naming conventions of all the other colors.

liach:
```suggestion
	FIELD field_33533 IRON_PINK Lnet/minecraft/class_3620;
```
Since we already have `IRON_GRAY`

liach:
Fyi marking outdated as git isn't smart enough to fix the merge conflict with the deepslate map color

mappings/net/minecraft/block/MapColor.mapping
haykam821: I think `RAW_IRON_PINK` is good.

```suggestion
	FIELD field_33533 RAW_IRON_PINK Lnet/minecraft/class_3620;
```

mappings/net/minecraft/data/server/RecipesProvider.mapping
haykam821: I would call those 'small compacting' recipes, since I believe only a few recipes like copper and quartz use them. For now, only raw metal has a specific function for reversible compacting recipes.

YanisBft (approved):

liach (approved):

PR #2351, opened 2021-04-27
Map text methods and params

### LINKED ISSUE #2353: Separator text fields suggest they are the only separators rather than the defaults
Despite separators being customizable as of Minecraft snapshot 21w15a, the naming of two `LiteralText` fields with the contents `, ` suggests that they are the only separators for text, rather than simply the default.

https://github.com/FabricMC/yarn/blob/a99ec0c4f06ce1a1309378b18685b8102800dccf/mappings/net/minecraft/text/Texts.mapping#L3-L4

apple502j:
This includes separator changes from 21w15a, and OrderedText renames (to better explain new methods.)

Fixes #2353

### DIFF
--- a/mappings/net/minecraft/text/ScoreText.mapping
+++ b/mappings/net/minecraft/text/ScoreText.mapping
@@ -2,7 +2,7 @@ CLASS net/minecraft/class_2578 net/minecraft/text/ScoreText
 	FIELD field_11785 objective Ljava/lang/String;
 	FIELD field_11786 selector Lnet/minecraft/class_2300;
 	FIELD field_11787 name Ljava/lang/String;
-	FIELD field_33290 EVERYTHING Ljava/lang/String;
+	FIELD field_33290 SENDER_PLACEHOLDER Ljava/lang/String;
 	METHOD <init> (Ljava/lang/String;Ljava/lang/String;)V
 		ARG 1 name
 		ARG 2 objective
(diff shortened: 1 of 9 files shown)


liach (approved):

mappings/net/minecraft/text/ScoreText.mapping
```diff
@@ -2,6 +2,7 @@ CLASS net/minecraft/class_2578 net/minecraft/text/ScoreText
 	FIELD field_11785 objective Ljava/lang/String;
 	FIELD field_11786 selector Lnet/minecraft/class_2300;
 	FIELD field_11787 name Ljava/lang/String;
+	FIELD field_33290 EVERYTHING Ljava/lang/String;
```
enbrain: This field doesn't seem to be a wildcard. `ScoreText.parse` compares the input with this field to determine if it should use the sender name.

enbrain:
```suggestion
	FIELD field_33290 SENDER_PLACEHOLDER Ljava/lang/String;
```

apple502j: Ahh thanks

enbrain (approved):

enbrain (approved):

Player3324 (approved):

PR #2355, opened 2021-04-28
21w17a stuff

apple502j:
Maps new recipe methods, some params, yaw/pitch setters and `EntityDestroyS2CPacket` (apparently Mojang made it singular)

Any better names for `validateAndSetX`/`offerReversibleCompactingRecipesWithXItemGroup`? Couldn't overload because param types are the same.

### DIFF
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -376,6 +376,12 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 		COMMENT axis-aligned integer box is fully loaded in the world.
 	METHOD method_35049 getRemovalReason ()Lnet/minecraft/class_1297$class_5529;
 	METHOD method_36209 onRemoved ()V
+	METHOD method_36454 getYaw ()F
+	METHOD method_36455 getPitch ()F
+	METHOD method_36456 validateAndSetYaw (F)V
+		ARG 1 yaw
+	METHOD method_36457 validateAndSetPitch (F)V
+		ARG 1 pitch
 	METHOD method_5621 getMountedHeightOffset ()D
 	METHOD method_5622 onBlockCollision (Lnet/minecraft/class_2680;)V
 		ARG 1 state
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -378,9 +378,9 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	METHOD method_36209 onRemoved ()V
 	METHOD method_36454 getYaw ()F
 	METHOD method_36455 getPitch ()F
-	METHOD method_36456 validateAndSetYaw (F)V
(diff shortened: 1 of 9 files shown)


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -376,6 +376,12 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 		COMMENT axis-aligned integer box is fully loaded in the world.
 	METHOD method_35049 getRemovalReason ()Lnet/minecraft/class_1297$class_5529;
 	METHOD method_36209 onRemoved ()V
+	METHOD method_36454 getYaw ()F
+	METHOD method_36455 getPitch ()F
+	METHOD method_36456 validateAndSetYaw (F)V
```
liach:
```suggestion
	METHOD method_36456 setYaw (F)V
```
Only your old name if there is a setter where yaw doesn't need to be validated. I recall the angles should always be wrapped.

For out-of-bounds input, are they always wrapped to within +- 180 degrees? Or how are they handled?

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -376,6 +376,12 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
...
+		ARG 1 yaw
+	METHOD method_36457 validateAndSetPitch (F)V
```
liach:
```suggestion
	METHOD method_36457 setPitch (F)V
```

liach (approved):

mappings/net/minecraft/network/packet/s2c/play/EntityDestroyS2CPacket.mapping
```diff
@@ -0,0 +1,7 @@
+CLASS net/minecraft/class_2716 net/minecraft/network/packet/s2c/play/EntityDestroyS2CPacket
```
enbrain, line 1: We need to rename `ClientPlayPacketListener.onEntitiesDestroy` as well.

mappings/net/minecraft/entity/Entity.mapping
apple502j: Throws exception for invalid inputs. I tried to name it `setYaw`  but there's `method_5636`.

enbrain: Judging from the implementation `LivingEntity.setYaw`, `method_5636` is more like `setBodyYaw` than `setYaw`.

apple502j: Hmm maybe that's right. (silverfish and endermite override 5636 and calls both 36456 and LivingEntity.method_5636, just to note)

liach: Yeah, minecraft (living, at least I think) entities do have head and body yaws. Need to distinguish those.

enbrain (approved):

YanisBft (approved):

liach (approved):

PR #2358, opened 2021-04-28
PGP sign maven artifacts

modmuss50:
Jar signing will require a bit more work as my signing plugin expects AbstractArchiveTask right now.


Player3324 (approved):

modmuss50:
Checks passed, not a lot more I can do to test, will merge and report back.

modmuss50:
Didnt work, got some issues that I will need to debug. Possibly due to using gradle 7 here.

I will revert this for now.

PR #2363, opened 2021-04-30
Map the parameter of equals methods

enbrain:


liach:
Hmm, I wonder if we can do this in batch with stitch's name inference service.

liach (approved):

YanisBft (approved):

haykam821 (approved):

PR #2366, opened 2021-04-30
Refactor coordinate names to follow conventions

enbrain:
This PR makes them follow the convention:
> Coordinates can be named `x`, `y`, and `z` when it's clear what they represent. If clarification is needed, add a word in
front of the coordinate (`velocityX`, not `xVelocity`).

https://github.com/FabricMC/yarn/blob/daabbb8bff164ccac9adbddb54a66d020bac5669/CONVENTIONS.md

### DIFF
--- a/mappings/net/minecraft/client/render/chunk/ChunkRendererRegion.mapping
+++ b/mappings/net/minecraft/client/render/chunk/ChunkRendererRegion.mapping
@@ -1,10 +1,10 @@
 CLASS net/minecraft/class_853 net/minecraft/client/render/chunk/ChunkRendererRegion
 	FIELD field_4481 offset Lnet/minecraft/class_2338;
-	FIELD field_4482 zSize I
+	FIELD field_4482 sizeZ I
 	FIELD field_4483 chunks [[Lnet/minecraft/class_2818;
-	FIELD field_4484 ySize I
+	FIELD field_4484 sizeY I
 	FIELD field_4485 fluidStates [Lnet/minecraft/class_3610;
-	FIELD field_4486 xSize I
+	FIELD field_4486 sizeX I
 	FIELD field_4487 chunkZOffset I
 	FIELD field_4488 chunkXOffset I
 	FIELD field_4489 blockStates [Lnet/minecraft/class_2680;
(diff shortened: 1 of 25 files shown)


YanisBft (approved):

mappings/net/minecraft/client/render/chunk/ChunkRendererRegion.mapping
```diff
@@ -1,10 +1,10 @@
 CLASS net/minecraft/class_853 net/minecraft/client/render/chunk/ChunkRendererRegion
 	FIELD field_4481 offset Lnet/minecraft/class_2338;
-	FIELD field_4482 zSize I
+	FIELD field_4482 sizeZ I
 	FIELD field_4483 chunks [[Lnet/minecraft/class_2818;
-	FIELD field_4484 ySize I
+	FIELD field_4484 sizeY I
 	FIELD field_4485 fluidStates [Lnet/minecraft/class_3610;
-	FIELD field_4486 xSize I
+	FIELD field_4486 sizeX I
```
liach, line 7: Hmm, for these ones I am not quite sure. We don't usually view these 3 sizes as a whole (we don't use them together in a 3 vector or pass them together into a method)

mappings/net/minecraft/entity/ai/FuzzyPositions.mapping
```diff
@@ -41,8 +41,8 @@ CLASS net/minecraft/class_5535 net/minecraft/entity/ai/FuzzyPositions
 		ARG 1 horizontalRange
 		ARG 2 verticalRange
 		ARG 3 startHeight
-		ARG 4 xDirection
-		ARG 6 zDirection
+		ARG 4 directionX
+		ARG 6 directionZ
```
liach, line 45: Sorry, my bad names! should have named them the way you do especially given the doc says they should be interpreted together as one "direction"

mappings/net/minecraft/client/render/chunk/ChunkRendererRegion.mapping
liach: After thinking, this rename is fine. We can just view them as a 3-vector representing a size, and that name is very good.

liach (approved):

PR #2368, opened 2021-04-30
Unpick entity flags and pufferfish states

YanisBft:


enbrain (approved):

liach (approved):
Just realized we need to unify these flag access method names, but that's out of scope here.

PR #2369, opened 2021-04-30
Fix swapped names

enbrain:


YanisBft (approved):

liach (approved):
matcher i guess

Player3324:
Did anyone check how/when this happened? Especially the field name shouldn't get mixed up in the matching process.

liach:
For the world border center one, it is just that shnupbups messed it up (the getters are written not in the expected x first z second order! weird indeed!) now checking the language definition one.

liach:
For the language definition one, it appears to have been between 21w08b and 21w10a. The fields and parameters seemed to have swapped order? Or did mojang rename the deobfed methods?

Player3324:
Need to see if it happens again, may have been a manual mismatch.

liach:
For that language definition one, I'd really say matcher is doing a good job. From a glimpse, all things matched are the same; the only change is the names of the deobfuscated methods (like the first method still return first field, 2nd method still return 2nd field, param order stayed, but the two methods swapped names). Need to dig into that more; maybe we just assumed the deobf'd names are right when we matched initially, and matcher simply carries on our assumptions?

Player3324:
https://github.com/sfPlayer1/Matcher/commit/1c6b8cbbd21b4409235a697fbc276847ceef8294 may have improved things either way, I'd wait for another clear instance before spending too much time investigating

PR #2382, opened 2021-05-05
Map new methods added to the AbstractBlock.Settings class in 21w18a

haykam821:
This pull request should be confirmed to not break Fabric API's [`FabricBlockSettings` methods](https://github.com/FabricMC/fabric/blob/daa367217679ffae237a7216154255bd840edc47/fabric-object-builder-api-v1/src/main/java/net/fabricmc/fabric/api/object/builder/v1/block/FabricBlockSettings.java#L249-L257) which have the same names.


modmuss50 (approved):

liach (approved):

enbrain (approved):

Player3324 (approved):
The two methods are public, so API should(TM) be fine

PR #2384, opened 2021-05-06
Drop invalid mappings

enbrain:
Done via Enigma menu.


liach (approved):

modmuss50:
I imagine these are left over from an update? If not I don’t think it would be too hard to run it on a PR build to ensure PRs aren’t including such mappings

Issue #1565, opened 2020-07-09
BooleanBiFunction -> CombinationRule


liach:
It is used exclusively combining voxel shapes or bit set voxel sets.

liach:
https://github.com/FabricMC/yarn/commit/1520549e3dea23f6804e46d6b929ca3640cf7b7e#diff-aa2bc7ebf401f48a9c4f9f6c1baae28e
Looking at this, the field names are from some legacy unobfuscated names. I wonder if we should revisit these...

PR #2386, opened 2021-05-07
Map util stuff

apple502j:
Question: does renaming `BooleanBiFunction.SAME` to `XNOR` make sense?


liach:
> 
> 
> Question: does renaming `BooleanBiFunction.SAME` to `XNOR` make sense?

See https://github.com/FabricMC/yarn/issues/1565#issuecomment-655853243. This `BooleanBiFunction` class is like a rule for combination for voxel shapes.

mappings/net/minecraft/util/UserCache.mapping
```diff
@@ -21,8 +24,22 @@ CLASS net/minecraft/class_3312 net/minecraft/util/UserCache
 	METHOD method_14515 findByName (Ljava/lang/String;)Lcom/mojang/authlib/GameProfile;
 		ARG 1 name
 	METHOD method_14516 getLastAccessedEntries (I)Ljava/util/stream/Stream;
+		ARG 1 limit
 	METHOD method_14517 load ()Ljava/util/List;
 	METHOD method_14518 save ()V
+	METHOD method_30164 add (Lnet/minecraft/class_3312$class_3313;)V
+		ARG 1 entry
+	METHOD method_30165 serialize (Lnet/minecraft/class_3312$class_3313;Ljava/text/DateFormat;)Lcom/google/gson/JsonElement;
```
liach: We use toJson/fromJson for such static methods; see https://github.com/FabricMC/yarn/blob/21w18a/CONVENTIONS.md#serialization

Player3324:
IMO XNOR etc are too abstract and don't really make sense for boolean operands (not bit vectors)

mappings/net/minecraft/util/snooper/Snooper.mapping
```diff
@@ -12,6 +14,11 @@ CLASS net/minecraft/class_1276 net/minecraft/util/snooper/Snooper
 		ARG 1 urlPath
 		ARG 2 listener
 		ARG 3 startTime
+	METHOD method_35022 ([ILjava/lang/String;)V
+		ARG 2 flag
+	METHOD method_35024 getInfo ()Ljava/util/Map;
+	METHOD method_35030 addInfo ()V
+	METHOD method_35032 addJVMFlags ()V
```
liach:
```suggestion
	METHOD method_35032 addJvmFlags ()V
```

liach (changes requested):

Player3324:
BooleanBiFunction btw has a few potential improvements (probably better for a separate PR/issue?):
- AND -> BOTH
- OR -> ANY
- NOT_OR -> NEITHER
- NOT_AND -> NOT_BOTH
- NOT_SAME -> DIFFERENT
- CAUSES -> NOT_ONLY_FIRST
- CAUSED_BY -> NOT_ONLY_SECOND

The first/second field names kind of conflict with a/b param naming in this PR

mappings/net/minecraft/util/UserCache.mapping
apple502j: It doesn't "create an object of the method owner's type" though; it takes/returns `UserCache.Entry`, not UserCache

mappings/net/minecraft/util/snooper/Snooper.mapping
```diff
@@ -1,4 +1,6 @@
 CLASS net/minecraft/class_1276 net/minecraft/util/snooper/Snooper
+	FIELD field_29957 SNOOPER_URL Ljava/lang/String;
+	FIELD field_29959 SNOOPER_VERSION I
```
enbrain:
```suggestion
	FIELD field_29957 BASE_URL Ljava/lang/String;
	FIELD field_29959 VERSION I
```
I think `SNOOPER` is redundant as it's obvious from the enclosing class name.

mappings/net/minecraft/util/UserCache.mapping
liach: Then `entryToJson` `entryFromJson` are still more specific and concise than `deserialize` `serialize`

enbrain (approved):

liach (approved):

PR #2388, opened 2021-05-08
Rename the BlockEntity#getSquaredRenderDistance method to getRenderDistance

haykam821:


liach (approved):

liach:
Is this used by anything other than beacon in vanilla? Maybe they have changed how this is used indeed.

haykam821:
It's used by beacons and end gateways (256 blocks) and structure blocks (96 blocks), and is specifically used in the generic block entity render dispatcher where it is passed to another method which squares the value. It is missing in 1.17 snapshots though.

Player3324 (approved):

PR #2398, opened 2021-05-13
Update to Filament 0.3.0

Juuxel:
The new Filament version contains the two tasks dealing with Unpick definitions. Now they also support up-to-date checks and running in parallel.


liach (approved):
Sure

YanisBft (approved):

PR #2402, opened 2021-05-15
framebuffer, screenshots, gl info

liach:
FixedDataSetter -> GlInfoConsumer
ScreenshotUtils -> Screenshooter (no longer a utility class)


enbrain (approved):

Player3324 (approved):

liach:
@OroArmor Since I see you wondering about the rename, here's the reason :point_up: it is no longer a utility class but an object class

OroArmor:
ah ok. i didnt know its functionality changed

liach:
Feel free to suggest a better name for `Screenshooter`.

Fyi, it holds a few utility methods for regular screenshots and panoramas; it is used as an object only for taking ultra-detailed raster images, which is available through a method in `MinecraftClient` that is not used anywhere in vanilla code, like the method for taking panorama. Any ideas?

haykam821:
I would suggest `ScreenshotRenderer` if the main ability is to render screenshots without necessarily writing them to a file, and otherwise `ScreenshotHelper`.

i509VCB:
Enlightened name imo would just be `Screenshot`.

Given a class named `Screenshot` I'd think that `finish` could become `take` since you take a screen shot.

liach:
Hmm, I think another way to name this is to look back at the `MinecraftClient` method that uses this. The data comes from some gl call, and I recall that gl call has a descriptive name, and was more concise than `screenshot`

PR #2405, opened 2021-05-15
Bump dependencies

liach:
Namely enigma and cfr


YanisBft:
Should wait for j16 enigma imo

liach:
Indeed, this build doesn't introduce too many new stuff; and enum improvements aren't really needed since now cfr is decompiling enums as regular classes somehow

PR #2410, opened 2021-05-16
Ping/Pong clientbound packets

ghost:


liach:
Conflict with #2401

Technici4n:
Given that you commented a mojang name on #2319 (now removed), I think this PR can be closed.

liach:
Imo seeing mojang name isn't the problem. We just need to have a way to derive a name that is valid when the mojang name is not present.

Technici4n:
#2401 is better anyway imo: it documents the added classes and puts them in the correct package

liach:
Also imagine trusting mojang names: [when we only see the mojang name](https://github.com/FabricMC/yarn/pull/2402/files#diff-59c80d9e9483f75d19b12dd6ded5a33037e2599cc8d9e462d6dbb92625d43e2c) vs [what something actually is](https://github.com/FabricMC/yarn/pull/2402/files#diff-d144574802a49a2fef2739995d2630df88ad5e7ec83a76e9a3bbc0ef37bb9a7eR8-R9).

We do need a mapping system for dfu and brigadier. They are literally unchecked, undocumented, misleading mojang names :laughing:

Technici4n:
~~The solution for dfu is just to delete it and operate on mutable NBT trees instead.~~ 😆

Player3324:
Please note that the license of the Mojang provided mappings is very restrictive when it comes to relevant uses in Yarn and similar projects. It is not an option to potentially risk violating their copyright, Yarn needs to remain untainted and free. Copyright protection extends beyond 1:1 replication, analog copies or other kinds of lookalikes are protected as well.

As such please close the PR if you based these names off Mojang's and avoid that approach in the future. If they are original work please state so. We'd be happy to see future contributions in the future either way, provided they are untainted in the above way.

@liach @Technici4n please try to be a little more welcoming and keep on topic

liach:
Sorry but the contents of this conflicts with what's in the current yarn. If you think your field names of `syncId` are better feel free to open another pr.

Technici4n:
Of course, my bad. If the contribution is not tainted by mojmap then it is welcome... I shouldn't have assumed it is tainted. \:)

PR #2412, opened 2021-05-19
21w20a blocks

apple502j:
Maps PowderSnowCauldronBlock, fire collision things, etc. Relatively small.

### DIFF
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -69,6 +69,15 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
 	METHOD method_35258 dropStacks (Lnet/minecraft/class_2680;Lnet/minecraft/class_47$class_48;)V
 		ARG 0 state
 		ARG 1 lootContext
+	METHOD method_36992 dropStack (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;Lnet/minecraft/class_1799;)V
+		ARG 0 world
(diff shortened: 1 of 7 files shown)


mappings/net/minecraft/block/Block.mapping
```diff
@@ -69,6 +69,15 @@ CLASS net/minecraft/class_2248 net/minecraft/block/Block
...
 		ARG 0 state
 		ARG 1 lootContext
+	METHOD method_36992 dropStack (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;Lnet/minecraft/class_1799;)V
+		ARG 0 world
+		ARG 1 pos
+		ARG 2 direction
+		ARG 3 stack
+	METHOD method_36993 dropStack (Lnet/minecraft/class_1937;Ljava/util/function/Supplier;Lnet/minecraft/class_1799;)V
+		ARG 0 world
+		ARG 1 itemEntityGetter
```
haykam821:
```suggestion
		ARG 1 itemEntitySupplier
```

haykam821 (changes requested):

enbrain (approved):

mappings/net/minecraft/block/Block.mapping
liach: Agree with haykam. This is to avoid extra item creation in case the drop shouldn't be spawned than a getter that retrieves an existing item.

liach (approved):
Just fix the `itemEntityGetter` name.

PR #2418, opened 2021-05-20
Block list checker for the allowed address resolver

liach:


liach:
@FabricMC/yarn-triage anyone mind reviewing this? Should be relatively small changes.

YanisBft (approved):

PR #2432, opened 2021-05-25
Map Minecraft logo outline method

jackassmc:
This was first added in 20w21a as method_29063, but has been method_29343 since 20w22a. It is used on the TitleScreen and the CreditsScreen.

### DIFF
--- a/mappings/net/minecraft/client/gui/DrawableHelper.mapping
+++ b/mappings/net/minecraft/client/gui/DrawableHelper.mapping
@@ -206,6 +206,10 @@ CLASS net/minecraft/class_332 net/minecraft/client/gui/DrawableHelper
 		ARG 3 x
 		ARG 4 y
 		ARG 5 color
+	METHOD method_29343 drawOutline (IILjava/util/function/BiConsumer;)V
+		ARG 1 x
+		ARG 2 y
+		ARG 3 renderCallback
 	METHOD method_33284 fillGradient (Lnet/minecraft/class_4587;IIIIIII)V
 		ARG 0 matrices
 		ARG 1 startX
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/client/gui/DrawableHelper.mapping
```diff
@@ -206,6 +206,10 @@ CLASS net/minecraft/class_332 net/minecraft/client/gui/DrawableHelper
 		ARG 3 x
 		ARG 4 y
 		ARG 5 color
+	METHOD method_29343 drawOutline (IILjava/util/function/BiConsumer;)V
```
enbrain:
```suggestion
	METHOD method_29343 drawWithOutline (IILjava/util/function/BiConsumer;)V
```
Since this method supplies the original position as well as outline positions, `drawWithOutline` is better imo.

enbrain (approved):

liach (approved):
I want to document the consumer a bit as otherwise its functionalities are not very straightforward.

mappings/net/minecraft/client/gui/DrawableHelper.mapping
```diff
@@ -206,6 +206,10 @@ CLASS net/minecraft/class_332 net/minecraft/client/gui/DrawableHelper
 		ARG 3 x
 		ARG 4 y
 		ARG 5 color
+	METHOD method_29343 drawWithOutline (IILjava/util/function/BiConsumer;)V
+		ARG 1 x
+		ARG 2 y
+		ARG 3 renderCallback
```
liach:
```suggestion
		ARG 3 renderAction
			COMMENT the action to render both the content and the outline, taking x and y positions as input
```

liach:
Heck, github eats suggestions when I forgot to add it to "start a review" while asking you whether to cancel when you click the button. Rewrote the suggestion.

liach:
@modmuss50 Mind allow running github actions for this build?

jackassmc:
I looked into this some more and it turns out that in `20w21a` `method_29063` is on `TitleScreen` and draws an outline but doesn't set its color to black, while in `20w22a` `method_29343` is on `DrawableHelper` and draws a black outline.

LambdAurora (approved):

YanisBft (approved):

PR #2433, opened 2021-05-26
Some function stuff

liach:


mappings/net/minecraft/server/function/CommandFunctionManager.mapping
```diff
@@ -1,40 +1,83 @@
...
+	FIELD field_33543 execution Lnet/minecraft/class_2991$class_6345;
+		COMMENT The active execution within this manager.
 	METHOD <init> (Lnet/minecraft/server/MinecraftServer;Lnet/minecraft/class_5349;)V
 		ARG 1 server
 		ARG 2 loader
-	METHOD method_12899 getTaggedFunctionSource ()Lnet/minecraft/class_2168;
+	METHOD method_12899 getCommandSource ()Lnet/minecraft/class_2168;
+		COMMENT {@return the command source to execute functions}
+		COMMENT
+		COMMENT <p>This command source {@linkplain ServerCommandSource#hasPermissionLevel(int)
```
apple502j, line 29: Is this affected by function-permission-level in server.properties?

liach: That is only used when parsing functions. I need to look into brigadier more for how exactly it works.

liach:
Hmm, so no one reviews this? This at least maps a nested class...

modmuss50 (approved):
LGTM, will merge. 👍

Issue #2395, opened 2021-05-13
Generating since tags for members


liach:
<details>
<summary> Context

[Source](https://discord.com/channels/507304429255393322/521545796882006027/842220158067539998)
</summary>

> haykam
 — 
Today at 8:55 PM
How would you feel about a `@since` tag automatically added based on when the intermediary was added?
It might be more annoying to have it on every field or method, but definitely on every class since those rarely get mismatched
Chocohead
 — 
Today at 9:02 PM
If you could generate it in the published Javadoc for everything but only classes in the source that would be a good middle ground
It'd be useful on methods and fields when you need it but annoying when you didn't

</details>

We are seeking to introduce `@since` tags for elements to indicate from when did they start appearing in intermediaries.

To accomplish this, the main difficulty is to have a system for indexing the start of appearance effectively, as currently these data have to be obtained by checking every intermediary tiny file, and it's not quite realistic to do that every time we build yarn locally.

My plan is to make a tracker for every existent intermediary number in yarn or intermediary. But only tracking numbers is not enough, as we want to record when an unobfuscated name has appeared in Minecraft as well.

Once this data is properly collected, rendering them is no big deal.

cc @haykam821 @Chocohead

liach:
@sfPlayer1 What way do you recommend so we can find out the introduction version of these intermediaries? Note that tracking counters may not be the most feasible when we encounter non-obfuscated names, which are not accounted by the intermediaries.

liach:
@modmuss50 Should we implement this feature? If yes, which way do you recommend to implement it so it's not performance/resource intensive? Imo checking every intermediary on every build may be a bit too costly, but generating a since table in intermediary repo for every version (intermediary repo is already growing too large, I heard) might have space concerns.

modmuss50:
I think it would be nice to have in the javadoc, but generating and storing that info is going to be a bit of a effort. The ovious solution to me would be to clone the intermediary repo when building yarn and run a tool over it to generate the info. I think we can make use of github actions caches to keep the clone time minimal, I dont imagine such a tool being that slow to run?

The yarn javadoc building is already really slow I wonder if there is anything that can be done to speed that up?

liach:
Can we make a "sinces" file that is updated with each yarn version? So when elements are lost, they are lost in the since file; otherwise, the new ones are added to a new version. Imo keeping such an extra table in yarn repo would be the most feasible (size around that of a built tiny mapping, I think)

[...]

PR #2437, opened 2021-05-28
Fix Entity.dealDamage

### LINKED ISSUE #1737: Entity.dealDamage -> Entity.applyEnchantmentEffects and more
Entity.dealDamage is actually used for apply the effects of enchantments after the entity already dealt the base damage. For example, in BeeEntity's attackEntityAsMob method, it calls attackEntityFrom first to dealt the actual damage and then call Entity.dealDamage afterwards. 

Upon closer look at Entity.dealDamage, it's mainly for applying  EnchantmentHelper.applyThornEnchantments and EnchantmentHelper.applyArthropodEnchantments. However, looking at these two method seems to show they are not just for Thorn or Bane of Arthropod but any enchantment that overrides onEntityDamaged method. So I could make an armor enchantment called Ballooning that makes the attacker get Levitation effect which would be applied in Entity.dealDamage. 

Thus why I believe Entity.dealDamage needs to be renamed to something that shows it is for applying any special enchantments. Furthermore, EnchantmentHelper.applyThornEnchantments and EnchantmentHelper.applyArthropodEnchantments should also be renamed as custom enchantments could be applied as well making it not specific to just Thorn or Bane of Arthropod.

I hope this helps and be sure to double check on everything I said too

enbrain:
Closes #1737

### DIFF
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -584,8 +584,9 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 		COMMENT @implNote Although this method is non-static, {@code this} keyword is
 		COMMENT not used anywhere in this method.
 		ARG 1 attacker
-			COMMENT the attacker; usually this entity, but may be a projectile's owner
-			COMMENT entity.
+			COMMENT the attacker; usually this entity, but may be a {@link
+			COMMENT net.minecraft.entity.projectile.ProjectileEntity#getOwner() projectile's
+			COMMENT owner entity}.
 		ARG 2 target
 	METHOD method_5724 updateVelocity (FLnet/minecraft/class_243;)V
 		ARG 1 speed
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -586,7 +586,7 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 		ARG 1 attacker
 			COMMENT the attacker; usually this entity, but may be a {@link
 			COMMENT net.minecraft.entity.projectile.ProjectileEntity#getOwner() projectile's
-			COMMENT owner entity}.
+			COMMENT owner entity}
 		ARG 2 target
 	METHOD method_5724 updateVelocity (FLnet/minecraft/class_243;)V
 		ARG 1 speed
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -573,7 +573,7 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	METHOD method_5721 isTouchingWaterOrRain ()Z
 	METHOD method_5722 isTeammate (Lnet/minecraft/class_1297;)Z
 		ARG 1 other
-	METHOD method_5723 dealDamage (Lnet/minecraft/class_1309;Lnet/minecraft/class_1297;)V
+	METHOD method_5723 applyEnchantmentEffects (Lnet/minecraft/class_1309;Lnet/minecraft/class_1297;)V
 		ARG 1 attacker
 		ARG 2 target
 	METHOD method_5724 updateVelocity (FLnet/minecraft/class_243;)V
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -576,7 +576,8 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	METHOD method_5723 applyDamageEffects (Lnet/minecraft/class_1309;Lnet/minecraft/class_1297;)V
 		COMMENT Applies damage effects to {@code attacker} or {@code target}.
 		COMMENT
-		COMMENT <p>Called when {@code attacker} damages {@code target}.
+		COMMENT <p>Called when {@code attacker} damages {@code target}. The attacker is
+		COMMENT usually this entity, but may be a projectile's owner entity.
 		COMMENT
 		COMMENT <p>Used to apply enchantment effects such as thorn or slowness of bane
 		COMMENT of arthropods.
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -579,8 +579,8 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 		COMMENT <p>Called when {@code attacker} damages {@code target}. The attacker is
 		COMMENT usually this entity, but may be a projectile's owner entity.
 		COMMENT
-		COMMENT <p>Used to apply enchantment effects such as thorn or slowness of bane
-		COMMENT of arthropods.
+		COMMENT <p>Used to apply damage effects based on enchantments, such
+		COMMENT as Thorns attacker damage or slowness from Bane of Arthropods.
 		COMMENT
 		COMMENT @implNote Although this method is non-static, {@code this} keyword is
 		COMMENT not used anywhere in this method.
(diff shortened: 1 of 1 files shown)


liach (changes requested):
This rename does indicate the relationship to enchantment, but it misses another important attribute, that these enchantment effects are only applied through a dealing of damage.

Imo `applyDamageEffects` might be good, as this callback can actually be used for any arbitrary damages dealt. Also note that this entity is dealing the damage, `attacker` is like this entity's owner, and `target` is the actual target of damage. This might worth documenting.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -573,7 +573,16 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	METHOD method_5721 isTouchingWaterOrRain ()Z
 	METHOD method_5722 isTeammate (Lnet/minecraft/class_1297;)Z
 		ARG 1 other
-	METHOD method_5723 dealDamage (Lnet/minecraft/class_1309;Lnet/minecraft/class_1297;)V
+	METHOD method_5723 applyDamageEffects (Lnet/minecraft/class_1309;Lnet/minecraft/class_1297;)V
+		COMMENT Applies damage effects to {@code attacker} or {@code target}.
+		COMMENT
+		COMMENT <p>Called when {@code attacker} damages {@code target}.
```
liach, line 579: Notice the attacker is usually this entity, but may be a projectile's owner entity, etc. I would prefer clarifying that in the javadoc as well

liach (approved):

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -573,7 +573,17 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
...
 	METHOD method_5722 isTeammate (Lnet/minecraft/class_1297;)Z
 		ARG 1 other
-	METHOD method_5723 dealDamage (Lnet/minecraft/class_1309;Lnet/minecraft/class_1297;)V
+	METHOD method_5723 applyDamageEffects (Lnet/minecraft/class_1309;Lnet/minecraft/class_1297;)V
+		COMMENT Applies damage effects to {@code attacker} or {@code target}.
+		COMMENT
+		COMMENT <p>Called when {@code attacker} damages {@code target}. The attacker is
+		COMMENT usually this entity, but may be a projectile's owner entity.
+		COMMENT
+		COMMENT <p>Used to apply enchantment effects such as thorn or slowness of bane
```
YanisBft:
```suggestion
		COMMENT <p>Used to apply enchantment effects such as thorn or slowness, or bane
```

enbrain: Bane of arthropods gives arthropods slowness when hit.

liach: Yeah it's like for bane of arthropods 4 I recall, though it's quite useless because if bane is that high the arthropod is usually killed.

YanisBft (approved):

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -573,7 +573,17 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
...
+		COMMENT of arthropods.
```
haykam821:
```suggestion
		COMMENT <p>Used to apply damage effects based on enchantments, such
		COMMENT as Thorns attacker damage or slowness from Bane of Arthropods.
```

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -573,7 +573,17 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
...
+		COMMENT
+		COMMENT @implNote Although this method is non-static, {@code this} keyword is
+		COMMENT not used anywhere in this method.
```
haykam821, line 585: If this should be documented, I think it should be processed similar to #2395.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -573,7 +573,17 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	METHOD method_5721 isTouchingWaterOrRain ()Z
 	METHOD method_5722 isTeammate (Lnet/minecraft/class_1297;)Z
 		ARG 1 other
-	METHOD method_5723 dealDamage (Lnet/minecraft/class_1309;Lnet/minecraft/class_1297;)V
+	METHOD method_5723 applyDamageEffects (Lnet/minecraft/class_1309;Lnet/minecraft/class_1297;)V
+		COMMENT Applies damage effects to {@code attacker} or {@code target}.
+		COMMENT
+		COMMENT <p>Called when {@code attacker} damages {@code target}. The attacker is
+		COMMENT usually this entity, but may be a projectile's owner entity.
```
haykam821: Would other deferred damage sources such as TNT not apply here?

haykam821 (changes requested):

mappings/net/minecraft/entity/Entity.mapping
enbrain: Only projectiles. TNT doesn't track its owner.

mappings/net/minecraft/entity/Entity.mapping
liach: Also not using `this` doesn't mean having this method as non-static is pointless: the subclasses can override this instance method while static methods can't be overridden.

mappings/net/minecraft/entity/Entity.mapping
liach: This is controlled by the entity code that calls this method. Worthy of documentation?

haykam821: TNT does track its owner:

Player661 was blown up by Player661

Is there a specific method that controls what the `attacker` is? I would document that here.

liach: Don't think so; I think that was controlled by damage source. This is more about enchantments.

Another fun fact I just discovered is that `PersistentProjectileEntity` calls the `onUserDamaged` and `onTargetDamaged` methods individually in `onEntityHit(EntityHitResult)`. :zany_face: please mojang

enbrain: > TNT does track its owner

You're right. I meant it doesn't track for enchantment effect purposes.

> Is there a specific method that controls what the `attacker` is? I would document that here.

`attacker` is either this entity or `ProjectileEntity#getOwner()` . I don't see any specific method for it.

enbrain: Maybe document `attacker` thing in parameter itself.

liach: > I don't see any specific method for it.

Hmm, they added a method for status effect cause. Hope they will use that method for both purposes soon :tm:
https://github.com/FabricMC/yarn/pull/2441/files#diff-0583232f2335075f834f611a33196ed09274d74e21f20a4c1c99ae70ff69abcdR25

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -576,15 +576,17 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
...
-		COMMENT usually this entity, but may be a projectile's owner entity.
+		COMMENT <p>Called when {@code attacker} damages {@code target}.
 		COMMENT
 		COMMENT <p>Used to apply damage effects based on enchantments, such
 		COMMENT as Thorns attacker damage or slowness from Bane of Arthropods.
 		COMMENT
 		COMMENT @implNote Although this method is non-static, {@code this} keyword is
 		COMMENT not used anywhere in this method.
 		ARG 1 attacker
+			COMMENT the attacker; usually this entity, but may be a {@link
```
liach:
```suggestion
			COMMENT the attacker; usually this entity, but may be a {@linkplain
```
otherwise "projectile's owner entity" is rendered as `projectile's owner entity`

liach (changes requested):

liach (approved):

mappings/net/minecraft/entity/Entity.mapping
enbrain: Fixed in https://github.com/FabricMC/yarn/pull/2437/commits/e6d23ab8c102f481f588cf44487ea6431feebdd8.

PR #2441, opened 2021-05-28
status effect source

liach:
Renamed `LootContext.getSupplier` to `LootContext.getTable` as well, legacy name


mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -450,7 +470,15 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_6091 travel (Lnet/minecraft/class_243;)V
 		ARG 1 movementInput
 	METHOD method_6092 addStatusEffect (Lnet/minecraft/class_1293;)Z
+		COMMENT Adds a status effect to this entity without specifying a source entity.
+		COMMENT
+		COMMENT <p>Consider calling {@link #addStatusEffect(StatusEffectInstance, Entity)}
+		COMMENT explicitly if the is a source entity associated to this status effect to add.
```
YanisBft: Please reword this sentence

liach:
```suggestion
		COMMENT if the {@code effect} is caused by or from an entity.
```

liach:
Squash pushed after fixing on github due to action failing for wrapper validation.

YanisBft (approved):

Juuxel (approved):

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -196,8 +196,14 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_24833 getBoundingBox (Lnet/minecraft/class_4050;)Lnet/minecraft/class_238;
 		ARG 1 pose
 	METHOD method_25937 displaySoulSpeedEffects ()V
-	METHOD method_26082 applyStatusEffect (Lnet/minecraft/class_1293;Lnet/minecraft/class_1297;)V
+	METHOD method_26082 setStatusEffect (Lnet/minecraft/class_1293;Lnet/minecraft/class_1297;)V
+		COMMENT Sets a status effect in this entity.
```
haykam821, line 200: I would document that this is for client-side effects for players. The name could possibly be more clear too.

haykam821 (changes requested):

mappings/net/minecraft/entity/LivingEntity.mapping
liach: Agreed, though I will put this in an `@apiNote` since when called on the server, this is effective too.

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -196,8 +196,21 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
...
+		COMMENT
+		COMMENT <p>This replaces the preexistent status effect of the same type on this entity.
```
haykam821:
```suggestion
		COMMENT <p>This replaces the pre-existing status effect of the same type on this entity.
```

haykam821 (changes requested):

mappings/net/minecraft/entity/LivingEntity.mapping
liach: I'm fairly sure it is a word. https://www.dictionary.com/browse/preexistent

haykam821: Would this be better?

```suggestion
		COMMENT <p>If this entity has a status effect of the same type already, it is replaced.
```

I'm not sure if status effect priority rules will apply here.

liach:
```suggestion
		COMMENT <p>The preexistent status effect of the same type on this entity, if there is one, is cleared.
```

liach:
Re haykam: it appears your concern is on the handling of the old/preexistent effect. I have clarified that the old one is just removed.

Juuxel:
The new JD is much better :+1:

PR #2444, opened 2021-05-28
Some collision stuff, fixes #2443

### LINKED ISSUE #2443: CollisionView.isBlockSpaceEmpty is the wrong way around
`CollisionView.isBlockSpaceEmpty` checks whether the provided Box _doesn't_ intersect any blocks.
I came across this in testing, but once you notice it, it's also clear from the method body or from its (one) usage.

It should be called something like `isBlockSpaceOccupied`.

liach:
Unpicked spliterator as BlockCollisionSpliterator uses it. Fixed #2443 and a few other wrong names.

We need a revisit into voxel shapes one day!

### DIFF
--- a/mappings/net/minecraft/util/shape/VoxelShape.mapping
+++ b/mappings/net/minecraft/util/shape/VoxelShape.mapping
@@ -59,3 +59,6 @@ CLASS net/minecraft/class_265 net/minecraft/util/shape/VoxelShape
 		ARG 11 z2
 	METHOD method_20538 getFace (Lnet/minecraft/class_2350;)Lnet/minecraft/class_265;
 		ARG 1 facing
+	METHOD method_33661 getClosestPointTo (Lnet/minecraft/class_243;)Ljava/util/Optional;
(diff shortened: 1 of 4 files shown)


mappings/net/minecraft/util/shape/VoxelShape.mapping
```diff
@@ -59,3 +59,6 @@ CLASS net/minecraft/class_265 net/minecraft/util/shape/VoxelShape
 		ARG 11 z2
 	METHOD method_20538 getFace (Lnet/minecraft/class_2350;)Lnet/minecraft/class_265;
 		ARG 1 facing
+	METHOD method_33661 getClosestPointTo (Lnet/minecraft/class_243;)Ljava/util/Optional;
+		COMMENT {@return the closest point in the shape to {@code target}}
+		ARG 1 target
```
Juuxel, line 64: When would this return `Optional.empty()`?

liach: When this shape is empty.

Juuxel (approved):

PR #2445, opened 2021-05-29
Build javadoc with JDK 17 to fix #2429

### LINKED ISSUE #2429: Javadocs output is entirely broken for method summary
Method summary on classes is entirely broken as the lines are not complete and then the content get shifted.

Here you can see it on a live javadoc: https://maven.fabricmc.net/docs/yarn-21w20a+build.9/net/minecraft/entity/passive/AnimalEntity.html

and a screenshot:
image

This happens for any classes.

modmuss50:


liach (changes requested):
Think we should update `publish.yml` as it runs gradle as well: https://github.com/FabricMC/yarn/blob/fa703582024de23a668b655d5e7f446b81dabeaf/.github/workflows/publish.yml#L31

modmuss50:
Yes, just editing this to get it to work first. will do the same in publish once its working :D

.github/workflows/build.yml
```diff
@@ -12,7 +12,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: gradle/wrapper-validation-action@v1
-      - run: ./gradlew build javadocJar checkMappings --stacktrace
+      - run: |
+          curl https://maven.fabricmc.net/net/java/jdk/openjdk-17-ea%2B24_linux-x64_bin.tar.gz --output java.tar.gz
+          tar -xf java.tar.gz
+      - run: ./gradlew build javadocJar checkMappings --stacktrace -Porg.gradle.java.installations.paths=./jdk-17
```
liach: Maybe we can debug by like:
```suggestion
      - run: ./gradlew javaToolchains build javadocJar checkMappings --stacktrace -Porg.gradle.java.installations.auto-download=false -Porg.gradle.java.installations.paths=./jdk-17
```

modmuss50:
Sure, was just setting up a vm to test this, shame there isnt a way to run actions locally

liach:
Have you verified that those methods are now fixed? Just look at the summary of any method that overrides a method without documentation. If yes then fantastic

modmuss50:
The output can be seen here: https://github.com/FabricMC/yarn/suites/2863112103/artifacts/63958204

modmuss50:
Seems fixed to me. Many thanks :)

liach (approved):

liach:
Yes, it does appear fixed in the artifact javadoc jar. Thanks! :heart:

PR #2451, opened 2021-05-30
Map two new classes used in profiler

apple502j:
Profiler itself is too confusing for me, so I didn't include them. One thing I know is: client/util/profiler is definitely used by servers.

SystemDetails is shared between profilers and crash reports.


mappings/net/minecraft/util/SystemDetails.mapping
```diff
@@ -0,0 +1,36 @@
...
+	FIELD field_33857 JAVA_VM_VERSION Ljava/lang/String;
+	FIELD field_33858 sections Ljava/util/Map;
+	METHOD method_37120 toProfilerString ()Ljava/lang/String;
+	METHOD method_37121 tryAddGroup (Ljava/lang/String;Ljava/lang/Runnable;)V
+		ARG 1 name
+		ARG 2 adder
+	METHOD method_37122 addSection (Ljava/lang/String;Ljava/lang/String;)V
+		ARG 1 name
+		ARG 2 value
+	METHOD method_37123 tryAddSection (Ljava/lang/String;Ljava/util/function/Supplier;)V
```
liach: Imo `addSection` suffices. There is no clear evidence that this is different from the other `addSection`, as that `String` version cannot throw an exception that needs to be handled.

mappings/net/minecraft/util/SystemDetails.mapping
```diff
@@ -0,0 +1,36 @@
...
+		ARG 1 name
+		ARG 2 valueSupplier
+	METHOD method_37124 toCrashReportString (Ljava/lang/StringBuilder;)V
```
liach: This is really adding to the string builder, so `writeTo` I recommend.

mappings/net/minecraft/util/SystemDetails.mapping
```diff
@@ -0,0 +1,36 @@
+CLASS net/minecraft/class_6396 net/minecraft/util/SystemDetails
+	FIELD field_33852 MEBIBYTE_DIVIDER J
+	FIELD field_33854 LOGGER Lorg/apache/logging/log4j/Logger;
+	FIELD field_33855 OPERATING_SYSTEM Ljava/lang/String;
+	FIELD field_33856 JAVA_VERSION Ljava/lang/String;
+	FIELD field_33857 JAVA_VM_VERSION Ljava/lang/String;
+	FIELD field_33858 sections Ljava/util/Map;
+	METHOD method_37120 toProfilerString ()Ljava/lang/String;
```
liach: Don't think this is that profiler oriented. Let me check my local stuff real quick

mappings/net/minecraft/util/SystemDetails.mapping
```diff
@@ -0,0 +1,36 @@
+CLASS net/minecraft/class_6396 net/minecraft/util/SystemDetails
+	FIELD field_33852 MEBIBYTE_DIVIDER J
```
liach: Imo `MEBI` suffices. In addition, there is another field which is like `GIGA`, used to convert Hz to GHz. Please map that.

mappings/net/minecraft/util/SystemDetails.mapping
```diff
@@ -0,0 +1,36 @@
...
+	FIELD field_33854 LOGGER Lorg/apache/logging/log4j/Logger;
+	FIELD field_33855 OPERATING_SYSTEM Ljava/lang/String;
+	FIELD field_33856 JAVA_VERSION Ljava/lang/String;
+	FIELD field_33857 JAVA_VM_VERSION Ljava/lang/String;
```
liach:
```suggestion
	FIELD field_33857 JVM_VERSION Ljava/lang/String;
```
This abbreviation is more conventional.

mappings/net/minecraft/util/SystemDetails.mapping
apple502j: At least it is used by profilers. Maybe I'll just call it "toString".

liach: Hmm, I called it `collect` locally as this is indeed an action of building a string from the sections.

Btw you shouldn't call it `toString` as that will cause an override of `java.lang.Object#toString()`, which the original method doesn't do.

apple502j: Yeah. (Currently there are two classes that have "toString" mapped to intermediary names. Is that a bug?)

modmuss50 (approved):

liach (approved):

PR #2452, opened 2021-05-30
Slot.isEnabled and CommandItemSlot -> StackReference

emilyploszaj:
`Slot.doDrawHoveringEffects` has been one of the most infuriatingly mismapped methods that has bothered me for a long time. I looked it up and it was mapped in 2016 and has not been touched since. For those unfamiliar, `Slot.doDrawHoveringEffects` is called on a slot to determine if it should be rendered at all, and if the player can interact with it. If this returns false it is impossible to tell it exists for a player. Thus, I have renamed to `Slot.isEnabled`. I'd also be fine with names such as `Slot.isVisible`, even though it also disables interaction.

The second part of this PR reworks the interface `CommandItemSlot` to become `StackReference`. This includes all method names associated. While a large portion of this interface used to and still is interacting with commands, it is used in other places that make absolutely no sense. Namely, it is passed as a mutable cursor stack from screen handlers to `ItemStack.onClicked`. It is now clear that it is intended to be used as a mutable stack reference, as all it does is get and set an item stack. This name will also allow modders to more easily find this interface, which I can imagine being useful in certain situations rather than reinventing the wheel.


Gegy (approved):

LambdAurora (approved):

Juuxel (approved):

liach (approved):

liach:
And yes, as the one who named this `CommandItemSlot`, I did intend someone to rename it once that class got a wider usage.

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -322,9 +322,9 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	METHOD method_32316 canFreeze ()Z
 	METHOD method_32317 setFrozenTicks (I)V
 		ARG 1 frozenTicks
-	METHOD method_32318 getCommandItemSlot (I)Lnet/minecraft/class_5630;
-		COMMENT Obtains an item slot for command modification purpose. Used by commands
-		COMMENT like {@code /loot} or {@code /replaceitem}.
+	METHOD method_32318 getStackReference (I)Lnet/minecraft/class_5630;
+		COMMENT Obtains a stack reference to be modified. Used by commands like
+		COMMENT {@code /loot} or {@code /replaceitem}.
```
apple502j: Maybe fix this line as well? replaceitem command is now item.

modmuss50 (approved):

PR #2453, opened 2021-05-31
Fix swapped names in Matrix3f and Matrix4f

enbrain:
It has been incorrect since the first commit that introduced these names. I have confirmed that it's human error.

... I'm the one who introduced these names. Sorry!

### DIFF
--- a/mappings/net/minecraft/util/math/Matrix3f.mapping
+++ b/mappings/net/minecraft/util/math/Matrix3f.mapping
@@ -56,50 +56,50 @@ CLASS net/minecraft/class_4581 net/minecraft/util/math/Matrix3f
 	METHOD method_35260 (Lnet/minecraft/class_4581;Lnet/minecraft/class_1158;)V
 		ARG 0 matrix
 		ARG 1 quaternion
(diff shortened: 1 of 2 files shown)


Player3324:
Can we maybe rename these to readColumn/RowMajor and note the misnaming in the javadoc? It looks like something to introduce sneaky bugs otherwise and Major is more descriptive than First IMO.

haykam821:
It should be fine since these methods are relatively unused and have only been named incorrectly in snapshots since [21w14a](https://github.com/FabricMC/yarn/pull/2297). I do agree that they should be use the term 'major' rather than 'first' though.

Player3324 (approved):
If you want to you could potentially describe what column major etc mean in Javadoc, fine either way

liach:
So row major fills up a row before it moves on to the second row, I assume?

Player3324:
Yes, see https://en.wikipedia.org/wiki/Row-_and_column-major_order - OpenGL is column major btw

modmuss50 (approved):

PR #2455, opened 2021-05-31
Map entity methods and params

apple502j:
Should be compatible with other PRs.

### DIFF
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -345,6 +345,8 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 		COMMENT
 		COMMENT @implNote The vanilla implementation always returns {@code null}.
 	METHOD method_33189 occludeVibrationSignals ()Z
+	METHOD method_33331 (Lnet/minecraft/class_4048;Lnet/minecraft/class_243;)V
+		ARG 2 pos
 	METHOD method_33332 calculateBoundingBox ()Lnet/minecraft/class_238;
 	METHOD method_33567 requestTeleportAndDismount (DDD)V
 		ARG 1 destX
@@ -398,6 +400,8 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 		ARG 2 pos
 	METHOD method_36974 tryCheckBlockCollision ()V
 	METHOD method_36975 playExtinguishSound ()V
+	METHOD method_37215 playAmethystChimeSound (Lnet/minecraft/class_2680;)V
+		ARG 1 state
 	METHOD method_5621 getMountedHeightOffset ()D
 	METHOD method_5622 onBlockCollision (Lnet/minecraft/class_2680;)V
 		ARG 1 state
@@ -558,6 +562,7 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 		ARG 1 status
 	METHOD method_5712 playStepSound (Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;)V
 		ARG 1 pos
+		ARG 2 state
 	METHOD method_5713 checkWaterState ()V
 	METHOD method_5715 isSneaking ()Z
 	METHOD method_5716 updateKilledAdvancementCriterion (Lnet/minecraft/class_1297;ILnet/minecraft/class_1282;)V
--- a/mappings/net/minecraft/entity/Entity.mapping
+++ b/mappings/net/minecraft/entity/Entity.mapping
@@ -9,7 +9,7 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 	FIELD field_23807 inanimate Z
 	FIELD field_25599 submergedFluidTag Lnet/minecraft/class_3494;
 	FIELD field_25750 trackedPosition Lnet/minecraft/class_243;
-	FIELD field_26994 prevAge I
+	FIELD field_26994 lastChimeAge I
 	FIELD field_26995 removalReason Lnet/minecraft/class_1297$class_5529;
 	FIELD field_26996 entityChangeListener Lnet/minecraft/class_5569;
 	FIELD field_27857 inPowderSnow Z
--- a/mappings/net/minecraft/entity/ExperienceOrbEntity.mapping
+++ b/mappings/net/minecraft/entity/ExperienceOrbEntity.mapping
@@ -1,6 +1,7 @@
 CLASS net/minecraft/class_1303 net/minecraft/entity/ExperienceOrbEntity
 	FIELD field_27009 pickingCount I
 	FIELD field_30055 DESPAWN_AGE I
+	FIELD field_30056 TICKS_PER_SECOND I
 	FIELD field_6159 amount I
 	FIELD field_6161 health I
 	FIELD field_6162 target Lnet/minecraft/class_1657;
@@ -11,6 +12,8 @@ CLASS net/minecraft/class_1303 net/minecraft/entity/ExperienceOrbEntity
 		ARG 4 y
 		ARG 6 z
 		ARG 8 amount
+	METHOD method_31492 (IILnet/minecraft/class_1303;)Z
+		ARG 2 orb
 	METHOD method_31493 spawn (Lnet/minecraft/class_3218;Lnet/minecraft/class_243;I)V
 		ARG 0 world
 		ARG 1 pos
--- a/mappings/net/minecraft/entity/ExperienceOrbEntity.mapping
+++ b/mappings/net/minecraft/entity/ExperienceOrbEntity.mapping
@@ -1,7 +1,7 @@
 CLASS net/minecraft/class_1303 net/minecraft/entity/ExperienceOrbEntity
 	FIELD field_27009 pickingCount I
 	FIELD field_30055 DESPAWN_AGE I
-	FIELD field_30056 TICKS_PER_SECOND I
+	FIELD field_30056 EXPENSIVE_UPDATE_INTERVAL I
 	FIELD field_6159 amount I
 	FIELD field_6161 health I
 	FIELD field_6162 target Lnet/minecraft/class_1657;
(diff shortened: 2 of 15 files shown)


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -398,6 +400,8 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 		ARG 2 pos
 	METHOD method_36974 tryCheckBlockCollision ()V
 	METHOD method_36975 playExtinguishSound ()V
+	METHOD method_37215 playAmethystChimeSound (Lnet/minecraft/class_2680;)V
```
liach, line 403: Need to rename `Entity.prevAge` field to something like `lastChimeAge` because it's exclusively used in this method.

mappings/net/minecraft/entity/ExperienceOrbEntity.mapping
```diff
@@ -1,6 +1,7 @@
 CLASS net/minecraft/class_1303 net/minecraft/entity/ExperienceOrbEntity
 	FIELD field_27009 pickingCount I
 	FIELD field_30055 DESPAWN_AGE I
+	FIELD field_30056 TICKS_PER_SECOND I
```
liach: This is more like an expensive update interval in ticks. Also `onEverySecond` may be named like `expensiveUpdate`? Feel free to debate.

mappings/net/minecraft/entity/ItemEntity.mapping
```diff
@@ -1,5 +1,7 @@
 CLASS net/minecraft/class_1542 net/minecraft/entity/ItemEntity
 	FIELD field_30456 DESPAWN_AGE I
+	FIELD field_30457 CANNOT_PICK_UP I
+	FIELD field_30458 NEVER_DESPAWN I
```
liach:
```suggestion
	FIELD field_30457 CANNOT_PICK_UP_DELAY I
	FIELD field_30458 NEVER_DESPAWN_AGE I
```
they apply to age and delay respectively.

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -271,6 +275,8 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_36977 damageHelmet (Lnet/minecraft/class_1282;F)V
 		ARG 1 source
 		ARG 2 amount
+	METHOD method_37119 tickItemStack (Lnet/minecraft/class_1799;)V
```
liach:
```suggestion
	METHOD method_37119 tickItemStackUsage (Lnet/minecraft/class_1799;)V
```
Since in server player entity, it triggers an advancement criteria trigger and this is called when the given stack is used.

mappings/net/minecraft/entity/boss/dragon/EnderDragonEntity.mapping
```diff
@@ -2,6 +2,9 @@ CLASS net/minecraft/class_1510 net/minecraft/entity/boss/dragon/EnderDragonEntit
 	FIELD field_18120 CLOSE_PLAYER_PREDICATE Lnet/minecraft/class_4051;
 	FIELD field_20865 yawAcceleration F
 	FIELD field_30428 MAX_HEALTH I
+	FIELD field_30430 TAKEOFF_THRESHOLD F
```
liach, line 5: Might document this, like this is a percentage of full health, and once a dragon takes this much damage while sitting it will takeoff

mappings/net/minecraft/entity/ExperienceOrbEntity.mapping
apple502j: I think we can rename that. Oh wait, that stupid name was mine... 😛

liach (approved):

liach:
@apple502j Mind fix the merge conflict on github real quick? You just arrange the mappings so that `tickItemStackUsage` comes immediately before the newly added `addStatusEffect` (ranked by their intermediaries)

apple502j:
@liach Okay I'll do that, plus 1.17-pre3

apple502j:
Update: Mapped lambda param I found while checking for falling block change.
Mapped "no retaliation" field and methods for goats.
Mapped stone floor and no light check methods.

mappings/net/minecraft/entity/mob/WaterCreatureEntity.mapping
```diff
@@ -1,8 +1,15 @@
...
 	METHOD method_35174 canSpawnUnderground (Lnet/minecraft/class_1299;Lnet/minecraft/class_5425;Lnet/minecraft/class_3730;Lnet/minecraft/class_2338;Ljava/util/Random;)Z
 		ARG 0 entityType
+		ARG 1 world
 		ARG 2 spawnReason
 		ARG 3 pos
 		ARG 4 random
+	METHOD method_37359 hasNoLight (Lnet/minecraft/class_5425;Lnet/minecraft/class_2338;)Z
+		ARG 0 world
+		ARG 1 pos
+	METHOD method_37360 hasStoneFloor (Lnet/minecraft/class_2338;Lnet/minecraft/class_5425;)Z
```
liach, line 11: i heard that the spawn requirement isn't the stone floor, but a few layers of stones below the floor?

apple502j: I don't think the code reads so.

Spawning condition from the code:
1. for-loop 5 times:
2. If the block is BASE_STONE_OVERWORLD, return true.
3. If the block is NOT water, return false.
4. If the loop ends (without finding one), return false.

The loop ends when it finds stone underwater. (It can technically return true for 1 layer of stone without water, but SpawnRestriction has IN_WATER.) So I think this can be called "stone floor".

mappings/net/minecraft/entity/ExperienceOrbEntity.mapping
```diff
@@ -27,10 +30,11 @@ CLASS net/minecraft/class_1303 net/minecraft/entity/ExperienceOrbEntity
 		ARG 2 amount
 	METHOD method_31497 merge (Lnet/minecraft/class_1303;)V
 		ARG 1 other
-	METHOD method_31498 onEverySecond ()V
-		COMMENT Called every second (every 20 ticks).
+	METHOD method_31498 expensiveUpdate ()V
```
haykam821, line 33: What does this method actually do? The name and most of the documentation are not helpful in describing the actual implementation of the method.

mappings/net/minecraft/entity/mob/WaterCreatureEntity.mapping
liach: Right, mojang says within 5 blocks below the spawning space

liach (approved):

mappings/net/minecraft/entity/damage/DamageSource.mapping
```diff
@@ -5,6 +5,7 @@ CLASS net/minecraft/class_1282 net/minecraft/entity/damage/DamageSource
 	FIELD field_28100 fallingBlock Z
 	FIELD field_28399 fromFalling Z
 	FIELD field_28400 STALAGMITE Lnet/minecraft/class_1282;
+	FIELD field_34004 noRetaliation Z
```
haykam821: Probably better to fit in with the `Angerable` interface, unless this field is more specific.

```suggestion
	FIELD field_34004 angering Z
```

mappings/net/minecraft/entity/mob/WaterCreatureEntity.mapping
haykam821: This sounds good to put in an `@implNote`.

haykam821 (changes requested):

mappings/net/minecraft/entity/ExperienceOrbEntity.mapping
liach: It performs checks and updates that are not as critical to each tick, such as moving to player, merging.

mappings/net/minecraft/entity/damage/DamageSource.mapping
apple502j: false means it should anger and true means it shouldn't.

Plus, this is independent of Angerable. Angerable is only implemented for mobs that are neutral; this affects any mob, like a Skeleton.

The code simply does not `setAttacker` when it `shouldNotRetaliate`. (That also means that if goat rams a dog, the poor dog's death message will not show the goat, if i'm correct.)

liach: Hmm, if this `noRetaliation` is too long, maybe consider naming it like `neutral` and document that entities taking this damage will stay neutral than retaliate?

mappings/net/minecraft/entity/damage/DamageSource.mapping
```diff
@@ -5,7 +5,12 @@ CLASS net/minecraft/class_1282 net/minecraft/entity/damage/DamageSource
 	FIELD field_28100 fallingBlock Z
 	FIELD field_28399 fromFalling Z
 	FIELD field_28400 STALAGMITE Lnet/minecraft/class_1282;
-	FIELD field_34004 noRetaliation Z
+	FIELD field_34004 neutral Z
+		COMMENT Whether the attacked entity should be neutral to the attacker.
+		COMMENT
+		COMMENT @implNote This is used by goats to prevent rammed mobs from retaliating.
```
liach:
```suggestion
		COMMENT @apiNote This is used by goats to prevent rammed mobs from retaliating.
```
This is about how this is used than how this field is implemented. implnote is usually for method body's code or the field initializer code.

YanisBft (approved):

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -223,6 +225,8 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_27071 shouldDropLoot ()Z
 	METHOD method_27302 shouldDisplaySoulSpeedEffects ()Z
 	METHOD method_27303 isOnSoulSpeedBlock ()Z
+	METHOD method_28305 (Lnet/minecraft/class_2487;Lnet/minecraft/class_2520;)V
+		ARG 1 nbt
```
enbrain:
```suggestion
		ARG 1 brain
```
Naming it `nbt` shadows the outer variable (the parameter of `writeCustomDataToNbt`).

mappings/net/minecraft/entity/mob/WaterCreatureEntity.mapping
```diff
@@ -1,8 +1,19 @@
...
+		COMMENT Returns if the position has a stone floor.
+		COMMENT
+		COMMENT @implNote This method checks if a block in the {@link net.minecraft.tags.BlockTags#BASE_STONE_OVERWORLD}
```
enbrain:
```suggestion
		COMMENT @implNote This method checks if a block in the {@link net.minecraft.tag.BlockTags#BASE_STONE_OVERWORLD}
```

mappings/net/minecraft/entity/damage/DamageSource.mapping
```diff
@@ -5,6 +5,12 @@ CLASS net/minecraft/class_1282 net/minecraft/entity/damage/DamageSource
 	FIELD field_28100 fallingBlock Z
 	FIELD field_28399 fromFalling Z
 	FIELD field_28400 STALAGMITE Lnet/minecraft/class_1282;
+	FIELD field_34004 neutral Z
+		COMMENT Whether the attacked entity should be neutral to the attacker.
+		COMMENT
+		COMMENT @apiNote This is used by goats to prevent rammed mobs from retaliating.
+		COMMENT Damages that are neutral do not cause {#link net.minecraft.entity.LivingEntity#setAttacker}
```
enbrain:
```suggestion
		COMMENT Damages that are neutral do not cause {@link LivingEntity#setAttacker(LivingEntity)}
```

enbrain (changes requested):

enbrain (approved):

PR #2457, opened 2021-06-01
Some world gen mappings (pre-releases)

mschae23:
I mapped a few world generation things for the 1.17 pre-releases.


liach (approved):

mappings/net/minecraft/world/gen/feature/RootSystemFeatureConfig.mapping
```diff
@@ -11,6 +11,7 @@ CLASS net/minecraft/class_5926 net/minecraft/world/gen/feature/RootSystemFeature
 	FIELD field_29276 hangingRootVerticalSpan I
 	FIELD field_29277 hangingRootStateProvider Lnet/minecraft/class_4651;
 	FIELD field_29278 hangingRootPlacementAttempts I
+	FIELD field_33616 allowedVerticalWaterForTree I
```
YanisBft, line 14: Note that `requiredVerticalSpaceForTree` field is used instead of this one in the codec, which is a bug

YanisBft (approved):

PR #2459, opened 2021-06-02
Map perfcommand (incomplete) and fix zip compressor mappings

liach:
the profiler system does need a revisit later.

though the perf command mapping isn't complete, it is still more complete than that in #2436.

fixed a few bad mappings and added docs in zip compressor from #2451, mainly private fields and parameters so users shouldn't hurt too bad


haykam821 (approved):

liach:
Wait a sec, need to fix package of perf command.

YanisBft (approved):

PR #2461, opened 2021-06-02
return inlining, up to enchantment package

liach:
see #2454

### DIFF
--- a/mappings/net/minecraft/block/BlockWithEntity.mapping
+++ b/mappings/net/minecraft/block/BlockWithEntity.mapping
@@ -1,6 +1,6 @@
 CLASS net/minecraft/class_2237 net/minecraft/block/BlockWithEntity
 	METHOD method_31618 checkType (Lnet/minecraft/class_2591;Lnet/minecraft/class_2591;Lnet/minecraft/class_5558;)Lnet/minecraft/class_5558;
-		COMMENT {@return the ticker if the given type and expected type are the same, or {@code null} if they are different.}
+		COMMENT {@return the ticker if the given type and expected type are the same, or {@code null} if they are different}
 		ARG 0 givenType
 		ARG 1 expectedType
 		ARG 2 ticker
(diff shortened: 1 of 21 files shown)


liach:
Also this can be treated as a prototype, so we can find out how to better implement return inlining for the other packages by examining the problems here and finding consistent solutions. Pinging @haykam821 who made the inlining request.

mappings/net/minecraft/block/BlockWithEntity.mapping
```diff
@@ -1,6 +1,6 @@
 CLASS net/minecraft/class_2237 net/minecraft/block/BlockWithEntity
 	METHOD method_31618 checkType (Lnet/minecraft/class_2591;Lnet/minecraft/class_2591;Lnet/minecraft/class_5558;)Lnet/minecraft/class_5558;
-		COMMENT Returns the ticker if the given type and expected type are the same, or null if they are different.
+		COMMENT {@return the ticker if the given type and expected type are the same, or {@code null} if they are different.}
```
liach:
```suggestion
		COMMENT {@return the ticker if the given type and expected type are the same, or {@code null} if they are different}
```

mappings/net/minecraft/client/render/entity/LivingEntityRenderer.mapping
```diff
@@ -22,9 +22,8 @@ CLASS net/minecraft/class_922 net/minecraft/client/render/entity/LivingEntityRen
 		ARG 3 translucent
 		ARG 4 showOutline
 	METHOD method_25450 isShaking (Lnet/minecraft/class_1309;)Z
-		COMMENT Returns if this entity is shaking in the way a zombie villager,
+		COMMENT {@return if this entity is shaking}. Such as in the way a zombie villager,
```
liach:
```suggestion
		COMMENT {@return if this entity is shaking} Specifically, in the way a zombie villager,
```

YanisBft (approved):

enbrain (approved):

PR #2462, opened 2021-06-02
Map some BeeEntity constants

IotaBread:
`field_30284` is related to the minimum chance of being hurt because the bee has stung (`this.random.nextInt(MathHelper.clamp(1200 - this.ticksSinceSting, 1, 1200)) == 0` in `#mobTick`, 1200 is the field), but I can't think of a good name for it.

One of either `field_30274` and `field_30275` is the pollination delay and the find hive (`FindHiveGoal`) delay, but we won't be able to know which is which until a value changes.

The names of 30291 and 30293 don't convince me, open to suggestions

### DIFF
--- a/mappings/net/minecraft/entity/passive/BeeEntity.mapping
+++ b/mappings/net/minecraft/entity/passive/BeeEntity.mapping
@@ -32,7 +32,7 @@ CLASS net/minecraft/class_4466 net/minecraft/entity/passive/BeeEntity
 	FIELD field_30286 POLLINATION_FAIL_TICKS I
 		COMMENT The duration in ticks when a bee's pollination is considered failed.
 	FIELD field_30288 MAX_POLLINATED_CROPS I
-	FIELD field_30289 STING_POSION_DURATION_NORMAL_DIFFICULTY I
+	FIELD field_30289 STING_POISON_DURATION_NORMAL_DIFFICULTY I
 	FIELD field_30290 STING_POISON_DURATION_HARD_DIFFICULTY I
 	FIELD field_30291 MAX_REACH_DISTANCE I
 		COMMENT The maximum distance a bee will wander from the hive before getting lost.
--- a/mappings/net/minecraft/entity/passive/BeeEntity.mapping
+++ b/mappings/net/minecraft/entity/passive/BeeEntity.mapping
@@ -34,11 +34,9 @@ CLASS net/minecraft/class_4466 net/minecraft/entity/passive/BeeEntity
 	FIELD field_30288 MAX_POLLINATED_CROPS I
 	FIELD field_30289 NORMAL_DIFFICULTY_STING_POISON_DURATION I
 	FIELD field_30290 HARD_DIFFICULTY_STING_POISON_DURATION I
-	FIELD field_30291 MAX_REACH_DISTANCE I
-		COMMENT The maximum distance a bee will wander from the hive before getting lost.
-		COMMENT Also the maximum distance to a flower.
-	FIELD field_30293 MAX_DISTANCE_FROM_HIVE I
-		COMMENT The maximum distance a bee should reach before having to return to the hive.
+	FIELD field_30291 FLOWER_SCAN_RANGE I
+		COMMENT The range of scan for flowers that can be navigated to.
+	FIELD field_30293 HIVE_SCAN_RANGE I
 	METHOD method_21769 addParticle (Lnet/minecraft/class_1937;DDDDDLnet/minecraft/class_2394;)V
 		ARG 1 world
 		ARG 2 lastX
--- a/mappings/net/minecraft/entity/passive/BeeEntity.mapping
+++ b/mappings/net/minecraft/entity/passive/BeeEntity.mapping
@@ -32,8 +32,8 @@ CLASS net/minecraft/class_4466 net/minecraft/entity/passive/BeeEntity
 	FIELD field_30286 POLLINATION_FAIL_TICKS I
 		COMMENT The duration in ticks when a bee's pollination is considered failed.
 	FIELD field_30288 MAX_POLLINATED_CROPS I
-	FIELD field_30289 STING_POISON_DURATION_NORMAL_DIFFICULTY I
-	FIELD field_30290 STING_POISON_DURATION_HARD_DIFFICULTY I
+	FIELD field_30289 NORMAL_DIFFICULTY_STING_POISON_DURATION I
+	FIELD field_30290 HARD_DIFFICULTY_STING_POISON_DURATION I
 	FIELD field_30291 MAX_REACH_DISTANCE I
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/entity/passive/BeeEntity.mapping
```diff
@@ -31,6 +31,14 @@ CLASS net/minecraft/class_4466 net/minecraft/entity/passive/BeeEntity
 		COMMENT A bee will start moving to a flower once this time in ticks has passed from a pollination.
 	FIELD field_30286 POLLINATION_FAIL_TICKS I
 		COMMENT The duration in ticks when a bee's pollination is considered failed.
+	FIELD field_30288 MAX_POLLINATED_CROPS I
+	FIELD field_30289 STING_POSION_DURATION_NORMAL_DIFFICULTY I
```
haykam821:
```suggestion
	FIELD field_30289 STING_POISON_DURATION_NORMAL_DIFFICULTY I
```

haykam821 (changes requested):

intact:
field_30284 is more or less death timer (after bee's attack succeeds)

mappings/net/minecraft/entity/passive/BeeEntity.mapping
```diff
@@ -31,6 +31,14 @@ CLASS net/minecraft/class_4466 net/minecraft/entity/passive/BeeEntity
 		COMMENT A bee will start moving to a flower once this time in ticks has passed from a pollination.
 	FIELD field_30286 POLLINATION_FAIL_TICKS I
 		COMMENT The duration in ticks when a bee's pollination is considered failed.
+	FIELD field_30288 MAX_POLLINATED_CROPS I
+	FIELD field_30289 STING_POISON_DURATION_NORMAL_DIFFICULTY I
+	FIELD field_30290 STING_POISON_DURATION_HARD_DIFFICULTY I
+	FIELD field_30291 MAX_REACH_DISTANCE I
+		COMMENT The maximum distance a bee will wander from the hive before getting lost.
+		COMMENT Also the maximum distance to a flower.
+	FIELD field_30293 MAX_DISTANCE_FROM_HIVE I
```
liach: these sound like max and min wander range

mappings/net/minecraft/entity/passive/BeeEntity.mapping
```diff
@@ -31,6 +31,14 @@ CLASS net/minecraft/class_4466 net/minecraft/entity/passive/BeeEntity
 		COMMENT A bee will start moving to a flower once this time in ticks has passed from a pollination.
 	FIELD field_30286 POLLINATION_FAIL_TICKS I
 		COMMENT The duration in ticks when a bee's pollination is considered failed.
+	FIELD field_30288 MAX_POLLINATED_CROPS I
+	FIELD field_30289 STING_POISON_DURATION_NORMAL_DIFFICULTY I
+	FIELD field_30290 STING_POISON_DURATION_HARD_DIFFICULTY I
```
liach: please make the difficulty prefix than suffix, according to yarn guidelines as these are durations than difficulties

IotaBread: Should it be with or without `DIFFICULTY`?

liach: either, but usually we know peaceful easy hard are difficulties

mappings/net/minecraft/entity/passive/BeeEntity.mapping
IotaBread: I think the javadoc of 30291 could be changed to "The maximum distance to a position before it is considered too far" 
because of the explanation below.

30291 is used only in `#isTooFar`, which is used in `MoveToFlowerGoal` and `MoveToHiveGoal`
`MoveToFlowerGoal`: when the bee isn't navigating checks whether the `#flowerPos` is too far, if it is, `#flowerPos` is set to null, if it isn't, the bee navigates to the flower.
`MoveToHiveGoal`: when the bee isn't navigating checks whether the `#hivePos` is more than 16 blocks away (`field_30293`), if it is and it is too far (`#isTooFar`) sets the bee as lost, if it isn't too far navigates back to the hive

liach: so it's the scan range for flower and hive respectively. if so consider naming them to show they are similar

liach (approved):

Juuxel (approved):

YanisBft (approved):

mappings/net/minecraft/entity/passive/BeeEntity.mapping
```diff
@@ -31,6 +31,12 @@ CLASS net/minecraft/class_4466 net/minecraft/entity/passive/BeeEntity
 		COMMENT A bee will start moving to a flower once this time in ticks has passed from a pollination.
 	FIELD field_30286 POLLINATION_FAIL_TICKS I
 		COMMENT The duration in ticks when a bee's pollination is considered failed.
+	FIELD field_30288 MAX_POLLINATED_CROPS I
+	FIELD field_30289 NORMAL_DIFFICULTY_STING_POISON_DURATION I
+	FIELD field_30290 HARD_DIFFICULTY_STING_POISON_DURATION I
+	FIELD field_30291 FLOWER_SCAN_RANGE I
+		COMMENT The range of scan for flowers that can be navigated to.
+	FIELD field_30293 HIVE_SCAN_RANGE I
```
haykam821:
```suggestion
	FIELD field_30293 MIN_HIVE_RETURN_DISTANCE I
		COMMENT The minimum distance that bees will immediately return to their hive at.
```

mappings/net/minecraft/entity/passive/BeeEntity.mapping
```diff
@@ -31,6 +31,12 @@ CLASS net/minecraft/class_4466 net/minecraft/entity/passive/BeeEntity
 		COMMENT A bee will start moving to a flower once this time in ticks has passed from a pollination.
 	FIELD field_30286 POLLINATION_FAIL_TICKS I
 		COMMENT The duration in ticks when a bee's pollination is considered failed.
+	FIELD field_30288 MAX_POLLINATED_CROPS I
+	FIELD field_30289 NORMAL_DIFFICULTY_STING_POISON_DURATION I
+	FIELD field_30290 HARD_DIFFICULTY_STING_POISON_DURATION I
+	FIELD field_30291 FLOWER_SCAN_RANGE I
+		COMMENT The range of scan for flowers that can be navigated to.
```
haykam821: I don't see how this relates to flowers specifically:

```suggestion
	FIELD field_30291 TOO_FAR_DISTANCE I
		COMMENT The minimum distance that bees lose their hive or flower position at.
```

haykam821 (changes requested):
I do not think the distance-related constants are correct.

modmuss50 (approved):

PR #2466, opened 2021-06-04
1.17-pre4 client stuff

apple502j:
Main changes:
- Blocklist changed once again. isBlocked is isAllowed now.
- Keyboard: while the color yellow may suggest a warning, it seems like this is used in non-warning situations; thus renamed. `processDebugKeys` is probably F3 related but it looks like a dead code.
- Some rendering maps, probably related to glowing signs
- Mapped init
- Mapped lambda params
- etc
- Did I do too much?


apple502j:
LOL, seems like I overworked too much, the prerelease version bumped 😛

mappings/net/minecraft/client/font/TextHandler.mapping
```diff
@@ -115,11 +115,11 @@ CLASS net/minecraft/class_5225 net/minecraft/client/font/TextHandler
 	METHOD method_35714 (Lnet/minecraft/class_5348;ILnet/minecraft/class_2583;Lnet/minecraft/class_5348;)Ljava/util/List;
 		ARG 2 maxWidth
 		ARG 3 style
-	METHOD method_35715 (Ljava/lang/String;ILnet/minecraft/class_2583;)I
+	METHOD method_35715 getLengthWithWidthLimit (Ljava/lang/String;ILnet/minecraft/class_2583;)I
```
liach: I suggest calling these `getLimitedStringLength` and `limitString` as you are imposing a width limit. if this text handler is exclusively for string widths, you can rename the class to show this feature.

mappings/net/minecraft/client/render/BlockBreakingInfo.mapping
```diff
@@ -15,3 +17,4 @@ CLASS net/minecraft/class_3191 net/minecraft/client/render/BlockBreakingInfo
 		ARG 1 lastUpdateTick
 	METHOD method_13990 getLastUpdateTick ()I
 	METHOD method_13991 getPos ()Lnet/minecraft/class_2338;
+	METHOD method_34868 getActorNetworkId ()I
```
liach:
```suggestion
	METHOD method_34868 getActorId ()I
```
Since nowadays we call the entity's network id getter as `getId`

mappings/net/minecraft/client/render/WorldRenderer.mapping
```diff
@@ -292,6 +296,8 @@ CLASS net/minecraft/class_761 net/minecraft/client/render/WorldRenderer
 	METHOD method_34811 getChunkCount ()D
 	METHOD method_34812 getViewDistance ()D
 	METHOD method_35774 reloadTransparencyShader ()V
+	METHOD method_35775 captureFrustum ()V
+	METHOD method_35776 killFrustum ()V
```
liach, line 300: Hmm, do we usually use discard/drop more often? I think we use kill more often in an entity-related context, but I am not sure.

mappings/net/minecraft/client/render/block/entity/BlockEntityRenderer.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS net/minecraft/class_827 net/minecraft/client/render/block/entity/BlockEntityRenderer
+	METHOD method_33893 getRenderedDistance ()I
```
liach:
```suggestion
	METHOD method_33893 getRenderDistance ()I
```
I think in minecraft, the distance at which further objects will not render is called "render distance" as seen in the options.

mappings/net/minecraft/client/render/WorldRenderer.mapping
apple502j: There is a string in Keyboard that logged "killed frustum" after this is called. Maybe we can rename; or we can keep.

liach: good name in this case!

mappings/net/minecraft/client/render/block/entity/SignBlockEntityRenderer.mapping
```diff
@@ -2,6 +2,8 @@ CLASS net/minecraft/class_837 net/minecraft/client/render/block/entity/SignBlock
 	FIELD field_27754 typeToModel Ljava/util/Map;
 	FIELD field_27755 textRenderer Lnet/minecraft/class_327;
 	FIELD field_32830 STICK Ljava/lang/String;
+	FIELD field_33962 GLOWING_BLACK_COLOR I
+	FIELD field_33963 RENDERED_DISTANCE I
```
liach:
```suggestion
	FIELD field_33963 RENDER_DISTANCE I
```

mappings/net/minecraft/client/render/chunk/ChunkRendererRegion.mapping
```diff
@@ -16,8 +16,13 @@ CLASS net/minecraft/class_853 net/minecraft/client/render/chunk/ChunkRendererReg
 		ARG 4 chunks
 		ARG 5 startPos
 		ARG 6 endPos
+	METHOD method_30000 areAllSectionsEmptyBetween (Lnet/minecraft/class_2338;Lnet/minecraft/class_2338;II[[Lnet/minecraft/class_2818;)Z
```
liach:
```suggestion
	METHOD method_30000 isEmptyBetween (Lnet/minecraft/class_2338;Lnet/minecraft/class_2338;II[[Lnet/minecraft/class_2818;)Z
```

liach (approved):
Should be good, I think.

modmuss50 (approved):
Awesome 👍

PR #2478, opened 2021-06-07
Sampler changes

liach:
Trying to map sampler. this is still a work in progress and is subject to change; many things still need to be named/renamed. Feel free to suggest general models on these classes.


liach:
Now I have fully mapped the recorder and sampler parts, which is the focus of this pr. I didn't investigate much into the profilers, though.

YanisBft (approved):

haykam821 (approved):

liach:
@modmuss50 mind review this? This is the last big patch of unmapped classes in 1.17.

PR #2479, opened 2021-06-07
Correct visibleOnly method name -> ignoreVisibility

supersaiyansubtlety:
`visibleOnly` actually sets `respectsVisibility` to `false`
```java
    public TargetPredicate visibleOnly() {
        this.respectsVisibility = false;
        return this;
    }
```
later in `test`, you can see that `respectsVisibility` is correctly named, as it precedes the visibility check:
```java
                if (this.respectsVisibility && baseEntity instanceof MobEntity && !((MobEntity)baseEntity).getVisibilityCache().canSee(targetEntity)) {
                    return false;
                }
```
so `visibleOnly`'s name implies it does the inverse of what it actually does.


liach (approved):
Think this should be right. Anyways I found that this target predicate really needs a revamp (same for ai sensor), so this should be good for now. Thanks!

modmuss50 (approved):

PR #2481, opened 2021-06-07
Some client mappings

### LINKED ISSUE #1925: `getSelected` -> `getSelectedOrNull`
`EntryListWidget` contains a mapping `getSelected` for `method_25334`, which returns a `EntryListWidget.Entry`.

However, there also exists a `method_20159` which is a wrapper around `method_25334` returning an `Optional<EntryListWidget.Entry>` (via `return java.util.Optional.ofNullable(this.method_25334());`).

I propose changing the name of the former function to inform the user of the nullability of its return type, and name the second function `getSelected`.

For the sake of following convention, we might also consider naming `method_20159` `getOptionalSelected` or `getSelectedOrEmpty`.

The convention that exists so far might be sufficient, but I haven't actually found any other examples of a getter function having a non-nullable alternative, and explicitly mentioning the nullability of the return type seems beneficial for the programmer so they can more easily avoid making a mistake.

YanisBft:
- Constants in OcelotEntityModel
- Fix #1925 
- Little progress on #2174

### DIFF
--- a/mappings/net/minecraft/client/gui/widget/EntryListWidget.mapping
+++ b/mappings/net/minecraft/client/gui/widget/EntryListWidget.mapping
@@ -90,7 +90,7 @@ CLASS net/minecraft/class_350 net/minecraft/client/gui/widget/EntryListWidget
 	METHOD method_25333 setLeftPos (I)V
 		ARG 1 left
 	METHOD method_25334 getSelectedOrNull ()Lnet/minecraft/class_350$class_351;
-		COMMENT Returns the selected entry of this entry list, or {@code null} if there is not.
+		COMMENT {@return the selected entry of this entry list, or {@code null} if there is not.}
 	METHOD method_25335 moveSelection (Lnet/minecraft/class_350$class_5403;)V
 		ARG 1 direction
 	METHOD method_25337 getRowTop (I)I
--- a/mappings/net/minecraft/client/gui/widget/EntryListWidget.mapping
+++ b/mappings/net/minecraft/client/gui/widget/EntryListWidget.mapping
@@ -89,7 +89,8 @@ CLASS net/minecraft/class_350 net/minecraft/client/gui/widget/EntryListWidget
 		ARG 1 index
 	METHOD method_25333 setLeftPos (I)V
 		ARG 1 left
-	METHOD method_25334 getSelected ()Lnet/minecraft/class_350$class_351;
+	METHOD method_25334 getSelectedOrNull ()Lnet/minecraft/class_350$class_351;
+		COMMENT Returns the selected entry of this entry list, or {@code null} if there is not.
 	METHOD method_25335 moveSelection (Lnet/minecraft/class_350$class_5403;)V
 		ARG 1 direction
 	METHOD method_25337 getRowTop (I)I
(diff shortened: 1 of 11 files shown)


mappings/net/minecraft/client/gui/widget/EntryListWidget.mapping
```diff
@@ -89,7 +89,8 @@ CLASS net/minecraft/class_350 net/minecraft/client/gui/widget/EntryListWidget
 		ARG 1 index
 	METHOD method_25333 setLeftPos (I)V
 		ARG 1 left
-	METHOD method_25334 getSelected ()Lnet/minecraft/class_350$class_351;
+	METHOD method_25334 getSelectedOrNull ()Lnet/minecraft/class_350$class_351;
+		COMMENT Returns the selected entry of this entry list, or {@code null} if there is not.
```
IotaBread:
```suggestion
		COMMENT {@return the selected entry of this entry list, or {@code null} if there is not.}
```

mappings/net/minecraft/client/gui/widget/EntryListWidget.mapping
```diff
@@ -89,7 +89,8 @@ CLASS net/minecraft/class_350 net/minecraft/client/gui/widget/EntryListWidget
 		ARG 1 index
 	METHOD method_25333 setLeftPos (I)V
 		ARG 1 left
-	METHOD method_25334 getSelected ()Lnet/minecraft/class_350$class_351;
+	METHOD method_25334 getSelectedOrNull ()Lnet/minecraft/class_350$class_351;
+		COMMENT {@return the selected entry of this entry list, or {@code null} if there is not.}
```
liach:
```suggestion
		COMMENT {@return the selected entry of this entry list, or {@code null} if there is none}
```

mappings/net/minecraft/server/dedicated/ServerPropertiesHandler.mapping
```diff
@@ -49,8 +49,6 @@ CLASS net/minecraft/class_3806 net/minecraft/server/dedicated/ServerPropertiesHa
 	FIELD field_26899 textFilteringConfig Ljava/lang/String;
 	FIELD field_27845 requireResourcePack Z
 	FIELD field_33432 resourcePackPrompt Ljava/lang/String;
-	METHOD <init> (Ljava/util/Properties;)V
-		ARG 1 properties
```
liach, line 53: Is this just a byproduct of dropping invalid mappings?

YanisBft: Yep it is

mappings/net/minecraft/client/gui/widget/EntryListWidget.mapping
liach: Just fyi, the default behavior of inline return tag is it adds an extra comma, so in the `{@return}` you just put what you would put in a regular block `@return` tag.

mappings/net/minecraft/server/dedicated/ServerPropertiesHandler.mapping
liach: If this is, you can keep it. If other prs have the same change, it's acceptable as git can handle such merges properly.

liach (approved):

IotaBread (approved):

PR #2484, opened 2021-06-08
Add and remove release/snapshot labels when updating branch

YanisBft:
I don't know if it works but I hope so

### DIFF
--- a/.github/workflows/update-base.yml
+++ b/.github/workflows/update-base.yml
@@ -34,14 +34,14 @@ jobs:
                     // Updates the pull request with the latest upstream changes.
                     await github.pulls.updateBranch({ owner, repo, pull_number })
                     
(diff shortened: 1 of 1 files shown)


modmuss50 (approved):

liach:
how about jokes and combat tests

haykam821:
`base` here is the default branch of Yarn:

https://github.com/FabricMC/yarn/blob/8fc8903adae1c02ddb42585882fa488f328e52d3/.github/workflows/update-base.yml#L21-L22

Jokes and combat tests shouldn't be the default Yarn branch, so this shouldn't be an issue.

haykam821 (approved):
It might be worth it in a later pull request to query the labels added once and store modifications locally so they can be updated in one API call.

PR #2489, opened 2021-06-10
Add mappings to StructurePoolElement

frqnny:
This class is used in Data and Generator classes in net.minecraft.structure in order to make structure pools via code. Mapping it will allow modders to understand what these methods do, instead of having to dig around vanilla Data and Generator classes for which ones to use.

The naming follows names like Lists#newArrayList.

### DIFF
--- a/mappings/net/minecraft/structure/pool/StructurePoolElement.mapping
+++ b/mappings/net/minecraft/structure/pool/StructurePoolElement.mapping
@@ -30,3 +30,30 @@ CLASS net/minecraft/class_3784 net/minecraft/structure/pool/StructurePoolElement
 		ARG 3 rotation
 	METHOD method_16757 getType ()Lnet/minecraft/class_3816;
 	METHOD method_19308 getGroundLevelDelta ()I
+	METHOD method_30421 newFeaturePoolElement (Lnet/minecraft/class_2975;)Ljava/util/function/Function;
+		ARG 0 processors
+	METHOD method_30422 (Lnet/minecraft/class_2975;Lnet/minecraft/class_3785$class_3786;)Lnet/minecraft/class_3776;
+		ARG 1 projection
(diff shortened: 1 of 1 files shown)


kvverti:
Technically speaking, these methods return factories rather than the objects themselves.

frqnny:
`StructurePools.register(
                new StructurePool(
                        BASE,
                        new Identifier("empty"),
                        ImmutableList.of(
                                new Pair<>(StructurePoolElement.newProcessedSinglePoolElement(MoStructures.MODID + ":lighthouse/base", MoStructures.PIRATE_SHIP_LIST), 1)
                        ),
                        StructurePool.Projection.RIGID
                )
        );`

This is how you would use one of these methods. IMO it makes more sense for a modder to do this. 
Otherwise what names should I try giving them? They are already pretty long, appending "factory" would probably be a bad idea.

mappings/net/minecraft/structure/pool/StructurePoolElement.mapping
```diff
@@ -30,3 +30,30 @@ CLASS net/minecraft/class_3784 net/minecraft/structure/pool/StructurePoolElement
 		ARG 3 rotation
 	METHOD method_16757 getType ()Lnet/minecraft/class_3816;
 	METHOD method_19308 getGroundLevelDelta ()I
+	METHOD method_30421 newFeaturePoolElement (Lnet/minecraft/class_2975;)Ljava/util/function/Function;
```
liach: Do we use names like `ofFeature` `ofLegacySingle` `ofProcessedLegacySingle`

frqnny: I think so. I'll update my PR to use those names then. Glad their shorter 😄

liach (approved):

enbrain (approved):

YanisBft (approved):

Issue #2389, opened 2021-05-08
BooleanBiFunction field names are inconsistent


Player3324:
BooleanBiFunction has a weird mix of pseudo-logic-op, descriptive or weird names.

Making them all more descriptive could involve the following changes:
- AND -> BOTH
- OR -> ANY
- NOT_OR -> NEITHER
- NOT_AND -> NOT_BOTH
- NOT_SAME -> DIFFERENT
- CAUSES -> NOT_ONLY_FIRST
- CAUSED_BY -> NOT_ONLY_SECOND

I don't think binary logic operator names are particularly good for boolean parameters, contrary to working with bit vectors, they also require a certain level familiarity to be useful. If it was desirable however, they should use their proper terms like NOR, NAND etc instead of NOT_OR etc.

The AND/OR/NAND/etc terms could be put in comments.

liach:
> contrary to working with bit vectors

Hmm, but their main usage in vanilla is for value merging in voxel shapes, mainly bit set voxel set

PR #2493, opened 2021-06-11
Document the BooleanBiFunction interface

haykam821:
Relates to #2389


liach (changes requested):
Imo we should only document it as a part of the voxel shape system, as this (actually a boolean binary operator) is exclusively used in the voxel shape combination logic there. I would put this on hold.

mappings/net/minecraft/util/function/BooleanBiFunction.mapping
```diff
@@ -1,17 +1,34 @@
...
 	FIELD field_16895 TRUE Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that always returns {@code true}.
 	FIELD field_16896 AND Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if both arguments are {@code true}.
 	FIELD field_16897 FALSE Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that always returns {@code false}.
 	FIELD field_16898 NOT_SECOND Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if the second argument is {@code false}.
 	FIELD field_16899 CAUSED_BY Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if the first argument is {@code true} and the second argument is {@code false}.
```
liach:
```suggestion
		COMMENT A {@link BooleanBiFunction} that returns {@code true} if the first argument is {@code true} or the second argument is {@code false}.
```

mappings/net/minecraft/util/function/BooleanBiFunction.mapping
```diff
@@ -1,17 +1,34 @@
...
 	FIELD field_16885 NOT_OR Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if neither argument is {@code true}.
 	FIELD field_16886 ONLY_FIRST Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if only the first argument is {@code true}.
 	FIELD field_16887 SECOND Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if the second argument is {@code true}.
 	FIELD field_16888 NOT_AND Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if either argument is {@code false}.
 	FIELD field_16889 CAUSES Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if the first argument is {@code false} and the second argument is {@code true}.
```
liach:
```suggestion
		COMMENT A {@link BooleanBiFunction} that returns {@code true} if the first argument is {@code false} or the second argument is {@code true}.
```

mappings/net/minecraft/util/function/BooleanBiFunction.mapping
```diff
@@ -1,17 +1,34 @@
 CLASS net/minecraft/class_247 net/minecraft/util/function/BooleanBiFunction
+	COMMENT A function that is provided two booleans and returns one boolean.
 	FIELD field_1366 OR Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if either argument is {@code true}.
 	FIELD field_16885 NOT_OR Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if neither argument is {@code true}.
 	FIELD field_16886 ONLY_FIRST Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if only the first argument is {@code true}.
 	FIELD field_16887 SECOND Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if the second argument is {@code true}.
```
liach, line 10:
```suggestion
		COMMENT A {@link BooleanBiFunction} that returns the second argument.
```

mappings/net/minecraft/util/function/BooleanBiFunction.mapping
```diff
@@ -1,17 +1,34 @@
...
 	FIELD field_16888 NOT_AND Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if either argument is {@code false}.
 	FIELD field_16889 CAUSES Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if the first argument is {@code false} and the second argument is {@code true}.
 	FIELD field_16890 SAME Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if both arguments are {@code true} or both arguments are {@code false}.
 	FIELD field_16891 FIRST Lnet/minecraft/class_247;
+		COMMENT A {@link BooleanBiFunction} that returns {@code true} if the first argument is {@code true}.
```
liach, line 18:
```suggestion
		COMMENT A {@link BooleanBiFunction} that returns the first argument.
```

liach (changes requested):
After looking, I think we should have this for now; though this is mainly used for combination in bit set voxel set, having something is better.

The main problems right now is you messed up the cause logic.

Additional suggestions: name the parameters in `apply` `a` `b` and use `a` `b` than "first argument" "second argument"; simplify the wording for `FIRST` and `SECOND`

mappings/net/minecraft/util/function/BooleanBiFunction.mapping
haykam821: I think that it is more clear to state the output in terms of the input without explicitly rewriting the expression in sentence form. This may be applicable to other, more complex data types, but not booleans.

mappings/net/minecraft/util/function/BooleanBiFunction.mapping
haykam821: See [here](https://github.com/FabricMC/yarn/pull/2493#discussion_r654747421)

mappings/net/minecraft/util/function/BooleanBiFunction.mapping
liach: If you want accuracy, should we use `returns ... if and only if` in place of all `returns ... if`?

haykam821: It could be `returns ... exclusively if` as well. I assumed all other cases returning `false` would be implied but I can clarify it if necessary.

liach: Then the current form should be good. Just make sure you fix the "causes" and "caused by" cases. a -> b is a causes b, or `!a || b` (like `a ? b : true`).

mappings/net/minecraft/util/function/BooleanBiFunction.mapping
liach:
```suggestion
		COMMENT A {@link BooleanBiFunction} that returns {@code true} if the first argument is {@code true} or the second argument is {@code false}. Equivalent to {@code second ? first : true}.
```

mappings/net/minecraft/util/function/BooleanBiFunction.mapping
liach:
```suggestion
		COMMENT A {@link BooleanBiFunction} that returns {@code true} if the first argument is {@code false} or the second argument is {@code true}. Equivalent to {@code first ? second : true}.
```

liach (changes requested):

liach:
Also, can you consider naming the parameters of the `apply(boolean, boolean)` method?

liach:
Still, I am just requesting fixing `and` with `or` for `CAUSED` and `CAUSED_BY`. In those ones, if the premise is false, then the statement is vacuously true. @haykam821 mind check that out? See https://en.wikipedia.org/wiki/Vacuous_truth#Scope_of_the_concept (that first sentence is a good summary)

liach:
While the rest of the changes are always good, you should fix the logical errors in the `CAUSED` and `CAUSED_BY` docs.

liach:
@haykam821 Mind apply https://github.com/FabricMC/yarn/pull/2493#discussion_r654740944 and https://github.com/FabricMC/yarn/pull/2493#discussion_r654740925 suggestions? These are logical errors in your documentation. Otherwise, this is good to go.

haykam821:
I used the following script to verify these inputs:

```js
(() => {
    const inputs = [
        [false, false],
        [false, true],
        [true, false],
        [true, true]
    ];
    const biFunction = ((a, b) => !a || b);

    return inputs.reduce((acc, [a, b]) => {
        const label = (a + "")[0] + (b + "")[0];
        acc[label] = biFunction(a, b);
        return acc;
    }, {});
})();
```

liach (approved):

PR #2494, opened 2021-06-13
change SplashScreen to SplashOverlay

alphaqu:
SplashScreen does not extend Screen but instead Overlay.

### DIFF
--- a/mappings/net/minecraft/client/gui/screen/SplashOverlay.mapping
+++ b/mappings/net/minecraft/client/gui/screen/SplashOverlay.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_425 net/minecraft/client/gui/screen/SplashScreen
+CLASS net/minecraft/class_425 net/minecraft/client/gui/screen/SplashOverlay
 	FIELD field_17767 reload Lnet/minecraft/class_4011;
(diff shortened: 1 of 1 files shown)


liach (approved):

enbrain (approved):

haykam821 (approved):

modmuss50 (commented):
I suggest we revert this change, as its not an overlay at all. SplashScreen is fine.

liach:
Hmm, it does work like an overlay when like, select a language and click done, or when a server-sent resource pack is loaded?

Also see https://github.com/FabricMC/yarn/pull/1992#discussion_r567935086

Juuxel:
"Splash screen" is an established term on its own, however, and that's exactly what this class is. See: https://github.com/FabricMC/yarn/pull/1992#issuecomment-770912109

alphaqu:
Yes but it causes confusion when searching and its technically not a screen at all. On a usage level it might be called a Screen but i think Yarn should be going for clarity about what the Class actually is.

alphaqu:
Or it may also be that Overlay should be renamed to Splash or maybe even SplashScreen

Player3324:
It's more of a loading progress bar overlay these days that gets used in a few locations.

PR #2495, opened 2021-06-14
Map SplashOverlay.renderProgressBar

jackassmc:
`x1` and `y1` could also be called `topLeftX` and `topLeftY`, `x2` and `y2` could be `bottomRightX` and `bottomRightY`.
Or instead of `x1` `y1` `x2` `y2` it could be `left` `top` `right` `bottom` but the order didn't feel right.

The fields were intended to be unpicked but unpicker doesn't support fields yet.

### DIFF
--- a/mappings/net/minecraft/client/gui/screen/SplashOverlay.mapping
+++ b/mappings/net/minecraft/client/gui/screen/SplashOverlay.mapping
@@ -8,6 +8,8 @@ CLASS net/minecraft/class_425 net/minecraft/client/gui/screen/SplashOverlay
 	FIELD field_18220 reloadStartTime J
 	FIELD field_2483 LOGO Lnet/minecraft/class_2960;
 	FIELD field_25041 BRAND_ARGB Ljava/util/function/IntSupplier;
+	FIELD field_32247 RELOAD_COMPLETE_FADE_DURATION J
+	FIELD field_32248 RELOAD_START_FADE_DURATION J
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/client/gui/screen/SplashOverlay.mapping
```diff
@@ -17,6 +19,10 @@ CLASS net/minecraft/class_425 net/minecraft/client/gui/screen/SplashOverlay
 		ARG 4 reloading
 	METHOD method_18103 renderProgressBar (Lnet/minecraft/class_4587;IIIIF)V
 		ARG 1 matrices
+		ARG 2 x1
+		ARG 3 y1
+		ARG 4 x2
+		ARG 5 y2
```
liach: Hmm, how about `min`/`max` (`start`/`end`) `X`/`Y`? Or if we stick with numbers, `01` or `12` suffixes?

jackassmc: Looking at it without looking at the function made me realize that number based names are not very clear.
I like `min`/`max`, should it be `minX` or `xMin`?

Juuxel: `minX` is the Yarn convention for that.

liach (approved):

YanisBft (approved):

PR #2499, opened 2021-06-18
Fix the update base script calling a non-existent contains function

haykam821:


YanisBft (approved):
Thank you, I don't know why I saw "contains" in the docs

haykam821:
Hopefully the script can be hardened to comment an error message with stack trace and remove the label as well. This should be good for now to fix #2484.

liach (approved):
:smile:

modmuss50:
Really need to split this out into a proper custom action with some tests as its a right pita to test atm

PR #2257, opened 2021-04-02
NBT refactor: step 3

LambdAurora:
Follow up to my other NBT PRs (like #2090)

This one fixes parameter names and some method names.

`ItemStack` remains untouched for the main part. If renames have to happen for `ItemStack` it will be in another smaller PR.

PR #2427, opened 2021-05-21
Backport 21w20a mappings to 1.16.5

LambdAurora:
As said in the title, another episode to the saga.

Main motivation is the NBT and resource reloader refactor.

I tried to reject anything that doesn't make sense in 1.16.5 (unpicked values in Javadocs, or 1.17 specific stuff) with an exception: `Material#MOSS`, despite moss not being in the game at the time, such being unused, they share the same intermediary, which means it would be safe to assume Mojang was already working on moss at the time and leaked its material in earlier code bases.

Hopefully won't be too much of a pain to review.

PR #2511, opened 2021-06-21
ItemStack refactor and documentation.

LambdAurora:
A long overdue refactor!

Also includes some javadocs.

This is a followup to #2257.

If merged, if this could be cherry picked into the 1.17 branch too, it would be much appreciated.


Technici4n:
Are we sure about this? I like the separation between the tag of the stack, and its serialized version (NBT).

LambdAurora:
> Are we sure about this? I like the separation between the tag of the stack, and its serialized version (NBT).

Though it's where there's the peak conflict between Item Tags and Item custom NBT.

And there's prior art from 2019, it's a long overdue refactor.

Technici4n:
You never encounter a type conflict because you can just do `NbtCompound tag = stack.getTag()` nowadays. The distinction with `NbtCompound nbt = stack.toNbt()` is quite clear to me. It's a lot clearer than `stack.getNbt()` vs `stack.toNbt()`.

LambdAurora:
> You never encounter a type conflict because you can just do `NbtCompound tag = stack.getTag()` nowadays. The distinction with `NbtCompound nbt = stack.toNbt()` is quite clear to me. It's a lot clearer than `stack.getNbt()` vs `stack.toNbt()`.

There's no `toNbt` method, there's `writeNbt(NbtCompound)` though.

Juuxel:
I'd leave the item stack extra data compound as "tag", as that's its specific name also in the serialised form (whether in JSON or in NBT).

mappings/net/minecraft/item/ItemStack.mapping
```diff
@@ -209,20 +251,27 @@ CLASS net/minecraft/class_1799 net/minecraft/item/ItemStack
...
 		ARG 1 amount
 	METHOD method_7972 copy ()Lnet/minecraft/class_1799;
+		COMMENT Creates and returns a copy of this item stack.
 	METHOD method_7973 areEqual (Lnet/minecraft/class_1799;Lnet/minecraft/class_1799;)Z
 		ARG 0 left
 		ARG 1 right
 	METHOD method_7974 setDamage (I)V
 		ARG 1 damage
-	METHOD method_7975 areTagsEqual (Lnet/minecraft/class_1799;Lnet/minecraft/class_1799;)Z
+	METHOD method_7975 areNbtEquals (Lnet/minecraft/class_1799;Lnet/minecraft/class_1799;)Z
```
dexman545: This compares 2 objects for equality, not the current object and another - it should be `areNbtEqual` or `areNbtsEqual`

liach (commented):
After thinking, using `nbt` throughout isn't bad.

This pr still has a few inconsistencies. I suggest:
 - `removeSubNbt` -> `removeNbt`
 - `putNbtElement` -> `setNbt` (distinguished with an extra `String` parameter)
given the getter names you chose.

mappings/net/minecraft/item/ItemStack.mapping
```diff
@@ -24,7 +25,7 @@ CLASS net/minecraft/class_1799 net/minecraft/item/ItemStack
 	FIELD field_8037 EMPTY Lnet/minecraft/class_1799;
 	FIELD field_8038 item Lnet/minecraft/class_1792;
 	FIELD field_8039 lastDestroyPos Lnet/minecraft/class_2694;
-	FIELD field_8040 tag Lnet/minecraft/class_2487;
+	FIELD field_8040 nbt Lnet/minecraft/class_2487;
```
liach, line 102: If this rename is made, should document that the nbt is stored in `Tag` sub compound within the nbt form of an item stack.

Also, doesn't mojang have a constant for `Tag` when they have those for `CanDestroy` etc?

mappings/net/minecraft/item/ItemStack.mapping
```diff
@@ -92,7 +93,15 @@ CLASS net/minecraft/class_1799 net/minecraft/item/ItemStack
 	METHOD method_7910 finishUsing (Lnet/minecraft/class_1937;Lnet/minecraft/class_1309;)Lnet/minecraft/class_1799;
 		ARG 1 world
 		ARG 2 user
-	METHOD method_7911 getOrCreateSubTag (Ljava/lang/String;)Lnet/minecraft/class_2487;
+	METHOD method_7911 getOrCreateNbt (Ljava/lang/String;)Lnet/minecraft/class_2487;
+		COMMENT Returns the {@linkplain NbtCompound} at the specified key in this item stack's custom NBT, or creates it if absent.
```
liach:
```suggestion
		COMMENT {@return the compound NBT at the specified key in this item stack's NBT, or a new compound if absent} The modifications to the new compound are not stored into the stack while those to the existing compound are, so always call  {@link setNbt(String, NbtElement)} to save the changes you made.
```

A few points from this sample suggestion:
1. Don't use `{@linkplain NbtCompound}`, which will render like "[NbtCompound](http://example.com)" than "[`NbtCompound`](http://example.com)".
2. You don't have to add create extraneous links when the type referred to is part of the descriptor, such as param or return types. Javadoc links that for you
2. The create new behavior IMO is quite confusing, as I have fallen into the trap before (since modification on existing tag objects are saved, you would tend to think so for new tag objects)

LambdAurora: Uh, the modification part is quite weird.
Especially since the method code is:
```java
      if (this.nbt != null && this.nbt.contains(key, NbtElement.COMPOUND_TYPE)) {
         return this.nbt.getCompound(key);
      } else {
         var nbtCompound = new NbtCompound();
         this.setNbt(key, nbtCompound);
         return nbtCompound;
      }
```
So from that, it should be saved for new NBT compound objects.

liach: It does call `setNbt` now? Thanks mojang! then might note to users that this modifies the nbt and for pure reading purposes, use the nullable getter one.

Technici4n: subTag is much clearer than just "nbt" which can now mean the serialized stack, the top level tag, or a subtag...

Technici4n:
Tbh I don't get the practical problem this PR solves? There is no conflict with the new names for the types...

mounderfod:
I feel that mid-1.17 is a bad time to refactor, it would be better if this had happened during the 1.17 snapshots but oh well, it may be worth waiting for 1.18 snapshots which are likely to be soon?

mounderfod:
> Tbh I don't get the practical problem this PR solves? There is no conflict with the new names for the types...

Consistency I guess? It doesn't seem worth the hassle for devs since the conflicts are dealt with.

liach (changes requested):
tag vs nbt aside, this one definitely need to fix inconsistency between `Nbt` `NbtElement` `SubNbt` discrepancies that the original name all conform to `SubTag`.

mappings/net/minecraft/item/ItemStack.mapping
```diff
@@ -139,20 +148,33 @@ CLASS net/minecraft/class_1799 net/minecraft/item/ItemStack
...
 	METHOD method_7938 hasCustomName ()Z
 	METHOD method_7939 setCount (I)V
+		COMMENT Sets the count of items in this item stack.
 		ARG 1 count
+			COMMENT the count of items
 	METHOD method_7940 canDestroy (Lnet/minecraft/class_5415;Lnet/minecraft/class_2694;)Z
 		ARG 1 tagManager
 		ARG 2 pos
-	METHOD method_7941 getSubTag (Ljava/lang/String;)Lnet/minecraft/class_2487;
+	METHOD method_7941 getNbt (Ljava/lang/String;)Lnet/minecraft/class_2487;
```
mounderfod: It seems weird to repeat these for `Tag` and `SubTag`. Why not `SubNbt` or similar?
This happens a bunch throughout the class?

mappings/net/minecraft/item/ItemStack.mapping
```diff
@@ -194,10 +228,18 @@ CLASS net/minecraft/class_1799 net/minecraft/item/ItemStack
 		ARG 3 breakCallback
 	METHOD method_7957 updateEmptyState ()V
 	METHOD method_7958 hasGlint ()Z
-	METHOD method_7959 putSubTag (Ljava/lang/String;Lnet/minecraft/class_2520;)V
+	METHOD method_7959 putNbtElement (Ljava/lang/String;Lnet/minecraft/class_2520;)V
```
mounderfod: (this has already been mentioned) `Element` hasn't been used anywhere else - why here?

mounderfod (commented):
A few quirks - I could be misunderstanding the code but it would be wise to avoid ambiguity in a PR which is designed to resolve just that.

mappings/net/minecraft/item/ItemStack.mapping
LambdAurora: Because the argument taken here is a NbtElement, while most of the other methods take a NbtCompound in.

mounderfod: Hmm, OK. If there's a technical reason for it being different, then I'm happy to approve it if others do.

LambdAurora:
> I feel that mid-1.17 is a bad time to refactor, it would be better if this had happened during the 1.17 snapshots but oh well, it may be worth waiting for 1.18 snapshots which are likely to be soon?

Uh... Where were you when #2427 was submitted then?

It's not uncommon for those to happen, and Fabric has tools to ease those renames for modders, especially with the `migrateMappings` task in Fabric Loom.

Even though this PR still has some consistency issues that I want to fix (haven't done it yet), but once those are fixed, it'll be way more consistent with the current state of NBT in Yarn.

mappings/net/minecraft/item/ItemStack.mapping
LambdAurora: When I worked on that PR, I was told that the `Sub` part isn't really necessary given the key parameter.
Javadocs are present to reduce the confusion.

FoundationGames:
> I'd leave the item stack extra data compound as "tag", as that's its specific name also in the serialised form (whether in JSON or in NBT).

@Juuxel 
World is referenced many times in the code as "level" yet we don't call it that, as it's not a very good name 
As such the whole point of this PR is to remove the ambiguity between item NBT and item tags, and not having those methods use the name "tag" (despite them returning NbtElements) accomplishes this, no matter whether serialized item stacks use "tag" (which is bad on Mojang's part since it contributes to that same ambiguity)

mappings/net/minecraft/item/ItemStack.mapping
liach: and now you have 3 types of names for methods working with sub elements

liach:
@modmuss50 What is your stance on this?

liach:
Instead of adding five `@see`s for each nbt method's javadoc, I suggest making a table on nbt operations in the class header doc (like under an `<h2 id="nbt-operations">` section) and replace those `@see` to individual methods with a `@see <a href="nbt-operations">Item Stack NBT Operations</a>`. This would allow us to better describe that ones with string parameters are working with sub tags.

Technici4n:
Why not use `subNbt`? `nbt` is not much shorter, but it's a lot less clear. Javadoc is no excuse for ambiguous names.

LambdAurora:
> Why not use `subNbt`? `nbt` is not much shorter, but it's a lot less clear. Javadoc is no excuse for ambiguous names.

@Technici4n switched to sub NBT for the related methods in the latest commit.

> Instead of adding five `@sees` for each nbt method's javadoc, I suggest making a table on nbt operations in the class header doc (like under an `<h2 id="nbt-operations">` section) and replace those `@see` to individual methods with a `@see <a href="nbt-operations">Item Stack NBT Operations</a>`. This would allow us to better describe that ones with string parameters are working with sub tags.

@liach applied your suggestion, it does look way better.

Here's a rendered view of the NBT operations documentation:
image

mappings/net/minecraft/item/ItemStack.mapping
LambdAurora: Mojang doesn't seem to have a constant for the `tag` key in the serialized NBT.

Technici4n:
Alright, I think I can live with this change... Will have to update `ResourceKey` accordingly! :smile:

liach (approved):
Looks good, at least this has consistency (as opposed to the irregular usage of sub/element etc before)

Technici4n (approved):

liach:
Since this is sent around a time when names are mostly stable, the changes probably will be accepted later if it is ever accepted.

Technici4n:
> Since this is sent around a time when names are mostly stable, the changes probably will be accepted later if it is ever accepted.

I have only one thing to say about this: `migrateMappings` exists. 😄

LambdAurora:
I find that care of stability in names quite sudden and new. There's a lot of PRs that was similarly sized, or even bigger, that got merged in a release cycle. I don't see why it should be different here.

liach:
Oh, I am just relaying modmuss' points here. Also @Technici4n I think this one is still somewhat controversial, mind ask people on discord to come here and vote on this?

dexman545 (approved):
More documentation is always a nice bonus.

YanisBft (approved):

lenrik1589 (approved):
This PR makes sense together with #2090 of which i do not approve   (as it just doesn't make sense to me as rename aiming to rid package of the "Tag" part while changing it to "NamedBinaryTag", that rename doesn't sound as good as original either) bu the comments seem fine

modmuss50 (approved):
I approve of the changes, howerver im a bit unsure about merging this in now as it will affect quite a few people.

The 1.16 backport was fine as its at the end of stable release where as with this people will be forced to adopt these changes when they go to 1.17.1 shortly.

The issue is none of us know how long its going to be until we get a nice chance to merge this in (1.18 snapshots) so it could be left sitting here for a while.

This higlights the need for cross version mappings, but thats clearly way out of scope to do here and now.

LambdAurora:
> im a bit unsure about merging this in now as it will affect quite a few people.
> 
> The 1.16 backport was fine as its at the end of stable release where as with this people will be forced to adopt these changes when they go to 1.17.1 shortly.
> 
> The issue is none of us know how long its going to be until we get a nice chance to merge this in (1.18 snapshots) so it could be left sitting here for a while.
> 
> This higlights the need for cross version mappings, but thats clearly way out of scope to do here and now.

The issue with letting that PR sit is you could end up with a forgotten PR. And I don't really make yarn PRs to let them wait a long time while they are trying to improve readability, reduce ambiguity, and increase consistency.

I mean, if people work on 1.18 stuff they'll be forced to adopt these changes too? So what's the issue?

It's not the first time there's refactors in a release cycle. If you tell me that we should consider mappings "frozen" in a release cycle then why even bother?

As far as I know the `migrateMappings` task in loom was specifically made for those cases, why not take advantage of having this tool?

I honestly don't understand why this suddenly becomes an issue while for the past years there was no issues with such kind of PRs.

modmuss50:
Generally we have merged refactors during snapshots, I'm sure someone could spend 20 mins to find a PR that wasnt merged during a snapshot but that would be an exeption to the unwritten rule. I think a good compromise will be to wait untill after 1.17.1 is releases allowing people to go to that without this change if they so wish.

A lot of people cannot or choose not to use migrate mappings for a number of reasons especially with its current state with java 16 code.

We just need to decide what people want. Going forward it might be best to write down some clear giudelines so this dilemma doesnt happen again in the future.

Juuxel (approved):
This is fine now that the docs specify the difference between the Yarn name and the serialised form.

As for `migrateMappings`, I'd note that it also doesn't work for anyone not using Java. For example, Kotlin mods like Adorn are really annoying to migrate by hand so I'd prefer big refactors being done during major version updates.

LambdAurora:
> I'd prefer big refactors being done during major version updates.

The thing I don't understand about this is big refactors from yarn POV and Minecraft's codebase POV are very likely to happen during those, and delaying stuff to those is just piling more changes which easily overwhelm modders.

Personally I prefer to be able to do step by step (hence why I usually follow snapshots) than all at once and getting demotivated because I end up lost.

modmuss50 (approved):
Right, if no one has any objections lets get this in now otherwise we will never. Last call otherwise this goes in in 24 hours.

In the future please try to keep refactors to snapshots 👍

LambdAurora:
Any update? It's been more than 24 hours. :P

Player3324 (approved):

modmuss50:
No one has comaplined, there has been ample amount of time to do so, so merging it. 👍

PR #2513, opened 2021-06-22
Revamp control to be more accurate

liach:
Fixes stepAngleTowards to be clampAngle, 3 mob entity methods. These should be the main one impacting modders, but the old names were exceedingly incorrect and hence I decided to fix them.

I initially wanted to just map the shulker look control class; turns out the whole look control system has some wrong names.


liach:
anyone mind review this?

modmuss50 (approved):

mappings/net/minecraft/entity/mob/MobEntity.mapping
```diff
@@ -48,7 +48,12 @@ CLASS net/minecraft/class_1308 net/minecraft/entity/mob/MobEntity
 		ARG 1 id
 	METHOD method_19540 setAttacking (Z)V
 		ARG 1 attacking
-	METHOD method_20240 getLookYawSpeed ()I
+	METHOD method_20240 getMaxLookYawChange ()I
+		COMMENT {@return the maximum degrees which the yaw can change when looking}
+		COMMENT
+		COMMENT <p>This is used by the look control.
+		COMMENT
+		COMMENT <p>The default return value is {@code 40}.
```
enbrain:
```suggestion
	METHOD method_20240 getMaxLookYawChange ()I
		COMMENT {@return the maximum degrees which the yaw can change when looking}
		COMMENT
		COMMENT <p>This is used by the look control.
		COMMENT
		COMMENT <p>The default return value is {@code 10}.
```

liach: Looks right, must have been bad copypasta

enbrain (approved):

PR #2517, opened 2021-06-24
Inline Java compatibility declaration

Ab20zipYT:


liach:
Please do not spam projects with such meaningless pull requests. The merge of these 2 into one line isn't that clear, and since java 9, the preferred way to declare a version to follow is to use `-version` flag for all java compile tasks.

This however can serve as a note that we can bump the java version for constant documentation to 16.

PR #2521, opened 2021-06-25
Document EntityVelocityUpdateS2CPacket and ExplosionS2CPacket

enbrain:


mappings/net/minecraft/network/packet/s2c/play/EntityVelocityUpdateS2CPacket.mapping
```diff
@@ -1,4 +1,9 @@
 CLASS net/minecraft/class_2743 net/minecraft/network/packet/s2c/play/EntityVelocityUpdateS2CPacket
+	COMMENT Sent when a server modifies an entity's velocity.
+	COMMENT
+	COMMENT <p>If the entity is a player, {@link
+	COMMENT net.minecraft.network.packet.s2c.play.ExplosionS2CPacket} can be used as
```
liach, line 5: Is the substitution 100% safe without any other side effects? Sounds quite weird if it is

liach (approved):

mappings/net/minecraft/network/packet/s2c/play/EntityVelocityUpdateS2CPacket.mapping
enbrain: It's safe from my understanding. 

When the client receives the explosion packet, it does:
1. Play sound at the given position
2. Add particle at the given position
3. Replace blocks at the given positions (`affectedBlocks`) with air
4. Update the player's velocity

(1.) and (2.) can be nullified with the position which is too far away to hear or render.
(3.) can be nullified with the empty list for `affectedBlocks`.

Also, some pvp servers do use explosion packet for knockback for anti-cheat purpose and don't seem to have issues.

YanisBft (approved):

PR #2522, opened 2021-06-26
Some NativeImage methods

DreenDex:


liach (commented):
I suggest just `getRed` `getGreen` `getBlue`, as the luminance values are effectively red/green/blue if the image is luminance or luminance-alpha.

liach (approved):

YanisBft (approved):

PR #2526, opened 2021-06-27
MinecraftClient.openScreen -> setScreen

### LINKED ISSUE #1597: MinecraftClient.openScreen -> setScreen

Juuxel:
Closes #1597. The issue has the main reasoning, but I'd also mention that an error string within the method calls it `setScreen`.

This is also a semi-impactful rename, though less than #2511 so I don't think it needs to be delayed to after a release.

Also fixes a tiny javadoc error in `Item.Settings`.


mappings/net/minecraft/client/MinecraftClient.mapping
```diff
@@ -200,12 +200,15 @@ CLASS net/minecraft/class_310 net/minecraft/client/MinecraftClient
...
-	METHOD method_1507 openScreen (Lnet/minecraft/class_437;)V
-		COMMENT Opens a new screen, changing the current screen if needed.
+	METHOD method_1507 setScreen (Lnet/minecraft/class_437;)V
+		COMMENT Sets the current screen to a new screen.
 		COMMENT
 		COMMENT <p>If the screen being opened is {@code null} and the client is not in game, the title screen will be opened.
-		COMMENT If the currently opened screen is {@code null} and player is dead then the death screen will be opened.
-		COMMENT Otherwise the currently open screen will be closed.
+		COMMENT If the currently opened screen is {@code null} and player is dead, then the death screen will be opened.
+		COMMENT Otherwise, the currently open screen will be closed.
```
liach: I suggest replace `currently` with `previous` to make it more clear (not confuse current with the passed parameter)

mappings/net/minecraft/client/MinecraftClient.mapping
```diff
@@ -204,8 +204,8 @@ CLASS net/minecraft/class_310 net/minecraft/client/MinecraftClient
 		COMMENT Sets the current screen to a new screen.
 		COMMENT
 		COMMENT <p>If the screen being opened is {@code null} and the client is not in game, the title screen will be opened.
-		COMMENT If the currently opened screen is {@code null} and player is dead, then the death screen will be opened.
-		COMMENT Otherwise, the currently open screen will be closed.
+		COMMENT If the previous screen is {@code null} and the player is dead, then the death screen will be opened.
```
liach: So the death screen would override the provided parameter screen when the player is dead, I assume?

mappings/net/minecraft/client/MinecraftClient.mapping
```diff
@@ -200,12 +200,15 @@ CLASS net/minecraft/class_310 net/minecraft/client/MinecraftClient
...
+	METHOD method_1507 setScreen (Lnet/minecraft/class_437;)V
+		COMMENT Sets the current screen to a new screen.
 		COMMENT
 		COMMENT <p>If the screen being opened is {@code null} and the client is not in game, the title screen will be opened.
-		COMMENT If the currently opened screen is {@code null} and player is dead then the death screen will be opened.
-		COMMENT Otherwise the currently open screen will be closed.
+		COMMENT If the previous screen is {@code null} and the player is dead, then the death screen will be opened.
+		COMMENT Otherwise, the previous screen will be closed.
+		COMMENT
+		COMMENT <p>If there is an open screen when the current screen is changed, {@link Screen#removed()} will
```
liach: Consider merging this and the previous sentence, both of which talk about screen removal.

mappings/net/minecraft/client/MinecraftClient.mapping
Juuxel: No, see the line above - this only applies when the new screen is null.

liach: Hmm, so a table format would be better:
```
<p>If the new screen is {@code null}:
<ul>
<li>if the client is not in game, the title screen will be opened;</li>
<li>if the previous screen is {@code null} and the player is dead, then the death screen will be opened.</li>
</ul>
```

Juuxel:
@liach I changed the docs a bit again, and also fixed a factual error about the death screen.

Rendered:
rendered JD

(The double `@Nullable` happens elsewhere too, not sure what causes it)

liach (approved):

liach:
> (The double @Nullable happens elsewhere too, not sure what causes it)

Fyi this is because the jetbrains nullable is both type-use and parameter capable. Hence, when mappingpoet generates `@Nullable Type value`, javadoc would interpret that nullable to be used for both cases, and hence it's rendered twice.

FoundationGames (approved):

enbrain (approved):

modmuss50 (approved):
I like this change, as this is qutie impactful I think its best we hold off on merging it untill we are also happy to merge: https://github.com/FabricMC/yarn/pull/2511

YanisBft (approved):

PR #2535, opened 2021-06-29
1.17.1-pre2/pre3 stuff

apple502j:
Methods related to villager dropping items, some map changes, and validation of integers. Still not quite sure with the naming `setPreviousTrackedSlotMutable`; it's basically `setPreviousTrackedSlot` without `copy()`ing the `stack`.

Update: now has 1.17.1-pre3 mappings, including `placeAt` argument number fix, `TallPlantBlock.withWaterloggedState` (which returns BlockState with waterlogged state if `pos` has water), a util for macOS window handling, `ATTACKABLE_TARGET_PREDICATE_IGNORE_VISIBILITY_OR_DISTANCE_SCALING` and `testAttackableTargetPredicateIgnoreVisibility`, (suggest shorter names please), etc.

Also maps the oddest method of all: `method_37459` (`isOdd`).

### DIFF
--- a/mappings/net/minecraft/screen/ScreenHandler.mapping
+++ b/mappings/net/minecraft/screen/ScreenHandler.mapping
@@ -95,6 +95,9 @@ CLASS net/minecraft/class_1703 net/minecraft/screen/ScreenHandler
 	METHOD method_37420 updateToClient ()V
 	METHOD method_37421 getRevision ()I
 	METHOD method_37422 nextRevision ()I
+	METHOD method_37449 setPreviousTrackedSlotMutable (ILnet/minecraft/class_1799;)V
+		ARG 1 slot
+		ARG 2 stack
 	METHOD method_7591 packQuickCraftData (II)I
 		ARG 0 quickCraftStage
 		ARG 1 buttonId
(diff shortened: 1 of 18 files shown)


mappings/net/minecraft/screen/ScreenHandler.mapping
```diff
@@ -95,6 +95,9 @@ CLASS net/minecraft/class_1703 net/minecraft/screen/ScreenHandler
 	METHOD method_37420 updateToClient ()V
 	METHOD method_37421 getRevision ()I
 	METHOD method_37422 nextRevision ()I
+	METHOD method_37449 setPreviousTrackedSlotMutable (ILnet/minecraft/class_1799;)V
```
liach, line 98: why mutable suffix?

apple502j: Conflicts with existing `setPreviousTrackedSlot`. (This method does not copy the stack, the existing one does.)

liach: Hmm, I don't see why they create a different setter and how this really affects anything beyond reducing object allocation. This name should be good.

liach (approved):

liach (approved):
good up to this point, hope you don't add more since some you mapped were old contents in fact

YanisBft (approved):

PR #2545, opened 2021-07-04
Rename BakedModel#getSprite to more intuitive name

Ueaj-Kerman:
getSprite is a somewhat confusing name, it's used for block break particles, some fluid stuff, and the "ItemBillboardParticle" (the particle effect used to show barrier blocks & now light blocks).

### DIFF
--- a/mappings/net/minecraft/client/render/model/BakedModel.mapping
+++ b/mappings/net/minecraft/client/render/model/BakedModel.mapping
@@ -12,7 +12,7 @@ CLASS net/minecraft/class_1087 net/minecraft/client/render/model/BakedModel
 	METHOD method_4708 useAmbientOcclusion ()Z
 	METHOD method_4709 getTransformation ()Lnet/minecraft/class_809;
 	METHOD method_4710 getOverrides ()Lnet/minecraft/class_806;
-	METHOD method_4711 getSprite ()Lnet/minecraft/class_1058;
+	METHOD method_4711 getModelIcon ()Lnet/minecraft/class_1058;
 		COMMENT {@return a texture that represents the model}
 		COMMENT <p>
 		COMMENT For example, this is used for the block break particles.
--- a/mappings/net/minecraft/client/render/model/BakedModel.mapping
+++ b/mappings/net/minecraft/client/render/model/BakedModel.mapping
@@ -12,9 +12,9 @@ CLASS net/minecraft/class_1087 net/minecraft/client/render/model/BakedModel
 	METHOD method_4708 useAmbientOcclusion ()Z
 	METHOD method_4709 getTransformation ()Lnet/minecraft/class_809;
 	METHOD method_4710 getOverrides ()Lnet/minecraft/class_806;
-	METHOD method_4711 getModelIcon ()Lnet/minecraft/class_1058;
+	METHOD method_4711 getParticleSprite ()Lnet/minecraft/class_1058;
 		COMMENT {@return a texture that represents the model}
 		COMMENT <p>
-		COMMENT For example, this is used for the block break particles.
+		COMMENT This is primarily used in particles. For example, block break particles use this sprite.
 	METHOD method_4712 hasDepth ()Z
 	METHOD method_4713 isBuiltin ()Z
--- a/mappings/net/minecraft/client/render/model/BakedModel.mapping
+++ b/mappings/net/minecraft/client/render/model/BakedModel.mapping
@@ -13,7 +13,7 @@ CLASS net/minecraft/class_1087 net/minecraft/client/render/model/BakedModel
 	METHOD method_4709 getTransformation ()Lnet/minecraft/class_809;
 	METHOD method_4710 getOverrides ()Lnet/minecraft/class_806;
 	METHOD method_4711 getSprite ()Lnet/minecraft/class_1058;
-		COMMENT A texture that represents the model.
+		COMMENT {@return a texture that represents the model}
 		COMMENT <p>
 		COMMENT For example, this is used for the block break particles.
 	METHOD method_4712 hasDepth ()Z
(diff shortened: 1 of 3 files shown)


mappings/net/minecraft/client/render/model/BakedModel.mapping
```diff
@@ -13,5 +13,8 @@ CLASS net/minecraft/class_1087 net/minecraft/client/render/model/BakedModel
 	METHOD method_4709 getTransformation ()Lnet/minecraft/class_809;
 	METHOD method_4710 getOverrides ()Lnet/minecraft/class_806;
 	METHOD method_4711 getSprite ()Lnet/minecraft/class_1058;
```
Ueaj-Kerman:
```suggestion
	METHOD method_4711 getModelIcon ()Lnet/minecraft/class_1058;
```
perhaps?

liach: `MinecraftClient.checkGameData` for context

However, looking at `InGameOverlayRenderer#renderOverlays` and `InGameHud#renderPortalOverlay`, the sprite is used in full than just for an icon/particle? Also fluid renderer seems to use the sprite wholesale too

mappings/net/minecraft/client/render/model/BakedModel.mapping
```diff
@@ -13,5 +13,8 @@ CLASS net/minecraft/class_1087 net/minecraft/client/render/model/BakedModel
...
+		COMMENT A texture that represents the model.
```
liach:
```suggestion
		COMMENT {@return a texture that represents the model}
```

mappings/net/minecraft/client/render/model/BakedModel.mapping
Ueaj-Kerman: It's like a "representation" of the model, the most important texture, icon just felt most fitting.

liach: Good idea! But if you do rename, remember to update the associated `getSprite` methods in `BlockModels` and `ItemModels` to `getIcon` as well (I think `getIcon` is enough, no need for `Model` in name)

liach:
Hmm, nowadays can blocks/items have other icons besides that from the models? If not, then `getIcon` is enough

Ueaj-Kerman:
the other methods in the class have `Model` in the name, like `getModel`

mappings/net/minecraft/client/render/model/BakedModel.mapping
```diff
@@ -12,6 +12,9 @@ CLASS net/minecraft/class_1087 net/minecraft/client/render/model/BakedModel
 	METHOD method_4708 useAmbientOcclusion ()Z
 	METHOD method_4709 getTransformation ()Lnet/minecraft/class_809;
 	METHOD method_4710 getOverrides ()Lnet/minecraft/class_806;
-	METHOD method_4711 getSprite ()Lnet/minecraft/class_1058;
+	METHOD method_4711 getModelIcon ()Lnet/minecraft/class_1058;
```
liach: > the other methods in the class have `Model` in the name, like `getModel`

but not for this one

```suggestion
	METHOD method_4711 getIcon ()Lnet/minecraft/class_1058;
```

Ueaj-Kerman: the `Model` here is here for a different reason, it's there to reduce ambiguity. Just getIcon by itself is too terse imo

liach: Sure, we can let people comment on this suggestion later

liach (approved):

Technici4n (approved):
Good idea, this always confused me, until it crashed cause I returned `null`. :smile:

Player3324:
IMO Icon is a bad name for these. It is meant to be a representative texture, not an abstract/descriptive icon.

liach:
This does act more like an avatar/profile picture. Icon somewhat implies the image is "small", which is not quite true here

Player3324:
"avatar/profile picture" is wrong there imo, it is usually meant to blend in with the model as a whole. Like you wouldn't use a fancy rendering of a grass block like a product image, but ideally some texture that mirrors the micro structure of the model including the colors used by it. The fragments of the particle texture are supposed to look like bits of the block.

liach:
Then "icon" is better as icon does mean a representative. If you can suggest a better name, go for it; if not, I think we can just add more documentation on how this icon isn't zoomed out/small

Player3324:
No an icon is more descriptive than representative, like a program icon usually doesn't look like the application at all. This makes it a bad term IMO, especially compared to "(representative) sprite" which is just what we call the sub textures used for blocks.

Technici4n:
`getRepresentativeSprite()` would be fine I guess.

liach:
Hmm, how about `asSprite` or `asIcon`? these name show that the sprite/icon isn't part of the model, but more like an alternative form of the model.

Player3324:
It is still a property of the model the way it is structured, even if it is an alternative form as well in some sense -> get is more appropriate. It basically has the quads and the sprite form. 

The model json calls the entry "particle", so getParticleSprite may be another obvious choice.

Ueaj-Kerman:
I was originally going to use particle, it works, but it seems to be and can be used for more than that. Icon still feels right in this case, Representative is a little long but it works too, Delegate is another option, or Summary?

Player3324:
By representative I meant more or less "lookalike" (as the actual model), which is quite different from icon/summary etc. that imply a more abstract image

YanisBft (approved):

Ueaj-Kerman:
changed it to `getParticleSprite`

Juuxel:
@Devan-Kerman Could you fix the merge conflicts?

Ueaj-Kerman:
yea, done

modmuss50 (approved):

Issue #2548, opened 2021-07-04
Recipe cleanups


liach:
When looking at #2544, I find quite a few more problem with recipes in general:
1. The distinction of "recipe" and "crafting". Crafting, in vanilla, strictly refers to the recipes in the crafting table, while we have mixed names, such as `Recipe.craft`.
2. In addition, we assume "recipe" can cover "crafting", as seen in `ShapedRecipe`, whose minecraft ID is `crafting_shaped`, and should be named `ShapedCraftingRecipe`.
3. `CuttingRecipe` has nothing related to cutting; it is more or so a recipe that takes a single input ingredient, and can be of any recipe type once it implements the proper marker interface.
4. There are a few more legacy names in the recipe package, such as inconsistent getters, `deserialize` in `RecipeManager`.

I plan to delay these changes into 1.18, as 1.17 is out here and mojang does plan to release 1.18 quite soon. The documentation and less impactful renames (params, private members) may go into 1.17 though.

PR #2549, opened 2021-07-04
Recipe docs

### LINKED ISSUE #2544: SpecialRecipeSerializer.id -> factory
It's a `Function<Identifier, T>` (where `T` is the handled recipe type). Might be an incorrect match?

liach:
Document recipes a bit, and clean up outdated docs. I think a few renames are necessary for later; see #2548.

There is also renames to fix #2544 and a few other less impactful ones (removing `Client` suffix after the getter is now universal)

### DIFF
--- a/mappings/net/minecraft/recipe/RecipeManager.mapping
+++ b/mappings/net/minecraft/recipe/RecipeManager.mapping
@@ -1,42 +1,126 @@
 CLASS net/minecraft/class_1863 net/minecraft/recipe/RecipeManager
+	COMMENT A recipe manager allows easier use of recipes, such as finding matches and
+	COMMENT remainders. It is also integrated with a recipe loader, which loads recipes
+	COMMENT from data packs' JSON files.
 	FIELD field_19359 GSON Lcom/google/gson/Gson;
 	FIELD field_9023 recipes Ljava/util/Map;
 	FIELD field_9024 errored Z
+		COMMENT This isn't quite indicating an errored state; its value is only set to
+		COMMENT {@code false} and is never {@code true}, and isn't used anywhere.
 	FIELD field_9027 LOGGER Lorg/apache/logging/log4j/Logger;
 	METHOD method_17717 getAllOfType (Lnet/minecraft/class_3956;)Ljava/util/Map;
 		ARG 1 type
 	METHOD method_17718 (Lnet/minecraft/class_3956;Lnet/minecraft/class_1937;Lnet/minecraft/class_1263;Lnet/minecraft/class_1860;)Ljava/util/stream/Stream;
 		ARG 3 recipe
 	METHOD method_17720 deserialize (Lnet/minecraft/class_2960;Lcom/google/gson/JsonObject;)Lnet/minecraft/class_1860;
+		COMMENT Reads a recipe from a JSON object.
+		COMMENT
+		COMMENT @implNote Even though a recipe's {@linkplain Recipe#getSerializer() serializer}
+		COMMENT is stored in a {@code type} field in the JSON format and referred so in this
+		COMMENT method, its registry has key {@code minecraft:root/minecraft:recipe_serializer}
+		COMMENT and is thus named.
+		COMMENT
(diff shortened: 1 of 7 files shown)


mappings/net/minecraft/recipe/SpecialRecipeSerializer.mapping
```diff
@@ -1,4 +1,10 @@
 CLASS net/minecraft/class_1866 net/minecraft/recipe/SpecialRecipeSerializer
-	FIELD field_9046 id Ljava/util/function/Function;
+	COMMENT A serializer for hardcoded recipes. The recipes with this serializer don't
+	COMMENT transport any extra data besides their ID when read from JSON or synchronized
+	COMMENT over network.
+	COMMENT
+	COMMENT <p>The name "special" comes from the fact that in vanilla, recipes using this
+	COMMENT serializer have IDs starting with {@code crafting_special_}.
```
Juuxel:
```suggestion
	COMMENT <p>The name "special" comes from the fact that in vanilla, recipes using this
	COMMENT serializer have IDs starting with {@code crafting_special_}. All of their logic and ingredients
	COMMENT are also defined in code, which distinguishes them from "non-special" recipes.
```

mappings/net/minecraft/recipe/RecipeSerializer.mapping
```diff
@@ -1,4 +1,15 @@
 CLASS net/minecraft/class_1865 net/minecraft/recipe/RecipeSerializer
+	COMMENT The recipe serializer controls the serialization and deserialization of
+	COMMENT recipe content. The us the serialization of recipes' IDs.
```
YanisBft: Missing words

mappings/net/minecraft/recipe/RecipeSerializer.mapping
```diff
@@ -22,11 +33,37 @@ CLASS net/minecraft/class_1865 net/minecraft/recipe/RecipeSerializer
 		ARG 0 id
 		ARG 1 serializer
 	METHOD method_8121 read (Lnet/minecraft/class_2960;Lcom/google/gson/JsonObject;)Lnet/minecraft/class_1860;
+		COMMENT Reads a recipe from a JSON object.
+		COMMENT
+		COMMENT @implNote If this throws any excpetion besides {@link com.google.gson.JsonParseException}
```
YanisBft:
```suggestion
		COMMENT @implNote If this throws any exception besides {@link com.google.gson.JsonParseException}
```

YanisBft (approved):

mappings/net/minecraft/recipe/RecipeSerializer.mapping
liach:
```suggestion
	COMMENT recipe content. The caller should handle the serialization of recipes' IDs.
```

mappings/net/minecraft/recipe/RecipeManager.mapping
```diff
@@ -1,42 +1,126 @@
 CLASS net/minecraft/class_1863 net/minecraft/recipe/RecipeManager
+	COMMENT A recipe manager allows easier use of recipes, such as finding matches and
+	COMMENT remainders. It is also integrated with a recipe loader, which loads recipes
+	COMMENT from data packs' JSON files.
 	FIELD field_19359 GSON Lcom/google/gson/Gson;
 	FIELD field_9023 recipes Ljava/util/Map;
 	FIELD field_9024 errored Z
+		COMMENT This isn't quite indicating an errored state; its value is only set to
+		COMMENT {@code false} and is never {@code true}, and isn't used anywhere.
```
Juuxel, line 9: Probably an old field that lost its uses along the way

Juuxel (approved):
Great work!

mappings/net/minecraft/recipe/RecipeManager.mapping
liach: Very likely, but the current use imo is best dissociated from its past use, as otherwise it may be bug prone

PR #2559, opened 2021-07-05
Misc structure gen mappings

frqnny:
Some mappings around Structure Gen and Jigsaw Structure.

The one that took me the most time to figure out was modifyBoundingBox. What it seems to do is shift the bounding box  of structure pool elements. I don't get why, but I do want to point out that it may affect jigsaw connecting. If it's too vague for triage's liking I can just take it out. 

Also notable that StructurePoolBasedGenerator#generate basically iterates through and calls generatePiece, so I'm welcome to changing it to something like generatePieces. 

the field JIGSAW_STRUCTURES was inaccurate- it also included Strongholds which are not jigsaws. This field is for those that modify the land like villages. Maybe it could be LAND_ADJUSTING_STRUCTURES if you guys think that'd fit better.

### DIFF
--- a/mappings/net/minecraft/structure/pool/StructurePoolBasedGenerator.mapping
+++ b/mappings/net/minecraft/structure/pool/StructurePoolBasedGenerator.mapping
@@ -1,5 +1,14 @@
 CLASS net/minecraft/class_3778 net/minecraft/structure/pool/StructurePoolBasedGenerator
 	FIELD field_16665 LOGGER Lorg/apache/logging/log4j/Logger;
+	METHOD method_30419 generate (Lnet/minecraft/class_5455;Lnet/minecraft/class_3812;Lnet/minecraft/class_3778$class_3779;Lnet/minecraft/class_2794;Lnet/minecraft/class_3485;Lnet/minecraft/class_2338;Lnet/minecraft/class_6130;Ljava/util/Random;ZZLnet/minecraft/class_5539;)V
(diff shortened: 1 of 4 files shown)


mappings/net/minecraft/structure/pool/StructurePoolBasedGenerator.mapping
```diff
@@ -1,5 +1,14 @@
 CLASS net/minecraft/class_3778 net/minecraft/structure/pool/StructurePoolBasedGenerator
 	FIELD field_16665 LOGGER Lorg/apache/logging/log4j/Logger;
+	METHOD method_30419 generate (Lnet/minecraft/class_5455;Lnet/minecraft/class_3812;Lnet/minecraft/class_3778$class_3779;Lnet/minecraft/class_2794;Lnet/minecraft/class_3485;Lnet/minecraft/class_2338;Lnet/minecraft/class_6130;Ljava/util/Random;ZZLnet/minecraft/class_5539;)V
```
liach, line 3: Mind update `StructurePoolBasedFeatureConfig`'s javadoc's `method_30419` to `generate` instead? https://github.com/FabricMC/yarn/pull/2559/checks#step:6:3543

liach (approved):
:heart: many thanks

Juuxel (approved):

YanisBft (approved):

PR #2561, opened 2021-07-05
FollowTargetGoal -> ActiveTargetGoal

### LINKED ISSUE #2560: FollowTargetGoal is not accurate
**Class Description**
`FollowTargetGoal` is the standard tracking goal used for entity AI. It first locates a nearby entity matching a predicate (if any); if one is found, the goal assigns it as the target of the `FollowTargetGoal`'s owner.

**Issue**
My issue with this name is that the goal does not actually make the entity "follow" anything, and "follow" implies action. Because the goal is based around `findClosestTarget` and setting a `target` on the mob, I think `TrackClosestTargetGoal` might be more appropriate (or something to that degree). The parent class is already named `TrackTargetGoal`, so adding a "closest" (which is pretty much the only thing `FollowTargetGoal` does) makes sense.

**Other Thoughts**
I'm guessing the "Follow" part of the name was taken from the `getFollowRange` (`GENERIC_FOLLOW_RANGE` attribute) method called inside the goal, but that seems to do more with the size of the goal rather than what it actually does.

frqnny:
Closes #2560, see what draylar put there.

adding on, "nearest" wouldn't fit because it does call a bunch of methods that also use "closest"

### DIFF
--- a/mappings/net/minecraft/entity/ai/goal/EnableableActiveTargetGoal.mapping
+++ b/mappings/net/minecraft/entity/ai/goal/EnableableActiveTargetGoal.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_3760 net/minecraft/entity/ai/goal/DisableableFollowTargetGoal
+CLASS net/minecraft/class_3760 net/minecraft/entity/ai/goal/EnableableActiveTargetGoal
 	FIELD field_17281 enabled Z
 	METHOD <init> (Lnet/minecraft/class_3763;Ljava/lang/Class;IZZLjava/util/function/Predicate;)V
 		ARG 1 actor
--- a/mappings/net/minecraft/entity/ai/goal/TameableTargetGoal.mapping
+++ b/mappings/net/minecraft/entity/ai/goal/TameableTargetGoal.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_1404 net/minecraft/entity/ai/goal/FollowTargetIfTamedGoal
+CLASS net/minecraft/class_1404 net/minecraft/entity/ai/goal/TameableTargetGoal
 	FIELD field_6656 tameable Lnet/minecraft/class_1321;
 	METHOD <init> (Lnet/minecraft/class_1321;Ljava/lang/Class;ZLjava/util/function/Predicate;)V
 		ARG 1 tameable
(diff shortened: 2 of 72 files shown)


liach:
Imo nearest isn't the principal feature of this goal. The distinguishing feature is this is that the goal finds targets by an entity class.

liach (changes requested):
You didn't check the subclasses. Definitely need to rename `DisableableFollowTargetGoal` and `FollowTargetIfTamedGoal` if this is rename is done.

Now I think `ActiveTargetGoal` (find target actively) may be better, as this is what hostile mobs use to target players

frqnny:
I went ahead and renamed some sub classes. For entity-specific ones like Spider, I just wrote it "TargetGoal" since it would be "SpiderEntity.TargetGoal"

mappings/net/minecraft/entity/ai/goal/EnableableActiveTargetGoal.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_3760 net/minecraft/entity/ai/goal/DisableableFollowTargetGoal
+CLASS net/minecraft/class_3760 net/minecraft/entity/ai/goal/EnableableActiveTargetGoal
```
liach: How about `Toggleable` if "enableable" and "disableable" appear too cursed

mappings/net/minecraft/entity/ai/goal/TameableTargetGoal.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_1404 net/minecraft/entity/ai/goal/FollowTargetIfTamedGoal
+CLASS net/minecraft/class_1404 net/minecraft/entity/ai/goal/TameableTargetGoal
```
liach: image
Should be `UntamedActiveTargetGoal`

mappings/net/minecraft/entity/passive/PolarBearEntity.mapping
```diff
@@ -19,7 +19,7 @@ CLASS net/minecraft/class_1456 net/minecraft/entity/passive/PolarBearEntity
 	METHOD method_6602 playWarningSound ()V
 	METHOD method_6603 setWarning (Z)V
 		ARG 1 warning
-	CLASS class_1457 FollowPlayersGoal
+	CLASS class_1457 PlayerTargetGoal
```
liach: This one is like targeting to protect baby (when there is baby close around) while the other one is general "anger at" target goal

liach (approved):
Thank you for these names :heart:

liach:
@modmuss50 How feasible is this bug-fixing one? Though this may affect mods using the old ai system (assuming they aren't using brains), these renames do accuracy describe the actual roles of these goals.

modmuss50:
I dont think these classes are used that widely, so changing them in 1.17 seems fine to me. Does anyone have any objections to this?

modmuss50 (approved):

YanisBft (approved):

PR #2563, opened 2021-07-05
Update toolchain and fix minor javadoc problems

### LINKED ISSUE #2518: Bump java version for constants (unpick) source set
Since minecraft itself is already targetting java 16, there is no need for us to stay on java 8 for either the package or constant source sets, whose gradle configuration was added at unpick addition before minecraft's java bump.

This is a reminder extracted from https://github.com/FabricMC/yarn/pull/2517#issuecomment-867545537.

liach:
Fixes #2518 and resolves gradle warnings, including "Execution optimizations have been disabled" and "Deprecated Gradle features were used in this build". Updated tiny remapper to get rid of the indy string concat spam as well.

Also simplifies the intermediary v2 extraction by merging it with download and the javadoc final mapping click to copy script copying.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -703,7 +703,6 @@ task sourcesJar(type: Jar, dependsOn: classes) {
 }
 
 // Only build jars for package infos if we need to actually expose stuff like annotation in the future.
-// Also need to downgrade java in that case!
 
 build.dependsOn constantsJar
 
@@ -812,15 +811,13 @@ task genFakeSource(type: JavaExec, dependsOn: ["mergeV2", "mapNamedJar"]) {
 }
 
 task decompileCFR(type: JavaExec, dependsOn: [mapNamedJar]) {
-	doFirst {
-		classpath = configurations.decompileClasspath
-	}
 	mainClass.set "org.benf.cfr.reader.Main"
 
 	args namedJar.getAbsolutePath(), "--outputdir", file("namedSrc").absolutePath
 
 	doFirst {
 		file("namedSrc").deleteDir()
+		classpath = configurations.decompileClasspath
 	}
 }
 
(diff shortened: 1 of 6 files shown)


YanisBft (approved):

liach:
Will mark this wip. Seems gradle now (both 7.0 and 7.1.1) has a resource leak that becomes obvious when the javadoc task is ran, so I will check what exactly is going on.

liach:
Apparently things are fine; it's just that my gradle.properties in my userprofile .gradle directory overrides these. It seems that there is a delay in javadoc task due to the gradle daemon hanging after the tasks are completed (including after the copy on click script is copied), but I am not very sure why that happens and it's not that critical.

build.gradle
```diff
@@ -655,7 +659,7 @@ task unpickIntermediaryJar(type: JavaExec, dependsOn: [mapIntermediaryJar, "cons
 	outputs.upToDateWhen { false }
 	group "unpick"
 
-	main "daomephsta.unpick.cli.Main"
+	mainClass.set "daomephsta.unpick.cli.Main"
```
modmuss50: mainClass = "daomephsta.unpick.cli.Main"

modmuss50 (changes requested):

build.gradle
liach: Hmm, I encountered errors when I used that regular format at first. Think the error must have been caused by something else, as gradle can report weird errors indeed (such as when I forgot to replace `finalizedBy` with `doLast` and gradle says "cannot convert a work result to a task")

liach:
On a side note, I found that the ram available to gradle doesn't quite relate to ram used by compile/javadoc, as gradle starts a new javadoc process. Hence I decided to give javadoc tool more ram and tune gradle ram back down to 1GB (java mission control shows gradle usually uses about 600-700 MB, and javadoc usually uses up to like 1.5GB)

PR #2564, opened 2021-07-05
Rename *Criterion.test methods to trigger

### LINKED ISSUE #2269: Criterion `trigger`/`test` inconsistency
`AbstractCriterion` has a method, `test(ServerPlayerEntity player, Predicate<T> tester)`.

This is protected, and is used by subclasses in methods which look like this (in `BeeNestDestroyedCriterion`):

```java
public void test(ServerPlayerEntity player, Block block, ItemStack stack, int beeCount) {
    this.test(player, (conditions) -> { return conditions.test(block, stack, beeCount); });
}
```

or (in `BrewedPotionCriterion`):

```java
public void trigger(ServerPlayerEntity player, Potion potion) {
   this.test(player, (conditions) -> { return conditions.matches(potion); });
}
```

all subclasses of `AbstractCriterion` have a method like this - some call it `trigger`, and some call it `test`. The name should be consistent, as all these methods do basically the same thing.

YanisBft:
As [liach said](https://github.com/FabricMC/yarn/issues/2269#issuecomment-813069618), these methods actually handle everything that should happen when a criterion is trigged. They do not return booleans, so it doesn't make sense to name them `test`.

Closes #2269

### DIFF
--- a/mappings/net/minecraft/advancement/criterion/CriterionProgress.mapping
+++ b/mappings/net/minecraft/advancement/criterion/CriterionProgress.mapping
@@ -5,7 +5,7 @@ CLASS net/minecraft/class_178 net/minecraft/advancement/criterion/CriterionProgr
 	METHOD method_784 isObtained ()Z
 	METHOD method_785 fromPacket (Lnet/minecraft/class_2540;)Lnet/minecraft/class_178;
 		ARG 0 buf
-	METHOD method_786 getObtainedDate ()Ljava/util/Date;
+	METHOD method_786 getObtained ()Ljava/util/Date;
 	METHOD method_787 toPacket (Lnet/minecraft/class_2540;)V
 		ARG 1 buf
 	METHOD method_788 obtainedAt (Ljava/lang/String;)Lnet/minecraft/class_178;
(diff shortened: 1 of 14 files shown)


liach (approved):

mappings/net/minecraft/advancement/criterion/CriterionProgress.mapping
```diff
@@ -5,7 +5,7 @@ CLASS net/minecraft/class_178 net/minecraft/advancement/criterion/CriterionProgr
 	METHOD method_784 isObtained ()Z
 	METHOD method_785 fromPacket (Lnet/minecraft/class_2540;)Lnet/minecraft/class_178;
 		ARG 0 buf
-	METHOD method_786 getObtainedDate ()Ljava/util/Date;
+	METHOD method_786 getObtained ()Ljava/util/Date;
```
Juuxel: Hmm, this might be confused with `isObtained` above. I'd rather rename the field to `obtainedDate`.

liach: I think the toString of this class was like `"obtained: " + (obtained == null ? "false" : obtained)` and the field hence named.

YanisBft: Yeah the toString shows that it's "obtained", but I agree with Juuz, obtainedDate is less confusing

Juuxel (approved):

PR #2568, opened 2021-07-07
Travel javadocs

frqnny:
When I was messing with this at first, I thought it was the X Y Z coordinates you wanted to travel to. next i thought it was velocity. until recently, i found they were actually speed. This javadoc intends to clarify this to unfortunate souls who also happen to see this and get confused as to why it's passing down a Vec3d.

Idk how to make javadocs so it may be some weird checkstyle or I may be doing something wrong.

### DIFF
--- a/mappings/net/minecraft/entity/LivingEntity.mapping
+++ b/mappings/net/minecraft/entity/LivingEntity.mapping
@@ -484,7 +484,7 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_6090 takeShieldHit (Lnet/minecraft/class_1309;)V
 		ARG 1 attacker
 	METHOD method_6091 travel (Lnet/minecraft/class_243;)V
-		COMMENT Allows you do velocity and speed calculations. This is useful for custom vehicle behavior, or speed custom entity movement.
+		COMMENT Allows you do velocity and speed calculations. This is useful for custom vehicle behavior, or custom entity movement. This is not to be confused with AI.
 		COMMENT For an example of custom vehicle behavior, @see HorseBaseEntity#travel
 		COMMENT For an example of flying entities, @see FlyingEntity#travel
 		ARG 1 movementInput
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -484,7 +484,11 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_6090 takeShieldHit (Lnet/minecraft/class_1309;)V
 		ARG 1 attacker
 	METHOD method_6091 travel (Lnet/minecraft/class_243;)V
+		COMMENT Allows you to do certain speed and velocity calculations. This is useful for custom vehicle behavior, or custom entity movement. This is not to be confused with AI.
+		COMMENT For an example of custom vehicle behavior, @see HorseBaseEntity#travel
+		COMMENT For an example of flying entities, @see FlyingEntity#travel
```
liach:
```suggestion
		COMMENT
		COMMENT <p>See vanilla examples of {@linkplain net.minecraft.entity.passive.HorseBaseEntity#travel
		COMMENT custom horse vehicle} and {@linkplain net.minecraft.entity.mob.FlyingEntity#travel
		COMMENT flying entities}.
```

liach (commented):
Should best mention that this is how server players move.

frqnny:
What do you mean? ServerPlayerEntity does not have a custom travel implementation.
 Are you referencing that it's a server-sided calculation?

liach:
> ServerPlayerEntity does not have a custom travel implementation.

I recall the serve play network handler receives client movement inputs and then call the `travel` method to apply these inputs. This feature IMO may be worth a mention.

frqnny:
I think you're wrong about this, since there is no mention of travel in the ServerPlayNetworkHandler

liach (approved):

liach:
> I think you're wrong about this, since there is no mention of travel in the ServerPlayNetworkHandler

Right, it calls `Entity.move` instead

YanisBft (approved):

PR #2578, opened 2021-07-13
Map Mojang's GLFW constants

IotaBread:
Map GLFW constants in Minecraft code. They are just a copy of the constants in `org/lwjgl/glfw/GLFW` (Mojang didn't even add every used one: `GLFW_KEY_WORLD_` 1 & 2 are used, but they have no Minecraft equivalent). Also map some lambda parameters

Should I replace the unpick definitions for key codes with these ones?


mappings/net/minecraft/client/util/InputUtil.mapping
```diff
@@ -55,5 +176,14 @@ CLASS net/minecraft/class_3675 net/minecraft/client/util/InputUtil
 			ARG 2 keyCode
 		METHOD method_1447 createFromCode (I)Lnet/minecraft/class_3675$class_306;
 			ARG 1 code
+		METHOD method_27447 (Ljava/lang/Integer;Ljava/lang/String;)Lnet/minecraft/class_2561;
+			ARG 0 buttonCode
+			ARG 1 translationKey
 		METHOD method_27448 (I)Lnet/minecraft/class_3675$class_306;
 			ARG 1 code
+		METHOD method_27449 (Ljava/lang/Integer;Ljava/lang/String;)Lnet/minecraft/class_2561;
+			ARG 0 scancode
```
YanisBft:
```suggestion
			ARG 0 scanCode
```

YanisBft (approved):

liach:
I think we can just stay with the current unpick configuration. You said that the Minecraft ones are incomplete.

liach (approved):

PR #2582, opened 2021-07-14
1.18-es1 JFR stuff

apple502j:
~~Still wip.~~ Should be ready for review.

Notes 1: JfrCommand is definitely available on both dedicated and integrated. Not sure about ResetChunksCommand, but is probably available on integrated too.

Notes 2: JFR packages use `/util/profiling/jfr` since there are already several deobfscuated classes related to JFR under this. Should we move them to be consistent with `/util/profiler` naming?

Notes 3: ResetChunksCommand has a strange code that assigns a variable but never uses. Unfortunately that's also the only place that constructs class_6469 (subclass of ProtoChunk). Naming ideas anyone?

### DIFF
--- a/mappings/net/minecraft/util/profiling/jfr/JfrProfiler.mapping
+++ b/mappings/net/minecraft/util/profiling/jfr/JfrProfiler.mapping
@@ -23,5 +23,5 @@ CLASS net/minecraft/class_6419 net/minecraft/util/profiling/jfr/JfrProfiler
 		METHOD recordingStateChanged (Ljdk/jfr/Recording;)V
 			ARG 1 recording
 	CLASS class_6420 InstanceType
-		METHOD method_37500 getInstanceType (Lnet/minecraft/server/MinecraftServer;)Lnet/minecraft/class_6419$class_6420;
+		METHOD method_37500 get (Lnet/minecraft/server/MinecraftServer;)Lnet/minecraft/class_6419$class_6420;
 			ARG 0 server
--- a/mappings/net/minecraft/util/profiling/jfr/JfrProfiler.mapping
+++ b/mappings/net/minecraft/util/profiling/jfr/JfrProfiler.mapping
@@ -1,0 +1,27 @@
+CLASS net/minecraft/class_6419 net/minecraft/util/profiling/jfr/JfrProfiler
+	FIELD field_34063 MINECRAFT Ljava/lang/String;
+	FIELD field_34064 WORLD_GENERATION Ljava/lang/String;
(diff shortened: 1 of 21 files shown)


mappings/net/minecraft/util/profiling/jfr/JfrProfiler.mapping
```diff
@@ -0,0 +1,27 @@
...
+	METHOD method_37495 start (Ljava/nio/file/Path;Lnet/minecraft/class_6419$class_6420;)Z
+		ARG 0 path
+		ARG 1 instanceType
+	METHOD method_37496 (Ljdk/jfr/Recording;)Z
+		ARG 0 recording
+	METHOD method_37498 isProfiling ()Z
+	CLASS 1
+		METHOD recordingStateChanged (Ljdk/jfr/Recording;)V
+			ARG 1 recording
+	CLASS class_6420 InstanceType
```
apple502j, line 25: fyi, InstanceType is `CLIENT`/`SERVER`. Its implementation is more like integrated/dedicated (this code will never run on MinecraftClient iirc) but `ServerType.CLIENT` is kinda weird.

modmuss50:
Just so you are aware, Im happy to merge mappings in for these versions. However its quite possible/likely that future 1.18 snapshots wont be matched from this version so these names might get lost, of course you can manually bring them over, or possibly write some stuff to bring them forward. But untill the real snapshots start I wouldnt be too bothered about spending hours mapping it out.

apple502j:
Well that's fine; I can probably try to bring them over - this is wip after all. I wish I had more RAM :P

modmuss50:
Sounds good, if possible ill try and save the intermediary names (and thus the mappings for this) but dont want to promise anything, its all a bit of an unknown 👍

mappings/net/minecraft/util/profiling/jfr/JfrProfiler.mapping
liach: `net.fabricmc.api.EnvType` noises :laughing:

liach:
Also afaik jfr command isn't registered to the dispatchers. For note 2, I don't know how the jfr works out yet and cannot yet determine the effects of package moves on it.

apple502j:
From what I know, there are marked as DontObfscuate because the JFR requires it to have certain property names and mojang probaby did it by marking the entire class. So I don't think moving between packages causes huge trouble. *[citation needed]*

modmuss50 (approved):
Looks good, 👍

PR #2588, opened 2021-07-17
Mapped class_6479

Mysterious-Dev:
This PR mapped class_6479 and method_37821.


liach (approved):
Sure, we can have this backbones for now... Though I hoped you could map the `execute` and parameters

modmuss50:
Looks like this whole class has been mapped by another PR. Thanks anyway 👍

PR #2591, opened 2021-07-18
Map stuff in LightmapTextureManager.

LambdAurora:
Especially the constants as I'm a bit annoyed to use the unmapped names when modding.

What would be cool is also to unpick those in the code base, but I have no idea how to write unpick definitions.

### DIFF
--- a/mappings/net/minecraft/client/render/LightmapTextureManager.mapping
+++ b/mappings/net/minecraft/client/render/LightmapTextureManager.mapping
@@ -1,5 +1,5 @@
 CLASS net/minecraft/class_765 net/minecraft/client/render/LightmapTextureManager
-	COMMENT Represents the lightmap texture manager.
+	COMMENT The lightmap texture manager maintains a texture containing the RGBA overlay for each of the 16&times;16 sky and block light combinations.
 	COMMENT <p>
 	COMMENT Also contains some utilities to pack and unpack lightmap coordinates from sky and block light values,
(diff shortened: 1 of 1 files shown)


haykam821 (approved):

mappings/net/minecraft/client/render/LightmapTextureManager.mapping
```diff
@@ -1,4 +1,18 @@
 CLASS net/minecraft/class_765 net/minecraft/client/render/LightmapTextureManager
+	COMMENT Represents the lightmap texture manager.
```
liach:
```suggestion
	COMMENT The lightmap texture manager maintains a texture containing the RGBA overlay for each of the 16&times;16 sky and block light combinations.
```
The general idea is to summarize what the manager actually does. Feel free to write in your own words.

liach (approved):

mappings/net/minecraft/client/render/LightmapTextureManager.mapping
LambdAurora: I don't know what to write instead tbh, so I'll just include that suggestion directly.

YanisBft (approved):

Daomephsta:
@LambdAurora 
> I have no idea how to write unpick definitions

https://github.com/FabricMC/unpick/wiki/Unpick-Format
Yarn uses v2

haykam821 (approved):

PR #2595, opened 2021-07-19
Unpick lightmap constants.

LambdAurora:
With some bonus method mappings that I encountered while unpicking.

### DIFF
--- a/unpick-definitions/lightmap_coordinates.unpick
+++ b/unpick-definitions/lightmap_coordinates.unpick
@@ -122,7 +122,7 @@ target_method net/minecraft/client/render/entity/EntityRenderer render (Lnet/min
 	param 5 lightmap_coordinates
 
 target_method net/minecraft/client/render/entity/ExperienceOrbEntityRenderer vertex (Lnet/minecraft/client/render/VertexConsumer;Lnet/minecraft/util/math/Matrix4f;Lnet/minecraft/util/math/Matrix3f;FFIIIFFI)V
-	param 9 lightmap_coordinates
+	param 10 lightmap_coordinates
 
 target_method net/minecraft/client/render/entity/FishingBobberEntityRenderer vertex (Lnet/minecraft/client/render/VertexConsumer;Lnet/minecraft/util/math/Matrix4f;Lnet/minecraft/util/math/Matrix3f;IFIII)V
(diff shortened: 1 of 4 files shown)


unpick-definitions/lightmap_coordinates.unpick
```diff
@@ -0,0 +1,222 @@
...
+
+target_method net/minecraft/client/render/entity/EntityRenderer getLight (Lnet/minecraft/entity/Entity;F)I
+	return lightmap_coordinates
+target_method net/minecraft/client/render/entity/EntityRenderer renderLabelIfPresent (Lnet/minecraft/entity/Entity;Lnet/minecraft/text/Text;Lnet/minecraft/client/util/math/MatrixStack;Lnet/minecraft/client/render/VertexConsumerProvider;I)V
+	param 4 lightmap_coordinates
+target_method net/minecraft/client/render/entity/EntityRenderer render (Lnet/minecraft/entity/Entity;FFLnet/minecraft/client/util/math/MatrixStack;Lnet/minecraft/client/render/VertexConsumerProvider;I)V
+	param 5 lightmap_coordinates
+
+target_method net/minecraft/client/render/entity/ExperienceOrbEntityRenderer vertex (Lnet/minecraft/client/render/VertexConsumer;Lnet/minecraft/util/math/Matrix4f;Lnet/minecraft/util/math/Matrix3f;FFIIIFFI)V
+	param 9 lightmap_coordinates
```
liach:
```suggestion
	param 10 lightmap_coordinates
```

liach (commented):
Cannot find all the typos, but I think a few typos won't cause major troubles

unpick-definitions/lightmap_coordinates.unpick
LambdAurora: The arguments numbering is quite confusing, once it starts at 0, once it starts at 1, it's not consistent at all in Enigma mappings and it leads to quite the confusion.

liach: In unpick format, it is the index of the argument in the argument array in the descriptor than the actual slot of the variable, I think, so you don't apply an offset for instance methods. Forgot how it treats long/double though.

liach (approved):

YanisBft (approved):

PR #2596, opened 2021-07-20
1.17.1 util.math stuff

apple502j:
There are also some methods that I vaguely know what it is meant to be, but cannot come up with a good name...


mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -6,6 +6,13 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
 	FIELD field_15726 RANDOM Ljava/util/Random;
 	FIELD field_15727 ARCSINE_TABLE [D
 	FIELD field_15728 SMALLEST_FRACTION_FREE_DOUBLE D
+	FIELD field_29844 PI F
+	FIELD field_29845 HALF_PI F
+	FIELD field_29846 TAU F
```
enbrain, line 11: I'd document that it's `2 * PI`.

enbrain (approved):

mappings/net/minecraft/util/math/ChunkPos.mapping
```diff
@@ -35,6 +35,8 @@ CLASS net/minecraft/class_1923 net/minecraft/util/math/ChunkPos
 		ARG 1 offsetX
 		ARG 2 y
 		ARG 3 offsetZ
+	METHOD method_37232 toLong (Lnet/minecraft/class_2338;)J
```
liach, line 38: Hmm, how is this different from the instance method toLong on block pos

liach: Hmm, this one converts a block pos to a chunk long, guess this name is fine

mappings/net/minecraft/util/math/Direction.mapping
```diff
@@ -88,6 +88,8 @@ CLASS net/minecraft/class_2350 net/minecraft/util/math/Direction
 		ARG 0 matrix
 		ARG 1 direction
 	METHOD method_23955 getUnitVector ()Lnet/minecraft/class_1160;
+	METHOD method_30928 pointsTo (F)Z
+		ARG 1 degrees
```
liach:
```suggestion
		ARG 1 yaw
```

mappings/net/minecraft/util/math/Direction.mapping
```diff
@@ -88,6 +88,8 @@ CLASS net/minecraft/class_2350 net/minecraft/util/math/Direction
 		ARG 0 matrix
 		ARG 1 direction
 	METHOD method_23955 getUnitVector ()Lnet/minecraft/class_1160;
+	METHOD method_30928 pointsTo (F)Z
```
liach, line 91: This rather returns if the current direction forms an acute angle with the given yaw value. Better note that this always return `false` for vertical directions.

YanisBft (approved):

mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -9,8 +9,10 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
 	FIELD field_29844 PI F
 	FIELD field_29845 HALF_PI F
 	FIELD field_29846 TAU F
+		COMMENT Tau is equal to {@code 2 * PI}.
 	FIELD field_29847 RADIANS_PER_DEGREE F
 	FIELD field_29848 DEGREES_PER_RADIAN F
+	FIELD field_29849 APPROXIMATE_EQUALITY_TOLERANCE F
```
liach: We usually call this 
```suggestion
	FIELD field_29849 EPSILON F
```
Look up &epsilon; in math.

liach (approved):

enbrain (approved):

Juuxel:
`canMakeAcuteAngle` sounds really awkward. Really, while the acute angle property of the method is true, a more user-friendly interpretation of it is whether the yaw rotation roughly faces this direction.

apple502j:
Hmm, cannot come up with a good name.

Juuxel:
IMO `pointsTo` was good already.

liach (approved):

PR #2599, opened 2021-07-20
World gen mappings for 1.18 experimental snapshot 2

mschae23:
I tried mapping some of the things used by the new multi noise biome source.

I know they are unlikely to stay once we get normal snapshots, but they could be useful anyway.


mappings/net/minecraft/world/biome/source/util/MultiNoiseUtil.mapping
```diff
@@ -0,0 +1,123 @@
...
+		ARG 6 offset
+	METHOD method_37625 createMixedNoisePoint (Lnet/minecraft/class_6452$class_6454;Lnet/minecraft/class_6452$class_6454;Lnet/minecraft/class_6452$class_6454;Lnet/minecraft/class_6452$class_6454;Lnet/minecraft/class_6452$class_6454;Lnet/minecraft/class_6452$class_6454;F)Lnet/minecraft/class_6452$class_4762;
+		ARG 0 temperature
+		ARG 1 humidity
+		ARG 2 continentalness
+		ARG 3 erosion
+		ARG 4 depth
+		ARG 5 weirdness
+		ARG 6 offset
+	CLASS class_4762 MixedNoisePoint
```
liach: This is no longer a point but multidimensional as now each parameter accepts a range. Feel free to rename.

mappings/net/minecraft/world/biome/source/util/MultiNoiseUtil.mapping
```diff
@@ -0,0 +1,123 @@
...
+		METHOD method_37644 (Lnet/minecraft/class_6452$class_6460;)F
+			ARG 1 point
+		METHOD method_37646 getHumidity ()Lnet/minecraft/class_6452$class_6454;
+		METHOD method_37648 getContinentalness ()Lnet/minecraft/class_6452$class_6454;
+		METHOD method_37650 getErosion ()Lnet/minecraft/class_6452$class_6454;
+		METHOD method_37652 getDepth ()Lnet/minecraft/class_6452$class_6454;
+		METHOD method_37654 getWeirdness ()Lnet/minecraft/class_6452$class_6454;
+		METHOD method_37656 getOffset ()F
+		METHOD method_37658 getParameters ()Ljava/util/List;
+	CLASS class_6454 ParameterConfig
```
liach: This should be a min and a max range. The methods get the distance from a point to the range or from another range to the range, though the code isn't straightforward up front

mschae23: I looked at `method_37627` when mapping it, and I thought that it may be a range, but I wasn't sure. I'll rename that.

mappings/net/minecraft/world/biome/source/util/MultiNoiseUtil.mapping
mschae23: What would be a good name then? Something like `MixedNoiseRange` doesn't work, because it isn't a range itself.

liach: Hmm, biome class calls this a `parameter` in the codec format, doesn't it?

mschae23: So something like `NoiseParameters`?

liach (approved):
We can leave the noise point for now. Since modmuss says we won't port from the exp snapshots,  we can use it as a playground or just fix the critical problems if we do port from this version.

modmuss50:
I didnt say that we wont port from them, I said I dont want to promise it. I need to talk to player about figuring out if we can generate intermediaies from 2 match files as in theory its possible.

If that fails we can think again to see if its possible to save the names, but again it might be impossible so dont want to guarantee it.

mschae23:
> If that fails we can think again to see if its possible to save the names, but again it might be impossible so dont want to guarantee it.

If 1.18 snapshots don't get matched from this version, I can also just manually port my mappings.

Also, what should we name `method_37727`? I accidentally named it `addFixedValue`, because I thought it was the method that took a float and created a `FixedFloatFunction`. But it actually just directly calls the private `add` method, probably for future-proofing.
I considered `add2`, but that isn't a good name. Maybe something like `addFunction`?

**Edit**: Or, even better, rename the private method, so that all of the public `add` methods can have the same name.

mappings/net/minecraft/world/biome/source/util/MultiNoiseUtil.mapping
mschae23: I thought about this, and, technically, wouldn't it be a hypercube now? https://en.wikipedia.org/wiki/Hypercube

liach: Sure, a hypercube is a good name

mappings/net/minecraft/world/biome/source/util/MultiNoiseUtil.mapping
```diff
@@ -1,46 +1,55 @@
 CLASS net/minecraft/class_6452 net/minecraft/world/biome/source/util/MultiNoiseUtil
-	METHOD method_37620 createParameterConfig (F)Lnet/minecraft/class_6452$class_6454;
-		ARG 0 parameter
-	METHOD method_37621 createParameterConfig (FF)Lnet/minecraft/class_6452$class_6454;
-		ARG 0 parameter1
-		ARG 1 parameter2
+	FIELD field_34168 PARAMETER_NUMBER I
```
liach: Can be `HYPERCUBE_DIMENSION` if you choose to call those "hypercubes"

mappings/net/minecraft/world/biome/source/util/MultiNoiseUtil.mapping
```diff
@@ -1,46 +1,55 @@
...
 		ARG 6 offset
-	CLASS class_4762 MixedNoisePoint
-		COMMENT Represents a point in a multi-dimensional cartesian plane. Mixed-noise
-		COMMENT biome generator picks the closest noise point from its selected point
-		COMMENT and choose the biome associated to that closest point. Another factor,
-		COMMENT rarity potential, favors larger differences in values instead, contrary
-		COMMENT to other point values.
+	CLASS class_4762 NoiseHypercube
+		COMMENT Represents a hypercube in a multi-dimensional cartesian plane. The multi-noise
+		COMMENT biome source picks the closest noise hpercube from its selected point
```
liach:
```suggestion
		COMMENT biome source picks the closest noise hypercube from its selected point
```

mappings/net/minecraft/world/biome/source/util/MultiNoiseUtil.mapping
```diff
@@ -1,46 +1,55 @@
...
+		COMMENT and chooses the biome associated to it. Another factor, rarity potential,
+		COMMENT favors larger differences in values instead, contrary to other range values.
```
liach: Hmm, don't think this factor exists any more. I think they now use an offset, where when they sample, the offset is always 0, and the biome target cube points have a non-0 offset

mschae23: I think that's how `offset` worked before as well, though.

liach: Can just delete this "Another factor, ..." sentence then

mschae23: I think "another factor" doesn't mean the name of it here, but that it's another thing that contributes to the biome layout.

liach:
```suggestion
```

mappings/net/minecraft/world/biome/source/util/MultiNoiseUtil.mapping
```diff
@@ -0,0 +1,197 @@
...
+		ARG 1 humidity
+		ARG 2 continentalness
+		ARG 3 erosion
+		ARG 4 depth
+		ARG 5 weirdness
+		ARG 6 offset
+	CLASS class_4762 NoiseHypercube
+		COMMENT Represents a hypercube in a multi-dimensional cartesian plane. The multi-noise
+		COMMENT biome source picks the closest noise hypercube from its selected point
+		COMMENT and chooses the biome associated to it. Another factor, rarity potential,
```
liach:
```suggestion
		COMMENT and chooses the biome associated to it.
```

liach:
I am pretty sure now this is modeled to be a simple hypercube so it can be easily plugged in to be solved (probably with least squares, didn't check). The sentence is now redundant and should be removed.

mschae23:
> I am pretty sure now this is modeled to be a simple hypercube so it can be easily plugged in to be solved (probably with least squares, didn't check). The sentence is now redundant and should be removed.

Actually, maybe the sentence was wrong before 1.18 too. There, offset is used in the same way as now. But it doesn't "favor larger differences", because setting offset to something that's not `0` makes the biome smaller, not larger.

mappings/net/minecraft/world/biome/source/BiomeSource.mapping
```diff
@@ -14,6 +14,16 @@ CLASS net/minecraft/class_1966 net/minecraft/world/biome/source/BiomeSource
 		ARG 1 seed
 	METHOD method_28442 getCodec ()Lcom/mojang/serialization/Codec;
 	METHOD method_28443 getBiomes ()Ljava/util/List;
+	METHOD method_37612 getTerrainParameters (II)[D
+		COMMENT Returns an array containing an {@code offset} and a {@code factor} value
+		COMMENT for terrain generation.
+		COMMENT
+		COMMENT @return an array with two terrain parameters.
```
liach:
```suggestion
		COMMENT {@return an array containing an {@code offset} and a {@code factor} value
		COMMENT for terrain generation}
```

mschae23:
So, is this good then?

PR #2601, opened 2021-07-21
Eliminated ArmorMaterial and ToolMaterial overlap

### LINKED ISSUE #2600: ArmorMaterial and ToolMaterial interface overlap
The interfaces `ArmorMaterial` and `ToolMaterial` have overlapping method names meaning trying to implement both under one singular "material" instance at a time results in the inability to differentiate between the two. This is often desirable as even the vanilla materials have discrepancies between the armor and tool materials (the enchantability of iron armor is different than that of iron tools). Doing so would be a small quality of life improvement over needing a separate `ArmorMaterial` instance and `ToolMaterial` instance for every set of custom tools and armor.

The overlap is as follows:
- Both share a method called `getDurability` however the one in `ArmorMaterial` accepts an `EquipmentSlot` whereas the one in `ToolMaterial` does not. Technically I guess these two don't clash but I can imagine it is a little bit confusing implementing both on the same class at the same time.
- Both share a method called `getEnchantability` with the exact same signatures.
- Both share a method called `getRepairIngredient` with the exact same signatures.

My recommendation is to rename these methods in their respective interfaces as follows:
- `getDurability` -> `getArmorDurability` and `getToolDurability` respectively.
- `getEnchantability` -> `getArmorEnchantability` and `getToolEnchantability` respectively.
- `getRepairIngredient` -> `getArmorRepairIngredient` and `getToolRepairIngredient` respectively.

I understand that internally within each interface the addition of these clarifiers is redundant, but in actual practice I believe it would be beneficial to differentiate these methods.

I'm creating an issue here because I have literally no idea how to make a PR for yarn and no idea how to suggest mapping changes etc. If someone would like to give me a quick rundown, or if there's a document you can point me to explaining the process I'd be more than happy to go in and make a PR for this.

ghost:
Closes #2600 

Further information about why can be found in the linked issue.

There was discussion with @haykam821 on Discord about it not being great practice to change mappings solely to eliminate overlap and I would generally agree with that, however it seems like this is a fairly reasonable implementation choice that could become standard for developers, and the only alternative I can think of involves writing a separate class with helper methods to convert to the two interfaces depending on what's explicitly necessary and I think that's a little bit overkill for developers to implement for something so simple. That being said if it is decided that this PR shouldn't be merged for that reason, then maybe it makes sense to have a helper class like I described somewhere in fapi to aid in creating large volumes of armor and tool sets.

### DIFF
--- a/mappings/net/minecraft/item/ArmorMaterial.mapping
+++ b/mappings/net/minecraft/item/ArmorMaterial.mapping
@@ -1,11 +1,11 @@
 CLASS net/minecraft/class_1741 net/minecraft/item/ArmorMaterial
 	METHOD method_24355 getKnockbackResistance ()F
 	METHOD method_7694 getName ()Ljava/lang/String;
-	METHOD method_7695 getRepairIngredient ()Lnet/minecraft/class_1856;
-	METHOD method_7696 getDurability (Lnet/minecraft/class_1304;)I
+	METHOD method_7695 getArmorRepairIngredient ()Lnet/minecraft/class_1856;
+	METHOD method_7696 getArmorDurability (Lnet/minecraft/class_1304;)I
 		ARG 1 slot
 	METHOD method_7697 getProtectionAmount (Lnet/minecraft/class_1304;)I
 		ARG 1 slot
 	METHOD method_7698 getEquipSound ()Lnet/minecraft/class_3414;
-	METHOD method_7699 getEnchantability ()I
+	METHOD method_7699 getArmorEnchantability ()I
 	METHOD method_7700 getToughness ()F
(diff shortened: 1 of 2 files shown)


liach:
Hmm, I think we come back to the old inheritance vs holding debate, like on whether we should carry the armor/tool material for an item through creating a subclass implementing both or a class containing getters for both.

My suggestion is that you for now can create a pojo or builder that takes some parameters to maximize the shared data and the built object has getters for the armor and tool materials.

liach:
@modmuss50 How do you think of these names? I personally think if you do create a class for definition of materials, you should make them hold two fields of `ArmorMaterial` and `ToolMaterial` types than having it implement both interfaces.

modmuss50:
It wouldnt be my choice to make them one class, but I can see why people would. Removing the clash is a good idea, but something for 1.18 imo? Not sure what you guys think?

Shnupbups:
Probably too late in the 1.18 release cycle for a refactor of this magnitude now, let alone in 1.17, so if we are to truly consider this it'd likely be for 1.19 snapshots.

Regardless, this could be a good change. Personally I'm fine with the idea of having one class that serves as both a `ToolMaterial` and `ArmorMaterial`, though it's probably best practice to keep them separate in mods. Having the option to combine them could be useful for some people though.

PR #2602, opened 2021-07-21
ChestStateManager -> ViewerCountManager

### LINKED ISSUE #2573: ChestStateManager -> ContainerStateManager
It manages the viewers of a container block, like the current javadoc states:
https://github.com/FabricMC/yarn/blob/8d68cc22a9cc6451fb6d79130c62dcc3d12fb2b7/mappings/net/minecraft/block/entity/ChestStateManager.mapping#L1-L2

However, calling it `ChestStateManager` is a bit of a misnomer as it has nothing specific for chests. It's also used for barrels, which are another type of container altogether. The current javadoc phrasing "chest-like" works well there, but it's awkward for naming the class itself.

The methods `onChestOpened`/`onChestClosed`/`openChest`/`closeChest` should also refer to a "container" instead of a "chest".

YanisBft:
Closes #2573


Juuxel (approved):

mappings/net/minecraft/block/entity/ContainerStateManager.mapping
```diff
@@ -1,36 +1,38 @@
...
 		ARG 1 world
 		ARG 2 pos
-	METHOD method_31681 onChestOpened (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;)V
-		COMMENT Run when this chest is opened (when the viewer count becomes nonzero).
+	METHOD method_31681 onContainerOpened (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;)V
+		COMMENT Run when this container is opened (when the viewer count becomes nonzero).
 		ARG 1 world
 		ARG 2 pos
 		ARG 3 state
 	METHOD method_31682 onInteracted (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;II)V
```
liach: No, this should be `onViewerCountUpdated`, as this is called after chest open/close or when server determines the viewer count should update (such as because the viewer has fallen further or been pushed away by water).

In addition, should we rename `updateViewerCount` to `onScheduledTick` since all three users call this on scheduled block ticks to check which users have since been far away and should be removed from the viewer list?

liach (commented):
In fact, this is a container viewership tracker than any "state manager". I would prefer a rename that straightforwardly points out what this is than a mediocre "container state manager".

It should document that `onXxx` are callback methods implemented by users; they call open/closeChest for user interaction-based open/closes, and call the `updateViewerCount` (maybe `onScheduledTick` to check if some viewers have since become invalid) to ensure the whole functionality works.

mappings/net/minecraft/block/entity/ContainerStateManager.mapping
valoeghese: this isn't even being changed by the pr lol

valoeghese (approved):
change ScreenHandler to Container while you're at it :wink:

mappings/net/minecraft/block/entity/ContainerStateManager.mapping
liach: No point of changing "chest" to "container" if we keep these wrong stuff, and the class name (which is inaccurate) and the javadocs are  changed by this pr already

mappings/net/minecraft/block/entity/ContainerStateManager.mapping
```diff
@@ -1,36 +1,38 @@
-CLASS net/minecraft/class_5561 net/minecraft/block/entity/ChestStateManager
-	COMMENT Handles the viewer count for chest-like block entities.
+CLASS net/minecraft/class_5561 net/minecraft/block/entity/ContainerStateManager
+	COMMENT Handles the viewer count for container block entities, like chests,
+	COMMENT ender chests and barrels.
```
haykam821:
```suggestion
	COMMENT ender chests, and barrels.
```

haykam821 (changes requested):

Technici4n (commented):
I'd suggest `ViewerCountManager` or similar, pleaser avoid container... :smile:

Juuxel:
`ViewerCountManager` might actually be more accurate, I wouldn't mind that. For anyone unfamiliar with this class, it counts how many players are using/viewing a container's contents and calls open/close listeners based on that. (For example, for game events/sculk sensors)

liach:
A `ViewerCountManager` is definitely the best name; this is only a utility used by chest block entity implementations, and only to track viewer counts based on viewer's open/close packets and distance to the block entity.

haykam821 (approved):

Juuxel (approved):

mappings/net/minecraft/block/entity/ViewerCountManager.mapping
```diff
@@ -1,36 +1,38 @@
...
 		ARG 1 world
 		ARG 2 pos
-	METHOD method_31681 onChestOpened (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;)V
-		COMMENT Run when this chest is opened (when the viewer count becomes nonzero).
+	METHOD method_31681 onContainerOpen (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;)V
+		COMMENT Run when this container is opened (when the viewer count becomes nonzero).
 		ARG 1 world
 		ARG 2 pos
 		ARG 3 state
 	METHOD method_31682 onInteracted (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;II)V
```
liach: Still, this actually updates the viewer count, even if open/close methods are not called.
```suggestion
	METHOD method_31682 onViewerCountUpdate (Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;Lnet/minecraft/class_2680;II)V
```

mappings/net/minecraft/block/entity/ViewerCountManager.mapping
```diff
@@ -1,36 +1,38 @@
...
-		COMMENT Run when a player interacts with this chest.
+		COMMENT Run when a player interacts with this container.
```
liach:
```suggestion
		COMMENT Called when the viewer count updates, such as when a player interact with this container
		COMMENT or when {@linkplain #updateViewerCount distance-based checks} are run.
```

liach (approved):

enbrain (approved):

PR #2608, opened 2021-07-25
Remove all instances of  'ZERO WIDTH NON-JOINER' (U+200C) present in mappings.

alcatrazEscapee:
Please no invisible unicode characters.

For reference, other non-ASCII characters in Yarn are:

- One “§” U+00A7 Section Sign
https://github.com/FabricMC/yarn/blob/fd85633178de1fb3c2b82654f868f6f1370f0277/mappings/net/minecraft/client/font/TextVisitFactory.mapping#L38
- Four “×” U+00D7 Multiplication Sign
https://github.com/FabricMC/yarn/blob/fd85633178de1fb3c2b82654f868f6f1370f0277/mappings/net/minecraft/world/biome/source/BiomeCoords.mapping#L8-L9

### DIFF
--- a/mappings/net/minecraft/state/property/Properties.mapping
+++ b/mappings/net/minecraft/state/property/Properties.mapping
@@ -35,7 +35,7 @@ CLASS net/minecraft/class_2741 net/minecraft/state/property/Properties
 	FIELD field_12494 DELAY Lnet/minecraft/class_2758;
 		COMMENT A property that specifies the delay a repeater will apply.
 	FIELD field_12495 NORTH_WIRE_CONNECTION Lnet/minecraft/class_2754;
-		COMMENT A property that specifies how redstone wire attaches to the north.‌
+		COMMENT A property that specifies how redstone wire attaches to the north.
 	FIELD field_12496 AXIS Lnet/minecraft/class_2754;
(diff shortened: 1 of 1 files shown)


haykam821 (approved):

modmuss50:
Idk how this even happens. Removing the other two might not be a bad idea, but this change is good 👍

modmuss50 (approved):

haykam821:
We could include this in Javadoc linting as well.

liach:
Afaik, the javadoc tool's output html include such no-break whitespaces. Maybe we just happened to copy the contents straight from the javadoc html and inadvertently included such spaces as a result.

> other non-ASCII characters

Just curious, what's wrong with other non-ASCII? Yarn already enforces utf-8 for compilation and production.

If we do need to remove those characters, we can use `&sect;` for &sect; and `&times;` for &times; (which I am already doing).

alcatrazEscapee:
> Just curious, what's wrong with other non-ASCII?

Nothing I'm aware of. I just had already compiled this information and decided to put it here in case anyone else wanted to know if there were any other odd unicode characters present (or in other words, as a justification for this PRs completeness).

PR #2609, opened 2021-07-27
Mappings for 1.18 mountain surface builders

mschae23:
This PR adds mappings for 1.18 mountain surface builders.

It renames `NewMountainSurfaceBuilder` to `AbstractMountainSurfaceBuilder`.


liach:
Should match the id of the surface builders than biomes, as they are surface builders than biomes. Can the surface builders by used by any biome? If yes, then the first two renames has even less reason.

mschae23:
> Can the surface builders by used by any biome? If yes, then the first two renames has even less reason.

They can. I thought that this change could be good for the same reason that the biomes were renamed, to reduce confusion between old mountains and new mountains. But if the name should be like the registry ID, I can revert the renames.

liach:
Hmm, matching registry ID isn't perfect too if the another instance of the class can be assigned to another registry ID. The best way is still to approach by the builder's behavior and be descriptive; if it makes the terrain "extreme" then the renames are perfectly fine.

mschae23:
> The best way is still to approach by the builder's behavior and be descriptive; if it makes the terrain "extreme" then the renames are perfectly fine.

The surface builder changes the blocks at the surface. The `mountain` one makes it have sometimes grass, sometimes stone. I don't know how I would describe that in the name.

liach:
Then I think keeping the old `MountainSurfaceBuilder` name should be fine, as we can use the registry id as a second-hand evidence.

mschae23:
> Then I think keeping the old `MountainSurfaceBuilder` name should be fine, as we can use the registry id as a second-hand evidence.

Sure! I initially renamed them because I wanted to also rename `NewMountainSurfaceBuilder` to `MountainSurfaceBuilder`, but `AbstractMountainSurfaceBuilder` is better for that class anyway.

liach:
Hmm, does this "update base" functionality work for experimental versions, I wonder?... Rip, it doesn't

mschae23:
> Hmm, does this "update base" functionality work for experimental versions, I wonder?... Rip, it doesn't

I fixed it now.

liach:
Mind change branch to snapshot 4?

mschae23:
> Mind change branch to snapshot 4?

Done.

mappings/net/minecraft/world/gen/surfacebuilder/StoneShoreSurfaceBuilder.mapping
```diff
@@ -0,0 +1,8 @@
+CLASS net/minecraft/class_6485 net/minecraft/world/gen/surfacebuilder/StoneShoreSurfaceBuilder
```
apple502j: This is also used by stony peaks (as parent) iirc - I mapped it `StoneSurfaceBuilder`.

mschae23:
Updated to experimental snapshot 5.

modmuss50 (approved):

PR #2611, opened 2021-07-28
Document whether inclusive or not

enbrain:


mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -262,9 +267,16 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
 		ARG 1 min
 		ARG 2 max
 	METHOD method_32751 nextBetween (Ljava/util/Random;II)I
```
Juuxel, line 269: I wonder if this (or maybe `nextInt`) deserves a better name at some point? This is like an "unsafe" version of `nextInt`, which is like an inclusive version of `Random.nextInt` with that extra guard for the range.

mappings/net/minecraft/util/math/intprovider/UniformIntProvider.mapping
```diff
@@ -11,7 +11,9 @@ CLASS net/minecraft/class_6019 net/minecraft/util/math/intprovider/UniformIntPro
 		ARG 0 instance
 	METHOD method_35017 create (II)Lnet/minecraft/class_6019;
 		ARG 0 min
+			COMMENT the minimum value, inclusive
 		ARG 1 max
+			COMMENT the maximum value, inclusive
```
Juuxel, line 16: Hmm, this'd already be clear if the class used the codec names (`min_inclusive` and `max_inclusive`) instead of just `min` and `max`. (Same for `BiasedToBottomIntProvider`.)

mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -262,9 +267,16 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
...
+		COMMENT {@return a random, uniformly distributed integer value between {@code
+		COMMENT min} and {@code max}}
```
liach: Can we use math notations like `value in {@code [min, max]}` where `[]` indicates inclusive ranges while `()` indicates exclusive ranges?

modmuss50 (approved):

liach (approved):

modmuss50:
Please make sure to assign me when stuff is ready to merge, sadly it used to make it float to the top of my github homepage but doing so might make it so I see things a bit quicker 👍

liach:
Thanks for the tip. Since player merges prs as well, I stopped assigning as a result. Will assign for the future. :+1:

PR #2619, opened 2021-07-29
Improve ChunkStatus mappings

jakobkmar:


liach:
Hmm, you renamed non-obfuscated members, which we don't usually do and aren't sure of its safety (should be handled by remappping though)

jakobkmar:
which members where non-obfuscated?

liach:
The 3 "doWork" methods that appear red but are actually just straight from mojang names. It's said to be a bug in proguard that they don't obfuscate these if these methods are used by lambda metafactories.

jakobkmar:
I will revert the two mappings of `doWork` as soon as I have access to my PC again.

Also, is it okay that I remapped `function` to `convertCallback` or is that one a ProGuard compiler bug and non-obfuscated as well?

liach:
Naming any parameter is fine; those red names are not always there but only suggested by enigma.

jakobkmar:
I reverted the mappings for the `doWork` functions.

liach (approved):

YanisBft (approved):

PR #2571, opened 2021-07-09
Use automapped names for documented fields

YanisBft:
The latest version of Enigma fixes javadocs on automapped fields (thanks 2xsaiko)
Closes #1493

PR #2621, opened 2021-08-02
fix inherited names not being mapped for AbstractNbtList

ghost:
theyre now mapped to the same name, it fixes issue

not sure why this was broken but it works now i think
also enigma refuses to save mappings from and to the same name so

### DIFF
--- a/mappings/net/minecraft/item/ItemStack.mapping
+++ b/mappings/net/minecraft/item/ItemStack.mapping
@@ -164,7 +164,9 @@ CLASS net/minecraft/class_1799 net/minecraft/item/ItemStack
 		ARG 1 otherStack
 	METHOD method_31578 isItemBarVisible ()Z
 	METHOD method_31579 getItemBarStep ()I
+		COMMENT {@return the length of the filled section of the durability bar in pixels (out of 13)}
 	METHOD method_31580 getItemBarColor ()I
+		COMMENT {@return the color of the filled section of the durability bar}
 	METHOD method_32347 getTooltipData ()Ljava/util/Optional;
 	METHOD method_33262 onItemEntityDestroyed (Lnet/minecraft/class_1542;)V
 		ARG 1 entity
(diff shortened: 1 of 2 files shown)


liach:
Is this a bug with the produced tiny mappings? If yes fixing the tiny mapping production in yarn would be better (for example, #2571 is broke because of the way tiny mappings are produced as of now)

ghost:
yeah i think so cuz its not renaming methods with inherited mappings if the inherited names arent remapped if that makes sense

liach:
Hmm, those methods are literally from `java.util.List`: [`void add(int, Object)`](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#add-int-E-), [`Object remove(int)`](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#remove-int-), and [`Object set(int, Object)`](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#set-int-E-). And they do appear correct in the `decompileCfr` results.

What exact problem are you having? Broken name in mod development, or just red text in enigma?

ghost:
when the mappings are generated for impl classes like NbtByteArray, it doesnt inherit those add remove set names. this is most likely intentional but it can cause issues as i found.
here is the merged reordered v2 mappings. left is latest published right is generated after this commit
image
on the right the names are correct, on the left they arent.

ghost:
also decompilecfr might be wrong, the method names in NbtByteArray are fixed after this in the generated named jar

liach:
Yeah, problem confirmed in decompilecfr, but this should be fixed in enigma/toolchain, as this is part of the problems with proper synthetic override name propagation. The tool should propagate the names from non-minecraft library classes as well.

liach:
Namely, we call this name inheritance with with https://github.com/FabricMC/Enigma/blob/master/enigma-cli/src/main/java/cuchaz/enigma/command/MapSpecializedMethodsCommand.java (see #977). Probably need to check if https://github.com/FabricMC/Enigma/blob/master/enigma/src/main/java/cuchaz/enigma/analysis/index/BridgeMethodIndex.java has this bug.

ghost:
yes. remapping the unobf name is not an ideal solution, a fix for the mappings generator needs to be found.

ghost:
the changes this pr makes arent a good solution. moving this to an issue.

ghost:
#2623

Issue #2616, opened 2021-07-29
AzaleaBlock.generator should be uppercase


haykam821:
Since the `AzaleaBlock.generator` field is both static and final, its name should be completely uppercase:

```java
private static final AzaleaSaplingGenerator generator = new AzaleaSaplingGenerator();
```

haykam821:
Fixed by #2624

Issue #2615, opened 2021-07-29
EntitySelectorOptions.options should be uppercase


haykam821:
Since the `EntitySelectorOptions.options` field is both static and final, its name should be completely uppercase:

```java
private static final Map<String, EntitySelectorOptions.SelectorOption> options = Maps.newHashMap();
```

haykam821:
Fixed by #2624

Issue #2590, opened 2021-07-17
`TestFunctions.testClasses` -> `TEST_CLASSES`


Juuxel:
It's a static final field.

enbrain:
It would be great if fabric-filament can detect this, since the same mistake is happening in other places such as https://github.com/FabricMC/yarn/pull/2308#discussion_r611027554.

Juuxel:
I added something like that to Filament locally, and it produced [a whole bunch of field naming errors](https://gist.github.com/Juuxel/0cddb23ecf02e191e60fdb50c32aee37).

I'm not 100% sure if the field naming checks should be erroring since it could make snapshot ports annoying. (imagine Mojang moving a field from static final to instance final etc.) @modmuss50 Do you have an opinion on this since you're usually the one porting?

enbrain:
Shouldn't we consider these fields as constants? We probably need to update CONVENTIONS.md for clarification on these cases.

## non-final, but static and effectively final

```
lint: net/minecraft/util/math/Vec3f.POSITIVE_Z: non-static or non-final field starts with uppercase character 'P'
lint: net/minecraft/util/math/Vec3f.POSITIVE_X: non-static or non-final field starts with uppercase character 'P'
lint: net/minecraft/util/math/Vec3f.POSITIVE_Y: non-static or non-final field starts with uppercase character 'P'
lint: net/minecraft/util/math/Vec3f.ZERO: non-static or non-final field starts with uppercase character 'Z'
lint: net/minecraft/util/math/Vec3f.NEGATIVE_X: non-static or non-final field starts with uppercase character 'N'
lint: net/minecraft/util/math/Vec3f.NEGATIVE_Y: non-static or non-final field starts with uppercase character 'N'
lint: net/minecraft/util/math/Vec3f.NEGATIVE_Z: non-static or non-final field starts with uppercase character 'N'
```

Although they are non-final, they are static and their value is never changed after the initialization. 

## non-static, but final and effectively static

```
lint: net/minecraft/client/render/debug/ChunkLoadingDebugRenderer.LOADING_DATA_CHUNK_RANGE: non-static or non-final field starts with uppercase character 'L'
lint: net/minecraft/entity/passive/FoxEntity.CalmDownGoal.WORRIABLE_ENTITY_PREDICATE: non-static or non-final field starts with uppercase character 'W'
[...]

PR #2624, opened 2021-08-04
Use upper snake case with static final fields

IotaBread:
I think I fixed all constants that were using camel case


haykam821:
This pull request fixes #2616 but does not fix #2615.

enbrain:
Also linking to https://github.com/FabricMC/yarn/issues/2590 which contains a list of non upper snake static final fields.

enbrain (approved):

IotaBread:
Please let me know if there are any static final fields left

haykam821 (approved):

PR #2630, opened 2021-08-11
1.18-exp3 spawning

apple502j:
Yep yep, some conflicts with #2629, fixing...

Some interesting stuff:

- `SUCCESSFUL_SPAWN_CHANCE` is 0.24 (= only 24% of spawn attempts actually spawn)
- there's now per-chunk per-player? mob cap. Named it density cap, any better names?
- `getMobSpawnablePlayers` takes chunkPos and returns list of non-spectator players within 128 block radius of the chunk.

### DIFF
--- a/mappings/net/minecraft/server/world/ThreadedAnvilChunkStorage.mapping
+++ b/mappings/net/minecraft/server/world/ThreadedAnvilChunkStorage.mapping
@@ -164,6 +164,10 @@ CLASS net/minecraft/class_3898 net/minecraft/server/world/ThreadedAnvilChunkStor
 	METHOD method_34869 getChebyshevDistance (Lnet/minecraft/class_1923;Lnet/minecraft/class_1297;)I
 		ARG 0 chunkPos
 		ARG 1 entity
+	METHOD method_37831 (Lnet/minecraft/class_1923;Lnet/minecraft/class_3222;)Z
+		ARG 1 player
+	METHOD method_37832 getMobSpawnablePlayers (Lnet/minecraft/class_1923;)Ljava/util/List;
+		ARG 1 chunkPos
 	CLASS class_3208 EntityTracker
 		COMMENT An entity tracker governs which players' clients can see an entity. Each
 		COMMENT tracker corresponds to one entity in a server world and is mapped from the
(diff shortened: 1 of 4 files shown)


mappings/net/minecraft/server/world/ThreadedAnvilChunkStorage.mapping
```diff
@@ -164,6 +164,10 @@ CLASS net/minecraft/class_3898 net/minecraft/server/world/ThreadedAnvilChunkStor
 	METHOD method_34869 getChebyshevDistance (Lnet/minecraft/class_1923;Lnet/minecraft/class_1297;)I
 		ARG 0 chunkPos
 		ARG 1 entity
+	METHOD method_37831 (Lnet/minecraft/class_1923;Lnet/minecraft/class_3222;)Z
+		ARG 1 player
+	METHOD method_37832 getMobSpawnablePlayers (Lnet/minecraft/class_1923;)Ljava/util/List;
```
liach, line 169: This return `List<EntityLike>`, though it only returns players in the current implementation. Should it be named `Players` or `Entities`/ `Targets`?

mappings/net/minecraft/world/SpawnHelper.mapping
```diff
@@ -124,16 +131,24 @@ CLASS net/minecraft/class_1948 net/minecraft/world/SpawnHelper
 		FIELD field_24398 cachedPos Lnet/minecraft/class_2338;
 		FIELD field_24399 cachedEntityType Lnet/minecraft/class_1299;
 		FIELD field_24400 cachedDensityMass D
+		FIELD field_34297 densityCapper Lnet/minecraft/class_6480;
+		FIELD field_34298 biome Lnet/minecraft/class_1959;
```
liach:
```suggestion
		FIELD field_34298 cachedBiome Lnet/minecraft/class_1959;
```

liach:
`DensityCap` Imo is good enough.

mappings/net/minecraft/server/world/ThreadedAnvilChunkStorage.mapping
apple502j: I don't think it makes sense for this method to ever return non-players. Random mob spawn only happens around players, and not other mobs - and this method uses operations only applicable to players, such as checking gamemode. Maybe in the future it may change, but when that happens it'll probably fail in the matching process and become a new method.

liach: Yeah, this is totally fine. We can always revisit this later; maybe they just limited the return type to prevent access to unnecessary functionalities.

liach (approved):

liach:
Mind change branch to snapshot 4?

apple502j:
@liach Done.

PR #2631, opened 2021-08-11
Fixed A Mapping Mismatch in A Constructor Method

MarbleGateKeeper:
The mapping of constructor method of net.minecraft.block.Material does not match correctly.

### DIFF
--- a/mappings/net/minecraft/block/Material.mapping
+++ b/mappings/net/minecraft/block/Material.mapping
@@ -85,7 +85,7 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 		ARG 4 blocksMovement
 		ARG 5 blocksLight
 		ARG 6 burnable
-		ARG 7 breakByHand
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/block/Material.mapping
```diff
@@ -84,8 +84,8 @@ CLASS net/minecraft/class_3614 net/minecraft/block/Material
 		ARG 3 solid
 		ARG 4 blocksMovement
 		ARG 5 blocksLight
-		ARG 6 breakByHand
-		ARG 7 burnable
+		ARG 6 burnable
+		ARG 7 breakByHand
```
Juuxel: Isn't this the value for `replaceable`? If so, it should be named that.
```suggestion
		ARG 7 replaceable
```

YanisBft (approved):

Juuxel (approved):

liach (approved):

PR #2639, opened 2021-08-13
Fix typos

fredster33:


fredster33:
Uh, okay? Legitimately though that this was valid, but I won't go up against project conventions.

modmuss50:
There are a number of reasons why we are quite wary against large batches of PRs like this. It’s become quite a common issue for us. It sounds like you are trying to be helpful, but not everyone is doing it for the same reason sadly.

Thanks for understanding.

fredster33:
All right, thanks for providing a response.

PR #2642, opened 2021-08-17
1.18-exp4 stuff

apple502j:
Maps methods used by templates and new ore generation.

- `bury` prob needs javadoc , it's more like a side-effect-ful `canGenerate`: if it cannot generate, false, and if it can, then increase height by `deltaY` blocks (and since it is used with negative delta, bury by `-deltaY` blocks) and return true.
- `largeCopperOre` is true only in dripstone caves.
- `getLowestCornerSurfaceHeight` is just that; get min surface height of 4 corners from heightmap, and return lowest value. "width" and "height" here is deltaX and deltaZ, though I am not sure which one is better. This is used to check if the area (or, the corners) are filled with water in temple generation.

### DIFF
--- a/mappings/net/minecraft/structure/StructurePieceWithDimensions.mapping
+++ b/mappings/net/minecraft/structure/StructurePieceWithDimensions.mapping
@@ -15,3 +15,7 @@ CLASS net/minecraft/class_3418 net/minecraft/structure/StructurePieceWithDimensi
 	METHOD method_14839 (Lnet/minecraft/class_1936;Lnet/minecraft/class_3341;I)Z
 		ARG 1 world
 		ARG 2 boundingBox
+	METHOD method_37865 bury (Lnet/minecraft/class_1936;Lnet/minecraft/class_3341;I)Z
+		ARG 1 world
+		ARG 2 boundingBox
+		ARG 3 deltaY
--- a/mappings/net/minecraft/world/gen/feature/StructureFeature.mapping
+++ b/mappings/net/minecraft/world/gen/feature/StructureFeature.mapping
@@ -94,6 +94,12 @@ CLASS net/minecraft/class_3195 net/minecraft/world/gen/feature/StructureFeature
 	METHOD method_28664 init ()V
 	METHOD method_28665 getCodec ()Lcom/mojang/serialization/Codec;
 	METHOD method_36420 getUndergroundWaterCreatureSpawns ()Lnet/minecraft/class_6012;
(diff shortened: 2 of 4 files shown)


Juuxel:
> "width" and "height" here is deltaX and deltaZ

Hm, using "height" for Z values is a bit odd. The delta names might be better?

modmuss50 (approved):

mappings/net/minecraft/world/gen/feature/StructureFeature.mapping
```diff
@@ -94,6 +94,12 @@ CLASS net/minecraft/class_3195 net/minecraft/world/gen/feature/StructureFeature
 	METHOD method_28664 init ()V
 	METHOD method_28665 getCodec ()Lcom/mojang/serialization/Codec;
 	METHOD method_36420 getUndergroundWaterCreatureSpawns ()Lnet/minecraft/class_6012;
+	METHOD method_37864 getLowestCornerSurfaceHeight (Lnet/minecraft/class_2794;IILnet/minecraft/class_1923;Lnet/minecraft/class_5539;)I
```
liach:
```suggestion
	METHOD method_37864 getLowestCornerInGroundHeight (Lnet/minecraft/class_2794;IILnet/minecraft/class_1923;Lnet/minecraft/class_5539;)I
```
for consistency with the calls made by this method

mappings/net/minecraft/structure/StructurePieceWithDimensions.mapping
```diff
@@ -15,3 +15,7 @@ CLASS net/minecraft/class_3418 net/minecraft/structure/StructurePieceWithDimensi
 	METHOD method_14839 (Lnet/minecraft/class_1936;Lnet/minecraft/class_3341;I)Z
 		ARG 1 world
 		ARG 2 boundingBox
+	METHOD method_37865 bury (Lnet/minecraft/class_1936;Lnet/minecraft/class_3341;I)Z
```
liach: This isn't exactly burying. This shifts the piece vertically to cover the lowest y in the x-z area specified by its bounding box.

You should best name this together with `method_14839`. These two methods are quite similar; they both choose the `hPos` for this piece. 14839 chooses by calculating the average height in the area for `hPos`, while this one finds the minimum height in the area for `hPos`. In the end, both shifts the bounding box vertically so the new `box.getMinY` is `hPos` (`+ deltaY` for this one)

liach: After this analysis, I believe this class is better named `ShiftableStructurePiece` as this is its main functionality; the two methods can be named `adjustToAverageHeight` and `adjustToMinHeight` (these are general ideas, feel free to come up with better ones)

liach (approved):

PR #2646, opened 2021-08-23
Fix the player parameter of methods used for playing sounds in the World class being unclear

haykam821:


liach (approved):
If we want to be clear, we can document that when the player is passed, the player's client already played the sound and notified the server, so we don't try to ask that player's client to play the sound again :smile:

haykam821:
I think that this is good since the confusion is mostly from a server-only standpoint. When the same code is being ran on both the server and client, what you mentioned will come into effect.

liach:
hmm, what is the main usages of this? in handling client input or when server does things automatically like by schedule?

haykam821:
Vanilla uses these methods for playing sounds on both the client and server. On the client, the sound is played if `except` is the client's player. On the server, the sound is played to all nearby players except `except`. This causes the sound to be played to all players for an action that is called on both the server and client, like block interactions.

liach:
then this should be like an actor/sender as it plays sound on client.

YanisBft (approved):

liach:
If this parameter rename turns out unsuitable, we can always improve it without affecting mapping users later. So why not.

PR #2657, opened 2021-09-01
1.18-exp6 stuff

apple502j:
Some shared constants and biome stuff.

### DIFF
--- a/mappings/net/minecraft/world/biome/source/util/VanillaBiomeParameters.mapping
+++ b/mappings/net/minecraft/world/biome/source/util/VanillaBiomeParameters.mapping
@@ -77,6 +77,11 @@ CLASS net/minecraft/class_6461 net/minecraft/world/biome/source/util/VanillaBiom
 		ARG 1 parameters
 	METHOD method_37718 writeRiverBiomes (Lcom/google/common/collect/ImmutableList$Builder;Lnet/minecraft/class_6452$class_6454;)V
 		ARG 1 parameters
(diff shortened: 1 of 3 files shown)


mappings/net/minecraft/world/biome/source/util/VanillaBiomeParameters.mapping
```diff
@@ -77,6 +77,11 @@ CLASS net/minecraft/class_6461 net/minecraft/world/biome/source/util/VanillaBiom
 		ARG 1 parameters
 	METHOD method_37718 writeRiverBiomes (Lcom/google/common/collect/ImmutableList$Builder;Lnet/minecraft/class_6452$class_6454;)V
 		ARG 1 parameters
+	METHOD method_37846 getShoreOrNearInlandBiome (IILnet/minecraft/class_6452$class_6454;Lnet/minecraft/class_5321;)Lnet/minecraft/class_5321;
```
liach, line 80: This one really checks special cases and uses shattered savanna than the fallback in those cases.

liach (approved):

mappings/net/minecraft/world/biome/source/util/VanillaBiomeParameters.mapping
mschae23: I think I always named it something like `getBiomeOrShatteredSavanna` in my PRs.

liach: Yeah, out of all its usages, one of it is for picking a biome for shore and near-inland; a few others are near-inland, and one usage is like make a call and never made use of the result biome

apple502j:
I'll probably make a new PR with just some SharedConstants edit. Biome stuff is full of weird stuff and weirdness stuff...

PR #2662, opened 2021-09-03
Update Quaternion.mapping

Xeloboyo:


liach (approved):
Should be right. This is quite weird, as the code of these methods differ from most other existing methods in Quaternion

liach:
@modmuss50 mind review?

modmuss50 (approved):
looks fine to me, its a bit beyond me what this does tbh unless anyone has any objections it can be merged. Easy enought to fix later if required 👍

YanisBft (approved):

PR #2680, opened 2021-09-16
Allow regular builds to produce artifacts (notably javadocs) for inspection

liach:
A piece of vestige left in the ci java version bump.

The source level for the yarn javadoc/package/constants will stay at 16 until mojang bumps (which I would expect to happen very quickly as 16 is already EOL now)


modmuss50 (approved):
Good spot

PR #2682, opened 2021-09-16
21w37a JFR forwardport

apple502j:
Forwardported (is that a word?) the JFR mappings from #2582 - you can diff the patch file of this and that PRs in WinMerge, ignore `(field|method)_\d+` and compare, I guess.

JFR uses a lot of records; while they do feature mojang names in methods the names used here are all from #2582 PR. There are new things like `path` arg becoming `reader` due to type change, `fromEvent`/`toStatistics` etc.

Method for forwardporting: Ctrl+C from the original patch.  Took 90min or so.


modmuss50:
Just a heads up Player has said he is looking into matching and porting the exp mappings over, so it might not be a bad idea to hold on porting any more over.

Glad to see the record remapping seems to be working, I think there is still room for improvement, possibly expanding engima's format to combine the field and method mappings into one, and then writing out the 2 when exporting to other formats. Would need some discussion first however.

apple502j:
Now this PR is 100% new-in-21w37a stuff

mappings/net/minecraft/util/profiling/jfr/JfrProfiler.mapping
```diff
@@ -6,13 +6,35 @@ CLASS net/minecraft/class_6505 net/minecraft/util/profiling/jfr/JfrProfiler
...
+		ARG 2 recording
 	METHOD method_37983 start (Ljava/io/Reader;Lnet/minecraft/class_6505$class_6506;)Z
+		ARG 0 reader
 		ARG 1 instanceType
 	METHOD method_37984 isProfiling ()Z
 	CLASS class_6506 InstanceType
+		FIELD field_34414 type Ljava/lang/String;
+		METHOD <init> (Ljava/lang/String;ILjava/lang/String;)V
+			ARG 3 type
+		METHOD method_37986 getType ()Ljava/lang/String;
```
enbrain:
```suggestion
		FIELD field_34414 name Ljava/lang/String;
		METHOD <init> (Ljava/lang/String;ILjava/lang/String;)V
			ARG 3 name
		METHOD method_37986 getName ()Ljava/lang/String;
```
Duplicating the name (i.e. `type` in `instanceType.getType()`) should be avoided.

modmuss50 (approved):

enbrain (approved):

PR #2683, opened 2021-09-16
Manually adding signature back for records to aid decompilers and javadoc gen

liach:
This does seem to replace the signature for classes beside `class_6561` (somehow cannot find it in intermediary jar on my side). On a side note, due to a cfr bug, this signature won't show up in the decompiled code yet.

@modmuss50 Is this approach feasible? I want to have this fixed so we can reliably get type parameters for javadocs (and probably mods, too!)


modmuss50:
It seems like something we could do with in loom as well, I havent looked into the problem area much but if this is the solution, id make this a json and include it in the extras folder in the mapping jar (see unpick) would then allow loom to use it.

Having said that users of the mojang mappings shouldnt miss out on this so something else might be needed.

liach:
Done. Now they are exported to `extras/record_signatures.json` in intermediary names.

Chocohead:
A long time ago it was discussed for a Tiny V2.1 format which would support adding annotations (such as `@OverrideOnly`) and generics/signatures (originally for older versions which lack generics at all) to the code as a way of allowing changes such as this. Never happened of course but it's an option

liach:
> A long time ago it was discussed for a Tiny V2.1 format which would support adding annotations (such as `@OverrideOnly`) and generics/signatures (originally for older versions which lack generics at all) to the code as a way of allowing changes such as this. Never happened of course but it's an option

If you have the link to the spec, I guess I can try implement it in yarn, though the main purpose of this is to get javadoc working for yarn again

modmuss50:
For now I would skip adding it to the extras as it’s not a major deal to add it later once the loom side is done (not working for other mappings is an issue as well).

This seems a reasonable quick fix while something more final is done.

Chocohead:
> 
> 
> > A long time ago it was discussed for a Tiny V2.1 format which would support adding annotations (such as `@OverrideOnly`) and generics/signatures (originally for older versions which lack generics at all) to the code as a way of allowing changes such as this. Never happened of course but it's an option
> 
> If you have the link to the spec, I guess I can try implement it in yarn, though the main purpose of this is to get javadoc working for yarn again

There was never anything concrete decided on to get to making a spec. Annotations I don't think went beyond talking about it (although `a` is the obvious key letter), for generics there was a pair of options:
```
c	MyClass
	f	myField
		g	java/util/List
			g	java/util/List
				g	[I
```
or
```
c	MyClass
	f	myField
		s	Ljava/util/List<Ljava/util/List<[I>;>;
```
Depending on whether it was better to just write the whole signature or make something a little more immediately understandable (at the expensive of building it into a signature to apply of course)

liach:
Hmm, I don't see significant harm in shipping this json. We can leave it for loom to use and remove it at any time.

modmuss50 (approved):
Fine for a quick hack, dont depend on the json as it might well go/be changed.

PR #2689, opened 2021-09-17
21w37a misc stuff

apple502j:
Took 2h 35m 18s (according to Gradle) to map; should not conflict with other PRs. Didn't map some of the worldgen and chunk stuff.

And I can finally make a joke about "I mapped random things" :P

PR #2690, opened 2021-09-17
Biome generation mappings for 21w37a

mschae23:
Added mappings for biome generation, which was changed a lot again compared to the experimental snapshots.

I also mapped the new noise block state providers.


modmuss50 (approved):

mappings/net/minecraft/util/math/Range.mapping
```diff
@@ -0,0 +1,19 @@
+CLASS net/minecraft/class_6497 net/minecraft/util/math/Range
```
liach: Potential conflict with #2689

mappings/net/minecraft/world/gen/MultiNoiseParameters.mapping
```diff
@@ -0,0 +1,21 @@
+CLASS net/minecraft/class_2088 net/minecraft/world/gen/MultiNoiseParameters
+	FIELD field_24727 CODEC Lcom/mojang/serialization/Codec;
+	FIELD field_34626 temperature Lnet/minecraft/class_5216$class_5487;
+	FIELD field_34627 humidity Lnet/minecraft/class_5216$class_5487;
+	FIELD field_34628 continentalness Lnet/minecraft/class_5216$class_5487;
+	FIELD field_34629 erosion Lnet/minecraft/class_5216$class_5487;
+	FIELD field_34630 weirdness Lnet/minecraft/class_5216$class_5487;
+	FIELD field_34631 shift Lnet/minecraft/class_5216$class_5487;
```
liach, line 8: I recall this was named `offset` before as this is a penalty on the biome (to make it more rare). What's the reasoning of `shift`, just curious?

mappings/net/minecraft/world/gen/stateprovider/DualNoise2DBlockStateProvider.mapping
```diff
@@ -0,0 +1,16 @@
+CLASS net/minecraft/class_6578 net/minecraft/world/gen/stateprovider/DualNoise2DBlockStateProvider
```
liach: I recommend changing `2D` to `2d` in the class name

mappings/net/minecraft/world/gen/MultiNoiseParameters.mapping
mschae23: Offset still exists. This is the noise that was previously called "location offset", but since the name is now in the codec, I renamed it to that.

mappings/net/minecraft/world/gen/stateprovider/DualNoise2DBlockStateProvider.mapping
mschae23: True, that would be more readable. I think I'll just leave out the "2D", because they're actually 3D (which is probably unintended, but it's still confusing if the class is named 2D in that case).

mappings/net/minecraft/util/math/Range.mapping
mschae23: I personally prefer my own mappings for that class, but I'll remove them to not get conflicts.

mappings/net/minecraft/class_6497.mapping
```diff
@@ -0,0 +1 @@
+CLASS net/minecraft/class_6497
```
apple502j: Just curious, how did this get here? (does unmapping everything in Enigma do this?)

mschae23: Not sure, I saw that when I committed it too. I just used the "reset to obfuscated" thing on everything.

modmuss50 (approved):

mappings/net/minecraft/world/gen/stateprovider/NoiseCutoffBlockStateProvider.mapping
```diff
@@ -0,0 +1,16 @@
+CLASS net/minecraft/class_6581 net/minecraft/world/gen/stateprovider/NoiseCutoffBlockStateProvider
```
liach: Its registry name is `noise_threshold_provider`, so I think maybe using `Threshold` than `Cutoff` in name may be better.

mschae23: It was `cutoff` in 21w37a, I didn't see the rename.

liach: Yeah, things look good beyond that.

liach (approved):

PR #2691, opened 2021-09-17
Map Entity#getWorld

ghost:
Previously this method only existed in `ServerPlayerEntity` as `getServerWorld` however now all entities have a `World getWorld()` method and `ServerPlayerEntity` overloads it to `ServerWorld getWorld()`


enbrain:
That method is mapped in #2685.

ghost:
My bad, was looking for `ServerPlayerEntity` when checking because that was the class I was using in my code 😅

PR #2697, opened 2021-09-18
Terrain generation mappings for 21w37a

mschae23:
Started mapping terrain generation classes in 21w37a.


liach (approved):
looks good on glance

modmuss50 (approved):

PR #2699, opened 2021-09-18
Only run publish github action if repo owner is fabricmc

### LINKED ISSUE #2696: publish script shouldn't try to run on forks
This causes "check failure" to appear on commits pushed on fork branches.

image

https://github.com/FabricMC/yarn/blob/21w37a/.github/workflows/publish.yml should be fixed by adding `if: github.repository_owner == 'FabricMC'` to L7.

### LINKED ISSUE #2698: The job ID is incorrect for the Publish workflow
The Publish workflow has a job ID of `build`, like the Build workflow, instead of a similar one like `publish`:

https://github.com/FabricMC/yarn/blob/59e55e43fe37f8a6e8d426576e12f198c98877d3/.github/workflows/publish.yml#L5

liach:
Closes #2696 and closes #2698

Example of a skipped publish action: https://github.com/liach/yarn/runs/3640663999

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,7 +2,7 @@ name: Publish
 on: [push]
 concurrency: ci-${{ github.ref }}
 jobs:
-  build:
+  publish:
     if: ${{ github.repository_owner == 'FabricMC' }}
     runs-on: ubuntu-20.04
     container:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -3,6 +3,7 @@ on: [push]
 concurrency: ci-${{ github.ref }}
 jobs:
   build:
+    if: ${{ github.repository_owner == 'FabricMC' }}
     runs-on: ubuntu-20.04
     container:
       image: openjdk:17-jdk


liach:
@apple502j how does this look?

apple502j (approved):
LGTM!

YanisBft (approved):

Juuxel (approved):

.github/workflows/publish.yml
```diff
@@ -2,7 +2,8 @@ name: Publish
 on: [push]
 concurrency: ci-${{ github.ref }}
 jobs:
-  build:
+  publish:
+    if: ${{ github.repository_owner == 'FabricMC' }}
```
modmuss50, line 6: I dont think you need the `${{`

```suggestion
    if: github.repository_owner == 'FabricMC'
```

Should do it as far as I can tell, im not 100% sure what you have here will work.

modmuss50 (changes requested):

.github/workflows/publish.yml
liach: this syntax is from https://github.com/FabricMC/yarn/blob/544daf01e8e119b20b527b0a8a6780ceb2746575/.github/workflows/build.yml#L17 and it works for my projects in my organization

if we do change, i recommend changing both places

modmuss50: Ok sure, blame me then 👍 lets go with this.

PR #2703, opened 2021-09-23
21w38a JFR

apple502j:
Note: class_6507 is cursed and will be mapped when intermediary is fixed


modmuss50 (approved):

enbrain (approved):

mappings/net/minecraft/util/profiling/jfr/BaseJfrProfiler.mapping
```diff
@@ -0,0 +1,28 @@
+CLASS net/minecraft/class_6611 net/minecraft/util/profiling/jfr/BaseJfrProfiler
```
liach: Since this is an api class actually used by users, I would recommend calling this `FlightProfiler` or something that don't prompt people to use `JfrProfiler` impl class over this class.

liach (approved):

PR #2705, opened 2021-09-23
21w38a telemetry

apple502j:


mappings/net/minecraft/client/util/telemetry/TelemetryFields.mapping
```diff
@@ -0,0 +1,24 @@
+CLASS net/minecraft/class_6610 net/minecraft/client/util/telemetry/TelemetryFields
```
enbrain: Maybe `TelemetryConstants`? This class contains an event type (`WorldLoaded`) and a formatter, which I don't think are fields.

mappings/net/minecraft/client/util/Session.mapping
```diff
@@ -3,19 +3,30 @@ CLASS net/minecraft/class_320 net/minecraft/client/util/Session
...
+		ARG 5 clientId
+		ARG 6 accountType
 	METHOD method_1673 getUuid ()Ljava/lang/String;
 	METHOD method_1674 getAccessToken ()Ljava/lang/String;
 	METHOD method_1675 getSessionId ()Ljava/lang/String;
 	METHOD method_1676 getUsername ()Ljava/lang/String;
 	METHOD method_1677 getProfile ()Lcom/mojang/authlib/GameProfile;
 	METHOD method_35718 getAccountType ()Lnet/minecraft/class_320$class_321;
+	METHOD method_38740 getClientId ()Ljava/util/Optional;
+	METHOD method_38741 getXuid ()Ljava/util/Optional;
```
liach, line 22: What's the source of this "xuid"? They just call it `UserId` in string literals.

ghost: That's what it is refered to in the blog post:
https://www.minecraft.net/en-us/article/minecraft-snapshot-21w38a

liach: Hmm, a search shows it probably means xbox user id. I'd still call it `UserId` because I haven't seen xuid in the literal strings

apple502j: It's weird. The command argument defined in `Main` is `xuid`, and the name in other places are `userId`. Maybe I should rename.

apple502j: Renamed param to `userId` to reduce confusion (and just in case they want to support mojang account). I think we should still call it xuid in `Session` because 1) that's the name of the cmd argument that is passed to init and 2) other account types may have other types of user id.

liach (approved):

enbrain (approved):

PR #2707, opened 2021-09-23
Yet another 21w38a stuff

apple502j:
Sound and util stuff.

### DIFF
--- a/mappings/net/minecraft/client/sound/SoundEngine.mapping
+++ b/mappings/net/minecraft/client/sound/SoundEngine.mapping
@@ -6,6 +6,8 @@ CLASS net/minecraft/class_4225 net/minecraft/client/sound/SoundEngine
 	FIELD field_19183 EMPTY_SOURCE_SET Lnet/minecraft/class_4225$class_4276;
 	FIELD field_19184 streamingSources Lnet/minecraft/class_4225$class_4276;
 	FIELD field_19185 staticSources Lnet/minecraft/class_4225$class_4276;
+	FIELD field_34946 disconnectExtensionPresent Z
+	FIELD field_34947 deviceSpecifier Ljava/lang/String;
 	METHOD method_19661 init (Ljava/lang/String;)V
 		ARG 1 deviceSpecifier
 	METHOD method_19662 release (Lnet/minecraft/class_4224;)V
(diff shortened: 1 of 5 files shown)


mappings/net/minecraft/util/Util.mapping
```diff
@@ -7,6 +7,9 @@ CLASS net/minecraft/class_156 net/minecraft/util/Util
 	FIELD field_24622 BOOTSTRAP_EXECUTOR Ljava/util/concurrent/ExecutorService;
 	FIELD field_25140 NIL_UUID Ljava/util/UUID;
 	FIELD field_33859 JAR_FILE_SYSTEM_PROVIDER Ljava/nio/file/spi/FileSystemProvider;
+	FIELD field_34891 MAX_BACKGROUND_THREADS I
```
liach: Maybe `MAX_PARALLELISM` which is used by java's documentation: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/ForkJoinPool.html#%3Cinit%3E(int)

```suggestion
	FIELD field_34891 MAX_PARALLELISM I
```

Your current name is wrong; it should be `MAX_MAX_BACKGROUND_THREADS`, but that corrected name sounds a bit weird. Hence this alternative.

mappings/net/minecraft/client/sound/SoundEngine.mapping
```diff
@@ -22,6 +24,9 @@ CLASS net/minecraft/class_4225 net/minecraft/client/sound/SoundEngine
 	METHOD method_38501 openDeviceOrFallback (Ljava/lang/String;)J
 		ARG 0 deviceSpecifier
 	METHOD method_38502 getSoundDevices ()Ljava/util/List;
+	METHOD method_38722 getCurrentDeviceName ()Ljava/lang/String;
```
liach, line 27:
```suggestion
	METHOD method_38722 getDeviceName ()Ljava/lang/String;
```
No point including `current` if there is no contrast with another "old" device, etc.

apple502j: There is, kinda; this will always return up-to-date info from the OS, and is used inside a reload code (to check if system default has changed). The name is not good, I guess, but "getDeviceName" sounds like a simple field getter when it isn't.

liach (approved):

modmuss50 (approved):

PR #2708, opened 2021-09-24
Move to name-proposal and update enigma/cfr

modmuss50:
Enimga/CFR update makes record decompiling better but still not perfect. Needs a bit more looking into.

Also moved and partly rewritten the name prosoal stuff that was in stitch into its own project: https://github.com/FabricMC/name-proposal.

I have spot checked the output mappings and they seem fine, still quite a lot of room for breakage however.


liach:
peeked at the new project and have a few recommendations in my mind. will send pr for this and forked javapoet when i get time

modmuss50:
@liach Thanks, this PR was also ment to kinda be a place to provide feedback on the new project, Its always a little be awkward to review new project.

👍

liach:
if you aim to move off stitch, feel free to check out my wip at https://github.com/liach/yarn/commit/c61ab73d64e2615a8dbd47bf3fe7442c421696c3

it is blocked by https://github.com/FabricMC/mapping-io/issues/6 for now

modmuss50:
Oh awesome, doing that was on my todo list. Might also be a good chance to move more stuff to filament if that would help make things a bit more maintainable.

Beyond that replacing stitch isn’t too far off, loom still uses it for jar merging, something I have plans for. And for v1 mapping support.

build.gradle
```diff
@@ -73,7 +74,7 @@ def unpickMetaFile = file("unpick-definitions/unpick.json")
 
 dependencies {
 	enigmaRuntime "cuchaz:enigma-swing:${project.enigma_version}:all"
-	enigmaRuntime "net.fabricmc:stitch:${project.stitch_version}"
```
liach, line 76: Beyonds the merging and name proposal, another thing we need to take care of is stitch's intermediary generation abilities.

gradle.properties
```diff
@@ -1,12 +1,13 @@
 # Done to increase the memory available to gradle.
 org.gradle.jvmargs=-Xmx1G
 
-enigma_version=1.4.2
+enigma_version=1.4.5
 stitch_version=0.6.1
 unpick_version=2.2.0
-cfr_version=0.0.6
+cfr_version=0.0.7
```
liach, line 7: I recall the newer cfr has problem with outer class record desugaring (not dumping with the record class dumper). Is that the case in enigma?

build.gradle
liach: Nvm this is enigma than buildscript dependency. Ignore my last comment.

liach:
The build failure might have something to do with https://github.com/FabricMC/fabric-loom/commit/98d8f3767253a1a3308542c2e896cbf5f4382033 which is addressed in #2710 as well.

liach (approved):

liach:
@modmuss50 mind pull latest 21w38a into your pr? i think build failed for that bad intermediary. also need to fix a merge conflict

gradle.properties
modmuss50: I havent noticed that, I hope this should be better than what we had before, im sure more fixes will be needed.

modmuss50:
Still seems to be failing, not quite sure why my changes would cause this?

gradle.properties
liach: if that becomes a proble.m, shouldn't be too hard to find a fix either

liach:
I've identified the cause to be missing outer class-based mapping reparations (like that one in loom); I should have identified this problem earlier based on that mappingpoet doesn't operate on intermediary or obf names at all, only yarn names.

liach:
@modmuss50 Mind take a look at this and keep it moving? I find the record inference stuff helpful and want to get this done soon :tm:

modmuss50:
@liach it now builds, will just check it works ok.

modmuss50:
Seems to work, I will merge this now. This has quite a high risk of breaking things so please let me know if you come across anything odd.

PR #2715, opened 2021-09-26
Improve ProjectileEntity

enbrain:
* `user` → `shooter`: for clarity.
* `modifierZ`, `modifierXYZ` → `speed`, `divergence`: to actually describe what they are.
* ~~`setVelocity` → `setProperties`: It also sets yaw and pitch. I'm not convinced that `setProperties` is ideal, but at least it's not wrong.~~
* `setProperties` → `setVelocity`: for consistency.


mappings/net/minecraft/entity/projectile/ProjectileEntity.mapping
```diff
@@ -30,11 +38,23 @@ CLASS net/minecraft/class_1676 net/minecraft/entity/projectile/ProjectileEntity
 		ARG 1 entity
 	METHOD method_7454 onEntityHit (Lnet/minecraft/class_3966;)V
 		ARG 1 entityHitResult
-	METHOD method_7485 setVelocity (DDDFF)V
+	METHOD method_7485 setProperties (DDDFF)V
```
liach: `setVelocity` imo is still better as the result is setting velocity; it's that the velocity is calculated in a somewhat quirky way.

mappings/net/minecraft/entity/projectile/ProjectileEntity.mapping
```diff
@@ -4,12 +4,20 @@ CLASS net/minecraft/class_1676 net/minecraft/entity/projectile/ProjectileEntity
 	FIELD field_28646 shot Z
 	FIELD field_33399 owner Lnet/minecraft/class_1297;
 	METHOD method_24919 setProperties (Lnet/minecraft/class_1297;FFFFF)V
```
liach: Imo this should be setting velocity as well. This just takes account of the shooter's velocity so the input velocity is relative.

mappings/net/minecraft/entity/projectile/ProjectileEntity.mapping
```diff
@@ -4,12 +4,20 @@ CLASS net/minecraft/class_1676 net/minecraft/entity/projectile/ProjectileEntity
...
 		ARG 3 yaw
+			COMMENT the yaw
 		ARG 4 roll
-		ARG 5 modifierZ
-		ARG 6 modifierXYZ
+			COMMENT the roll
+		ARG 5 speed
+			COMMENT the length of the velocity
+		ARG 6 divergence
+			COMMENT the multiplier for the random vector added to the direction
```
liach: Maybe mention this is like fuziness, where player usages have 1 and other mobs/tools have higher values like 12 for snow golem, 8 for witch, 6 for dispensers, and difficulty-adjusted value for skeleton shooting, etc.

mappings/net/minecraft/entity/projectile/ProjectileEntity.mapping
```diff
@@ -4,12 +4,20 @@ CLASS net/minecraft/class_1676 net/minecraft/entity/projectile/ProjectileEntity
...
 		ARG 2 pitch
+			COMMENT the pitch
 		ARG 3 yaw
+			COMMENT the yaw
 		ARG 4 roll
-		ARG 5 modifierZ
-		ARG 6 modifierXYZ
+			COMMENT the roll
+		ARG 5 speed
+			COMMENT the length of the velocity
```
liach: just say the speed, if you want to specific mention the velocity will be scaled to have this speed in 3d space than in x-z plane projection

liach (commented):
I recommend checking out `net/minecraft/block/dispenser/ProjectileDispenserBehavior`'s `getVariation` and `getForce` too.

mappings/net/minecraft/entity/projectile/ProjectileEntity.mapping
enbrain: I'm happy to rename it to `fuzziness` if it's clearer. Well, we got three candidates: `divergence`, `variation` and `fuzziness`.

liach (approved):

PR #2732, opened 2021-10-07
Some 21w40a worldgen mappings

mschae23:
Added mappings for world generation changes in 21w40a.


enbrain (approved):

mappings/net/minecraft/world/gen/blockpredicate/TrueBlockPredicate.mapping
```diff
@@ -0,0 +1,6 @@
+CLASS net/minecraft/class_6680 net/minecraft/world/gen/blockpredicate/TrueBlockPredicate
+	FIELD field_35158 INSTANCE Lnet/minecraft/class_6680;
```
enbrain, line 2:
```suggestion
	FIELD field_35158 instance Lnet/minecraft/class_6680;
```
It's only static, not final. Just what YanisBft commented in https://github.com/FabricMC/yarn/pull/2733#discussion_r725663401.

mschae23: I'd assume that's a mistake though. It's still effectively final anyway. There's nothing to change in the instance.

PR #2734, opened 2021-10-07
Fix parameter mis-order in BrewingRecipeRegistry

Machine-Maker:
The method `BrewingRecipeRegistry#craft(input, ingredient)` has the potion itemstack named `ingredient` but in the previous method `BrewingRecipeRegistry#hasPotionRecipe` the potion itemstack is `input`. They are inverted. One or the other should be swapped so the parameter names are consistent with what they represent in the context of potion brewing


enbrain (approved):
I'd name them `ingredient` and `potion` since both are input in some sense.

YanisBft (approved):

Machine-Maker:
Well, `potion` isn't 100% correct-sounding either, since it can just be a water bottle. And `input` is what's used to represent that slot in other methods already

PR #2764, opened 2021-10-16
AbstractBlock: add mapping for missing method

sarunint:


enbrain (approved):
We probably need to rename `getMaxModelOffset` to horizontal at some point.

modmuss50 (approved):

PR #2765, opened 2021-10-16
Spawn point tweaks and fixes

Juuxel:
- `spawnPointSet` -> `spawnForced` + getter
  - To match the NBT key, and because a non-forced spawn point (e.g. a bed) can be set
- Added various parameter mappings and some jd

### DIFF
--- a/mappings/net/minecraft/server/network/ServerPlayerEntity.mapping
+++ b/mappings/net/minecraft/server/network/ServerPlayerEntity.mapping
@@ -164,7 +164,7 @@ CLASS net/minecraft/class_3222 net/minecraft/server/network/ServerPlayerEntity
 		ARG 4 forced
 			COMMENT whether the new spawn point is {@linkplain #isSpawnForced() forced}
 		ARG 5 sendMessage
-			COMMENT if {@code true}, a chat message about the spawn point change will be sent
+			COMMENT if {@code true}, a game message about the spawn point change will be sent
 	METHOD method_26285 isBedTooFarAway (Lnet/minecraft/class_2338;Lnet/minecraft/class_2350;)Z
 		ARG 1 pos
 		ARG 2 direction
--- a/mappings/net/minecraft/server/network/ServerPlayerEntity.mapping
+++ b/mappings/net/minecraft/server/network/ServerPlayerEntity.mapping
@@ -39,7 +39,7 @@ CLASS net/minecraft/class_3222 net/minecraft/server/network/ServerPlayerEntity
 	FIELD field_13998 joinInvulnerabilityTicks I
 	FIELD field_23191 spawnPointDimension Lnet/minecraft/class_5321;
 	FIELD field_23192 spawnPointPosition Lnet/minecraft/class_2338;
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/server/network/ServerPlayerEntity.mapping
```diff
@@ -153,13 +153,18 @@ CLASS net/minecraft/class_3222 net/minecraft/server/network/ServerPlayerEntity
 		ARG 1 origin
 	METHOD method_26280 getSpawnPointPosition ()Lnet/minecraft/class_2338;
 	METHOD method_26281 getSpawnPointDimension ()Lnet/minecraft/class_5321;
-	METHOD method_26282 isSpawnPointSet ()Z
+	METHOD method_26282 isSpawnForced ()Z
```
liach, line 156: What exactly does forcing do? No fuzzy offset at spawn? Ignore beds/respawn anchors?

mappings/net/minecraft/server/network/ServerPlayerEntity.mapping
```diff
@@ -153,13 +153,18 @@ CLASS net/minecraft/class_3222 net/minecraft/server/network/ServerPlayerEntity
...
 		ARG 1 dimension
+			COMMENT the new spawn dimension
 		ARG 2 pos
+			COMMENT the new spawn point, or {@code null} if resetting to the world spawn
 		ARG 3 angle
-		ARG 4 spawnPointSet
+		ARG 4 forced
+			COMMENT whether the new spawn point is {@linkplain #isSpawnForced() forced}
 		ARG 5 sendMessage
+			COMMENT if {@code true}, a chat message about the spawn point change will be sent
```
liach:
```suggestion
			COMMENT if {@code true}, a game message about the spawn point change will be sent
```
we use game message for system non-hotbar messages

mappings/net/minecraft/server/network/ServerPlayerEntity.mapping
Juuxel: See `PlayerEntity.findRespawnPosition` (the first bool parameter, also mapped in this PR). The logic of that method is basically

1. If the block at the spawn point is a respawn anchor, handle that in `RespawnAnchorBlock`
2. If the block is a bed, handle that in `BedBlock`
3. Now it's an unknown block (from Mojang's perspective), so if not forced (not set with `/spawnpoint`), return `Optional.empty` (since there's been a bed or anchor before that has been destroyed!)
4. If forced and players can actually spawn at that position, spawn there.

liach: Ah, so forced is simply skipping respawn anchor and bed checks.

Juuxel: Not really skipping, but falling back if they fail. Their logic will still run if the blocks happen to be respawn anchors or beds. All this should be a bit clearer now since I added javadoc for that method.

enbrain (approved):

YanisBft (approved):

PR #2772, opened 2021-10-20
21w42a particles

YanisBft:


haykam821:
Could this pull request also map `field_35432`? This is the set of items that should have block marker particles for their respective blocks.

YanisBft:
> Could this pull request also map `field_35432`? This is the set of items that should have block marker particles for their respective blocks.

Done!

haykam821 (approved):

mappings/net/minecraft/client/particle/Particle.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS net/minecraft/class_703 net/minecraft/client/particle/Particle
+	FIELD field_28786 velocityMultiplier F
```
enbrain, line 2: Maybe `friction` since this field is applied every tick and its value is between 0 and 1.

enbrain (approved):

PR #2777, opened 2021-10-20
Rename DefaultBiomeCreator to OverworldBiomeCreator and add mappings for Nether and End counterparts

sarunint:
Looks like Nether's and End's biome creator has been extracted out to their own classes in 21w41a.


sarunint:
Duplicate of #2775

PR #2790, opened 2021-10-27
21w43a stuff

apple502j:
Does not include tickers & worldgen.


enbrain (changes requested):
It needs to fix the merge conflict. modmuss fixed the intermediary used in the `Option` class.

Looks good otherwise.

enbrain (approved):

modmuss50 (approved):

PR #2801, opened 2021-10-31
Refactor equipment sync stuff

enbrain:
The current names don't convey that they are sync stuff.


liach:
Most important entity data are already registered to the data tracker and synced. This change imo is bad unless there is already a non-synced (which non-syncing is the best behavior than an error) component of equipments.

enbrain:
> This change imo is bad unless there is already a non-synced (which non-syncing is the best behavior than an error) component of equipments.

There is a "non-synced" version called `LivingEntity#getEquippedStack`.

The method `method_30129` compares `getEquippedStack` and `method_30125` (or `method_30126`) to compute the difference between the last sent one and the current one. This difference is sent as an `EntityEquipmentUpdateS2CPacket`, not through the data tracker system.

liach (approved):

modmuss50:
Im slightly uneasy about this change, I will need to take a closer look at the code to gain an understanding of the change, but I think it boils down to if it really does mean something else for the majority of cases.

This rename is highly impactful, so before we go through with it I want to make sure its the right thing to do.

mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -255,22 +255,30 @@ CLASS net/minecraft/class_1309 net/minecraft/entity/LivingEntity
 	METHOD method_30120 (Ljava/util/List;Lnet/minecraft/class_1304;Lnet/minecraft/class_1799;)V
 		ARG 2 slot
 		ARG 3 stack
-	METHOD method_30121 swapHandStacks (Ljava/util/Map;)V
-		ARG 1 equipment
-	METHOD method_30122 setArmorInSlot (Lnet/minecraft/class_1304;Lnet/minecraft/class_1799;)V
+	METHOD method_30121 sendHandStackSwap (Ljava/util/Map;)V
```
liach: I prefer
```suggestion
	METHOD method_30121 checkHandStackSwap (Ljava/util/Map;)V
```
as this first checks and sends the swap only if the hands have updated

liach:
> This rename is highly impactful, so before we go through with it I want to make sure its the right thing to do.

That was my instinct upon seeing this as well, but it turns out all these methods and fields are all private. They are details around keeping track and sending equipment stacks to the client.

modmuss50:
Ah, if these are private and not the ones used by most people then it sounds a lot less impacful, I will take a look at the code before i jump to any more conclusions. 👍

liach:
someone mind take a look in enigma and review this?

liach:
@Shnupbups mind review this? This is a good change, though it takes some time to review all these contents

Shnupbups (approved):

Shnupbups:
Looks good to me!

modmuss50 (approved):

PR #2805, opened 2021-11-01
21w43a datafixer methods

apple502j:
At least the stuff I think I understand. Most params are not mapped because gosh it's so confusing


mappings/net/minecraft/datafixer/fix/ProtoChunkTickListFix.mapping
```diff
@@ -1,3 +1,7 @@
 CLASS net/minecraft/class_6740 net/minecraft/datafixer/fix/ProtoChunkTickListFix
+	FIELD field_35447 WATER_CONTAINING_BLOCK_IDS Lcom/google/common/collect/ImmutableSet;
```
enbrain: What is a water containing block? Is that different from a waterlogged block?

apple502j: The list contains kelp, seagrass, bubble columns, etc. Is there a word that describes these - blocks that are "always waterlogged"?

enbrain: Not sure, but I think `ALWAYS_WATERLOGGED_BLOCK_IDS` is more descriptive then.
```suggestion
	FIELD field_35447 ALWAYS_WATERLOGGED_BLOCK_IDS Lcom/google/common/collect/ImmutableSet;
```

enbrain (approved):

modmuss50 (approved):

PR #2817, opened 2021-11-09
Unpick key binding categories

Juuxel:
Note that this modifies the existing `key_codes.unpick` file and renames it to `key_bindings.unpick` since you can't have multiple target method definitions for the same target method.


enbrain (approved):
I wonder if something similar can be done for the `StickyKeyBinding`'s constructor.

Juuxel:
@enbrain I actually didn't see that class at all, added.

enbrain (approved):

modmuss50 (approved):

PR #2820, opened 2021-11-11
Fix 1.18-pre1 record signatures

enbrain:


modmuss50 (approved):

liach (approved):
Should be good looking at the before and after build results

PR #2737, opened 2021-10-08
NearestLivingEntitiesSensor Mappings + ChunkSimulationDistanceS2C Packet + Other misc mappings

aws404:
This PR includes mappings for:
- Changes to the NearestLivingEntitiesSensor, the sensor now stores data in the form of 'VisibleLivingEntityList' which has methods for streaming, iterating, testing and getting from the sorted list.
  - Currently placed in the root of the entity.ai.brain package, however I think it may fit elsewhere better?
- Adds mapping for the packet used to communicate the simulation distance from server -> client (used similar wording to the ChunkLoadDistanceS2CPacket)
- Misc. block predicate mappings
- Fix an ingredient mapping

PR #2825, opened 2021-11-12
Map the new LivingTargetCache class

enbrain:
Supersedes #2737

### DIFF
--- a/mappings/net/minecraft/entity/ai/brain/LivingTargetCache.mapping
+++ b/mappings/net/minecraft/entity/ai/brain/LivingTargetCache.mapping
@@ -1,9 +1,9 @@
-CLASS net/minecraft/class_6670 net/minecraft/entity/ai/brain/VisibleLivingEntityList
+CLASS net/minecraft/class_6670 net/minecraft/entity/ai/brain/LivingTargetCache
 	FIELD field_35104 EMPTY Lnet/minecraft/class_6670;
 	FIELD field_35105 entities Ljava/util/List;
-	FIELD field_35106 visibilityPredicate Ljava/util/function/Predicate;
+	FIELD field_35106 targetPredicate Ljava/util/function/Predicate;
 	METHOD <init> (Lnet/minecraft/class_1309;Ljava/util/List;)V
-		ARG 1 observer
+		ARG 1 owner
 		ARG 2 entities
 	METHOD method_38971 empty ()Lnet/minecraft/class_6670;
 	METHOD method_38972 contains (Lnet/minecraft/class_1309;)Z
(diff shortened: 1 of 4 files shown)


liach:
I feel `SensedEntityList` or `SensedEntities` would be better; this has nothing to do with `visible` given you can use any predicate.

enbrain:
Hmmm, `visible` is what mojang calls it. The following code is how the class is used.

```java
brain.remember(MemoryModuleType.MOBS, list);
brain.remember(MemoryModuleType.VISIBLE_MOBS, new class_6670(entity, list));
```

```java
public static final MemoryModuleType<List<LivingEntity>> MOBS = register("mobs");
public static final MemoryModuleType<class_6670> VISIBLE_MOBS = register("visible_mobs");
```

liach:
Hmm, after thinking, maybe when the `attackable` is false in the target predicate, it means the target predicate is choosing a visible target? But then "visible" becomes a confusing word, given mobs have visibility logic too (see `net.minecraft.entity.mob.MobVisibilityCache`, but `canSee(Entity)` is defined in `LivingEntity`). Preferably we should not mix up these "visibility" concepts either.

Hence, I still prefer a reference to "sense" than "see" (as we reserve "see" for whether an entity can connect to another entity in ray casting). Or we can perform a refactor to put all these terminologies in line with each other.

Hence, I would suggest names like `[Living](Sense/Target)Cache`. The word "target" needs a little bit clarification too, since in `LivingEntity` the two `canTarget` methods actually inquire if the given entity/type can be potentially chosen as *attack* target.

enbrain:
I agree that "visible" is confusing, but where is "cache" coming from? I don't see any cache logic in `class_6670`. Do you mean that a brain memory is a cache and `class_6670` is only used as a memory value?

liach:
Isn't this object created every few ticks and swiftly discarded later as the list of entities update? It seems to work like mob visibility cache, as if it has computed results, it returns it.

enbrain:
Ah, never mind. I checked again and it's indeed a cache. (They use `Object2BooleanOpenHashMap`.)

mappings/net/minecraft/entity/ai/brain/LivingTargetCache.mapping
```diff
@@ -0,0 +1,28 @@
+CLASS net/minecraft/class_6670 net/minecraft/entity/ai/brain/LivingTargetCache
+	FIELD field_35104 EMPTY Lnet/minecraft/class_6670;
+	FIELD field_35105 entities Ljava/util/List;
+	FIELD field_35106 targetPredicate Ljava/util/function/Predicate;
+	METHOD <init> (Lnet/minecraft/class_1309;Ljava/util/List;)V
+		ARG 1 owner
+		ARG 2 entities
+	METHOD method_38971 empty ()Lnet/minecraft/class_6670;
+	METHOD method_38972 contains (Lnet/minecraft/class_1309;)Z
```
liach, line 9: Should this be `canTarget` or just `contains`, as this first checks if the entity is in the list and then checks if it matches the target predicate?

liach (approved):

mappings/net/minecraft/entity/ai/brain/LivingTargetCache.mapping
enbrain: Since all operations in this class test the target predicate, I think this class represents a list of entities that only match the predicate.

modmuss50 (approved):

PR #2835, opened 2021-11-16
Fix PillagerOutpostFeature.MONSTER_SPAWNS having wrong descriptor

enbrain:


liach (approved):
Ah, I only noticed the weird lack of newline. Didn't expect this missing semicolon in review

PR #2844, opened 2021-11-18
Use "offhand" instead of "offHand"

### LINKED ISSUE #2818: LivingEntity.getOffHandStack -> getOffhandStack, PlayerInventory.offHand -> offhand
To match the capitalisation used elsewhere in Yarn and Minecraft (the game uses "offhand" in localised strings).

apple502j:
Resolves #2818

Question: should we remap `Hand.OFF_HAND` (an enum)?


liach (approved):

Juuxel:
> Question: should we remap Hand.OFF_HAND (an enum)?

No, because they have the internal name field that should match the field name. I think that's just Mojang being inconsistent: at least we can be consistent and choose the most common spelling.

Juuxel (approved):

enbrain (approved):

YanisBft (approved):

modmuss50:
Uhh, my bad for merging this refactor. I dont think it should have been merged this late into a dev cycle....

net.minecraft.util.Hand is also still wrong.

modmuss50:
Ah, just seen the comment above. Im not a big fan of this.

Juuxel:
> net.minecraft.util.Hand is also still wrong.

That's an enum so it can't really be renamed (`name()` would return the wrong value).

PR #2858, opened 2021-11-18
Update mappingpoet to include a warning for records' behavior under reflection

liach:
Warning text:
https://github.com/FabricMC/MappingPoet/blob/master/src/main/resources/record_warning.txt

```
<div class="deprecation-block"><span class="deprecated-label">Warning: This class is not a record at runtime!</span>
<div class="deprecation-comment">Due to Proguard processing, This class is not a record at
       runtime when inspected by reflection. In addition, if this record has any type parameter,
       inspecting generic types through reflection may have unexpected results or throw exceptions.
       See <a href="https://fabricmc.net/wiki/tutorial:reflection#records">
       The Fabric Wiki</a> for more details.</div>
</div>
```

See also https://fabricmc.net/wiki/tutorial:reflection#records for detailed explanation.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -889,7 +889,7 @@ javadoc {
 				'implSpec:a:Implementation Requirements:',
 				'implNote:a:Implementation Note:'
 		)
-		taglets "net.fabricmc.mappingpoet.jd.MappingTaglet"
+		taglets "net.fabricmc.mappingpoet.jd.RecordWarningTaglet", "net.fabricmc.mappingpoet.jd.MappingTaglet"
 		// taglet path, header, extra stylesheet settings deferred
 		it.use()
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,4 +10,4 @@ name_proposal_version=0.1.2
 # Javadoc generation/linking
 fabric_loader_version=0.11.6
 jetbrains_annotations_version=22.0.0
-mappingpoet_version=0.3.0
+mappingpoet_version=0.3.1


modmuss50:
Would it be a better idea to fix the records at runtime? Its already done in loom, should be quite easy to also fix it in loader when it remaps the jar?

The generic types might be a bit of a harder thing to do during prod runtime but that seems like a minor issue?

liach:
> Would it be a better idea to fix the records at runtime? Its already done in loom, should be quite easy to also fix it in loader when it remaps the jar?

Given we already have record components identified separately by matcher, it would not be hard at all.

> The generic types might be a bit of a harder thing to do during prod runtime but that seems like a minor issue?

Fixing this one is even easier: We just ship the record signatures json with fabric loader to apply these signature updates, and these crashes and errors are easily fixed. Also this is another argument to mojang for restoring record signatures as removing the signatures will cause reflection to throw exceptions :smile:

--------

So these comes down to whether we want light vs heavy modification to vanilla. We are currently on the lighter side by applying minimal patches so the game imitates vanilla as much as possible. (forge just decompiles and recompiles, which definitely changes a lot of stuff) It's up to you on whether to patch these or to leave the fabric environment more like vanilla, given both of these behaviors are observable with vanilla (tested in 21w37a, when they first introduced records)

modmuss50:
> Fixing this one is even easier: We just ship the record signatures json with fabric loader

No we cannot, as fabric loader is not version specific. We could include them with intermediary but that has the problem of beign hard to update once released. With this I think its best to just wait for mojang to fix it.

For the record components its trivial as like you said components have their own prefix. See how loom does this here: https://github.com/FabricMC/fabric-loom/blob/c380b3673974de065016843906b246f08a90d80a/src/main/java/net/fabricmc/loom/util/TinyRemapperHelper.java#L84-L86

liach:
ah, reminds me of the good old times when installer specifies yarn versions. i suggest leaving both as is and wait for mojang, and add this warning for now.

modmuss50:
Ok, I agree. Can you make it clear that its just at production runtime and not during development? I think it would be good to note that it doesnt behave the same in both environments.

liach:
I should also mention these problems only happen with reflection; otherwise modders may proceed without concern.

liach:
@modmuss50 Should I move the message file to yarn itself's extras so we can update it without updating mappingpoet? I will probably make turning on this warning and the taglet a system property if we do so.

modmuss50:
Yes, I dont think it makes much sense beign defined in MP

modmuss50:
This might have all been fixed for pre6: https://bugs.mojang.com/browse/MC-236947

liach:
interesting, will wait and see

modmuss50:
Signature and components are fixed in pre6. 👍

PR #2860, opened 2021-11-19
Map BufferedInputStream

apple502j:
I'm not even sure why mojang reinvented a wheel. Is this even different from the one in JDK, API-wise?

### DIFF
--- a/mappings/net/minecraft/util/BufferedInputStream.mapping
+++ b/mappings/net/minecraft/util/BufferedInputStream.mapping
@@ -1,0 +1,19 @@
+CLASS net/minecraft/class_6826 net/minecraft/util/BufferedInputStream
+	FIELD field_36209 DEFAULT_BUFFER_SIZE I
+	FIELD field_36210 stream Ljava/io/InputStream;
+	FIELD field_36211 buf [B
+	FIELD field_36212 count I
+	FIELD field_36213 pos I
+	METHOD <init> (Ljava/io/InputStream;)V
+		ARG 1 stream
+	METHOD <init> (Ljava/io/InputStream;I)V
+		ARG 1 stream
+		ARG 2 size
+	METHOD method_39779 getAvaliable ()I
+	METHOD method_39780 fill ()V
+	METHOD read ([BII)I
+		ARG 1 buf
+		ARG 2 offset
+		ARG 3 length
+	METHOD skip (J)J
+		ARG 1 n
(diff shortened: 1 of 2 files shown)


enbrain:
They use `BufferedOutputStream` along with `class_6826` in `ChunkStreamVersion` so definitely they know the `BufferedInputStream` in JDK. The code structure of `class_6826` is very similar to [the OpenJDK implementation](https://github.com/openjdk/jdk/blob/c79a485f1c3f9c0c3a79b8847fdcd50a141cd529/src/java.base/share/classes/java/io/BufferedInputStream.java) but some checks (such as whether the stream is open) are omitted. Might be an unsafe version that fixes a sneaky bug (in a very hacky way) or optimizes certain things?

enbrain (approved):

modmuss50:
Its not ideal having the same name IMO, it will make it easy to accidentally use this one over the java version. Without really knowing why this one is better (Is it safer? Is it faster?) its hard to give it another name tho.

enbrain:
It's not the best but something like `BufferedChunkInputStream` could work as a cheap solution.

mappings/net/minecraft/util/BufferedInputStream.mapping
```diff
@@ -0,0 +1,19 @@
...
+	FIELD field_36210 stream Ljava/io/InputStream;
+	FIELD field_36211 buf [B
+	FIELD field_36212 count I
+	FIELD field_36213 pos I
+	METHOD <init> (Ljava/io/InputStream;)V
+		ARG 1 stream
+	METHOD <init> (Ljava/io/InputStream;I)V
+		ARG 1 stream
+		ARG 2 size
+	METHOD method_39779 getAvaliable ()I
```
liach: I recommend
```suggestion
	METHOD method_39779 getAvailableBuffer ()I
```
for this only concerns about the buffer than the whole stream

mappings/net/minecraft/util/BufferedInputStream.mapping
```diff
@@ -0,0 +1,19 @@
+CLASS net/minecraft/class_6826 net/minecraft/util/BufferedInputStream
+	FIELD field_36209 DEFAULT_BUFFER_SIZE I
+	FIELD field_36210 stream Ljava/io/InputStream;
+	FIELD field_36211 buf [B
+	FIELD field_36212 count I
+	FIELD field_36213 pos I
```
liach:
```suggestion
	FIELD field_36212 end I
	FIELD field_36213 start I
```
Though those are jdk names, I find these ones more comprehensible.

liach (commented):
I recommend `FixedBufferInputStream` as compared to the jdk buffered input stream, this one:

### disadvantages
1. Does not support async operations
2. Does not support marking and jumping back

### advantages
1. Does not reallocate a new byte array every time `fill()` is called

------
I am suggesting this name based on this single advantage.

liach (approved):

modmuss50 (approved):

PR #2863, opened 2021-11-20
Document placement modifier

enbrain:


mappings/net/minecraft/world/gen/feature/PlacedFeature.mapping
```diff
@@ -18,6 +18,18 @@ CLASS net/minecraft/class_6796 net/minecraft/world/gen/feature/PlacedFeature
...
+		COMMENT placement modifiers to the given position.
+		COMMENT
+		COMMENT <p>The steps to calculate positions are as follows:
+		COMMENT <ul>
+		COMMENT   <li>applying the first modifier to the given position,
+		COMMENT   <li>applying the second modifier to each of the result positions,
+		COMMENT   <li>flattening the result positions,
+		COMMENT   <li>applying the third modifier to each of the result positions,
+		COMMENT   <li>flattening the result positions,
+		COMMENT   <li>and so on.
```
mschae23: Do you really need to explain how `flatMap` works here?

enbrain: Probably mentioning `flatMap` just works. I just wanted to avoid confusion with the 1.17 decorator. Decorator was a (linked-list backed) stack, which consumes the last set decorator first.

liach: Don't think a list is appropriate here as you are generally repeating one procedure. You should say something like:
"starting with a set of one position, for each placement modifier, we send each position in the set into the modifier and take union of the results to obtain the updated set of positions"

liach (approved):

Shnupbups (approved):

PR #2884, opened 2021-11-24
Rename NbtNull to NbtEnd

### LINKED ISSUE #2882: Rename `NbtNull` to `NbtEnd`
I've tried using it as a `null` value multiple times, but it will actually end the serialization in many cases. Renaming it will ensure that mods won't use it as a null value by mistake, like they could with JSON. See also AztechMC/Modern-Industrialization#161.

apple502j:
Resolves #2882

### DIFF
--- a/mappings/net/minecraft/nbt/NbtElement.mapping
+++ b/mappings/net/minecraft/nbt/NbtElement.mapping
@@ -1,9 +1,9 @@
 CLASS net/minecraft/class_2520 net/minecraft/nbt/NbtElement
 	COMMENT Represents an NBT element.
-	FIELD field_33250 NULL_TYPE B
-		COMMENT The numeric ID of an NBT null value. Is {@value}.
+	FIELD field_33250 END_TYPE B
+		COMMENT The numeric ID of an NBT end value. Is {@value}.
 		COMMENT
-		COMMENT @see NbtNull
+		COMMENT @see NbtEnd
 	FIELD field_33251 BYTE_TYPE B
 		COMMENT The numeric ID of an NBT byte value. Is {@value}.
 		COMMENT
--- a/mappings/net/minecraft/nbt/NbtEnd.mapping
+++ b/mappings/net/minecraft/nbt/NbtEnd.mapping
@@ -1,0 +1,7 @@
+CLASS net/minecraft/class_2491 net/minecraft/nbt/NbtEnd
+	COMMENT Represents the NBT end value.
+	COMMENT Defines the end of an {@link NbtCompound} object during serialization,
+	COMMENT and is the type of an empty {@link NbtList}.
+	FIELD field_21032 TYPE Lnet/minecraft/class_4614;
+	FIELD field_21033 INSTANCE Lnet/minecraft/class_2491;
+	FIELD field_33193 SIZE I
--- a/mappings/net/minecraft/nbt/NbtEnd.mapping
+++ b/mappings/net/minecraft/nbt/NbtEnd.mapping
@@ -4,4 +4,6 @@ CLASS net/minecraft/class_2491 net/minecraft/nbt/NbtEnd
 	COMMENT and is the type of an empty {@link NbtList}.
 	FIELD field_21032 TYPE Lnet/minecraft/class_4614;
 	FIELD field_21033 INSTANCE Lnet/minecraft/class_2491;
+		COMMENT A dummy instance of the NBT end. It will never appear nested in any parsed NBT
+		COMMENT structure and should never be used as NBT compound values or list elements.
 	FIELD field_33193 SIZE I
--- a/mappings/net/minecraft/nbt/NbtNull.mapping
+++ b/mappings/net/minecraft/nbt/NbtNull.mapping
@@ -1,8 +1,0 @@
-CLASS net/minecraft/class_2491 net/minecraft/nbt/NbtNull
-	COMMENT Represents the NBT null value.
-	COMMENT Defines the end of an NBT compound object,
-	COMMENT represents nonexistent values in an NBT compound object,
-	COMMENT and is the type of empty NBT lists.
-	FIELD field_21032 TYPE Lnet/minecraft/class_4614;
-	FIELD field_21033 INSTANCE Lnet/minecraft/class_2491;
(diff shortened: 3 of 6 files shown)


liach (approved):

LambdAurora (changes requested):
Please see:
> `NbtEnd` is misleading, see:
#2090 
https://github.com/FabricMC/yarn/pull/2090#discussion_r588860544
> 
> If you're doing low-level programming it is common to use null as a character to end the treatment of data, but it's not only treated as an end.

liach:
> Please see:
> 
> > `NbtEnd` is misleading, see:
> > #2090
> > [#2090 (comment)](https://github.com/FabricMC/yarn/pull/2090#discussion_r588860544)
> > If you're doing low-level programming it is common to use null as a character to end the treatment of data, but it's not only treated as an end.

The single fact that if you put such a value into the value of a compound NBT would lead to malformed NBT compound defeats your whole argument.

LambdAurora:
> > Please see:
> > > `NbtEnd` is misleading, see:
> > > #2090
> > > [#2090 (comment)](https://github.com/FabricMC/yarn/pull/2090#discussion_r588860544)
> > > If you're doing low-level programming it is common to use null as a character to end the treatment of data, but it's not only treated as an end.
> 
> The single fact that if you put such a value into the value of a compound NBT would lead to malformed NBT compound defeats your whole argument.

smh let's rename `NULL` to `END` in C then

liach:
Eh, this is how mojang literally calls it
image

the `END` name does address the issue that it shouldn't be misused within compounds (as they end compounds). `NbtNull` name fails to accomplish so. If you have other names that can address these issues:

1. Indicate it shouldn't be used in compounds as `null`s as in Java/JSON sense (the two things Minecraft modders interact most often with) but is actually a terminating character
2. Can be used as element type for empty list nbts

Then feel free to suggest these names. `Null` doesn't address the first issue.

Juuxel:
Yep, End makes the most sense in a Java context where `null` is just a value and not a terminator (in standard data structures anyway).

mappings/net/minecraft/nbt/NbtElement.mapping
```diff
@@ -1,9 +1,9 @@
 CLASS net/minecraft/class_2520 net/minecraft/nbt/NbtElement
 	COMMENT Represents an NBT element.
-	FIELD field_33250 NULL_TYPE B
-		COMMENT The numeric ID of an NBT null value. Is {@value}.
+	FIELD field_33250 END_TYPE B
```
Juuxel, line 3: This needs to be updated in the corresponding unpick file.

LambdAurora:
> Eh, this is how mojang literally calls it
> 
> the `END` name does address the issue that it shouldn't be misused within compounds (as they end compounds). `NbtNull` name fails to accomplish so. If you have other names that can address these issues:
> 
>     1. Indicate it shouldn't be used in compounds as `null`s as in Java/JSON sense (the two things Minecraft modders interact most often with) but is actually a terminating character
> 
>     2. Can be used as element type for empty list nbts
> 
> 
> Then feel free to suggest these names. `Null` doesn't address the first issue.

As far as I know we're not here to strictly follows Mojang's decade naming, and we already determined multiple times that early names from Mojang are terrible.

I really don't like `End` as it really does not address the empty type issue at all. 
What frustrates me is 16 people agreed on the previous PR, and no one went "oh yeah this is an issue" while I pointed out the rename in the original message and the reasoning.

liach:
> What frustrates me is 16 people agreed on the previous PR, and no one went "oh yeah this is an issue" while I pointed out the rename in the original message and the reasoning.

Eh, not a cool analogy. Reminds me of how, in the 1800s, Aaron Burr promoted a bill in NYC to provide sanitized water; everyone agreed except they didn't notice a clause that granted the creation of a bank.

You definitely aren't trying to be cunning like Burr, but over time we find an actual problem with some parts of that large refactor that we initially didn't anticipate. Tweaking this part does not mean invalidating the whole previous refactor.


----

Back on the `NbtEnd`/`NbtNull` itself, we should probably document that the singleton `INSTANCE` should never appear in an NBT structure; whenever the null/end type appears, it will only have a type (as in the nbt list type or the nbt compound end marker), and it shouldn't be ever used as a list element/compound element etc. This definitely should be documented disregarding whether we carry out this rename.

mappings/net/minecraft/nbt/NbtEnd.mapping
```diff
@@ -0,0 +1,7 @@
+CLASS net/minecraft/class_2491 net/minecraft/nbt/NbtEnd
+	COMMENT Represents the NBT end value.
+	COMMENT Defines the end of an {@link NbtCompound} object during serialization,
+	COMMENT and is the type of an empty {@link NbtList}.
+	FIELD field_21032 TYPE Lnet/minecraft/class_4614;
+	FIELD field_21033 INSTANCE Lnet/minecraft/class_2491;
```
liach, line 6:
```suggestion
	FIELD field_21033 INSTANCE Lnet/minecraft/class_2491;
		COMMENT A dummy instance of the NBT end. It will never appear nested in any parsed NBT
		COMMENT structure and should never be used as NBT compound values or list elements.
```

Technici4n:
> What frustrates me is 16 people agreed on the previous PR, and no one went "oh yeah this is an issue" while I pointed out the rename in the original message and the reasoning.

The rest of the tag -> nbt rename PR is good, and this end -> null rename was not the main focus of the PR. Just like the original `Tag` was later deemed suboptimal, I believe `NbtNull` to be suboptimal per the issues mentioned above, and we have the opportunity to change it, so let's just do it! 😄

========= (how do you do this line?)

Yes, some additional comment would definitely be welcome! :)

liach:
========= (how do you do this line?)

`----------`

rendered like:

----------

Juuxel (approved):

Shnupbups (approved):

modmuss50 (approved):

modmuss50:
I agree with the change, however it is quite late in in the day to rename it. I dont think many mods use this so it wont affect too many people. Im happy to merge this now, just want to allow for a bit of time for people to disagree with me however.

PR #2885, opened 2021-11-24
1.18-pre8 stuff

apple502j:


mappings/net/minecraft/text/TranslatableText.mapping
```diff
@@ -14,8 +14,9 @@ CLASS net/minecraft/class_2588 net/minecraft/text/TranslatableText
 		ARG 2 args
 	METHOD method_11022 getKey ()Ljava/lang/String;
 	METHOD method_11023 getArgs ()[Ljava/lang/Object;
-	METHOD method_11024 setTranslation (Ljava/lang/String;Ljava/util/function/Consumer;)V
+	METHOD method_11024 format (Ljava/lang/String;Ljava/util/function/Consumer;)V
```
apple502j: Note that this method will not set translation on its own anymore; it just formats the string with `this.args` and calls `partsConsumer`

Shnupbups (approved):

Juuxel (approved):

mappings/net/minecraft/text/TranslatableText.mapping
liach: It just feeds the translations to the consumers, but `format` doesn't seem that relevant either. I recommend `forEachPart`
```suggestion
	METHOD method_11024 forEachPart (Ljava/lang/String;Ljava/util/function/Consumer;)V
```

liach (approved):

PR #2895, opened 2021-12-01
Use suffixes for key binding fields in the GameOptions class

### LINKED ISSUE #2888: Key binding fields should not have 'key' prefix
The fields which have type `KeyBinding` in the `GameOptions` class don't follow this convention:

> To make code as easy to read as possible, keep names in the natural language order. For example, a class representing a chest block entity should be named `ChestBlockEntity` rather than `BlockEntityChest`. Though prefix naming may be helpful for grouping classes together in an IDE's tree view, reading and writing code is done much more often than browsing files.

Some examples that should be renamed are:
* `GameOptions.keyForward` -> `forwardKey`
* `GameOptions.keysHotbar` -> `hotbarKeys`
* `GameOptions.keysAll` -> `allKeys`

haykam821:
Fixes #2888


liach (approved):

Shnupbups (approved):

Juuxel (approved):
Hmm, I wonder if this'll break a lot of mods that access vanilla keybindings directly (that should be pretty rare though)

YanisBft (approved):

enbrain (approved):

modmuss50 (approved):

PR #2903, opened 2021-12-04
Rename incorrectly-named intersectsEntities methods

### LINKED ISSUE #2901: Entity intersection related methods are probably incorrect
It looks from impl and code usage (e.g. mob spawning related methods) that methods in EntityView, CollisionView etc that return whether an entity collides/intersects with other entities (e.g. `EntityView#intersectsEntities`) returns `false` when it intersects. Can anyone confirm this?

apple502j:
Fixes #2901 

Also renamed param for `EntityView#doesNotIntersectEntities` to match the CollisionView name - will javadoc in that PR.

### DIFF
--- a/mappings/net/minecraft/world/CollisionView.mapping
+++ b/mappings/net/minecraft/world/CollisionView.mapping
@@ -31,9 +31,9 @@ CLASS net/minecraft/class_1941 net/minecraft/world/CollisionView
 	METHOD method_8600 getCollisions (Lnet/minecraft/class_1297;Lnet/minecraft/class_238;)Ljava/lang/Iterable;
 		ARG 1 entity
 		ARG 2 box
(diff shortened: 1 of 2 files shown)


liach:
hmm, might be better to unify this with isSpaceEmpty shall they act similarly (it takes a box instead)

apple502j:
isSpaceEmpty, afaik, checks for blocks and world borders as well. This one is just for entities.

liach (approved):
Consider renaming `Entity.inanimate` #2920 too, as that is strongly related to this. For instance, this will not check dropped items but will check boats, minecarts, living entities; you can test by putting minecart/marker and regular armor stand/item with a long despawn delay `{Item:{id:"apple",Count:1b},PickupDelay:3000}` into water and see the frost walker behavior of each of those.

enbrain (approved):

modmuss50 (approved):

PR #2909, opened 2021-12-09
Minor Block javadoc improvements

Juuxel:
- Added mention of `AbstractBlock` as it contains many of the useful block methods
- "logic" is uncountable (and thus singular) in this context
- Fixed broken `{@link}` to a member
- Fixed broken grammar ("the possible states ... ~~is~~ *are*")

### DIFF
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -1,6 +1,6 @@
 CLASS net/minecraft/class_2248 net/minecraft/block/Block
 	COMMENT A block is a voxel in a {@linkplain World world}. {@link AbstractBlock},
-	COMMENT this class and its subclasses define all logic for those voxels.
+	COMMENT this class, and its subclasses define all logic for those voxels.
 	COMMENT
 	COMMENT <p>There is exactly one instance for every type of block. Every stone
 	COMMENT block for example in a world shares the same block instance. Each block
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/block/Block.mapping
```diff
@@ -1,6 +1,6 @@
 CLASS net/minecraft/class_2248 net/minecraft/block/Block
-	COMMENT A block is a voxel in a {@linkplain World world}.
-	COMMENT This class and its subclasses define all logics for those voxels.
+	COMMENT A block is a voxel in a {@linkplain World world}. {@link AbstractBlock},
+	COMMENT this class and its subclasses define all logic for those voxels.
```
haykam821:
```suggestion
	COMMENT this class, and its subclasses define all logic for those voxels.
```

Juuxel: It [works both with and without](https://en.wikipedia.org/wiki/Serial_comma) (unless you follow a specific style guide, which we don't).

Shnupbups: Personally I prefer the Oxford comma to be present

Shnupbups (approved):

mappings/net/minecraft/block/Block.mapping
liach: think american english requires the comma before and unless it's only two items, as seen in americn college test

Juuxel: It's Oxfordised now.

haykam821 (approved):

liach (approved):

enbrain (approved):

modmuss50 (approved):

PR #2943, opened 2021-12-20
Screen#{isPauseScreen -> shouldPause}

nnym:
1. The current name implies that whether the screen pauses the game is a permanent property; which it is not (it is checked every render tick).
2. We know that it's a screen.


haykam821 (approved):

Shnupbups (approved):

enbrain (approved):

Juuxel (approved):
I think method is used quite a lot in different mods, but this shouldn't be impactful for most mods. (So this falls into sort of semi-impactful refactors, I guess...) A good rename in any case.

modmuss50:
Using github search this does seem quite an impacful rename. More than others I have suggested holding back.

liach:
Is the definition of a "pause" pausing the integrated server?
I guess "render tick" is a frame render (like the frame in fps) than a client tick. Need to find time to check these to review.

Juuxel:
> Is the definition of a "pause" pausing the integrated server?

Yes, it pauses the game in singleplayer. (Probably not in LAN games even though it uses the integ server)

modmuss50 (approved):

PR #2948, opened 2021-12-29
Add mc version id into jar manifest

modmuss50:
To be used by loom to warn when the mc version does not match, right now it tries (and fails) to parse the loom version.


Juuxel (approved):

liach:
can we just do it for all jars produced (including decompiled code and javadoc) so we won't miss new jars' manifests in case we need but forgot to add?

YanisBft (approved):

PR #2950, opened 2021-12-30
Hoppers are not an example of SidedInventory

reoseah:
…because they are not sided inventories

### DIFF
--- a/mappings/net/minecraft/inventory/SidedInventory.mapping
+++ b/mappings/net/minecraft/inventory/SidedInventory.mapping
@@ -1,5 +1,5 @@
 CLASS net/minecraft/class_1278 net/minecraft/inventory/SidedInventory
-	COMMENT A special inventory interface for inventories that expose different slots for different sides, such as hoppers.
+	COMMENT A special inventory interface for inventories that expose different slots for different sides, such as furnaces.
 	METHOD method_5492 canInsert (ILnet/minecraft/class_1799;Lnet/minecraft/class_2350;)Z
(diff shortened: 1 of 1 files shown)


haykam821 (approved):

Shnupbups (approved):

liach:
Hmm, should we just remove "such as" clauses if they are available in subclasses? In IDE, you can navigate easily; in html javadoc, it offers all known (vanilla) implementations of an interface and direct subclasses. Those should be more reliable.

reoseah:
I think that class comment is fine as it, with "hopper" changed to "furnace". Everyone knows how furnace works with hoppers, so that's a good example.

What's not fine are comments on methods inside it, they are just rephrasing method name and parameter types:
```java
    /**
     * Gets the available slot positions that are reachable from a given side.
     */
    public int[] getAvailableSlots(Direction var1);

    /**
     * Determines whether the given stack can be inserted into this inventory at the specified slot position from the given direction.
     */
    public boolean canInsert(int var1, ItemStack var2, @Nullable Direction var3);

    /**
     * Determines whether the given stack can be removed from this inventory at the specified slot position from the given direction.
     */
    public boolean canExtract(int var1, ItemStack var2, Direction var3);
```

I would just remove these, they are purely tautological. But that's outside of the scope of this PR ;3

On `getAvailableSlots` comment could mention that it's not just *available* slots, but also an *order* in which hopers and most other users would try to insert/extract items.

liach (approved):

enbrain (approved):

PR #2952, opened 2022-01-01
Document what deprecated means in AbstractBlock

enbrain:


Shnupbups (approved):

Shnupbups:
How feasible would it be to replace the `@Deprecated` annotation entirely with `@OverrideOnly`? Outside the scope of this PR, sure, but another potential idea.

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -1,4 +1,14 @@
 CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
+	COMMENT An abstract class that defines some logic for {@link Block blocks}.
+	COMMENT
+	COMMENT <p>Deprecated methods in this class mean they should only be called
```
```suggestion
	COMMENT <p id="deprecated-methods">Deprecated methods in this class mean they should only be called
```

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -15,22 +25,27 @@ CLASS net/minecraft/class_4970 net/minecraft/block/AbstractBlock
 		ARG 1 settings
 	METHOD method_16841 getOffsetType ()Lnet/minecraft/class_4970$class_2250;
 	METHOD method_17454 createScreenHandlerFactory (Lnet/minecraft/class_2680;Lnet/minecraft/class_1937;Lnet/minecraft/class_2338;)Lnet/minecraft/class_3908;
+		COMMENT @deprecated Consider calling {@link AbstractBlockState#createScreenHandlerFactory} instead. See {@linkplain AbstractBlock the class javadoc}.
```
```suggestion
		COMMENT @deprecated Consider calling {@link AbstractBlockState#createScreenHandlerFactory} instead. See <a href="#deprecated-methods">the class javadoc</a>.
```
I recommend such javadoc (which JDK uses), linking to the exact part we need. Same for other methods.

liach (approved):

liach:
> How feasible would it be to replace the `@Deprecated` annotation entirely with `@OverrideOnly`? Outside the scope of this PR, sure, but another potential idea.

if we do, we need it to be opt-in, as mojang has other deprecations in like mathhelper with different puroses

YanisBft (approved):

Shnupbups:
> > How feasible would it be to replace the `@Deprecated` annotation entirely with `@OverrideOnly`? Outside the scope of this PR, sure, but another potential idea.
> 
> if we do, we need it to be opt-in, as mojang has other deprecations in like mathhelper with different puroses

Oh definitely it would be on a per case basis

modmuss50:
Yes, I would like to be able to remove and add annotations (Stuff like JB's contract might be nice to use). 

It could work similar to unpick, have a file that defines the modifications that gets included in the extras dir in release jar.

modmuss50 (approved):

PR #2961, opened 2022-01-17
Map the RedstoneConstants class

haykam821:
This is likely to be a class holding redstone constants for the following reasons:

- Redstone is one of the two major mechanics that could be considered large enough to have its own constants class, the other being light
- The `LightingProvider#field_31713` field already contains a max light constant
- Light does not have a prevalent concept of a default level, unlike block state properties
- The package is ordered between portals and state


Shnupbups (approved):

liach:
again, why wouldn't this be light constants (min level, max level, default block level) or chunk (size, offset, etc) constants?

mappings/net/minecraft/class_6148.mapping
```diff
@@ -1,7 +0,0 @@
-CLASS net/minecraft/class_6148
```
modmuss50, line 1: I think this should just be moved into the existing unused package next to where all of the package info's live.

modmuss50 (changes requested):

mappings/net/minecraft/class_6148.mapping
haykam821: Should the class keep these new mappings as well?

liach: Unless there is definitive proof that these are and only are constants for redstone values (also two of the fields have same value, dunno how you can tell them apart), they should not keep the mappings and the class name you give it.

Maybe something like the original https://github.com/liach/yarn/blob/83ec0c0aff1f3a94693e4333eb4ba04e6e490878/mappings/net/minecraft/unused/UnknownWorldConstants6148.mapping

gudenau:
I don't have an issue with this myself, it seems fairly logical and changing this stuff in the future will only break source compatibility due to constant inlining.

PR #2965, opened 2022-01-17
Rename Panda flag from 'scared' to 'sitting'

Machine-Maker:
When you send the entity metadata packet with the appropriate update to the client
the panda's behavior can be better described as sitting up rather than being scared

### DIFF
--- a/mappings/net/minecraft/entity/passive/PandaEntity.mapping
+++ b/mappings/net/minecraft/entity/passive/PandaEntity.mapping
@@ -5,7 +5,7 @@ CLASS net/minecraft/class_1440 net/minecraft/entity/passive/PandaEntity
 	FIELD field_30343 MAIN_GENE_MUTATION_CHANCE I
 	FIELD field_30344 SNEEZING_FLAG I
 	FIELD field_30345 PLAYING_FLAG I
-	FIELD field_30346 SCARED_FLAG I
+	FIELD field_30346 SITTING_FLAG I
 	FIELD field_30347 LYING_ON_BACK_FLAG I
 	FIELD field_30348 EATING_ANIMATION_INTERVAL I
 	FIELD field_30349 HIDDEN_GENE_MUTATION_CHANCE I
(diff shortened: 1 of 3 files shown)


Machine-Maker:
@haykam821 Why do you think this shouldn't be changed?

apple502j:
Current name seems like a bug. 
image

mappings/net/minecraft/entity/passive/PandaEntity.mapping
```diff
@@ -63,7 +63,7 @@ CLASS net/minecraft/class_1440 net/minecraft/entity/passive/PandaEntity
 		ARG 1 bitmask
 	METHOD method_6534 getScaredAnimationProgress (F)F
```
apple502j: I believe this method (plus a few that mention "scared", except `isScaredByThunderstorm`) also needs a change.

apple502j (changes requested):

Machine-Maker:
Ok, updated other methods and fields with the `sitting` moniker instead of `scared` (except for isScaredByThunderstorm)

mappings/net/minecraft/entity/passive/PandaEntity.mapping
```diff
@@ -63,15 +63,15 @@ CLASS net/minecraft/class_1440 net/minecraft/entity/passive/PandaEntity
...
```
apple502j: This is still not fixed?

Machine-Maker: yeah... not sure how I missed that. Fixed now. Only reference to pandas and "scared" should be the "isScaredByThunderstorm" method now.

apple502j:
@Machine-Maker Also, can you fix unpick-definitions/entity_flags.unpick‎ to reflect the rename? Should be easy, just change to SITTING

Machine-Maker:
Done @apple502j.

apple502j (approved):

modmuss50 (approved):

PR #2974, opened 2022-01-20
Rename the Texture class to TextureMap

haykam821:
This class represents a map of texture keys to identifiers rather than a single texture.


Shnupbups (approved):

liach:
hmm, what's the correspondence in the client.model package? might worth documenting

apple502j:
@liach ~~I believe it's `client.texture.AbstractTexture` and its subclasses.~~ wait no, that's for individual texture

Juuxel (approved):

modmuss50 (approved):

PR #2978, opened 2022-01-22
Fix some setters not matching fields and params

enbrain:


liach:
done with a program? make these renames have no conflicts and for things like setCurrentStack, don't think you should rename the param from stack to currentStack

enbrain:
> done with a program? make these renames have no conflicts

I first obtained the list of non-matching setters by these two regular expressions.

* `\sset([A-Za-z]+)\([A-Za-z]+\s([A-Za-z]+)\)\s*\{\n\s*this\.(?!\1)([A-Za-z])+\s=\s\2;\n\s*\}` for mismatched method and field
* `\sset([A-Za-z]+)\([A-Za-z]+\s(?!\1)([A-Za-z]+)\)\s*\{\n\s*this\.([A-Za-z])+\s=\s\2;\n\s*\}` for mismatched method and param

After that, I made changes manually using Enigma GUI. Is there any task to make extra sure there are no conflicts?

> for things like setCurrentStack, don't think you should rename the param from stack to currentStack

Sure, I will revert them.

liach:
Ah, I thought you just edited the mapping files directly. If enigma doesn't reject a rename, then a rename is fine.

enbrain:
I re-cloned the repository and confirmed that Enigma doesn't reject any of the renames just in case I accidentally edited mapping files directly. Fyi, if I intentionally make a conflict and try to run the `build` task, it fails at https://github.com/FabricMC/stitch/blob/22a2dfdeb6a7570596e4184aed3f8b6bcfeb894c/src/main/java/net/fabricmc/stitch/commands/CommandMergeTiny.java#L90.

Let me know if there are still bad renames in this PR.

mappings/net/minecraft/block/entity/LootableContainerBlockEntity.mapping
```diff
@@ -3,9 +3,9 @@ CLASS net/minecraft/class_2621 net/minecraft/block/entity/LootableContainerBlock
 	FIELD field_12037 lootTableId Lnet/minecraft/class_2960;
 	FIELD field_31352 LOOT_TABLE_SEED_KEY Ljava/lang/String;
 	FIELD field_31353 LOOT_TABLE_KEY Ljava/lang/String;
-	METHOD method_11281 setInvStackList (Lnet/minecraft/class_2371;)V
-		ARG 1 list
-	METHOD method_11282 getInvStackList ()Lnet/minecraft/class_2371;
+	METHOD method_11281 setInventory (Lnet/minecraft/class_2371;)V
+		ARG 1 inventory
+	METHOD method_11282 getInventory ()Lnet/minecraft/class_2371;
```
Juuxel: Not sure if I'd call this "inventory" since it's just a `DefaultedList<ItemStack>` used to implement the `Inventory` (which is the `LootableContainerBlockEntity` itself)

enbrain: I reverted it for now. Maybe I'll look into it someday.

modmuss50 (approved):

apple502j (approved):

Issue #2525, opened 2021-06-27
Screen.removed -> onClose


Juuxel:
See also #2524 for the current method called `onClose`.

It's called when a screen is changed and the old screen is thrown away. I feel like the word "close" is more accurate (or at least more descriptive) than "remove" here.

(Side note: This might be one of those old unobfuscated names that were ported to Yarn.)

PR #2987, opened 2022-01-24
Rename Screen#removed and Screen#onClose

### LINKED ISSUE #2524: Screen.onClose -> close
It actually also closes the screen by calling `client.openScreen(null)`.

NoahvdAa:
Resolves #2524 ~~and #2525~~


Juuxel:
Doing both of these renames in the PR has a chance for breakage where a mod compiles just fine but doesn't actually work anymore because they called `onClose` (now `close`) but it actually calls the new `onClose` (previously `removed`). IMO the `onClose` -> `close` one is a better one for accuracy. (Closing a screen does still "remove" it from being shown)

NoahvdAa:
Thanks, reverted the `onClose` rename. I assumed that would be taken care of by migrateMappings, but now that I actually think about it, that doesn't really make sense.

Juuxel (approved):

modmuss50 (approved):

PR #2990, opened 2022-01-28
Pattern name BACKSLASH_S_PATTERN does not indicate what the pattern does

CyborgCabbage:
The pattern BACKSLASH_S_PATTERN is very literally named after the regex. This is confusing as the pattern isn't detecting any backslashes or s, it detects white-space. I have therefore renamed it WHITESPACE_PATTERN.

This also made clear that the function getLastPlayerNameStart just gets the first character of the current word and doesn't have anything to do with player names specifically, so I have renamed that also.

Forgive me if I have done anything wrong, this is my first pull request.


enbrain (approved):

liach (approved):
Yeah that name was a fault on my end. See https://github.com/FabricMC/yarn/commit/a867725a7ffb22c82faef53dec46c4ef7605e36e#diff-010a3344a5274524edfa65c5ccf0eea4ffa19ae914085238acf1aba63c47d3bbL36

haykam821 (approved):

PR #2991, opened 2022-01-28
Rename Fertilizable to BoneMealable

SquidCooki2:
When a player refers to the act of using bone meal in-game, they would say:
"Let's bone meal that wheat field," instead of, "Let's _fertilize_ that wheat field."

This may also cause confusion with the fertilization/pollination feature of bees.


haykam821 (changes requested):
Personally, I'm not a fan of this refactor. Fertilization is unlikely to be confused with pollination by bees.

In addition, some names aren't changed, such as `BoneMealItem.useOnFertilizable` or `WorldEvents.PLANT_FERTILIZED` (and its documentation).

PR #2995, opened 2022-02-09
Plz explain

apple502j:
I need sleep, this PR is very WIP

PR #2998, opened 2022-02-10
22w06a resources

liach:
So now:
 - Resource reloads are handled by `SimpleResourceReload#start` than by the managers, much cleaner
 - The old resource manager impl probably stayed the same; now named `LifecycledResourceManagerImpl` to show it should be discarded when `close()` is called. Good for programming habits
 - The `ReloadableResourceManagerImpl` now keeps a backing lifecycled manager that actually supply the resources, like the tag manager/factory etc.
   - `close()` it is different from closing the lifecycled one, as it only replaces the backing manager with an empty one (like initial state)
   - I am tempted to rename this to `StaticTagManager`, maybe later
   - This is only used by the client; the server has a similar record structure, not mapped by this pr (as it touches dynamic registries)
 - Reloaders are now registered to `ReloadableResourceManagerImpl.addReloader` and the return value of `ServerResourceManager.getReloaders()`, the latter one needs some mixin. Easier to mod in general.

------------
For the task is daunting, we will start with small ones. ~~Conflicts with #2995, where it maps my `LifecycledResourceManager` and impl to `AutoCloseableResourceManager` `CombinedResourceManager` (not accurate IMO)~~

### DIFF
--- a/mappings/net/minecraft/resource/LifecycledResourceManagerImpl.mapping
+++ b/mappings/net/minecraft/resource/LifecycledResourceManagerImpl.mapping
@@ -4,7 +4,7 @@ CLASS net/minecraft/class_6861 net/minecraft/resource/LifecycledResourceManagerI
 	COMMENT <p>It handles resources by namespaces, hoping that most namespaces are
 	COMMENT defined in only few resource packs.
 	COMMENT
-	COMMENT @see NamespacedResourceManager
+	COMMENT @see NamespaceResourceManager
 	FIELD field_36389 subManagers Ljava/util/Map;
 	FIELD field_36390 packs Ljava/util/List;
 	METHOD <init> (Lnet/minecraft/class_3264;Ljava/util/List;)V
--- a/mappings/net/minecraft/resource/LifecycledResourceManagerImpl.mapping
+++ b/mappings/net/minecraft/resource/LifecycledResourceManagerImpl.mapping
@@ -1,0 +1,14 @@
+CLASS net/minecraft/class_6861 net/minecraft/resource/LifecycledResourceManagerImpl
+	COMMENT A basic implementation of resource manager with a lifecycle.
+	COMMENT
+	COMMENT <p>It handles resources by namespaces, hoping that most namespaces are
+	COMMENT defined in only few resource packs.
+	COMMENT
+	COMMENT @see NamespacedResourceManager
+	FIELD field_36389 subManagers Ljava/util/Map;
+	FIELD field_36390 packs Ljava/util/List;
+	METHOD <init> (Lnet/minecraft/class_3264;Ljava/util/List;)V
+		ARG 1 type
+		ARG 2 packs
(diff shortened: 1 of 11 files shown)


apple502j:
Your mappings are clearly better than mine so :)

apple502j (approved):

mappings/net/minecraft/resource/NamespaceResourceManager.mapping
```diff
@@ -1,4 +1,10 @@
 CLASS net/minecraft/class_3294 net/minecraft/resource/NamespaceResourceManager
+	COMMENT A resource manager that only loads resources for a specific namespace.
+	COMMENT This is an implementation detail for {@link ResourceManagerImpl}, based
+	COMMENT on the heuristic that most namespaces are only defined in few resource
+	COMMENT packs, making loading in those namespaces faster.
+	COMMENT
+	COMMENT @see ResourceManagerImpl
```
enbrain: Did you mean `LifecycledResourceManagerImpl`? I can't find `ResourceManagerImpl`.

mappings/net/minecraft/resource/LifecycledResourceManagerImpl.mapping
```diff
@@ -0,0 +1,14 @@
+CLASS net/minecraft/class_6861 net/minecraft/resource/LifecycledResourceManagerImpl
+	COMMENT A basic implementation of resource manager with a lifecycle.
+	COMMENT
+	COMMENT <p>It handles resources by namespaces, hoping that most namespaces are
+	COMMENT defined in only few resource packs.
+	COMMENT
+	COMMENT @see NamespacedResourceManager
```
enbrain: `NamespaceResourceManager`?

mappings/net/minecraft/resource/NamespaceResourceManager.mapping
liach:
```suggestion
	COMMENT @see LifecycledResourceManagerImpl
```

mappings/net/minecraft/resource/LifecycledResourceManagerImpl.mapping
liach:
```suggestion
	COMMENT @see NamespaceResourceManager
```

liach:
fyi @enbrain I need yet to fix javadoc gen, so couldn't detect those problems by eye

mappings/net/minecraft/resource/NamespaceResourceManager.mapping
```diff
@@ -1,4 +1,10 @@
 CLASS net/minecraft/class_3294 net/minecraft/resource/NamespaceResourceManager
+	COMMENT A resource manager that only loads resources for a specific namespace.
+	COMMENT This is an implementation detail for {@link ResourceManagerImpl}, based
```
enbrain: Same here.
```suggestion
	COMMENT This is an implementation detail for {@link LifecycledResourceManagerImpl}, based
```

enbrain (approved):

PR #2999, opened 2022-02-10
22w06a non-cursed stuff

apple502j:
Regional compliancies, etc. 6904 is considered cursed and is not mapped here.


liach (approved):
looks good on web

Gegy (approved):

modmuss50 (approved):

PR #3004, opened 2022-02-11
22w06a more cursed stuff

apple502j:
Mapped three cursed things - registry, tags, and data fixer. Here I explain the names. I've tried to be as accurate as I can, but the names are pretty inconsistent to say the least. (See below)

# Per-class docs (didn't bother to javadocify)
## TagPacketSerializer
`TagPacketSerializer` is used to serialize packets in networking. It doesn't serialize packets itself, but since the result is used in a packet this name should be fine. I did not name the deserializer `deserializeTags` - probably because of 1) the terminology "load" e.g. `EntryLoader` and 2) I believe it has a side effect. The closest one we have is "read", but we don't use "read" in static context if I remember correctly.

5748 docs is kept from older version.

## Codecs
Some names are stolen from exposed names. `rawIdChecked` checks if the raw ID is valid (`elementToRawId` returns `errorRawId` when invalid). `createNonEmptyEntryListChecker`/`nonEmptyEntryList` is like the existing one except for `RegistryEntryList`. `createEqualTypeChecker` creates a function that checks whether a collection's values are of the same type.

## ForwardingDynamicOps
Just named params. I should've done this in separate PR.

## RegistryOps
This is almost the same as 1.18.1 `RegistryOps`, but some code is moved to `RegistryLoader`.

## RegistryLoader
A wrapper of `EntryLoader`, iirc. `LoaderAccess` holds `EntryLoader` and DRM.

## DynamicRegistryManager
Ah yes. DRM. It's like a Bedrock Wither.

Terminologies:

- `Static` refers to normal (non-dynamic) registry.
- `Managed` refers to registries managed by `this`/the DRM.
- `Mutable` refers to `MutableRegistry`.
- `NonMutable` refers to `Registry` that is not `MutableRegistry`.
- Non-prefixed methods return registry from both Managed and Static. In some methods the prefix `All` is used to clarify the scope. (e.g. `streamAllRegistries`)
- `Synced` is self-explanatory.
- `createAndLoad` loads entries using `EntryLoader`, and `addEntriesToLoad` adds entries which will be loaded later.
- `DRM.Entry` is an entry of a `RegistryKey` of a registry and a registry.
- `MutableRegistryManager` is a manager of `MutableRegistry`. DRM itself is not mutable.
- `NonMutableRegistryManager` is the opposite.

## Registry
`streamEntryLists` renamed because "entry list" is used for different things. (It streams an entry with a tag as a key and tag values as a value.)

## RegistryCodecs
`Codecs` but for registries. (Should this be in dynamic package or registry package?) `createRegistryEntryListCodec` is self-explanatory. `createRegistryDecoder` creates a decoder that loads entries with `EntryLoader`. It also has its own record for some reason, which I kept the old `RegistryManagerEntry` naming.

## RegistryEntryListCodec
`entryListCodec` was renamed to `directEntryListCodec` to clarify it is a codec for `Direct` variant only. `entryListStorageCodec` is a codec for `RegistryEntryList#getStorage`. `alwaysSerializeAsList` in `createDirectEntryListCodec`, if true, will cause it to encode `[A]` as `[A]` instead of `A` and vice versa.

## RegistryKey
See javadoc and enbrain's review

# TODO
There are some inconsistencies to look at:

- How to name `RegistryKey<? extends Registry<E>>` - `registry`, `registryKey`, `registryRef` (used in most cases in the PR)
- `RegistryKey<E>` is named `key` or `registryKey` in almost all cases, but what's the corresponding name for a value? Some calls it `value`. Exposed mojang name calls it `element`. Some calls it `entry`.
- `Entry` should be reserved for `RegistryEntry` (and `DRM.Entry`), but some code does not follow this rule.
- `List<RegistryEntry<E>>` vs `RegistryEntryList<E>` - I have named the former `entries` and the latter `entryList` in many cases, or the former `directEntryList` in some cases where it was obvious it was dealing with serialized `RegistryEntryList.Direct<E>`


mappings/net/minecraft/util/registry/RegistryEntryListCodec.mapping
```diff
@@ -21,8 +21,8 @@ CLASS net/minecraft/class_6898 net/minecraft/util/registry/RegistryEntryListCode
 	METHOD method_40382 createDirectEntryListCodec (Lcom/mojang/serialization/Codec;Z)Lcom/mojang/serialization/Codec;
 		ARG 0 entryCodec
 		ARG 1 alwaysSerializeAsList
-			COMMENT If {@code false}, a list with 1 item will be serialized
-			COMMENT as that item, instead of the list.
+			COMMENT whetehr to always serialize the list as a list
```
Shnupbups:
```suggestion
			COMMENT whether to always serialize the list as a list
```

mappings/net/minecraft/util/registry/RegistryKey.mapping
```diff
@@ -55,3 +57,5 @@ CLASS net/minecraft/class_5321 net/minecraft/util/registry/RegistryKey
 		ARG 0 registry
 	METHOD method_39155 (Lnet/minecraft/class_5321;Lnet/minecraft/class_2960;)Lnet/minecraft/class_5321;
 		ARG 1 id
+	METHOD method_39752 matchingRegistry (Lnet/minecraft/class_5321;)Ljava/util/Optional;
```
haykam821: This feels strange, but it should be fine. I would be inclined to name it `ifMatchingRegistry` if that did not imply too heavily that it would act like `Optional.ifPresent`.

mappings/net/minecraft/util/registry/RegistryEntryListCodec.mapping
```diff
@@ -1,9 +1,44 @@
...
+		ARG 3 prefix
+	METHOD method_40380 (Lcom/mojang/datafixers/util/Either;)Ljava/util/List;
+		ARG 0 either
+	METHOD method_40381 (Lcom/mojang/datafixers/util/Pair;)Lcom/mojang/serialization/DataResult;
+		ARG 0 pair
+	METHOD method_40382 createDirectEntryListCodec (Lcom/mojang/serialization/Codec;Z)Lcom/mojang/serialization/Codec;
 		ARG 0 entryCodec
+		ARG 1 alwaysSerializeAsList
+			COMMENT whether to always serialize the list as a list
+			COMMENT instead of serializing as one entry if the length is {@code 0}
```
haykam821, line 27: Could this comment be added to other similar `alwaysSerializeAsList` parameters?

mappings/net/minecraft/util/registry/RegistryKey.mapping
liach: Notice `ifPresent` is a functional method that only executes the given consumer conditionally; this one is more like `filter` `map` of `Optional`, and IMO the current name is better.

mappings/net/minecraft/util/registry/DynamicRegistryManager.mapping
```diff
@@ -22,12 +26,15 @@ CLASS net/minecraft/class_5455 net/minecraft/util/registry/DynamicRegistryManage
...
+		ARG 0 ops
+		ARG 1 loaderAccess
+		ARG 2 info
+	METHOD method_30527 getOptionalManaged (Lnet/minecraft/class_5321;)Ljava/util/Optional;
 		COMMENT Retrieves a registry optionally from this manager.
 		ARG 1 key
 	METHOD method_30530 get (Lnet/minecraft/class_5321;)Lnet/minecraft/class_2378;
-		COMMENT Retrieves a registry from this manager, or throws an exception when the
-		COMMENT registry does not exist.
+		COMMENT Retrieves a registry from this manager or {@link Registry.REGISTRIES},
```
enbrain:
```suggestion
		COMMENT Retrieves a registry from this manager or {@link Registry#REGISTRIES},
```

mappings/net/minecraft/util/registry/DynamicRegistryManager.mapping
```diff
@@ -62,3 +116,34 @@ CLASS net/minecraft/class_5455 net/minecraft/util/registry/DynamicRegistryManage
 		METHOD comp_294 entryCodec ()Lcom/mojang/serialization/Codec;
 		METHOD comp_295 networkEntryCodec ()Lcom/mojang/serialization/Codec;
 		METHOD method_30537 isSynced ()Z
+	CLASS class_6890 NonMutableRegistryManager
```
enbrain: I think `Immutable` is more common than `NonMutable` and saves 1 character.

mappings/net/minecraft/util/registry/DynamicRegistryManager.mapping
```diff
@@ -62,3 +116,34 @@ CLASS net/minecraft/class_5455 net/minecraft/util/registry/DynamicRegistryManage
...
+	CLASS class_6892 Entry
+		METHOD method_40329 of (Lnet/minecraft/class_6880$class_6883;)Lnet/minecraft/class_5455$class_6892;
+			ARG 0 entry
+		METHOD method_40330 of (Ljava/util/Map$Entry;)Lnet/minecraft/class_5455$class_6892;
+			ARG 0 entry
+		METHOD method_40331 of (Lnet/minecraft/class_5321;Lnet/minecraft/class_2378;)Lnet/minecraft/class_5455$class_6892;
+			ARG 0 key
+			ARG 1 value
+		METHOD method_40332 freeze ()Lnet/minecraft/class_5455$class_6892;
+	CLASS class_6893 MutableRegistryManager
```
enbrain: `DynamicRegistryManager.MutableRegistryManager` is verbose imo. Just `Mutable` (like `BlockPos.Mutable`)? I don't think it's confusing with `MutableRegistry` if that's your concern.

mappings/net/minecraft/util/dynamic/LoaderBoundDynamicOps.mapping
```diff
@@ -0,0 +1,29 @@
+CLASS net/minecraft/class_6903 net/minecraft/util/dynamic/LoaderBoundDynamicOps
+	FIELD field_36488 loaderAccess Ljava/util/Optional;
+	FIELD field_36489 dynamicRegistryManager Lnet/minecraft/class_5455;
```
enbrain: The current yarn uses `registryManager` for the field names (such as `PlayerManager#registryManager`).

apple502j:
I'll rename a few stuff to match 1.18.1 names, so give me a sec...

mappings/net/minecraft/util/registry/RegistryCodecs.mapping
```diff
@@ -0,0 +1,74 @@
...
+		FIELD comp_290 key Lnet/minecraft/class_5321;
+		FIELD comp_291 rawId I
+		FIELD comp_292 value Ljava/lang/Object;
+		METHOD <init> (Lnet/minecraft/class_5321;ILjava/lang/Object;)V
+			ARG 1 key
+			ARG 2 rawId
+			ARG 3 entry
+		METHOD comp_290 key ()Lnet/minecraft/class_5321;
+		METHOD comp_291 rawId ()I
+		METHOD comp_292 value ()Ljava/lang/Object;
```
enbrain: Inconsistency between `entry` and `value`. (Is the constructor even mappable in Enigma?)

mappings/net/minecraft/util/registry/DynamicRegistryManager.mapping
```diff
@@ -36,17 +43,64 @@ CLASS net/minecraft/class_5455 net/minecraft/util/registry/DynamicRegistryManage
...
+		ARG 0 registryRef
+	METHOD method_40316 toNonMutable ()Lnet/minecraft/class_5455$class_6890;
+	METHOD method_40317 createSimpleRegistry (Lnet/minecraft/class_5321;)Lnet/minecraft/class_2385;
+		ARG 0 registryRef
+	METHOD method_40318 streamStaticRegistries ()Ljava/util/stream/Stream;
+	METHOD method_40319 getNetworkEntryCodec (Lnet/minecraft/class_5321;)Lcom/mojang/serialization/DataResult;
+		ARG 0 registryKey
+	METHOD method_40320 createCodec ()Lcom/mojang/serialization/Codec;
+	METHOD method_40321 streamSyncedManagedRegistries ()Ljava/util/stream/Stream;
+	METHOD method_40323 collectMutableRegistries ()Ljava/util/Map;
```
enbrain:
```suggestion
	METHOD method_40323 createMutableRegistries ()Ljava/util/Map;
```
This doesn't collect existing registries. I think `collect` is an implementation detail for creating empty mutable registries.

mappings/net/minecraft/util/registry/RegistryKey.mapping
enbrain: To me, it looks like a safe cast (I got an idea from the `Prism` optic, but `Optional` instead of `Either`). In its only one usage, `key.method_39752(ref)` casts from `RegistryKey<?>` to `RegistryKey<E>` if the `key` belongs to the `ref`, and yields `Optional.empty()` otherwise. Is `tryCast` a good name?

mappings/net/minecraft/util/dynamic/LoaderBoundDynamicOps.mapping
liach: imo this one is fine

mappings/net/minecraft/util/registry/RegistryKey.mapping
apple502j: Ahhh! That makes 100% sense.

mappings/net/minecraft/util/dynamic/LoaderBoundDynamicOps.mapping
apple502j: Renamed anyway :p

mappings/net/minecraft/util/registry/RegistryCodecs.mapping
apple502j: Hmm, weird. I'll remap it to `value`. `entry` is reserved for `RegistryEntry` imo

mappings/net/minecraft/util/registry/RegistryCodecs.mapping
```diff
@@ -0,0 +1,78 @@
+CLASS net/minecraft/class_6895 net/minecraft/util/registry/RegistryCodecs
+	METHOD method_40338 (Lnet/minecraft/class_2378;)Ljava/util/Map;
+		ARG 0 registry
+	METHOD method_40339 (Lnet/minecraft/class_2385;Lcom/mojang/serialization/Lifecycle;Lnet/minecraft/class_5321;Ljava/lang/Object;)V
+		ARG 2 key
+		ARG 3 value
+	METHOD method_40340 createRegistryEntryListCodec (Lnet/minecraft/class_5321;)Lcom/mojang/serialization/Codec;
```
Gegy: Very minor thought: shorter-style names might be nice here- e.g. `RegistryCodecs.entryList(...)` instead of `RegistryCodecs.createRegistryEntryListCodec(...)` where there is a bit of redundancy.
I imagine these are used quite often in record codec builders, so it might help to reduce noise? 🙂

Gegy (approved):

modmuss50 (approved):
Looks great at a glace, lets get this in before the next snapshot. Any problems please raise a PR or issue 👍

PR #3011, opened 2022-02-16
Fix broken javadoc jar

### LINKED ISSUE #2873: Use --ignore-source-errors option in javadoc
Per https://bugs.openjdk.java.net/browse/JDK-8268582 javadoc has such an option. Using this option to generate partial documentation is preferable to having blank content when Minecraft undergoes a big update and the fake source generated doesn't comply to the minimum compilation standards required by javadoc tool, such as package-private access, etc.

modmuss50:
- Build against JDK 18
- Update Gradle
- Update asm for unpick
- set ignore-source-errors as suggested in #2873

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,7 @@ jobs:
   build:
     strategy:
       matrix:
-        java: [17-jdk]
+        java: [17-jdk, 18-jdk]
     runs-on: ubuntu-20.04
     container:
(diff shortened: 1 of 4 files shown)


modmuss50:
Confirmed this fixes it, also seems to work on J17 (at least the version on docker atm) will stick to using JDK 18 as there is no reason no to. (You can still build and use yarn on J17)

PR #3021, opened 2022-02-18
Remove reference to wool

Machine-Maker:
The field name is `id`, the getter is `getId`, but the parameter name in the enum ctor is still `woolId` even tho that doesn't really have anything to do with wool still.


Shnupbups (approved):

modmuss50:
Uhh, was hoping update-base would rebase, I should make that a thing.

modmuss50 (approved):

PR #3026, opened 2022-02-19
Generate a build artifact consumable by loom.

modmuss50:
Pending on: https://github.com/FabricMC/fabric-loom/pull/599

This should make it a lot easier to try out yarn PRs via loom.


modmuss50:
Turns out you need to be authenticated with github to download build artifacts...

PR #3038, opened 2022-02-25
Map VeinCreator (class_6955)

NoahvdAa:
This class *seems* to be related to the creation of ore veins. Not sure about the parameter names, but what I have right now seems to be what they're used for.


mappings/net/minecraft/world/gen/VeinCreator.mapping
```diff
@@ -0,0 +1,21 @@
+CLASS net/minecraft/class_6955 net/minecraft/world/gen/VeinCreator
```
enbrain: I suggest `class_6955` → `OreVeinSampler` and `method_40548` → `create`. This way, `class_6955.method_40548` can be read as a method that returns a sampler (and that sampler samples a block state in a ore vein for a given position).

mappings/net/minecraft/world/gen/VeinCreator.mapping
```diff
@@ -0,0 +1,21 @@
...
+		ARG 0 veinToggle
+		ARG 1 veinRidged
+		ARG 2 veinGap
+		ARG 3 randomDeriver
+	CLASS class_6354 VeinType
+		FIELD field_33605 ore Lnet/minecraft/class_2680;
+		FIELD field_33606 stone Lnet/minecraft/class_2680;
+		FIELD field_33607 minY I
+		FIELD field_33608 maxY I
+		FIELD field_33668 rawBlock Lnet/minecraft/class_2680;
```
enbrain: I would be tempted to rename it `rawOreBlock` for clarity.

mappings/net/minecraft/world/gen/OreVeinSampler.mapping
```diff
@@ -0,0 +1,21 @@
+CLASS net/minecraft/class_6955 net/minecraft/world/gen/OreVeinSampler
+	FIELD field_36627 RAW_BLOCK_CHANCE F
```
enbrain:
```suggestion
	FIELD field_36627 RAW_ORE_BLOCK_CHANCE F
```

mappings/net/minecraft/world/gen/OreVeinSampler.mapping
```diff
@@ -0,0 +1,21 @@
...
+		ARG 0 veinToggle
+		ARG 1 veinRidged
+		ARG 2 veinGap
+		ARG 3 randomDeriver
+	CLASS class_6354 VeinType
+		FIELD field_33605 ore Lnet/minecraft/class_2680;
+		FIELD field_33606 stone Lnet/minecraft/class_2680;
+		FIELD field_33607 minY I
+		FIELD field_33608 maxY I
+		FIELD field_33668 rawBlock Lnet/minecraft/class_2680;
```
enbrain:
```suggestion
		FIELD field_33668 rawOreBlock Lnet/minecraft/class_2680;
```

mappings/net/minecraft/world/gen/OreVeinSampler.mapping
```diff
@@ -0,0 +1,21 @@
+CLASS net/minecraft/class_6955 net/minecraft/world/gen/OreVeinSampler
+	FIELD field_36627 RAW_BLOCK_CHANCE F
+	METHOD method_40547 (Lnet/minecraft/class_6910;Lnet/minecraft/class_2680;Lnet/minecraft/class_6574;Lnet/minecraft/class_6910;Lnet/minecraft/class_6910;Lnet/minecraft/class_6910$class_6912;)Lnet/minecraft/class_2680;
+		ARG 5 startPos
```
enbrain: Ore veins don't have a start position as far as I know. Instead, the chunk generator calls this lambda for every position not covered by the aquifer sampler. Notice that it returns `null` to use the fallback state/the chunk generator's default block state. Just `pos` or `noisePos`?

enbrain (changes requested):

enbrain (approved):

modmuss50 (approved):

PR #3039, opened 2022-03-03
Fix registry value being named key

Machine-Maker:
These represent a registry value, not the key

Needed `valuex` instead of `value` cause `value` is already defined in the scope (its in a computeIfAbsent call)


mappings/net/minecraft/util/registry/SimpleRegistry.mapping
```diff
@@ -43,7 +43,7 @@ CLASS net/minecraft/class_2370 net/minecraft/util/registry/SimpleRegistry
 	METHOD method_40267 (Lnet/minecraft/class_6880$class_6883;)Z
 		ARG 0 entry
 	METHOD method_40271 (Ljava/lang/Object;)Lnet/minecraft/class_6880$class_6883;
-		ARG 1 key
+		ARG 1 val
```
haykam821:
```suggestion
		ARG 1 valuex
```

haykam821 (changes requested):
Lambda parameters should be suffixed with `x` if they would otherwise conflict with other name.

mappings/net/minecraft/util/registry/SimpleRegistry.mapping
Machine-Maker: resolved

haykam821 (approved):

apple502j (approved):

liach (approved):

PR #3045, opened 2022-03-16
22w11a save loading

apple502j:
SaveLoading loads the saves, but the loaded save can be used for any purpose by what I named "save user". SaveUserFactory is implemented (as functional interface) by SaveLoader and a lambda that creates GenerationOptionsHolder.

GOH is an immutable object holding generator options, DRM, etc and is used client-side.

IntegratedServerLoader contains many of the logics copied from server.Main, plus some client-only things like backup screen creation.


mappings/net/minecraft/server/SaveLoading.mapping
```diff
@@ -0,0 +1,35 @@
...
+		ARG 2 throwable
+	METHOD method_42097 (Lnet/minecraft/class_5455$class_6890;Lnet/minecraft/class_7237$class_7239;Lnet/minecraft/class_6860;Ljava/lang/Object;Lnet/minecraft/class_5350;)Ljava/lang/Object;
+		ARG 4 dataPackContents
+	METHOD method_42098 load (Lnet/minecraft/class_7237$class_6906;Lnet/minecraft/class_7237$class_6907;Lnet/minecraft/class_7237$class_7239;Ljava/util/concurrent/Executor;Ljava/util/concurrent/Executor;)Ljava/util/concurrent/CompletableFuture;
+		ARG 0 functionLoaderConfig
+		ARG 1 savePropertiesSupplier
+		ARG 2 saveUserFactory
+		ARG 3 prepareExecutor
+		ARG 4 applyExecutor
+	CLASS class_6906 FunctionLoaderConfig
```
liach: More like `ServerConfig`, as this record is mainly used for server stuff than just for function stuff.

mappings/net/minecraft/server/SaveLoading.mapping
```diff
@@ -0,0 +1,35 @@
...
+		FIELD comp_361 commandEnvironment Lnet/minecraft/class_2170$class_5364;
+		FIELD comp_362 functionPermissionLevel I
+		FIELD comp_634 dataPacks Lnet/minecraft/class_7237$class_7238;
+		METHOD comp_361 commandEnvironment ()Lnet/minecraft/class_2170$class_5364;
+		METHOD comp_362 functionPermissionLevel ()I
+		METHOD comp_634 dataPacks ()Lnet/minecraft/class_7237$class_7238;
+	CLASS class_6907 SavePropertiesSupplier
```
liach: Looks like a load context supplier that supplies dynamic registries and the data D that will be fed to the apply block. "save properties" sounds somewhat like that class in level properties hierarchy.

mappings/net/minecraft/server/SaveLoading.mapping
```diff
@@ -0,0 +1,35 @@
...
+		METHOD get (Lnet/minecraft/class_3300;Lnet/minecraft/class_5359;)Lcom/mojang/datafixers/util/Pair;
+			ARG 1 resourceManager
+			ARG 2 dataPackSettings
+	CLASS class_7238 DataPacks
+		FIELD comp_635 manager Lnet/minecraft/class_3283;
+		FIELD comp_636 settings Lnet/minecraft/class_5359;
+		METHOD comp_635 manager ()Lnet/minecraft/class_3283;
+		METHOD comp_636 settings ()Lnet/minecraft/class_5359;
+		METHOD method_42099 load ()Lcom/mojang/datafixers/util/Pair;
+	CLASS class_7239 SaveUserFactory
```
liach: Can we call this `Applier` as "user" is somewhat reminiscent of the offline players?

mappings/net/minecraft/server/SaveLoader.mapping
```diff
@@ -6,4 +6,7 @@ CLASS net/minecraft/class_6904 net/minecraft/server/SaveLoader
 	METHOD comp_358 dynamicRegistryManager ()Lnet/minecraft/class_5455$class_6890;
 	METHOD comp_359 saveProperties ()Lnet/minecraft/class_5219;
 	METHOD method_40431 ofLoaded (Lnet/minecraft/class_7237$class_6906;Lnet/minecraft/class_7237$class_6907;Ljava/util/concurrent/Executor;Ljava/util/concurrent/Executor;)Ljava/util/concurrent/CompletableFuture;
```
liach: This is more accurately
```suggestion
	METHOD method_40431 load (Lnet/minecraft/class_7237$class_6906;Lnet/minecraft/class_7237$class_6907;Ljava/util/concurrent/Executor;Ljava/util/concurrent/Executor;)Ljava/util/concurrent/CompletableFuture;
```

mappings/net/minecraft/server/SaveLoading.mapping
apple502j: I think this name was from 1.18.2 when this was actually used to configure FunctionLoader. We can rename it, yeah.

mappings/net/minecraft/server/SaveLoading.mapping
apple502j: All vanilla usage use `SaveProperties` class as D, fyi (this is another name moved from 1.18.2 SaveLoader) - LoadContextSupplier is a good name, renaming.

mappings/net/minecraft/server/SaveLoading.mapping
apple502j: Sounds good

mappings/net/minecraft/server/SaveLoader.mapping
apple502j: Hmm. I remember this being a static method, checking if we used `load` in static context.

apple502j: Okay seems like `load` can be used both in static and non-static context (existing usage in resources etc). Renaming

liach: My rationale is that this works much like `SimpleResourceReload.create` where we return a future that has already been scheduled to the prepare and apply executors.

mappings/net/minecraft/server/SaveLoading.mapping
liach: Now `D` can be `GeneratorOptions` and `GeneratorOptionsHolder` as you've mentioned, `Pair<SavePropertiesLoader, DynamicRegistryManager.Immutable>`, and `SaveLoader`. Thanks for renaming!

liach (approved):

modmuss50 (approved):

PR #3046, opened 2022-03-17
22w11a registry

apple502j:
RegistryWrapper is a wrapper for a registry. Note that this is not a Registry. RegistryWrapperFactory is a factory class for the wrapper. This is currently used to control whether entry list creation should fail, succeed, or return empty one.


liach (approved):

apple502j:
Player suggested renaming this to "RegistryAccess". It is less impl-based; however given Mojang record names use this terminology to refer to DRM, I'm not sure if this is good (given this name will also be used in API.)

Also, it seems like current usage is focused on commands (the classes themselves can be used anywhere). Maybe we need to put it in a better package or rename the classes.

Player3324:
It is worth noting that it is RegistryAccess for commands specifically, not in general, which kind of detaches it from any other similar facilities elsewhere.

liach:
hmm, back in enigma, it appears to me to have a focus on controlling the tag lists. maybe we can rename it later if needed?

PR #3067, opened 2022-03-27
Chunk generator stuff

Machine-Maker:
got the structureOverrides name from the codec for the various chunk generators, field name was `structure_overrides`.

other names were just copied from the type of the field/param

### DIFF
--- a/mappings/net/minecraft/world/gen/chunk/ChunkGenerator.mapping
+++ b/mappings/net/minecraft/world/gen/chunk/ChunkGenerator.mapping
@@ -142,7 +142,7 @@ CLASS net/minecraft/class_2794 net/minecraft/world/gen/chunk/ChunkGenerator
 	METHOD method_40450 getDebugHudText (Ljava/util/List;Lnet/minecraft/class_7138;Lnet/minecraft/class_2338;)V
 		ARG 1 text
 	METHOD method_41039 streamStructureSets ()Ljava/util/stream/Stream;
-	METHOD method_41042 getStructuresRegistryForCodec (Lcom/mojang/serialization/codecs/RecordCodecBuilder$Instance;)Lcom/mojang/datafixers/Products$P1;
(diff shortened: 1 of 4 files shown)


Player3324:
There are some merge conflicts from an older PR. I believe structureRegistry is better than structuresRegistry by not having the extraneous plural. There are other conflicts where either name may or may not be better, please have a look.

Machine-Maker:
@sfPlayer1 addressed those issues I think. I also changed the references to `stronghold` in ChunkGenerator. those concentric ring positions can be used for custom structure sets in data packs and aren't just for strongholds

modmuss50 (approved):

mappings/net/minecraft/world/gen/chunk/ChunkGenerator.mapping
```diff
@@ -141,6 +142,8 @@ CLASS net/minecraft/class_2794 net/minecraft/world/gen/chunk/ChunkGenerator
 	METHOD method_40450 getDebugHudText (Ljava/util/List;Lnet/minecraft/class_7138;Lnet/minecraft/class_2338;)V
 		ARG 1 text
 	METHOD method_41039 streamStructureSets ()Ljava/util/stream/Stream;
+	METHOD method_41042 getStructuresRegistryForCodec (Lcom/mojang/serialization/codecs/RecordCodecBuilder$Instance;)Lcom/mojang/datafixers/Products$P1;
```
enbrain: Since this returns a record codec builder that defines how to get a structure set registry from the chunk generator,
```suggestion
	METHOD method_41042 createStructureSetRegistryGetter (Lcom/mojang/serialization/codecs/RecordCodecBuilder$Instance;)Lcom/mojang/datafixers/Products$P1;
```

enbrain (commented):
Looks good besides one method.

mappings/net/minecraft/world/gen/chunk/ChunkGenerator.mapping
Machine-Maker: yeah, I like that better. I wasn't 100% on this one for sure.

PR #3068, opened 2022-03-28
rename pauseWhenMobIdle to skipMobIdleCheck

SFort:
The variable is used twice
```
if (!pauseWhenMobIdle) {
    //stops when idle
} else if {
    //other checks which get skipped
}
```
```
if (pauseWhenMobIdle || mob.canSee(player)){
    //Mob moves twords player 
}
```
most entities appear to have it set to false with the Ravager having it set to true

due to the second use i don't think `skipMobIdleCheck` is entierly accurate but i also can't think of something that makes sense for both


frqnny (approved):

modmuss50:
Can we make sure this change happens in the snapshot branch first, then it can be backported as required.

SFort:
i take it the snapshot branch is the main one? (and should i open a diffrent pr for it, or no need?)

PR #3086, opened 2022-04-08
22w14a Mangrove mappings

DuffyScottC:
Adding a mapping for the MangroveSaplingGenerator and related fields.


mappings/net/minecraft/block/sapling/MangroveSaplingGenerator.mapping
```diff
@@ -0,0 +1,4 @@
+CLASS net/minecraft/class_7382 net/minecraft/block/sapling/MangroveSaplingGenerator
+    FIELD field_38758  randomFloat
```
liach, line 2: And this piece of mapping is incorrect format-wise.

liach (commented):
Sorry, but your mappings have been covered by others. Please use a legitimate tool like enigma to create mappings or at least build and validate them.

PR #3093, opened 2022-04-15
Update deps

modmuss50:
Update deps.
Move to spotless as the licenser plugin has deprecation warnings on gradle 7.4
Fix some line endings.


apple502j (approved):
thanks for fixing crlf mess

PR #3102, opened 2022-04-20
Partial mappings of new text api

0x3C50:
Thought I'd do this, since the new api seems kind of confusing

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -64,11 +64,11 @@ configurations {
 	javadocClasspath
 	decompileClasspath
 	mappingPoetJar {
-		extendsFrom asm
 		transitive = false
 	}
 	mappingPoet {
 		extendsFrom mappingPoetJar
+		extendsFrom asm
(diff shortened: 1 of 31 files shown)


enbrain:
Sorry to say, but the new mappings in this PR has been replaced by #3105 (which I think is much better).

About the static factory method renames,
* `A.toB()` should be used for methods that convert `A` to `B` (see `CONVENTIONS.md`)
* `text` suffix is redundant

0x3C50:
> Sorry to say, but the new mappings in this PR has been replaced by #3105 (which I think is much better).
> 
> About the static factory method renames,
> 
>     * `A.toB()` should be used for methods that convert `A` to `B` (see `CONVENTIONS.md`)
> 
>     * `text` suffix is redundant

I noticed while merging the commits, thought I'd help out a bit but turns out the others were quicker :sweat_smile:

PR #3105, opened 2022-04-20
Text renames

liach:
Text type instances -> TextContent, Nbt data factory -> NbtDataSource, Translatable -> Translation (in class)

Cleanup in mappings for their optimization refactor (good refactors)

We might need to replace more `translatable` with `translation`, I guess? It says `translation` than `translatable` in class `toString`, and I wonder about the `translatable` vs `translation` usages elsewhere.
image

Replaces #3102, sorry but that one is making wrong renames on factory methods and doesn't properly describe the new text model IMO

### DIFF
--- a/mappings/net/minecraft/text/NbtDataSource.mapping
+++ b/mappings/net/minecraft/text/NbtDataSource.mapping
@@ -1,4 +1,4 @@
 CLASS net/minecraft/class_7419 net/minecraft/text/NbtDataSource
-	COMMENT A data source for the nbt text content.
+	COMMENT A data source for the nbt text content. Unmodifiable.
 	METHOD method_10916 get (Lnet/minecraft/class_2168;)Ljava/util/stream/Stream;
 		ARG 1 source
--- a/mappings/net/minecraft/text/NbtDataSource.mapping
+++ b/mappings/net/minecraft/text/NbtDataSource.mapping
@@ -1,0 +1,4 @@
+CLASS net/minecraft/class_7419 net/minecraft/text/NbtDataSource
+	COMMENT A data source for the nbt text content.
+	METHOD method_10916 get (Lnet/minecraft/class_2168;)Ljava/util/stream/Stream;
+		ARG 1 source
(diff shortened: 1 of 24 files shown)


apple502j:
Should we close #3101 and #3102?

Shnupbups:
Hmm, should they be just `TranslationContent` or should they be `TranslationTextContent`?

apple502j:
`TranslatedContent`?

liach:
@Shnupbups @apple502j took both of your suggestions, probably `translatable` vs `translation` rename will be moved to another one

mappings/net/minecraft/text/NbtDataSource.mapping
```diff
@@ -0,0 +1,4 @@
+CLASS net/minecraft/class_7419 net/minecraft/text/NbtDataSource
+	COMMENT A data source for the nbt text content. Unmodifiable.
```
Juuxel:
```suggestion
	COMMENT A data source for the NBT text content. Unmodifiable.
```

mappings/net/minecraft/text/Texts.mapping
```diff
@@ -47,5 +47,9 @@ CLASS net/minecraft/class_2564 net/minecraft/text/Texts
 	METHOD method_37112 join (Ljava/util/Collection;Lnet/minecraft/class_2561;)Lnet/minecraft/class_2561;
 		ARG 0 texts
 		ARG 1 separator
+	METHOD method_43475 brokenReplaceTranslationKey (Lnet/minecraft/class_2561;Ljava/lang/String;Ljava/lang/String;)Lnet/minecraft/class_2561;
```
Juuxel, line 50: How is this broken?

liach: So it's like `brokenReplaceTranslationKey(Text text, ...)` and then `if (text instanceof TranslatableTextContent) ...`
Still, this is a good story on how they refactored all these stuff. This is still used by narrator.

Juuxel (approved):

enbrain (approved):

apple502j (approved):

modmuss50 (approved):

PR #3106, opened 2022-04-21
Fix AtomicSimpleRandom package

apple502j:
Whoops


haykam821 (approved):

liach:
Hmm, are all randoms now in math? I recall we had a few in world gen packages for their proximity to noise chunk generators and such. The move is good, and just make sure all such random number generators are moved altogether.

apple502j:
@liach yeah, all other randoms are already in util.math.random

Shnupbups (approved):

liach (approved):

PR #3108, opened 2022-04-21
22w16b entity

### LINKED ISSUE #3110: WardenBrain#tick is more like #updateActivities
I saw something similar with other new entities like tadpoles. 
Either way, the method is not at all how the brain is ticked (it's ticked earlier according to WardenEntity#mobTick)

apple502j:
Entity tracked position etc.

Resolves #3110

### DIFF
--- a/mappings/net/minecraft/entity/TrackedPosition.mapping
+++ b/mappings/net/minecraft/entity/TrackedPosition.mapping
@@ -1,0 +1,21 @@
+CLASS net/minecraft/class_7422 net/minecraft/entity/TrackedPosition
+	FIELD field_39015 COORDINATE_SCALE D
+	FIELD field_39016 pos Lnet/minecraft/class_243;
+	METHOD method_43487 toPacketValue (D)J
+		ARG 0 value
+	METHOD method_43488 toRealValue (J)D
+		ARG 0 value
+	METHOD method_43489 withDelta (JJJ)Lnet/minecraft/class_243;
+		ARG 1 x
+		ARG 3 y
+		ARG 5 z
+	METHOD method_43490 getDeltaX (Lnet/minecraft/class_243;)J
+		ARG 1 pos
+	METHOD method_43491 getDeltaY (Lnet/minecraft/class_243;)J
+		ARG 1 pos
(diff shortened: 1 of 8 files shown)


mappings/net/minecraft/entity/TrackedPosition.mapping
```diff
@@ -0,0 +1,21 @@
+CLASS net/minecraft/class_7422 net/minecraft/entity/TrackedPosition
+	FIELD field_39015 COORDINATE_SCALE D
+	FIELD field_39016 pos Lnet/minecraft/class_243;
+	METHOD method_43487 toPacketValue (D)J
+		ARG 0 value
+	METHOD method_43488 toRealValue (J)D
```
liach: Maybe `pack` and `unpack`, like for light data converted to long?

mappings/net/minecraft/entity/TrackedPosition.mapping
```diff
@@ -0,0 +1,21 @@
+CLASS net/minecraft/class_7422 net/minecraft/entity/TrackedPosition
```
Shnupbups, line 1: Why not `TrackedPos` to match `BlockPos` and `GlobalPos`?

Shnupbups (approved):

mappings/net/minecraft/entity/TrackedPosition.mapping
apple502j: Hmm, maybe.

apple502j: The stuff has always been called "tracked position" in yarn, though. Do we rename it?

frqnny: I am on phone, but if it is in double and not int then marking it as Position would also mark this difference between BlockPos and this new class.

apple502j: You can't use this as `Vec3d`/`Vec3i`/whatever anyway, this is just a wrapper with network syncing utils.

enbrain (approved):

liach (approved):

PR #3113, opened 2022-04-21
Fix some noise sampler stuff

enbrain:
* `permutations` → `permutation`: This array as a whole represents a single permutation on a set `{0, ..., 255}`.
* `getGradient` → `map`: This is a wrapper around the permutation array access.
* `fadeLocalX` → `fadeLocalY`: Matching went wrong in [21w05a](https://github.com/FabricMC/yarn/commit/323e6f8e29e7064360fecb0037e076c5e073dc72#diff-f6ebd0f1c3d2486618a72a10f316557704b999fc1a224180614472fc254f2809L28-L30).

### DIFF
--- a/mappings/net/minecraft/util/math/noise/PerlinNoiseSampler.mapping
+++ b/mappings/net/minecraft/util/math/noise/PerlinNoiseSampler.mapping
@@ -16,8 +16,8 @@ CLASS net/minecraft/class_3756 net/minecraft/util/math/noise/PerlinNoiseSampler
 		ARG 1 x
 		ARG 3 y
 		ARG 5 z
-	METHOD method_16449 permutate (I)I
-		ARG 1 hash
+	METHOD method_16449 map (I)I
+		ARG 1 input
 	METHOD method_16450 sample (IIIDDDD)D
 		ARG 1 sectionX
 		ARG 2 sectionY
--- a/mappings/net/minecraft/util/math/noise/PerlinNoiseSampler.mapping
+++ b/mappings/net/minecraft/util/math/noise/PerlinNoiseSampler.mapping
@@ -1,7 +1,7 @@
 CLASS net/minecraft/class_3756 net/minecraft/util/math/noise/PerlinNoiseSampler
 	FIELD field_16588 originZ D
 	FIELD field_16589 originY D
-	FIELD field_16590 permutations [B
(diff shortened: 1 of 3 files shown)


apple502j (approved):

mappings/net/minecraft/util/math/noise/PerlinNoiseSampler.mapping
```diff
@@ -16,7 +16,7 @@ CLASS net/minecraft/class_3756 net/minecraft/util/math/noise/PerlinNoiseSampler
 		ARG 1 x
 		ARG 3 y
 		ARG 5 z
-	METHOD method_16449 getGradient (I)I
+	METHOD method_16449 permutate (I)I
```
liach: Since this sends an input to the permutation map for an output, can this be called `map`? Also the parameter can be updated to `input` than `hash`

enbrain: Now that I think about it, `permutate` sounds like it takes a whole array. `hash` was wrong or inaccurate as the sequence of the calls in the `sample` method is the hash function.

mappings/net/minecraft/util/math/noise/PerlinNoiseSampler.mapping
```diff
@@ -16,7 +16,7 @@ CLASS net/minecraft/class_3756 net/minecraft/util/math/noise/PerlinNoiseSampler
...
 		ARG 1 hash
```
enbrain:
```suggestion
	METHOD method_16449 map (I)I
		ARG 1 input
```

mappings/net/minecraft/util/math/noise/SimplexNoiseSampler.mapping
```diff
@@ -23,7 +23,7 @@ CLASS net/minecraft/class_3541 net/minecraft/util/math/noise/SimplexNoiseSampler
 		ARG 4 y
 		ARG 6 z
 		ARG 8 distance
-	METHOD method_16456 getGradient (I)I
+	METHOD method_16456 permutate (I)I
 		ARG 1 hash
```
liach:
```suggestion
	METHOD method_16456 map (I)I
		ARG 1 input
```

liach (approved):

PR #3118, opened 2022-04-27
Move the ScreenTexts class to the net.minecraft.screen package

### LINKED ISSUE #3112: ScreenTexts is no longer client-side only
This is referenced by `SignBlockEntity` and used server-side.

haykam821:
Fixes #3112


enbrain (approved):

apple502j:
I believe this should be moved to `net.minecraft.text` instead.

haykam821:
@apple502j I disagree as this class is still clearly related to screens; for example, the text fields are used for common components of a GUI. Otherwise, it would just be a second `Texts` class.

apple502j:
Hmm, then maybe it's fine to keep it at screen package.

Juuxel:
I'd say either package is fine - this is related to both after all. `text` might be a bit more natural though since the `screen` package mostly deals with screen handlers.

apple502j:
Should we have a vote for this？

liach:
I think this is just some general ui prompt text. It probably works better in screen package as it's not core text infrastructure, but we can consider renaming the class as well.

Juuxel (approved):

modmuss50 (approved):

liach (approved):

PR #3124, opened 2022-04-28
22w17a stuff

apple502j:
Small PR for non-networking-related 22w17a changes.

### DIFF
--- a/mappings/net/minecraft/util/MessageHoldingException.mapping
+++ b/mappings/net/minecraft/util/MessageHoldingException.mapping
@@ -1,0 +1,8 @@
+CLASS net/minecraft/class_7437 net/minecraft/util/MessageHoldingException
+	FIELD field_39084 message Lnet/minecraft/class_2561;
+	METHOD <init> (Lnet/minecraft/class_2561;)V
+		ARG 1 message
+	METHOD <init> (Lnet/minecraft/class_2561;Ljava/lang/Throwable;)V
+		ARG 1 message
+		ARG 2 cause
+	METHOD method_43626 getMessage ()Lnet/minecraft/class_2561;
--- a/mappings/net/minecraft/util/MessageHoldingException.mapping
+++ b/mappings/net/minecraft/util/MessageHoldingException.mapping
@@ -1,8 +1,0 @@
-CLASS net/minecraft/class_7437 net/minecraft/util/MessageHoldingException
-	FIELD field_39084 message Lnet/minecraft/class_2561;
-	METHOD <init> (Lnet/minecraft/class_2561;)V
-		ARG 1 message
-	METHOD <init> (Lnet/minecraft/class_2561;Ljava/lang/Throwable;)V
-		ARG 1 message
-		ARG 2 cause
-	METHOD method_43626 getMessage ()Lnet/minecraft/class_2561;
--- a/mappings/net/minecraft/util/Util.mapping
+++ b/mappings/net/minecraft/util/Util.mapping
@@ -206,7 +206,7 @@ CLASS net/minecraft/class_156 net/minecraft/util/Util
 		COMMENT @see #combineSafe(List)
 		ARG 0 futures
 	METHOD method_43498 waitAndApply (Ljava/util/function/Function;Ljava/util/function/Predicate;)Ljava/lang/Object;
-		COMMENT Run tasks using the prepare-apply model.
+		COMMENT Runs tasks using the prepare-apply model.
 		COMMENT
 		COMMENT @apiNote This method takes a function that supplies an executor to use in the
 		COMMENT apply stage. Inside the function, callers should run the preparation,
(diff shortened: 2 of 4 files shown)


mappings/net/minecraft/util/MessageHoldingException.mapping
```diff
@@ -0,0 +1,8 @@
+CLASS net/minecraft/class_7437 net/minecraft/util/MessageHoldingException
+	FIELD field_39084 message Lnet/minecraft/class_2561;
+	METHOD <init> (Lnet/minecraft/class_2561;)V
+		ARG 1 message
+	METHOD <init> (Lnet/minecraft/class_2561;Ljava/lang/Throwable;)V
+		ARG 1 message
+		ARG 2 cause
+	METHOD method_43626 getMessage ()Lnet/minecraft/class_2561;
```
liach: Conflicts with [`String getMessage()` from `Throwable`](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/lang/Throwable.html#getMessage()). Recommended `getMessageText` or `getTextMessage`, and this class should be renamed like `TextifiedException` (a name that means this class is an exception with a text message, yet it's not a problem with text)

mappings/net/minecraft/util/Util.mapping
```diff
@@ -205,6 +205,11 @@ CLASS net/minecraft/class_156 net/minecraft/util/Util
 		COMMENT @see #combine(List)
 		COMMENT @see #combineSafe(List)
 		ARG 0 futures
+	METHOD method_43498 waitUntilDone (Ljava/util/function/Function;Ljava/util/function/Predicate;)Ljava/lang/Object;
+		ARG 0 queuedValueSupplier
```
liach:
```suggestion
		ARG 0 resultFactory
```
Also document this that the factory is `applyExecutor -> result`, that the async preparation tasks are supposed to be run when the factory returns the preliminary result, and the preparation tasks should send apply work to the given apply executor.

Otherwise, this would be extremely confusing. The rationale of this utility is that we thus can run tasks in the prepare-apply model when the game engine isn't in  the ticking mode yet.

mappings/net/minecraft/util/Util.mapping
```diff
@@ -205,6 +205,11 @@ CLASS net/minecraft/class_156 net/minecraft/util/Util
 		COMMENT @see #combine(List)
 		COMMENT @see #combineSafe(List)
 		ARG 0 futures
+	METHOD method_43498 waitUntilDone (Ljava/util/function/Function;Ljava/util/function/Predicate;)Ljava/lang/Object;
```
liach: This isn't just waiting; this is running tasks created when the function is called with an executor that add runnables to a queue for the current main thread to run. So, this can just run prepared tasks on the current main thread once any is done, to reduce computational waste. I recommend names like
```suggestion
	METHOD method_43498 waitAndApply (Ljava/util/function/Function;Ljava/util/function/Predicate;)Ljava/lang/Object;
```

mappings/net/minecraft/util/MessageHoldingException.mapping
apple502j: How did Enigma not prevent that!??

liach: Maybe it failed to scan classpath, but this is valid in jvm after all as they have different descriptors. Mixin uses such dangerous tricks, too.

apple502j:
I've added javadoc, not sure if this is 100% correct tho.

liach (approved):

mappings/net/minecraft/util/Util.mapping
```diff
@@ -205,6 +205,25 @@ CLASS net/minecraft/class_156 net/minecraft/util/Util
 		COMMENT @see #combine(List)
 		COMMENT @see #combineSafe(List)
 		ARG 0 futures
+	METHOD method_43498 waitAndApply (Ljava/util/function/Function;Ljava/util/function/Predicate;)Ljava/lang/Object;
+		COMMENT Run tasks using the prepare-apply model.
```
enbrain:
```suggestion
		COMMENT Runs tasks using the prepare-apply model.
```

mappings/net/minecraft/util/Util.mapping
```diff
@@ -205,6 +205,25 @@ CLASS net/minecraft/class_156 net/minecraft/util/Util
...
+		COMMENT
+		COMMENT @apiNote This method takes a function that supplies an executor to use in the
+		COMMENT apply stage. Inside the function, callers should run the preparation,
+		COMMENT and use the passed executor for applying.
+		ARG 0 resultFactory
+			COMMENT a function that takes the apply-stage executor and returns the preliminary result
+		ARG 1 donePredicate
+			COMMENT a predicate that, given the result, checks whether applying has finished
+	METHOD method_43499 waitAndApply (Ljava/util/function/Function;)Ljava/util/concurrent/CompletableFuture;
+		COMMENT Run tasks using the prepare-apply model, such as creation of a {@link
```
enbrain:
```suggestion
		COMMENT Runs tasks using the prepare-apply model, such as creation of a {@link
```

enbrain (approved):

PR #874, opened 2019-09-25
Fix clash of CommandBlockBlockEntity#getType with BlockEntity#getType

NeunEinser:

PR #3126, opened 2022-04-28
Fix LevelStorageException name conflict

apple502j:


liach (approved):

liach:
Apparently this enigma bug has been there for long; see #874

enbrain (approved):

PR #3135, opened 2022-05-04
Map changes to banner patterns in 22w18a

haykam821:


mappings/net/minecraft/block/entity/BannerPatterns.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS net/minecraft/class_7446 net/minecraft/block/entity/BannerPatterns
+	METHOD method_43721 initAndGetDefault (Lnet/minecraft/class_2378;)Lnet/minecraft/class_2582;
```
liach, line 2:
```suggestion
	METHOD method_43721 registerAndGetDefault (Lnet/minecraft/class_2378;)Lnet/minecraft/class_2582;
```
See `MaterialRule` `MaterialCondition` `DensityFunctionTypes` `PaintingVariants`

haykam821: The `initAndGetDefault` method name is used for dimension types, structure sets, world presets, and flat level generator presets. While the inconsistency should be fixed, I think it is out of scope for this pull request.

liach: Hmm, turns out the `Registry` ones use `registerAndGetDefault` and `BuiltinRegistries` ones use `initAndGetDefault`

apple502j: Kinda makes sense given DRM registries are dynamic (although I believe `BuiltingRegistries` is not the ones they register to)

apple502j (approved):

enbrain (approved):

PR #3136, opened 2022-05-04
Map message type and format

liach:
Mojang calls it "chat format" in its registry and "chat style format" in the blog, but I think `MessageType` and `MessageFormat` fits better.

Since this name differs from what's in the registry, I didn't map the key and the registry in the Registry class. In addition, we should consider making new packages for these chat stuff, such as a `chat` package. The `network` package is more for network infrastructures, while the `text` package is text infrastructures so I oppose a move to there.

### DIFF
--- a/mappings/net/minecraft/network/MessageType.mapping
+++ b/mappings/net/minecraft/network/MessageType.mapping
@@ -1,7 +1,7 @@
 CLASS net/minecraft/class_2556 net/minecraft/network/MessageType
 	COMMENT A message type, or chat type in registry, or chat style format per
 	COMMENT Minecraft blog, defines how a piece of text content should be rendered.
-	COMMENT It can direct messges to the chat hud, the in-game hud overlay, and the
+	COMMENT It can direct messages to the chat hud, the in-game hud overlay, and the
 	COMMENT narrator, and format them with the sender information.
 	FIELD comp_792 Ljava/util/Optional;
 		COMMENT the display rule of a message to the chat hud, or
@@ -10,14 +10,14 @@ CLASS net/minecraft/class_2556 net/minecraft/network/MessageType
 		COMMENT the display rule of a message to the in-game hud overlay, or
 		COMMENT {@code Optional.empty()} if it is not displayed
 	FIELD comp_794 Ljava/util/Optional;
-		COMMENT the narration rule of of a message, or {@code Optional.empty()} if it is
+		COMMENT the narration rule of a message, or {@code Optional.empty()} if it is
 		COMMENT not narrated
 	FIELD field_39227 CODEC Lcom/mojang/serialization/Codec;
 		COMMENT The codec for the message type.
 	METHOD method_43843 (Lcom/mojang/serialization/codecs/RecordCodecBuilder$Instance;)Lcom/mojang/datafixers/kinds/App;
 		ARG 0 instance
 	METHOD method_43844 registerAndGetDefault (Lnet/minecraft/class_2378;)Lnet/minecraft/class_2556;
-		COMMENT Registers the built-in message types and return the default "chat" type.
+		COMMENT Registers the built-in message types and returns the default "chat" type.
 		ARG 0 registry
 	METHOD method_43845 getKey (Ljava/lang/String;)Lnet/minecraft/class_5321;
 		ARG 0 id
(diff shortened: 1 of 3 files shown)


apple502j:
@liach whoops, I mapped both as well (see #3141 and #3142)
Can you have a cross-review or something like that?

my comments:
- I believe the exposed name for MessageFormat is `Decoration`. Also, I believe MessageFormat is also the name used by ICU framework, and these two have similar but very different functions. Given this is usually constructed by modders and ICU framework is well-known this conflict does not seem great.
- `Kind` vs `Priority`: imo `Kind` is better, I only chose `priority` as it was exposed, but "kind" makes more sense in NarrationManager context, maybe I'll steal that
- I need sleep, see ya later

mappings/net/minecraft/network/MessageType.mapping
```diff
@@ -1 +1,82 @@
 CLASS net/minecraft/class_2556 net/minecraft/network/MessageType
+	COMMENT A message type, or chat type in registry, or chat style format per
+	COMMENT Minecraft blog, defines how a piece of text content should be rendered.
+	COMMENT It can direct messges to the chat hud, the in-game hud overlay, and the
```
enbrain:
```suggestion
	COMMENT It can direct messages to the chat hud, the in-game hud overlay, and the
```

mappings/net/minecraft/network/MessageType.mapping
```diff
@@ -1 +1,82 @@
...
+	COMMENT narrator, and format them with the sender information.
+	FIELD comp_792 Ljava/util/Optional;
+		COMMENT the display rule of a message to the chat hud, or
+		COMMENT {@code Optional.empty()} if it is not displayed
+	FIELD comp_793 Ljava/util/Optional;
+		COMMENT the display rule of a message to the in-game hud overlay, or
+		COMMENT {@code Optional.empty()} if it is not displayed
+	FIELD comp_794 Ljava/util/Optional;
+		COMMENT the narration rule of of a message, or {@code Optional.empty()} if it is
```
enbrain:
```suggestion
		COMMENT the narration rule of a message, or {@code Optional.empty()} if it is
```

mappings/net/minecraft/network/MessageType.mapping
```diff
@@ -1 +1,82 @@
...
+		COMMENT not narrated
+	FIELD field_39227 CODEC Lcom/mojang/serialization/Codec;
+		COMMENT The codec for the message type.
+	METHOD method_43843 (Lcom/mojang/serialization/codecs/RecordCodecBuilder$Instance;)Lcom/mojang/datafixers/kinds/App;
+		ARG 0 instance
+	METHOD method_43844 registerAndGetDefault (Lnet/minecraft/class_2378;)Lnet/minecraft/class_2556;
+		COMMENT Registers the built-in message types and return the default "chat" type.
```
enbrain:
```suggestion
		COMMENT Registers the built-in message types and returns the default "chat" type.
```

mappings/net/minecraft/network/MessageType.mapping
```diff
@@ -1 +1,82 @@
...
+			COMMENT the formatting of a message, or {@code Optional.empty()} if the content
+			COMMENT is displayed directly
+		FIELD comp_796 Lnet/minecraft/class_2556$class_7466$class_7467;
+			COMMENT the narration kind, and if this narration can interrupt other ones
+		FIELD field_39233 CODEC Lcom/mojang/serialization/Codec;
+			COMMENT The codec for the message narraton rule.
+		METHOD method_43846 (Lcom/mojang/serialization/codecs/RecordCodecBuilder$Instance;)Lcom/mojang/datafixers/kinds/App;
+			ARG 0 instance
+		METHOD method_43847 of (Lnet/minecraft/class_7463;Lnet/minecraft/class_2556$class_7466$class_7467;)Lnet/minecraft/class_2556$class_7466;
+			COMMENT {@return a rule that narrate a message after formatting} The message
```
enbrain:
```suggestion
			COMMENT {@return a rule that narrates a message after formatting} The message
```

mappings/net/minecraft/network/MessageType.mapping
```diff
@@ -1 +1,82 @@
...
+			COMMENT is narrated as the given {@code type}, and the narration can include
+			COMMENT sender information in addition to the message content.
+			ARG 0 decoration
+			ARG 1 kind
+		METHOD method_43848 of (Lnet/minecraft/class_2556$class_7466$class_7467;)Lnet/minecraft/class_2556$class_7466;
+			COMMENT {@return a rule that narrate a message's content directly} The message
```
enbrain:
```suggestion
			COMMENT {@return a rule that narrates a message's content directly} The message
```

mappings/net/minecraft/network/MessageType.mapping
```diff
@@ -1 +1,82 @@
...
+			COMMENT is displayed directly
+		FIELD comp_796 Lnet/minecraft/class_2556$class_7466$class_7467;
+			COMMENT the narration kind, and if this narration can interrupt other ones
+		FIELD field_39233 CODEC Lcom/mojang/serialization/Codec;
+			COMMENT The codec for the message narraton rule.
+		METHOD method_43846 (Lcom/mojang/serialization/codecs/RecordCodecBuilder$Instance;)Lcom/mojang/datafixers/kinds/App;
+			ARG 0 instance
+		METHOD method_43847 of (Lnet/minecraft/class_7463;Lnet/minecraft/class_2556$class_7466$class_7467;)Lnet/minecraft/class_2556$class_7466;
+			COMMENT {@return a rule that narrate a message after formatting} The message
+			COMMENT is narrated as the given {@code type}, and the narration can include
```
enbrain: Did you mean `kind`?
```suggestion
			COMMENT is narrated as the given {@code kind}, and the narration can include
```

mappings/net/minecraft/network/MessageType.mapping
```diff
@@ -1 +1,82 @@
...
+			COMMENT sender information in addition to the message content.
+			ARG 0 decoration
+			ARG 1 kind
+		METHOD method_43848 of (Lnet/minecraft/class_2556$class_7466$class_7467;)Lnet/minecraft/class_2556$class_7466;
+			COMMENT {@return a rule that narrate a message's content directly} The message
+			COMMENT is narrated as the given {@code type}, without any sender-related
```
enbrain:
```suggestion
			COMMENT is narrated as the given {@code kind}, without any sender-related
```

enbrain (approved):

PR #3138, opened 2022-05-04
Remove broken "Iterator" mapping

liach:
This is apparently causing yarn build failures and must be fixed.


apple502j (approved):

Player3324:
I believe I already fixed it? No idea why it still shows as being there here, it is not in https://github.com/FabricMC/yarn/tree/22w18a/mappings/net/minecraft/client/search ?

liach:
Ah, I guess I checked the repo out too early.

Issue #3125, opened 2022-04-28
AbstractRandom -> Random


liach:
Nowadays, seems mojang moved pretty much all use cases of `java.util.Random` to the current `AbstractRandom` interface. I hence recommend simply renaming it to `Random`.

The JDK has some nice documentation for understanding random: https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/random/RandomGenerator.SplittableGenerator.html

liach:
@apple502j How do you think since you initially mapped these randoms? I feel we should rename the base class to `Random` and have a few other simplifications, as it's now the norm for the whole game codebase. In addition, for other terminologies, we can probably look at JDK, such as rename `derive` to `split` as in that javadoc link above.

apple502j:
@liach That sounds fine - still snapshot phase, we can make this change I think.

haykam821:
This issue was fixed by #3140.

PR #3140, opened 2022-05-04
Random renames

liach:
See #3125

A few renames should encourage modders to use the right types of random.


mappings/net/minecraft/util/math/random/AsyncRandom.mapping
```diff
@@ -1,4 +1,5 @@
-CLASS net/minecraft/class_7384 net/minecraft/util/math/random/BlockingSimpleRandom
+CLASS net/minecraft/class_7384 net/minecraft/util/math/random/AsyncRandom
+	COMMENT A random that can be shared by multiple threads safely.
```
apple502j, line 2: Is this really async, or just "async-safe"?

apple502j (changes requested):

mappings/net/minecraft/util/math/random/AsyncRandom.mapping
liach: This probably needs to be `ThreadSafeRandom` then

haykam821:
Would renaming `nextPredictable` to `nextTriangular` be in scope for this pull request?

liach:
Hmm, does that have a triangular density function?

apple502j:
@liach Yep.

image

enbrain:
Yes, `nextPredictable` has a triangular PDF. Let `g(x)` be the PDF of the uniform distribution `U(0, 1)`. The PDF of `nextPredictable(b, v)` is `f(x) = (1/v)∫[-∞,∞]g((x-b)/v+y)g(y)dy`.

* For `x < b - v`, `f(x) = 0`.
* For `b - v <= x < b`, `f(x) = (1/v)∫[0,1+(x-b)/v]g((x-b)/v+y)g(y)dy = (1/v)(1+(x-b)/v)`.
* For `b <= x <= b + v`, `f(x) = (1/v)∫[0,1-(x-b)/v]g((x-b)/v+y)g(y)dy = (1/v)(1-(x-b)/v)`.
* For `b + v < x`, `f(x) = 0`.

It's an isosceles triangle and matches the PDF in https://en.wikipedia.org/wiki/Triangular_distribution#Symmetric_triangular_distribution.

Shnupbups:
I support a rename of `nextPredictable` - I only named it that due to it being referred to as such in the snapshot blogpost, but a more accurate name would be appreciated I'm sure.

apple502j (approved):
Let's rename nextPredictable in another PR. This definitely has to be merged before pre-release stage begins, because this will be a quite impactful rename after that.

modmuss50 (approved):

PR #3141, opened 2022-05-04
22w18a texts and utils

apple502j:
Contains `Decoration` class (in `net.minecraft.text`) - given its relations with other text classes (and its name) I believe that's the best place to go. Also mapped JsonHelper methods, Style, etc.

### DIFF
--- a/mappings/net/minecraft/text/Text.mapping
+++ b/mappings/net/minecraft/text/Text.mapping
@@ -95,7 +95,7 @@ CLASS net/minecraft/class_2561 net/minecraft/text/Text
 			ARG 1 type
 			ARG 2 context
 			ARG 3 json
-		METHOD method_43474 convertLiteralToString (Ljava/lang/Object;)Ljava/lang/Object;
+		METHOD method_43474 optimizeArgument (Ljava/lang/Object;)Ljava/lang/Object;
 			ARG 0 text
 		METHOD method_43858 toSortedJsonString (Lnet/minecraft/class_2561;)Ljava/lang/String;
 			ARG 0 text
--- a/mappings/net/minecraft/text/Text.mapping
+++ b/mappings/net/minecraft/text/Text.mapping
@@ -95,6 +95,10 @@ CLASS net/minecraft/class_2561 net/minecraft/text/Text
 			ARG 1 type
 			ARG 2 context
 			ARG 3 json
+		METHOD method_43474 convertLiteralToString (Ljava/lang/Object;)Ljava/lang/Object;
+			ARG 0 text
+		METHOD method_43858 toSortedJsonString (Lnet/minecraft/class_2561;)Ljava/lang/String;
+			ARG 0 text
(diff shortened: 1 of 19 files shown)


apple502j:
Decoration is now documented, toText is renamed to apply per liach PR.

Here is my opinion on where this class should live: Definitely text package. Yes, this is only used in chats, but its concept of translatable text factory applies outside that. The class name is from exposed record field. If there exists a name that expresses clearly that this contains params and styling (and is not confusing like TextFactory) we should rename; however imo there is none. liach suggested using the name MessageFormat - noting this is kinda biased view as I’ve worked with localizations a lot, I disagree with that name because the well-known MessageFormat by ICU and java.text and this class are made for different purposes and has different functions.

mappings/net/minecraft/util/JsonHelper.mapping
```diff
@@ -212,3 +233,14 @@ CLASS net/minecraft/class_3518 net/minecraft/util/JsonHelper
 		ARG 1 element
 	METHOD method_37165 deserializeArray (Ljava/io/Reader;)Lcom/google/gson/JsonArray;
 		ARG 0 reader
+	METHOD method_43677 writeSorted (Lcom/google/gson/stream/JsonWriter;Lcom/google/gson/JsonElement;Ljava/util/Comparator;)V
+		ARG 0 writer
+		ARG 1 json
+		ARG 2 comparator
+	METHOD method_43678 sortEntries (Ljava/util/Collection;Ljava/util/Comparator;)Ljava/util/Collection;
```
enbrain:
```suggestion
	METHOD method_43678 sort (Ljava/util/Collection;Ljava/util/Comparator;)Ljava/util/Collection;
```

mappings/net/minecraft/util/JsonHelper.mapping
```diff
@@ -212,3 +233,14 @@ CLASS net/minecraft/class_3518 net/minecraft/util/JsonHelper
...
+		ARG 0 entries
+		ARG 1 comparator
+	METHOD method_43679 deserializeArray (Ljava/lang/String;)Lcom/google/gson/JsonArray;
+		ARG 0 content
+	METHOD method_43680 writeSorted (Lcom/google/gson/JsonElement;)Ljava/lang/String;
```
enbrain:
```suggestion
	METHOD method_43680 toSortedString (Lcom/google/gson/JsonElement;)Ljava/lang/String;
```

mappings/net/minecraft/text/Text.mapping
```diff
@@ -95,6 +95,10 @@ CLASS net/minecraft/class_2561 net/minecraft/text/Text
 			ARG 1 type
 			ARG 2 context
 			ARG 3 json
+		METHOD method_43474 convertLiteralToString (Ljava/lang/Object;)Ljava/lang/Object;
```
enbrain: Maybe `optimizeArgument`? This method is applied to translation arguments to remove unnecessary indirections.

apple502j:
```suggestion
		METHOD method_43474 optimizeArgument (Ljava/lang/Object;)Ljava/lang/Object;
```

enbrain (approved):

PR #3146, opened 2022-05-11
Fix name conflict with getPrimeSuspect

### LINKED ISSUE #3129: WardenAngerManager has two getPrimeSuspect
One returns nullable and one returns optional + LivingEntity check iirc.

apple502j:
Fixes #3129

Better names? The only difference from `getPrimeSuspect` is that the return type is `Optional<LivingEntity>` instead of `@Nullable Entity`.


enbrain:
`public`/`private` is also the difference. Would it make sense to use `getPrimeSuspect`/`getPrimeSuspectInternal`?

haykam821:
Does this method return an empty optional if the prime suspect is not a `LivingEntity`? I would name it `getLivingPrimeSuspect` with `@return` documentation for how the optional is empty when there is no suspect or a non-living suspect.

apple502j:
I'll use the internal suffix, then.

@haykam821 I don't like "living" prefix because without context this means non-discarded entities - although as of 18a wardens should never get angry at discarded entities.

haykam821:
The "living" prefix is not an issue for the class itself. I think there is enough of a distinction between entities that can live and non-discarded entities.

modmuss50 (approved):

PR #3153, opened 2022-05-12
Rename the SaplingGenerator class

### LINKED ISSUE #2866: SaplingGenerator is misleading
It does not generate a sapling. It generates a tree when a sapling grows naturally or when a bone meal item is used on a sapling.

enbrain:
Resolves #2866

I avoided "sapling" because it's also used by `AzaleaBlock`.


apple502j:
Package now becomes an issue, previously it was fine since sapling is a block, but "tree growth" does not sound like something that could go under block package. Not sure how to fix this though. It doesn’t quite fit in worldgen package.

haykam821:
Keep in mind that azaleas and propagules can still be considered saplings since they are in the `minecraft:saplings` block and item tags.

enbrain:
I will wait for a better name. I'm currently busy and am not going to work on this.

PR #3158, opened 2022-05-14
Rename ThreadedAnvilChunkStorage

apple502j:
This is from exposed mojang name, but the name is bad (not incorrect, but unfriendly). Being partially threaded and using Anvil chunk storage format is an implementation detail that does not warrant spending 13 letters. I replaced these 13 letters with another, coincidentally 13 letters, that explains much important parts of this class - being server-side and managing chunk **storage**. Hopefully this explains the relations with ServerChunkManager and other similarly named classes.


liach:
How about just calling it ChunkStorage instead?

apple502j:
@liach Hmm, that looks too short for an important class, and it's also weird if `VersionedChunkStorage`, etc was a parent of `ChunkStorage`.

`ServerChunkStorage` could work though.

Player3324:
IMO the new name isn't good enough, it does quite a bit more than storing chunks with some runtime state and various worldgen features. The relationship with ServerChunkManager needs to be considered as well, it is a similar dumping ground.

apple502j:
@sfPlayer1 I understand this handles more than storage (e.g. entity loading, storing chunk generator) but the main purpose is that. I don't really see a better name for this (aside from maybe removing Manager suffix). Javadocs should disambiguate if necessary.

PR #3173, opened 2022-05-20
Move LargeEntitySpawnHelper to entity

### LINKED ISSUE #3171: Move LargeEntitySpawnHelper to entity package
This is not a generic utility.

`ItemScatter` is in the util package for some reason too, but that one is shared by blocks, entities, and other things, so it's not that straightforward.

apple502j:
Resolves #3171


liach:
How similar is this to `world.SpawnHelper`? Would be good to put them together if they are similar.

apple502j:
@liach This is just a utility function to spawn one entity. SpawnHelper is a complex class handling all of the natural mob spawning logic. Thus it should stay in the world package.

haykam821 (approved):

modmuss50 (approved):

PR #3175, opened 2022-05-20
Unify initializer names for DRM registries

### LINKED ISSUE #3170: Registry initializer method names post-22w19a
Before 22w19a, the registry initializer methods were named like this:

- `initAndGetDefault` for non-DRM registries
- `registerAndGetDefault` for DRM registries

Since 22w19a the method is considered a functional interface named `BuiltinRegistries$Initializer`. The class name was chosen because `InitAndGetDefaulter` doesn't sound great. But this raises a question: what should the methods be named then?

Ideas:

- Keep current one
- Rename to `init`. (`initialize`, but yarn convention abbreviates that.) The "get default" part is an implementation detail and should be documented on the functional interface.
- Something else?

apple502j:
Resolves #3170 
BannerPatterns rename is to fix inconsistency with non-DRM registries


haykam821 (changes requested):
Many registries still have inconsistent names. Make sure that the multiple initializer methods used to initialize the structure set registry are renamed as well.

apple502j:
Hmm, perhaps naming it an initializer was a mistake, half of them don’t initialize in them...

haykam821:
The methods that do not register don't have to be unified or changed in this pull request.

apple502j:
Gonna do this another time.

PR #3178, opened 2022-05-20
Rename AsyncRandom and nextPredictable

### LINKED ISSUE #3180: Missing verb in the documentation for the Xoroshiro128PlusPlusRandom#next method
https://github.com/FabricMC/yarn/blob/2f8ce2e639fdfc6f8d873773c21e5cb130516607/mappings/net/minecraft/util/math/random/Xoroshiro128PlusPlusRandom.mapping#L17-L19

apple502j:
See #3140 reviews
Fixes #3180

### DIFF
--- a/mappings/net/minecraft/util/math/random/Random.mapping
+++ b/mappings/net/minecraft/util/math/random/Random.mapping
@@ -55,7 +55,7 @@ CLASS net/minecraft/class_5819 net/minecraft/util/math/random/Random
 	METHOD method_43058 nextDouble ()D
 	METHOD method_43059 nextGaussian ()D
 	METHOD method_43385 nextTriangular (DD)D
-		COMMENT {@return a random number between {@code mode - deviation} to {@code mode + deviation}
-		COMMENT with mode {@code mode}}
+		COMMENT {@return a random {@cpde double} between {@code mode - deviation} and
+		COMMENT {@code mode + deviation} (both inclusive) with mode {@code mode}}
 		ARG 1 mode
 		ARG 3 deviation
--- a/mappings/net/minecraft/util/math/random/Random.mapping
+++ b/mappings/net/minecraft/util/math/random/Random.mapping
@@ -9,7 +9,7 @@ CLASS net/minecraft/class_5819 net/minecraft/util/math/random/Random
 	COMMENT {@link java.util.Random}.</li>
 	COMMENT <li>{@link CheckedRandom}: Throws when used concurrently. Based on
 	COMMENT {@link java.util.Random}.</li>
-	COMMENT <li>{@link AsyncRandom}: Blocks the thread when used concurrently. Based on
+	COMMENT <li>{@link ThreadSafeRandom}: Blocks the thread when used concurrently. Based on
 	COMMENT {@link java.util.Random}.</li>
 	COMMENT <li>{@link Xoroshiro128PlusPlusRandom}: Silently breaks when used concurrently.
 	COMMENT Implements the Xoroshiro128++ algorithm.</li>
@@ -55,5 +55,7 @@ CLASS net/minecraft/class_5819 net/minecraft/util/math/random/Random
 	METHOD method_43058 nextDouble ()D
 	METHOD method_43059 nextGaussian ()D
 	METHOD method_43385 nextTriangular (DD)D
-		ARG 1 base
-		ARG 3 variance
+		COMMENT {@return a random number between {@code mode - deviation} to {@code mode + deviation}
+		COMMENT with mode {@code mode}}
+		ARG 1 mode
+		ARG 3 deviation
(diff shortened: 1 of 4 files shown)


mappings/net/minecraft/util/math/random/Random.mapping
```diff
@@ -54,6 +54,6 @@ CLASS net/minecraft/class_5819 net/minecraft/util/math/random/Random
 	METHOD method_43057 nextFloat ()F
 	METHOD method_43058 nextDouble ()D
 	METHOD method_43059 nextGaussian ()D
-	METHOD method_43385 nextPredictable (DD)D
+	METHOD method_43385 nextTriangular (DD)D
 		ARG 1 base
 		ARG 3 variance
```
liach: This isn't a variance, but more like a "radius", like the difference from the mean (should rename "base" too) to the upper and lower limit.

haykam821: Would `deviation` work?

```suggestion
		ARG 3 deviation
```

liach: > Would `deviation` work?

Sure, as long as it doesn't overlap with statistical terms.

liach:
Also remember to update the reference to `AsyncRandom` in `Random` interface documentation

mappings/net/minecraft/util/math/random/Random.mapping
haykam821: These are the typical statistical terms when describing a triangular distribution:

triangular-distribution

The parameters `mean` and `deviation` would align with these terms.

apple502j:
Note that for triangular distribution the peak is the mode. While in this case it equals the mean we should use commonly accepted terms.

mappings/net/minecraft/util/math/random/Random.mapping
```diff
@@ -54,6 +54,8 @@ CLASS net/minecraft/class_5819 net/minecraft/util/math/random/Random
 	METHOD method_43057 nextFloat ()F
 	METHOD method_43058 nextDouble ()D
 	METHOD method_43059 nextGaussian ()D
-	METHOD method_43385 nextPredictable (DD)D
-		ARG 1 base
-		ARG 3 variance
+	METHOD method_43385 nextTriangular (DD)D
+		COMMENT {@return a random {@cpde double} between {@code mode - deviation} and
```
haykam821:
```suggestion
		COMMENT {@return a random {@code double} between {@code mode - deviation} and
```

haykam821 (changes requested):

liach (approved):

mappings/net/minecraft/util/math/random/Random.mapping
```diff
@@ -54,6 +54,8 @@ CLASS net/minecraft/class_5819 net/minecraft/util/math/random/Random
 	METHOD method_43057 nextFloat ()F
 	METHOD method_43058 nextDouble ()D
 	METHOD method_43059 nextGaussian ()D
-	METHOD method_43385 nextPredictable (DD)D
-		ARG 1 base
-		ARG 3 variance
+	METHOD method_43385 nextTriangular (DD)D
+		COMMENT {@return a random {@code double} between {@code mode - deviation} and
+		COMMENT {@code mode + deviation} (both inclusive) with mode {@code mode}}
+		ARG 1 mode
```
haykam821: The `mode` parameter should be named `mean` to align with other Yarn parameter names. For example, the `MathHelper.nextGaussian` method has `mean` and `deviation` parameters.

```suggestion
		COMMENT {@return a random {@code double} between {@code mean - deviation} and
		COMMENT {@code mean + deviation} (both inclusive) with mean {@code mean}}
		ARG 1 mean
```

haykam821 (changes requested):

mappings/net/minecraft/util/math/random/Random.mapping
liach: It's simultaneously mode and mean.

haykam821: I think that Yarn should standardize on using mean.

liach: Or just `center` as this is the center of the triangular distribution, more straightforward to general users

apple502j: Triangular distribution in general does not have to be symmetric, and in that case the peak is not the mean. Here both works but this is fine.

Gaussian distribution's μ is the mean - never heard it as the mode, even though mathematically it is the same.

haykam821 (approved):

liach:
> Triangular distribution in general does not have to be symmetric, and in that case the peak is not the mean. Here both works but this is fine.

Then I recommend you document that this triangular distribution is symmetric, and thus its mean, median, and mode is the same value `mode` passed.

haykam821 (approved):

modmuss50 (approved):

PR #3183, opened 2022-05-22
Rename Structure and StructureType

### LINKED ISSUE #3179: Some classes use old configured structure feature naming
- [ ] `ConfiguredStructureFeatureTags` class
- [ ] `ConfiguredStructureFeatureTagProvider` class
- [ ] `configuredStructureFigure` parameter of the `StructureSets.register` method
- [ ] `StructureLocator#configuredStructureFeatureRegistry` field
- [ ] `configuredStructureFeature` parameter of the `StructureTypes.register` method

### LINKED ISSUE #3182: StructureType renames
There are two `StructureType`s in yarn, `structure.StructureType` (`s.ST`) and `world.gen.StructureType` (`w.g.s.ST`). This is already concerning.

`w.g.s.ST` is a type of a structure registered to `structure` dynamic registry, while `s.ST` is a type of a structure type registered to `structure_type` normal registry. This is also explained by the fact that `s.ST` is just a fancy `Supplier<Codec>`.

~~If we follow the conventions, `s.ST` should be named `StructureTypeType`, but since that's also bad I suggest renaming everything to `StructureCodecType` (registry and class). Also, given that this has nothing to do with `n.m.structure` it should belong to `w.g.s`.~~

Given `w.g.s.StructureType` is a singleton it should follow `X` and `Xs` convention (`Block`/`Blocks`) instead of `X` and `XType` (`BlockEntity`/`BlockEntityType`). Problems arise from `Structure` being taken by `s.Structure` - thus I propose naming it `GeneratedStructure`/`GeneratedStructures`. This isn't far too stretched from vanilla registry, which is `worldgen/structure`. (Subclass names can stay.) `s.StructureType` should be moved to `w.g.s`.

apple502j:
Resolves #3182

Structures really need a revisit. In my opinion the net.minecraft.structure package should be gone and
everything should be moved to net.minecraft.world.gen.structure. That'd be a very big change, though.
This commit basically reverts the entirety of #3079 and contains the changes necessary to keep consistency
(but not other new changes). I could make it bigger, but this needs an ASAP merge to stop the bad names from
going to the release.

Summary of the renames:

- `structure.Structure` became `structure.StructureTemplate`. See `/place template`, datafixes, etc.
- `structure.StructureType` became `world.gen.structure.StructureType`.
- `world.gen.structure.StructureType` became `world.gen.structure.Structure`.
- `world.gen.structure.StructureTypes` became `world.gen.structure.Structures`.
- Other names referencing the renamed classes by their old name
- And missed "feature" renames (Fixes #3179)


apple502j:
Note: not impactful (for now), since this PR only touches less frequently used `Structure` and classes that already received a rename. If it gets to the late pre-release cycle or actual release, it will be impactful.

Shnupbups (approved):

modmuss50 (approved):

frqnny (approved):
I love these changes. Adding javadocs IMO would be even more amazing for modders trying to figure out doing custom structure stuffs with code rather than pure datapack, but this can be set for another pr.

Hugman76 (approved):

YanisBft (approved):

modmuss50:
Conflicts need fixing, and then I think this is good to go.

haykam821 (approved):

PR #3193, opened 2022-05-26
Add `yarnCommon` task that opens enigma with only common classes

deirn:
Noticed that there are some classes that don't have `@Environment(EnvType.CLIENT)` in `net.minecraft.client` package, this would help find those.

I can't help naming these though, as I've been using mojmap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,8 @@ Compared to launching Enigma externally, the gradle task adds a name guesser plu
 ### `yarnUnpicked`
 Same as above, but unpicks the constants and launches Enigma with them. Can be a little bit slower to get going.
 
+### `yarnCommon`
+Same as `yarn`, but will only show common classes.
 
 ### `build`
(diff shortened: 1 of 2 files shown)


modmuss50 (approved):
Looks reasonable, If needed we can add something to fail the build for common classes that are in the client only package.

apple502j:
@modmuss50 Please don't make it fail the build for now. We have 10 or so of them, and I believe a few are rename-impactful. (This was found by grepping the non-client source for the import, so I might've missed some classes.)

apple502j (approved):

liach:
> @modmuss50 Please don't make it fail the build for now. We have 10 or so of them, and I believe a few are rename-impactful. (This was found by grepping the non-client source for the import, so I might've missed some classes.)

No worries, this doesn't add validation that prevents common classes in `client` packages or vice versa.

liach (approved):

PR #3197, opened 2022-05-30
ChatDecorator -> TextDecorator

liach:
This can decorate any text given a player. Outdated docs are removed, and we can pull them from git history if we ever need them.


apple502j:
Okay, fine, I guess I should've renamed this when commands got proper decoration.

apple502j (changes requested):
I weakly-support the rename but oppose removing big chunks of the documentation.

Yes, naming-wise, this decorator is universal. You can apply it for any texts. However, Yarn has always been "modder-first" mapping. While I'm sure accuracy is the top priority (and the name "chat decorator" no longer fits neatly with the fix of several bugs related to non-player-invoked commands), this does not mean everything has to be super-generalized. We write the implementation notes and method usages for reasons.

There is another "for modders" thing, and it's here. This interface in question. This is specifically provided by Mojang to enhance server-side mods. Mods are expected, and encouraged should I say, to implement this interface. And in most cases, they do so to decorate messages (most notably incoming chat messages). This is unlike any other existing classes or interfaces - `text.Decoration`, while used by messaging, is not usually the one modders use, and it did not have any interesting properties that are specific to the vanilla usage. This one has more details to explain than being `BiFunction<ServerPlayerEntity, Text, CompletableFuture<Text>>`.

This PR removes most of the implementation details modders are expected to understand - and useful notes for them; What this is used for, and several common pitfalls (preview requirement, being not recommended for censoring, idempotence requirement and solution). Yes, some of the javadocs were outdated. But some, not all - and it is not a proper reason to remove all instead of rewriting.

Where did these documentations go? Git history. But git history is only accessible via git. It won't show up in javadoc website, or more importantly, the javadoc jar. It's not moved to other locations, it's removed. I'd like to also say that in my opinion this is too long to list on the method documentation for `MinecraftServer#method_43929`, and it'd be inconvenient to link to it as well.

That said, I think if we were to rename this, `MessageDecorator` could be another option. We have precedents over calling these stuff "messages", and this interface has deep relations with messaging (more than that of `Decoration`, etc). Given the release is coming soon, I'd suggest splitting this PR to the rename part and the javadoc part; or, I could submit my own PR with `MessageDecorator` name.

PR #3202, opened 2022-05-31
Repackage NameGenerator

apple502j:
This class is used by logical server.


liach (approved):
It's like a debug-oriented thing. util sounds like a good package.

Juuxel (approved):

haykam821 (approved):

PR #3205, opened 2022-06-01
Fix wrong word in MessageDecorator javadoc

apple502j:
> what exactly do you mean?
should it not be pure i.e. completely deterministic given the inputs?

\- Technician, 2022


apple502j:
LOL, just after I open PR it gets a conflict. Closing.

PR #3212, opened 2022-06-15
Add and fix mappings for structure placements

Machine-Maker:
Something odd is going on with `StructurePlacement#defaultShouldGenerate`
All those private static methods should have the exact same param names, and its very clear that the order is `seed, salt, chunkX, chunkZ, frequency` based on the call to `FrequencyReductionMethod#shouldGenerate`. But inside defaultShouldGenerate the params are put in the wrong order into the `ChunkRandom#setRegionSeed`. setRegionSeed might be misnamed, but I think its params are in the right order based on a call to it in `RandomSpreadStructurePlacement#getStartChunk`.

### DIFF
--- a/mappings/net/minecraft/world/gen/chunk/placement/StructurePlacement.mapping
+++ b/mappings/net/minecraft/world/gen/chunk/placement/StructurePlacement.mapping
@@ -1,5 +1,6 @@
 CLASS net/minecraft/class_6874 net/minecraft/world/gen/chunk/placement/StructurePlacement
 	FIELD field_36428 TYPE_CODEC Lcom/mojang/serialization/Codec;
+	FIELD field_37775 ARBITRARY_SALT I
 	FIELD field_37776 locateOffset Lnet/minecraft/class_2382;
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/world/gen/chunk/placement/StructurePlacement.mapping
```diff
@@ -59,8 +65,11 @@ CLASS net/minecraft/class_6874 net/minecraft/world/gen/chunk/placement/Structure
 			ARG 5 x
 			ARG 6 z
 	CLASS class_7153 GenerationPredicate
-		METHOD shouldGenerate (JIIIF)Z
+		METHOD shouldGenerate shouldGenerate (JIIIF)Z
```
apple502j:
```suggestion
		METHOD shouldGenerate (JIIIF)Z
```

Machine-Maker: done

apple502j (approved):

modmuss50 (approved):

PR #3219, opened 2022-06-19
Fix Type-O? getChtOpacity -> getChatOpacity

yurisuika:
Think that about sums it up.


liach (approved):

modmuss50 (approved):

apple502j:
Cspell does not trigger with 3-letter unknown words, probably why I didn't find it

PR #3234, opened 2022-07-07
[1.18.2] Update Buildscript

thecatcore:


.editorconfig
```diff
@@ -1,2 +1,2 @@
 [*.{gradle,mapping,unpick}]
-indent_style = tab
+indent_style = tab
```
liach, line 2: Don't know why these file ending changes happen, which doesn't quite mirror the current yarn head.

thecatcore: hm, the file ending was changed between 1.18.2 and HEAD at some point it seems like if you compare the two branches :thinking: .

thecatcore: So I should revert them?

src/constants/java/net/fabricmc/yarn/constants/package-info.java
```diff
@@ -1,13 +1,9 @@
-/*
- * This file is free for everyone to use under the Creative Commons Zero license.
- */
```
modmuss50, line 3: Why are you removing the license header?

modmuss50 (changes requested):

modmuss50:
Im not sure what the gloal here is? A lot of these changes were just required for 1.19. I guess the PR isnt wrong, im just a little unsure what the benefit of backporting this to 1.18 is?

thecatcore:
Simply keeping old main release buildscript updated and working. (1.18.2 being the first branch I'm updating)
Here enigma, cfr update + yarnCommon task.

src/constants/java/net/fabricmc/yarn/constants/package-info.java
thecatcore: fixed

modmuss50:
They should still work though? Im a little bit worried about merging buildscript changes like this (especially on branches that are even older than this) as it may have unintended side effects?

I appreciate the effort, but without any obvious benefit I don't believe its worth the time. The time could be much better spent improving the mainline buildscript?

Please let me know if you think otherwise.

modmuss50:
Im going to close this PR as I dont think its something we want to do. I would be more happy to backport specific changes if/when required.

PR #3254, opened 2022-07-23
Fix entity mapping

### LINKED ISSUE #3195: chance field in AnimalMateGoal should be speed
Yarn version `1.18.2+build.3`

The `chance` field of most `Goal`s is typically used to describe whether or not the `Goal` should start according to a random chance. However, in `AnimalMateGoal` it is used to describe the speed at which the animal moves towards a found mate, not the chance of whether or not it should mate.

`chance` is not used as a random chance to start:

`AnimalMateGoal`
```java
public boolean canStart() {
    if (!this.animal.isInLove()) {
        return false;
    } else {
        this.mate = this.findMate();
        return this.mate != null;
    }
}
 ```

`chance` is passed as the `speed` parameter to the method `EntityNavigation#startMovingTo`:

`AnimalMateGoal`
```java
public void tick() {
    this.animal.getLookControl().lookAt(this.mate, 10.0F, (float)this.animal.getMaxLookPitchChange());
    this.animal.getNavigation().startMovingTo(this.mate, this.chance);
    ++this.timer;
    if (this.timer >= this.getTickCount(60) && this.animal.squaredDistanceTo(this.mate) < 9.0D) {
        this.breed();
    }
}
```

### LINKED ISSUE #3209: The EatSweetBerriesGoal class handles glow berries
- Class name should be generic to include all kinds of berries
- Foxes only pick berries as part of this goal; eating happens during movement (`tickMovement` method)
- The `eatSweetBerry` method should be renamed for the above two reasons
- The `EATING_TIME` field represents the number of ticks before picking finalizes

### LINKED ISSUE #3240: "nether portal cooldown" in Entity is used for end gateways as well

### LINKED ISSUE #3242: Entity#collides needs a better name
`Entity#collides` is confusing - it is used by three things: client-side entity targetting, boat placement check, and projectile collision.

There is also `Entity#isCollidable`, which is "can you stand on the entity without falling?" (`true` for boats and shulkers). `EntityView` methods check `intersectionChecked` field.

Any better name for this?

### LINKED ISSUE #3244: Entity#getItemsHand should be getHandItems
This returns item stacks, not hands.

### LINKED ISSUE #3245: Entity#shouldLeaveSwimmingPose -> isCrawling
While the behavior is technically the same (returns true if swimming but not touching water), its usage suggests this is not used for checking whether to leave swimming pose. This is used to modify the player's speed when the player is under such conditions.

The general community consensus is that the feature is called "crawling" - we should adopt this name.

### LINKED ISSUE #3246: Entity#hasPassengerType should use overload instead
This method does not take EntityType as the name suggests; instead it takes `Predicate<Entity>`. It should be named `hasPassenger` (overload with existing method taking Entity).

apple502j:
Resolves #3195
Resolves #3209
Resolves #3240 
Resolves #3242 
Resolves #3244 
Resolves #3245 
Resolves #3246

### DIFF
--- a/mappings/net/minecraft/entity/passive/FoxEntity.mapping
+++ b/mappings/net/minecraft/entity/passive/FoxEntity.mapping
@@ -122,7 +122,7 @@ CLASS net/minecraft/class_4019 net/minecraft/entity/passive/FoxEntity
 	CLASS class_4024 MateGoal
 		METHOD <init> (Lnet/minecraft/class_4019;D)V
 			ARG 2 chance
-	CLASS class_4025 EatSweetBerriesGoal
+	CLASS class_4025 EatBerriesGoal
 		FIELD field_17974 timer I
 		FIELD field_30336 EATING_TIME I
 		METHOD <init> (Lnet/minecraft/class_4019;DII)V
--- a/mappings/net/minecraft/entity/passive/FoxEntity.mapping
+++ b/mappings/net/minecraft/entity/passive/FoxEntity.mapping
@@ -129,7 +129,7 @@ CLASS net/minecraft/class_4019 net/minecraft/entity/passive/FoxEntity
 			ARG 2 speed
 			ARG 4 range
 			ARG 5 maxYDifference
-		METHOD method_18307 eatSweetBerry ()V
+		METHOD method_18307 eatBerries ()V
 		METHOD method_33586 pickGlowBerries (Lnet/minecraft/class_2680;)V
 			ARG 1 state
 		METHOD method_33587 pickSweetBerries (Lnet/minecraft/class_2680;)V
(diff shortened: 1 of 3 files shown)


mappings/net/minecraft/entity/passive/FoxEntity.mapping
```diff
@@ -122,7 +122,7 @@ CLASS net/minecraft/class_4019 net/minecraft/entity/passive/FoxEntity
 	CLASS class_4024 MateGoal
 		METHOD <init> (Lnet/minecraft/class_4019;D)V
 			ARG 2 chance
-	CLASS class_4025 EatSweetBerriesGoal
+	CLASS class_4025 EatBerriesGoal
```
Shnupbups, line 125: The `eatSweetBerry` method within this class should also be changed, as it applies to glow berries too.

Shnupbups (approved):

modmuss50 (approved):

PR #3252, opened 2022-07-20
1.19.1-pre6/rc2 networking

apple502j:
Big stuff includes message decorator caching, etc

PR #3259, opened 2022-07-25
Misleading ChatHud#render

IMB11:
TickDelta is incorrectly labeled, it is the *total amount of ticks in the render cycle that have passed.*

Grain of salt on that explanation, it's hard to explain the values because it's sporadic and does not have a clear pattern from a brief glance.


apple502j:
`tickDelta` is one of the parameters frequently used, so if we do need to fix it we should fix that in all places.

Besides, if it is "the amount of ticks that have passed since the last invokation(sic)", isn't it the very definition of delta? That for `T(n)` where `n` is the number of invocations and `T(n)` is the total ticks counted from the epoch, `tickDelta = T(n+1) - T(n)`.

Linguardium:
normally tickDelta is "the partial tick that has passed between render frames" where in this instance it doesnt match the tickDelta from the calling parent method but the tick counter in the parent class

apple502j:
Ah, that makes sense. I'll rename it then.

apple502j:
Done in #3252

PR #3260, opened 2022-07-26
1.19.1-rc3 networking

apple502j:


mappings/net/minecraft/server/filter/FilterMask.mapping
```diff
@@ -0,0 +1,25 @@
+CLASS net/minecraft/class_7649 net/minecraft/server/filter/FilterMask
```
apple502j: TODO: move to `network/message`, used on both sides

Issue #2594, opened 2021-07-19
MathHelper.lerpFromProgress -> map


Juuxel:
and `clampedLerpFromProgress` -> `clampedMap`.

These functions take five values `lerpValue, lerpStart, lerpEnd, start, end` and map the `lerpValue` from the first range (`lerpStart..lerpEnd`) to the second (`start..end`). Additionally, `clampedLerpFromProgress` clamps it to the new range.

`lerpFromProgress` is a very confusing name for this method and doesn't really describe what it does. Processing has an equivalent method, which they call [`map`](https://processing.org/reference/map_.html), and I think that is much more descriptive. In any case, we should document the method to clarify it.

Juuxel:
Also, `MathHelper.getLerpProgress(value, from, to)` can be considered normalising a number from a range `from..to` to 0..1. (Processing calls this [`norm`](https://processing.org/reference/norm_.html) for that reason.)

Though I'm not sure whether renaming it is that useful since you can also think of it as the lerp delta between `from` and `to`.

mschae23:
I support this rename. "lerp from progress" is really confusing, and I have to look up the method to see in what order the parameters are each time. `map` sounds a lot simpler.

PR #3278, opened 2022-08-14
Rename MathHelper.(clamped)lerpFromProgress -> (clamped)map, add docs

Juuxel:
- `MathHelper.lerpFromProgress` -> `map`
- `MathHelper.clampedLerpFromProgress` -> `clampedMap`
- Documented both methods and their overloads

`map` is the name that seems to be generally used for this function in game programming, see for example [Processing](https://processing.org/reference/map_.html) and [Unreal Engine](https://docs.unrealengine.com/5.0/en-US/BlueprintAPI/Math/Float/MapRangeUnclamped/).


apple502j (approved):
Seems not so impactful after quick search

modmuss50 (approved):

haykam821:
Fixes #2594

Juuxel:
Thanks haykam, I forgot I already had made an issue for this

Issue #3051, opened 2022-03-17
Transformation name


liach:
Now we get two `Transformation` class, one for JSON model perspective transforms and one for entity model animation. There are other similar names like `AffineTransform` that is about transformations in rendering. Might want to define "transform" to be one specific thing than being ambiguous as it is right now.

PR #3285, opened 2022-09-03
Math helper consts / Animation mappings update

Sollace:
This started with me wanting to just map the unknown fields in MathHelper, but then it led me to mapping the methods in there as well.

They're not used by anything (yet) so I've tried my best to document them based on what I can see them doing internally.

That also led me to working on the animation packages. Renamed some packages to be a little more consistent.

i.e. `Transformation` -> `BoneAnimation` (based on unobfuscated field names in `Animation`)

And also `Animation` -> `Timeline` (timeline being the technical term for a sequence of animations. Seems to remove some ambiguity of calling them both "animations")

Also renamed the corresponding fields in the same package (did not look outside of render/entity/animation - that can rather be another PR) and named some methods that were unmapped based on their usage and what I've learned from looking at the other animation classes whilst doing this PR.


mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -13,8 +13,11 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
 	FIELD field_29847 RADIANS_PER_DEGREE F
 	FIELD field_29848 DEGREES_PER_RADIAN F
 	FIELD field_29849 EPSILON F
+	FIELD field_29850 FLOOR_DOWNSHIFT I
+	FIELD field_29851 FLOOR_UPSHIFT F
 	FIELD field_29853 HALF_PI_RADIANS_SINE_TABLE_INDEX J
 	FIELD field_29856 DEGREES_TO_SINE_TABLE_INDEX F
+	FIELD field_29859 SINE_TABLE_SIZE I
```
enbrain, line 20: This field is the size of `ARCSINE_TABLE` and `COSINE_TABLE` (which have size `257`). `SINE_TABLE` has the size of `65536`.

mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -388,6 +416,12 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
 			COMMENT the fallback for unparsable {@code string}
 		ARG 2 min
 			COMMENT the minimum if the parsed value is too small
+	METHOD method_34950 getRandomIndex (Lnet/minecraft/class_5819;[D)I
+		COMMENT Picks a random value from a list
+		COMMENT
+		COMMENT @return The index of the random pick.
+		ARG 0 random
+		ARG 1 values
```
enbrain, line 424: This looks more like an array of cumulative probabilities than the associated values themselves.

mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -399,6 +433,11 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
 		ARG 0 n
 	METHOD method_34956 perlinFadeDerivative (D)D
 		ARG 0 value
+	METHOD method_34957 deviate (D)D
+		COMMENT Creates a random small, positive deviation from the passed in value.
```
enbrain, line 437: Not sure if I call it a positive deviation. `method_34957(216880)` returns `216879.99999996054`.

mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -399,6 +433,11 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
...
+		COMMENT
+		COMMENT @return The passed in value + some random amount < 0.01
```
enbrain, line 439: The absolute value of the random amount is much more smaller than `0.01` since it's multiplied by `1.0E-7`.

mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -360,6 +363,31 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
...
+	METHOD method_34940 generateInverseExponents (DDDDDDII)[D
+		ARG 0 scale1
+		ARG 2 normalization1
+		ARG 4 exponent1
+		ARG 6 scale2
+		ARG 8 normalization2
+		ARG 10 exponent2
+		ARG 12 startIndex
+		ARG 13 endIndex
+	METHOD method_34941 generateInverseExponents (DDDII)[D
```
enbrain, line 375: This looks like a normal distribution except you can specify the scale.

mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -360,6 +363,31 @@ CLASS net/minecraft/class_3532 net/minecraft/util/math/MathHelper
 		ARG 0 value
 		ARG 1 min
 		ARG 2 max
+	METHOD method_34940 generateInverseExponents (DDDDDDII)[D
```
enbrain, line 366: This looks like a mixture distribution of two normal distribution.

enbrain:
Transformation rename relates to #3051.

apple502j:
Would like to have class rename split in a separate PR

mappings/net/minecraft/util/math/MathHelper.mapping
Sollace: It's the size for the ARC**SINE** and CO**SINE** tables. I picked the name because they both have "SINE" in the name, not because it belongs to the other SINE table.

mappings/net/minecraft/util/math/MathHelper.mapping
Sollace: The times I ran it only seemed to return values larger than the input. Might be a selection biase I guess.

mappings/net/minecraft/util/math/MathHelper.mapping
Sollace: Seems I misinterpreted this function. Thought it was going `random.nextDouble(initialValue * 3000)` but it was actually using `initialValue * 3000` as the _seed_. So it's not clear what this method is any more.

Sollace:
@apple502j 
> Would like to have class rename split in a separate PR

https://github.com/FabricMC/yarn/pull/3287

mappings/net/minecraft/util/math/MathHelper.mapping
Juuxel: It's still confusing since you'd assume `SINE_TABLE_SIZE` is the size of `SINE_TABLE`.

PR #3296, opened 2022-09-18
Move filament into yarn repo.

modmuss50:
First PR to just move it, further improvements will be made later.


thecatcore:
Hmm, wondering how it will be possible to use the logic outside of yarn repo 🤔.

modmuss50:
Is this something you do? My short/medium term plan is to move most if not all of the logic out of the build.gradle into properly structured and tested tasks under build-logic.

As filament is a seperate repo it slows down the development quite a bit, yarn is quite uniquie as we mostly push the toolchain forward on one branch there isnt much benefit to us having a seperate repo + release for it.

If you do use filament and more importantly plan to keep up with yarn's toolchain changes then this might need reconsidering. 

Another option may be to also release the build logic as a standalone gradle plugin, I would need to invistage to see if this is possible.

Juuxel:
> Another option may be to also release the build logic as a standalone gradle plugin, I would need to invistage to see if this is possible.

Yep, it should be perfectly possible since included builds are just normal Gradle builds that you run alongside the "main" build. (There's nothing special about build-logic.) You could even still call it "Filament" for releasing purposes :wink:

modmuss50:
> You could even still call it "Filament" for releasing purposes

Thats a good idea, ill rename this from build-logic back to filament, will most likely sort out the publishing later once its a bit more finalised.

thecatcore:
Awesome then 👍

PR #3300, opened 2022-09-19
Move jar remapping to Filament

Juuxel:
TODO:
- [x] switch libraryDir to a file collection

### DIFF
--- a/filament/src/main/java/net/fabricmc/filament/task/MapJarTask.java
+++ b/filament/src/main/java/net/fabricmc/filament/task/MapJarTask.java
@@ -32,26 +32,13 @@ import net.fabricmc.tinyremapper.TinyRemapper;
 import net.fabricmc.tinyremapper.TinyUtils;
 
 public abstract class MapJarTask extends DefaultTask {
-	@InputFile
-	public abstract RegularFileProperty getInput();
-
-	@InputFile
-	public abstract RegularFileProperty getMappings();
-
-	@OutputFile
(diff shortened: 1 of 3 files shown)


filament/src/main/java/net/fabricmc/filament/task/MapJarTask.java
```diff
@@ -0,0 +1,131 @@
...
+
+import net.fabricmc.tinyremapper.OutputConsumerPath;
+import net.fabricmc.tinyremapper.TinyRemapper;
+import net.fabricmc.tinyremapper.TinyUtils;
+
+public abstract class MapJarTask extends DefaultTask {
+	@InputFile public abstract RegularFileProperty getInput();
+	@InputFile public abstract RegularFileProperty getMappings();
+	@OutputFile public abstract RegularFileProperty getOutput();
+	@Classpath public abstract ConfigurableFileCollection getClasspath();
```
modmuss50, line 38: I didnt know @Classpath is a thing 👍

modmuss50 (approved):
Looks great, thanks 👍

filament/src/main/java/net/fabricmc/filament/task/MapJarTask.java
Juuxel: Heh, see the table here: <https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:up_to_date_checks>

There are some other interesting ones like `@CompileClasspath` and `@Console` too

PR #3311, opened 2022-10-08
Map NbtLongArray.set

asablock:


apple502j:
This is already mapped by the superclass, I believe.

PR #3314, opened 2022-10-10
Map some item constants

YanisBft:


mappings/net/minecraft/item/WrittenBookItem.mapping
```diff
@@ -1,5 +1,9 @@
 CLASS net/minecraft/class_1843 net/minecraft/item/WrittenBookItem
+	FIELD field_30929 MAX_TITLE_SELECTION_LENGTH I
 	FIELD field_30930 MAX_TITLE_LENGTH I
+	FIELD field_30931 MAX_PAGE_SELECTION_LENGTH I
+	FIELD field_30932 MAX_PAGES_LENGTH I
```
liach: `MAX_PAGE_LENGTH`?

mappings/net/minecraft/item/BowItem.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS net/minecraft/class_1753 net/minecraft/item/BowItem
+	FIELD field_30855 TICKS_PER_SECOND I
```
liach, line 2: Is this the only usage of `20` in this class? This looks like a weird usage, and I expect the actual usage to be more specific.

mappings/net/minecraft/item/CrossbowItem.mapping
```diff
@@ -2,6 +2,9 @@ CLASS net/minecraft/class_1764 net/minecraft/item/CrossbowItem
 	FIELD field_30863 RANGE I
 	FIELD field_30864 CHARGED_KEY Ljava/lang/String;
 	FIELD field_30865 CHARGED_PROJECTILES_KEY Ljava/lang/String;
+	FIELD field_30866 DEFAULT_PULL_TIME I
+	FIELD field_30869 DEFAULT_SPEED F
+	FIELD field_30870 FIREWORK_ROCKET_SPEED F
```
enbrain, line 7:
```suggestion
	FIELD field_30869 ARROW_SPEED F
	FIELD field_30870 FIREWORK_ROCKET_SPEED F
```
Since a crossbow only accepts arrows or firework rockets.

mappings/net/minecraft/item/WrittenBookItem.mapping
enbrain: Does this really have something to do with selection? It seems that the game just uses the different length limit for edit and view.

mappings/net/minecraft/item/BowItem.mapping
enbrain: Could this be related to the literal `20` in `BowAttackGoal#tick`? This might be the minimum number of ticks `x` that maximizes `BowItem#getPullProgress(x)`.

mappings/net/minecraft/item/WrittenBookItem.mapping
YanisBft: It's the max length of the `Pages` nbt key, hence the name

enbrain: No. It's the max length of an element in the `Pages` nbt key. `getString` is an index access.

image

mappings/net/minecraft/item/CrossbowItem.mapping
YanisBft: The thing is there's no mention of arrows in the method using these constants, hence "default speed"
image

mappings/net/minecraft/item/WrittenBookItem.mapping
YanisBft: Totally misread the code, changed in latest commit

mappings/net/minecraft/item/BowItem.mapping
YanisBft: Didn't see the `20` in `BowAttackGoal#tick`, but they mean probably the same thing. "ticks per second" is a bad name, do you think of any better?

apple502j (approved):

mappings/net/minecraft/item/BowItem.mapping
enbrain: I can't think of any at the moment. I guess `TICKS_PER_SECOND` wouldn't hurt as it's a constant anyway.

mappings/net/minecraft/item/CrossbowItem.mapping
enbrain: Sure.

enbrain (approved):

mappings/net/minecraft/item/WrittenBookItem.mapping
YanisBft: You're right, changed

PR #3319, opened 2022-10-14
Update BirdNavigation.mapping, QOL patch

avetharun:
Added a mapping for a duplicate method for BirdNavigation::canEnterOpenDoors (QOL patch)

Reference: 
image

```java

    public boolean canEnterOpenDoors() {
        return this.nodeMaker.canEnterOpenDoors();
    }
    ...
    public boolean method_35129() {
        return this.nodeMaker.canEnterOpenDoors();
    }
```


Juuxel:
The names should generally reflect the usage in these cases, not the fact that they're duplicated.

See the various entity/player methods that check whether the player is sneaking but have different meanings.

avetharun:
> The names should generally reflect the usage in these cases, not the fact that they're duplicated.
> 
> See the various entity/player methods that check whether the player is sneaking but have different meanings.

I figured as much, but it's strange seeing a unnamed method and not knowing what it does.
Apologies, I'll cancel this PR now

PR #3338, opened 2022-10-21
22w42a networking

apple502j:
Includes packets, server networking, message changes, client networking, abuse report, etc.


mappings/net/minecraft/network/encryption/NetworkEncryptionUtils.mapping
```diff
@@ -31,6 +31,7 @@ CLASS net/minecraft/class_3515 net/minecraft/network/encryption/NetworkEncryptio
 	FIELD field_39039 RSA_PUBLIC_KEY_SUFFIX Ljava/lang/String;
 	FIELD field_39109 SHA256_WITH_RSA Ljava/lang/String;
 	FIELD field_39272 BASE64_ENCODER Ljava/util/Base64$Encoder;
+	FIELD field_40111 SHA_256_BITS I
```
haykam821: The field above doesn't separate `SHA-256` using an underscore:

```suggestion
	FIELD field_40111 SHA256_BITS I
```

mappings/net/minecraft/network/packet/s2c/play/ProfilelessChatMessageS2CPacket.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS net/minecraft/class_7827 net/minecraft/network/packet/s2c/play/ProfilelessChatMessageS2CPacket
```
haykam821, line 1: The term 'profileless' is a bit odd but I'm not familiar enough with the code to suggest an improvement.

mappings/net/minecraft/network/packet/s2c/play/PlayerRemovalS2CPacket.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS net/minecraft/class_7828 net/minecraft/network/packet/s2c/play/PlayerRemovalS2CPacket
```
haykam821: Shouldn't this be named `PlayerRemoveS2CPacket` to match other packet names?

mappings/net/minecraft/network/packet/s2c/play/ProfilelessChatMessageS2CPacket.mapping
apple502j: This is used for chat messages sent through command in a way that makes it unverifiable - such as `/execute as entity` or command block. They are not associated with a sender profile, thus its name. (Is in use since 1.19, so not a new name)

modmuss50 (approved):

PR #3340, opened 2022-10-21
22w42a worldgen

apple502j:


mappings/net/minecraft/world/gen/chunk/placement/StructurePlacement.mapping
```diff
@@ -27,6 +27,8 @@ CLASS net/minecraft/class_6874 net/minecraft/world/gen/chunk/placement/Structure
 		ARG 5 frequency
 	METHOD method_41636 getLocatePos (Lnet/minecraft/class_1923;)Lnet/minecraft/class_2338;
 		ARG 1 chunkPos
+	METHOD method_41637 buildCodec (Lcom/mojang/serialization/codecs/RecordCodecBuilder$Instance;)Lcom/mojang/datafixers/Products$P5;
+		ARG 0 instance
```
Juuxel, line 31: Do we generally call this `instance` in Yarn? That name is a bit vague since it follows DFU's naming pattern of calling all typeclass implementation classes `Instance` and placing them inside the target class.

(I've sometimes called it `builder` in my own code but that isn't exactly accurate either since it works with RecordCodecBuilders and is not an RCB itself)

apple502j: Hmm, there isn't a good alternative, and it's already used a million times... maybe it's ok?

mappings/net/minecraft/world/gen/chunk/BlendingData.mapping
```diff
@@ -56,15 +56,21 @@ CLASS net/minecraft/class_6749 net/minecraft/world/gen/chunk/BlendingData
...
 		ARG 2 chunkZ
+	METHOD method_39571 (Lcom/mojang/serialization/codecs/RecordCodecBuilder$Instance;)Lcom/mojang/datafixers/kinds/App;
+		ARG 0 instance
 	METHOD method_39572 initChunkBlendingData (Lnet/minecraft/class_2791;Ljava/util/Set;)V
 		ARG 1 chunk
 		ARG 2 newNoiseChunkDirections
 	METHOD method_39573 validate (Lnet/minecraft/class_6749;)Lcom/mojang/serialization/DataResult;
 		ARG 0 data
+	METHOD method_39574 ([D)V
+		ARG 0 heights2
```
mschae23, line 73: Is this the convention for situations like this?

mschae23 (approved):

mappings/net/minecraft/world/gen/chunk/placement/StructurePlacement.mapping
Juuxel: Yep, seems fine enough.

mappings/net/minecraft/world/gen/chunk/BlendingData.mapping
apple502j: This is a weird situation, but yes we append `2` or `3` in case of conflicts (most often seen as `buf2`).

modmuss50 (approved):

PR #3341, opened 2022-10-21
22w42a blocks and items

apple502j:
Renames `WallStandingBlockItem` because hanging signs don't stand.


haykam821 (approved):

Juuxel (approved):
The `WallStandingBlockItem` rename might be slightly impactful (for example Adorn uses it) but the old name is inaccurate so it can go.

YanisBft (approved):

Juuxel (approved):

PR #3344, opened 2022-10-22
Add documentation for ToolMaterial & ToolItem classes

Draylar:
`ToolMaterial`, `ToolMaterials`, and `ToolItem` documentation. I was able to duplicate a bit between this & the `ArmorMaterial` PR, so if any feedback is left on either PR for those methods, I will change it on both PRs. Thanks!


apple502j (commented):
Inline return maybe?

YanisBft (approved):

apple502j (approved):

PR #3345, opened 2022-10-22
Rename StackAdder to Entries

modmuss50:
Happy to hear other suggestions.


apple502j (approved):
Yeah, think this is better.

Shnupbups (approved):

PR #3352, opened 2022-10-22
22w42a client

### LINKED ISSUE #3282: `UnclampedModelPredicateProvider` should be renamed to `ClampedModelPredicateProvider`
It actually clamps the value of the `unclampedCall` method to `[0, 1]`.

### LINKED ISSUE #3303: BipedEntityModel.setAttributes -> copyBipedStateTo
It's an extended version of `copyStateTo` from a superclass:

The current name is a bit bad and sounds like you're setting the "attributes" (whatever that is) of `this` model to the parameter, while it's the opposite.

I'm thinking `copyBipedStateTo` because it's a biped-specific version of `copyStateTo`. It could also be `copyArmorStateTo` since it's only called from armor code iirc.

apple502j:
New client stuff (such as player list, remaining resource stuff, etc) along with package changes for mispackaged common source classes.

Resolves #3282
Resolves #3303


mappings/net/minecraft/client/particle/ParticleManager.mapping
```diff
@@ -91,10 +95,24 @@ CLASS net/minecraft/class_702 net/minecraft/client/particle/ParticleManager
 		ARG 2 count
 	METHOD method_34023 (Lnet/minecraft/class_5878;)V
 		ARG 1 group
+	METHOD method_45764 (Lnet/minecraft/class_2960;)Lnet/minecraft/class_2960;
+		ARG 0 textureId
+	METHOD method_45766 (Lnet/minecraft/class_3695;Ljava/util/concurrent/CompletableFuture;Ljava/util/concurrent/CompletableFuture;Ljava/lang/Void;)V
+		ARG 1 void_
```
YanisBft: Doesn't seem like the right index

apple502j: Blame Enigma :p

apple502j:
```suggestion
```

YanisBft (approved):

modmuss50 (approved):

PR #3359, opened 2022-10-25
Rename CommandBlockItem

apple502j:
This is also used by structure block and jigsaw block items.


haykam821 (approved):
This class could benefit from some simple documentation but the rename itself looks good.

Shnupbups (approved):

PR #3367, opened 2022-10-27
new ButtonWidget stuff

RedLime:
Written Builder and NarrationSupplier class
and field_25035(`EMPTY`) has been changed to `EMPTY_TOOLTIP` because now ButtonWidget class have several static variables


RedLime:
working on another classes wait a minutes please :)

PR #3379, opened 2022-11-03
22w44a worldgen

### LINKED ISSUE #3170: Registry initializer method names post-22w19a

### LINKED ISSUE #3177: HoglinStableData should be renamed to BastionHoglinStableData

apple502j:
Resolves #3177


Shnupbups (approved):
Looks good, but worth noting this doesn't fully resolve #3170 as many methods outside the world package use the `registerAndGetDefault` name still.

PR #3384, opened 2022-11-04
Refactor the program and shader classes

### LINKED ISSUE #3128: `Program` and `Shader` are confusing
Currently, `class_281` is named `net.minecraft.client.gl.Program` and `class_5944` is named `net.minecraft.client.render.Shader`.

`class_281` is actually a single shader stage in the graphics pipeline. In other words, it's a shader object that can be created by `glCreateShader`. Calling it "program" is only correct in the sense that it's a program that runs on a GPU. `class_5944` is actually a shader program (with some uniform variables). In other words, it's a program object that can be created by `glCreateProgram`. Abbreviating "shader program" as "shader" is plainly wrong.

The fact that complicates this stuff is that Minecraft calls `class_5944` "shader" in strings (and this is probably why yarn contributors decided to call it "shader" as well). The most notable one is the `RenderSystem#getShader` method, which is not obfuscated by Mojang. For that reason, simply swapping `Shader` and `Program` will introduce another confusion.

I propose renaming `Shader` to `ProgramObject` and renaming `Program` to `ShaderObject` with some warning javadoc that explicitly says Mojang's names are misleading. The `Object` suffix makes it clear that it is using OpenGL terms, not names Mojang uses.

Apparently, this has caused too much trouble for shader/rendering engine mod developers.

enbrain:
Resolves #3128. Highly impactful, especially the method renames in `GameRenderer` as they are API part of the rendering engine.

Main renames are:

* `nm.client.render.Shader` → `nm.client.gl.ShaderProgram`: See the issue. Moved to the `gl` package as it directly interacts with GL.
* `Program` → `ShaderStage`: See the issue.
* `GlShader` → `ShaderProgramSetupView`: This interface is only used for linking or deleting shader programs.
* `ShaderEffect` → `PostEffectProcessor`: This class applies multiple post-processing shader programs to the rendering result.
* `GameRenderer.SHADER_LOCATIONS` → `SUPER_SECRET_SETTING_PROGRAMS`: This field is an implementation of the Super Secret Settings feature that was removed in 1.9. The 1.8.9 equivalent is the `ab` field in the `bfk` class.


mappings/net/minecraft/client/gl/ShaderProgramParseException.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_2973 net/minecraft/client/gl/ShaderParseException
+CLASS net/minecraft/class_2973 net/minecraft/client/gl/ShaderProgramParseException
```
apple502j: I believe this actually has nothing to do with shaders. When I last checked, it was in common jar.

enbrain: Renamed to `InvalidHierarchicalFileException`. Let me know if there is a better name.

apple502j (approved):

modmuss50 (approved):

modmuss50:
Merged to try and clear the build failuire

Shnupbups (approved):

mappings/net/minecraft/client/render/BufferRenderer.mapping
```diff
@@ -3,8 +3,8 @@ CLASS net/minecraft/class_286 net/minecraft/client/render/BufferRenderer
 	COMMENT BufferBuilder}.
 	FIELD field_38982 currentVertexBuffer Lnet/minecraft/class_291;
 	METHOD method_34420 reset ()V
-	METHOD method_43433 drawWithShader (Lnet/minecraft/class_287$class_7433;)V
-		COMMENT Draws {@code buffer} using the shader specified with {@link
+	METHOD method_43433 drawWithGlobalProgram (Lnet/minecraft/class_287$class_7433;)V
```
apple502j, line 6: @enbrain I wonder if the former name is better, "shader" makes more sense in this context. Should we revert this?

enbrain: Can you elaborate why?

apple502j: "Shader" is less ambiguous than "program". After all, a program is a set of shaders.

enbrain: Imo, it's expected knowledge that "program" means a shader program in this context. If ambiguity is really an issue, `drawWithGlobalShaderProgram` would be better. The point of this refactor is that a mere "shader" should never be used to refer to a shader program.

On a side note, almost all content mods shouldn't call this method directly anyway. There are higher level API `Tessellator.draw` and `RenderLayer.draw`.

SplashAni:
what is Vertexer now

PR #3385, opened 2022-11-06
Rename ScreenHandler.transferSlot -> quickMove

### LINKED ISSUE #2944: ScreenHandler.transferSlot -> quickMoveFromSlot
This is only used for implementing `SlotActionType.QUICK_MOVE` in `internalOnSlotClick`, and in all classes where it's overridden, it implements the quick move logic. The `int index` parameter is the index of the slot where the items will be transferred out of (if possible).

This refactor could be impactful, but IMO it's needed so people can understand this better. This method should also be documented - especially the `ItemStack` return value.

## Return value (yes, this deserves its own section)
The returned item stack is **really weird**. The default implementation in `ScreenHandler` just always returns the original stack in the slot, which causes an infinite loop in `internalOnSlotClick` (see diagram below). The method is overridden in all `ScreenHandler` subclasses, however, and there the return value looks like this:

- If any items were moved out of the slot, it returns the original item stack in the slot.
- If no items could be moved, it returns `ItemStack.EMPTY`.

## Diagram

If this is unclear (and it probably is), here's a diagram of approximately what happens. This is simplified a bit to fit the diagram, but all the relevant details should be correct.

> Note: `onSlotClick` refers to `internalOnSlotClick`

diagram of quick moving items

Juuxel:
Fixes #2944.

The current name is very vague, and `quickMove` describes the behaviour more accurately. There also doesn't seem to be much opposition to it based on my issue's reactions.

The name comes from `SlotActionType.QUICK_MOVE`.


modmuss50 (approved):
LGTM, please allow extra time for others to leave their feedback.

YanisBft (approved):
Makes more sense

apple502j (approved):

enbrain (approved):

modmuss50:
No one seems to be objecting, I will merge it 👍

modmuss50 (approved):

PR #3387, opened 2022-11-07
Remove wrapper-validation-action as its unstable, replace with distributionSha256Sum

modmuss50:
See https://docs.gradle.org/current/userguide/gradle_wrapper.html


apple502j (approved):
Tbh, given the current PR practice I believe validation serves no purpose, but hey it's for security :tm:

modmuss50:
I agree, the build script downloads loads of unverified libraries/plugins in just the same was as the gradle wrapper. In the future I may look into improving that but its not a major issue.

PR #3390, opened 2022-11-09
class_7923 -> Registries, constants

Benonardo:
Class name and package may not be ideal


modmuss50 (approved):

mappings/net/minecraft/Registries.mapping
```diff
@@ -0,0 +1,71 @@
+CLASS net/minecraft/class_7923 net/minecraft/Registries
```
enbrain: I think this class should be in the `net/minecraft/util/registry` package.

enbrain (commented):
Can you map `field_41175`? The auto-proposed name `AIR` is extremely confusing.

mappings/net/minecraft/Registries.mapping
```diff
@@ -0,0 +1,69 @@
+CLASS net/minecraft/class_7923 net/minecraft/util/registry/Registries
```
enbrain: This file should be moved to `mappings/net/minecraft/util/registry/Registries.mapping`. I highly recommend you use Enigma when editing mappings.

Benonardo: Good idea, probably won't make the same mistake twice :+1:

enbrain (approved):

modmuss50 (approved):

PR #3391, opened 2022-11-09
22w45a gui stuff

MattiDragon:
Hope I didn't break git too bad here

### DIFF
--- a/mappings/net/minecraft/client/gui/screen/Screen.mapping
+++ b/mappings/net/minecraft/client/gui/screen/Screen.mapping
@@ -22,6 +22,7 @@ CLASS net/minecraft/class_437 net/minecraft/client/gui/screen/Screen
 	FIELD field_33822 narrator Lnet/minecraft/class_6385;
 	FIELD field_33823 elementNarrationStartTime J
 	FIELD field_33824 screenNarrationStartTime J
+	FIELD field_41105 tooltips Ljava/util/List;
 	METHOD <init> (Lnet/minecraft/class_2561;)V
 		ARG 1 title
 	METHOD method_25393 tick ()V
@@ -169,6 +170,17 @@ CLASS net/minecraft/class_437 net/minecraft/client/gui/screen/Screen
 	METHOD method_40041 hide ([Lnet/minecraft/class_339;)V
 		ARG 0 widgets
 	METHOD method_41843 clearAndInit ()V
+	METHOD method_47412 setTooltips (Lnet/minecraft/class_7919;)V
+		ARG 1 tooltip
+	METHOD method_47413 renderWithTooltips (Lnet/minecraft/class_4587;IIF)V
+		ARG 1 matrices
+		ARG 2 mouseX
+		ARG 3 mouseY
+		ARG 4 delta
+	METHOD method_47414 setTooltips (Ljava/util/List;)V
+		ARG 1 tooltips
+	METHOD method_47415 setTooltips (Lnet/minecraft/class_2561;)V
+		ARG 1 text
 	CLASS class_6390 SelectedElementNarrationData
 		FIELD field_33825 selectable Lnet/minecraft/class_6379;
 		FIELD field_33826 index I
--- a/mappings/net/minecraft/client/gui/screen/Screen.mapping
+++ b/mappings/net/minecraft/client/gui/screen/Screen.mapping
@@ -171,16 +171,16 @@ CLASS net/minecraft/class_437 net/minecraft/client/gui/screen/Screen
 		ARG 0 widgets
 	METHOD method_41843 clearAndInit ()V
 	METHOD method_47412 setTooltip (Lnet/minecraft/class_7919;)V
-		ARG 1 tooltips
+		ARG 1 tooltip
 	METHOD method_47413 renderWithTooltip (Lnet/minecraft/class_4587;IIF)V
 		ARG 1 matrices
 		ARG 2 mouseX
 		ARG 3 mouseY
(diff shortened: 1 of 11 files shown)


mappings/net/minecraft/client/gui/screen/Screen.mapping
```diff
@@ -169,6 +170,17 @@ CLASS net/minecraft/class_437 net/minecraft/client/gui/screen/Screen
 	METHOD method_40041 hide ([Lnet/minecraft/class_339;)V
 		ARG 0 widgets
 	METHOD method_41843 clearAndInit ()V
+	METHOD method_47412 setTooltips (Lnet/minecraft/class_7919;)V
+		ARG 1 tooltip
```
apple502j, line 174: Singular or plural?

apple502j (changes requested):

mappings/net/minecraft/client/gui/screen/Screen.mapping
MattiDragon: Enigma was rejecting singular so I assumed it was already in use

apple502j: @MattiDragon No, the param name

MattiDragon: Oh, yeah that's wrong

mappings/net/minecraft/client/gui/widget/ClickableWidget.mapping
```diff
@@ -78,3 +82,10 @@ CLASS net/minecraft/class_339 net/minecraft/client/gui/widget/ClickableWidget
 		ARG 1 x
 	METHOD method_46426 getX ()I
 	METHOD method_46427 getY ()I
+	METHOD method_47399 appendClickableNarrations (Lnet/minecraft/class_6382;)V
+		ARG 1 builder
+	METHOD method_47400 setTooltip (Lnet/minecraft/class_7919;)V
+		ARG 1 tooltip
+	METHOD method_47401 applyTooltip ()V
+	METHOD method_47402 setTooltipDelay (I)V
+		ARG 1 duration
```
enbrain:
```suggestion
		ARG 1 delay
```

mappings/net/minecraft/client/gui/screen/recipebook/RecipeBookWidget.mapping
```diff
@@ -91,3 +91,4 @@ CLASS net/minecraft/class_507 net/minecraft/client/gui/screen/recipebook/RecipeB
 	METHOD method_36485 isGuiOpen ()Z
 	METHOD method_37308 (Ljava/util/List;Lnet/minecraft/class_339;)V
 		ARG 1 button
+	METHOD method_47429 resetTooltip ()V
```
enbrain: Seems more like `updateTooltip`? It's called from `mouseClicked` when the toggle state is changed.

```suggestion
	METHOD method_47429 updateTooltip ()V
```

mappings/net/minecraft/client/gui/screen/Screen.mapping
enbrain: > Enigma was rejecting singular so I assumed it was already in use

Yes, `MultiplayerScreen#setTooltip` already exists. This method and the field should probably be named based on the fact that the method must be called every rendering frame to differentiate these two.

mappings/net/minecraft/client/gui/screen/recipebook/RecipeBookWidget.mapping
MattiDragon: I'm pretty sure it's called from a `reset` method, unless I confused something

mappings/net/minecraft/client/gui/screen/Screen.mapping
MattiDragon: @enbrain 
> the method must be called every rendering frame
This is true for the generic tooltip system too

mappings/net/minecraft/client/gui/screen/Screen.mapping
```diff
@@ -169,6 +170,17 @@ CLASS net/minecraft/class_437 net/minecraft/client/gui/screen/Screen
 	METHOD method_40041 hide ([Lnet/minecraft/class_339;)V
 		ARG 0 widgets
 	METHOD method_41843 clearAndInit ()V
+	METHOD method_47412 setTooltips (Lnet/minecraft/class_7919;)V
```
enbrain: @MattiDragon

> > the method must be called every rendering frame
>
> This is true for the generic tooltip system too

You are right. (I didn't notice `MultiplayerScreen` sets the `tooltip` to `null` at the head of `render`.) In that case, I'd rename `MultiplayerScreen#tooltip` to `MultiplayerScreen#multiplayerScreenTooltip` or similar and use singular for the one in `Screen`. It's confusing to use plural just because it conflicts.

haykam821 (changes requested):
The screen tooltip fields and methods should use the 'current tooltip' terminology since these tooltips are ephemeral and set during rendering. The widget tooltips, meanwhile, only need to be set once.

mappings/net/minecraft/client/gui/screen/Screen.mapping
MattiDragon: Pushed these changes, resolving

apple502j (approved):

enbrain (approved):

modmuss50 (approved):

PR #3396, opened 2022-11-09
Great Registry Repackaging

### LINKED ISSUE #3002: Top-level registry package
Since tags are greatly tied to registries now, it would be ideal to restructure registries to have a more promenient top-level package, like tags already do.

Here's a potential structure:

- `net.minecraft.registry`
  - The `Registry` class and its subclasses
  -  `.tag`
  - `.entry`
- `net.minecraft.world`
  - The `DynamicRegistryManager` class

apple502j:
Resolves #3002

Since this snapshot's changes mean people have to change imports anyway, why not do it now?

And question to folks: should we make subpackages? and if so how?


Shnupbups (approved):
Not that I oppose it exactly, but why repackage `tag` to `registry/tag`?

apple502j:
Tags are now managed by the registry (since 1.18.2).

YanisBft (approved):

haykam821:
Could entry and entry list classes be placed into a separate package?

Bixilon:
this PR can't be serious...

Shnupbups (approved):

haykam821 (approved):

md5:
Placing the entry classes in a separate package results in this error when trying to bootstrap during a JUnit test:

> java.lang.IllegalAccessError: class net.minecraft.registry.SimpleRegistry tried to access method 'void net.minecraft.registry.entry.RegistryEntry$Reference.setRegistryKey(net.minecraft.registry.RegistryKey)' (net.minecraft.registry.SimpleRegistry and net.minecraft.registry.entry.RegistryEntry$Reference are in unnamed module of loader 'app')

It looks like the `fabric-loader-junit` package allows me to get past that error, so perhaps it doesn't matter all that much for my use case, but I thought I'd bring it up since this mapping is resulting code that tries to access a package private member across package boundaries.

PR #3415, opened 2022-11-17
22w46a misc

apple502j:


YanisBft (approved):

mappings/net/minecraft/world/entity/LookupContinuityGetter.mapping
```diff
@@ -0,0 +1,9 @@
+CLASS net/minecraft/class_7927 net/minecraft/world/entity/LookupContinuityGetter
```
enbrain: Imo, `net/minecraft/util/LazyIterationConsumer` as modders could possibly use this class for other purposes. This is a consumer that accepts lazily evaluated elements in an iteration.

enbrain (approved):

enbrain:
Requesting review from YanisBft as new commit requires re-approval.

YanisBft (approved):

PR #3427, opened 2022-11-23
1.19.3-pre2 tooltip positioners

YanisBft:
On a side note, Tooltip class is in nm.client.gui.screen, shouldn't it be moved in the tooltip package?


apple502j:
Yeah, it should be moved.

apple502j (approved):

PR #3446, opened 2022-12-11
Revert some shader renames for ease of porting to 1.19.3

Juuxel:
Reverts small but impactful parts of #3384.

- `BufferRenderer.drawWithGlobalProgram` -> `drawWithShader`: The old name matches `RenderSystem.setShader` and it's also more clear IMO.
- `GameRenderer` shader programs and their getters. These are **very often** user-facing and cause a lot of headaches when porting. We also have to consider that not every developer can use `migrateMappings` (Kotlin users, for example).

There are also some less often used methods in `GameRenderer` that just refer to "programs" where "shader programs" would be more clear, but that's out of scope for this PR.

### DIFF
--- a/mappings/net/minecraft/client/render/BufferRenderer.mapping
+++ b/mappings/net/minecraft/client/render/BufferRenderer.mapping
@@ -3,7 +3,7 @@ CLASS net/minecraft/class_286 net/minecraft/client/render/BufferRenderer
 	COMMENT BufferBuilder}.
 	FIELD field_38982 currentVertexBuffer Lnet/minecraft/class_291;
 	METHOD method_34420 reset ()V
-	METHOD method_43433 drawWithGlobalProgram (Lnet/minecraft/class_287$class_7433;)V
+	METHOD method_43433 drawWithShader (Lnet/minecraft/class_287$class_7433;)V
 		COMMENT Draws {@code buffer} using the shader program specified with {@link
 		COMMENT com.mojang.blaze3d.systems.RenderSystem#setShader
 		COMMENT RenderSystem#setShader}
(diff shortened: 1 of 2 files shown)


modmuss50:
I dont think these renames should be reverted due to the "ease of porting" especially now, its going to cause even more confusion.

Reverting them because the new names are wrong/better is a diffrent story through. Ideally this should have been correct in the first place, if we are going to revert these names we need to be sure they are correct.

Juuxel:
> I dont think these renames should be reverted due to the "ease of porting" especially now, its going to cause even more confusion.

@modmuss50 Yep, that's no longer a valid reason since it's been a few weeks.

----

As for the names themselves, some of the renames like `drawWithGlobalProgram` lead to confusing mismatches between `RenderSystem` methods and `BufferRenderer` methods. You *set a shader* with `RenderSystem` (this is an unobf name!), but then use it by *drawing with the global program* with `BufferRenderer`, which was `drawWithShader` before. "Shader" also can't be confused for a shader stage in this context.

Others were completely pointless like the `GameRenderer` shader program renames because they can't be confused for shader stages - it's clear from the type that they're shader programs.

IMO a lot of the renames in the original PR made the code less clear by replacing "shader" with just "program" without any mention of shaders in many places, which is less accessible for people not familiar with shader code.

A lot of the renames were seemingly done just for the sake of eradicating the name "shader" from the mappings even where it had no risk of confusion and actually ended up leading to worse and less clear names. Those renames should be reverted to keep Yarn easy to understand and to reduce the inconsistency introduced in the original PR.

Juuxel:
There's not much sense in having this PR open anymore. See my comment above for some entries that could *at least* be renamed back to make them more clear and consistent with unobf names, but that would be another PR.

PR #3449, opened 2022-12-23
migrate from openjdk base image to eclipse-temurin

gdams:
https://hub.docker.com/_/openjdk has been deprecated for some time and will no longer get security updates


modmuss50 (approved):
LGTM, this is something we have been doing for the other projects 👍 

Thanks.

PR #3460, opened 2023-01-19
More 23w03a stuff

apple502j:


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -923,6 +923,7 @@ CLASS net/minecraft/class_1297 net/minecraft/entity/Entity
 		ARG 1 tickDelta
 	METHOD method_46395 extinguishWithSound ()V
 	METHOD method_46396 reinitDimensions ()V
+	METHOD method_48155 canSprint ()Z
```
Shnupbups: Does this not collide with `method_46743 canSprint ()Z` in `ClientPlayerEntity`?

apple502j:
```suggestion
```

mappings/net/minecraft/text/Text.mapping
```diff
@@ -62,6 +62,13 @@ CLASS net/minecraft/class_2561 net/minecraft/text/Text
 		COMMENT {@return whether the text contains {@code text}, without considering styles}
 		ARG 1 text
 	METHOD method_44746 withoutStyle ()Ljava/util/List;
+	METHOD method_48321 translatableWithFallback (Ljava/lang/String;Ljava/lang/String;)Lnet/minecraft/class_5250;
+		ARG 0 key
+		ARG 1 fallback
+	METHOD method_48322 translatableWithFallback (Ljava/lang/String;Ljava/lang/String;[Ljava/lang/Object;)Lnet/minecraft/class_5250;
```
Shnupbups, line 68: Could these not just be `translatable`? I feel the `fallback` argument name is enough to get that these are ones with fallback.

Shnupbups (approved):

mappings/net/minecraft/text/Text.mapping
apple502j: This is for user convenience - to allow calling `translatable(String, ...Object)` without weird casts.

Shnupbups: Ah, yeah that makes sense.

Shnupbups (approved):

PR #3470, opened 2023-02-01
Mappings for net/minecraft/entity

PiTheGuy:


modmuss50:
Hi, thanks for the PR. However I dont think we can accept this as it goes against our [contributing](https://github.com/FabricMC/yarn#contributing) guidelines. A lot of these names are hard/impossible to name without the use of the Mojang mappings. These names appear to match 1-1 with the ones from the Mojang mappings.

Sorry about this, but I think its also best I revert your previous PR as it appears to also contain a lot of mappings that match 1-1 with Mojang's. Please reach out to me on discord (or in this issue) if you wish to talk about this some more.

PR #3473, opened 2023-02-04
Fix wrong name of `method_10772`.

keuin:
Method `hasChannel` (`method_10772`) in class `net/minecraft/network/ClientConnection` (`class_2535`) has a wrong name. Its decompiled code generated by yarn is:

```java
public boolean hasChannel() {
    return this.channel == null;
}
```

The method returns true if the internal channel of `ClientConnection` is null. However the name contradicts to what the method has done.

I have read the name convention documented in `CONVENTIONS.md`, but it doesn't state how we should name an inversed getter method. This pull request renames `hasChannel` to `isChannelAbsent`. Please let me know if there is a better choice!


YanisBft (approved):

keuin:
Sorry, I thought the branch has been merged but it wasn't. I am re-opening the pull request.

Juuxel (approved):

PR #3493, opened 2023-02-16
Build script improvements

modmuss50:
- Gradle 8.1
- Move more to filament
- Task caching should be a lot better
- Configuration caching enabled


thecatcore:
Bump filament version?

RawDiamondMC:
Update to gradle 8.0.2?

thecatcore (approved):

filament/src/main/java/net/fabricmc/filament/task/enimga/ConvertMappingsTask.java
```diff
@@ -0,0 +1,34 @@
+package net.fabricmc.filament.task.enimga;
```
Juuxel: "enimga"

gradle.properties
```diff
@@ -1,16 +1,18 @@
 # Done to increase the memory available to gradle.
 org.gradle.jvmargs=-Xmx1G
+org.gradle.parallel=true
```
Juuxel, line 3: Does this matter? To my knowledge, Yarn doesn't have subprojects. This flag only affects multiproject builds. (See [this discord discussion](https://discord.com/channels/507304429255393322/592749519171616796/1021440447144546384))

Juuxel (changes requested):

gradle.properties
modmuss50: No, I double checked and it doesnt, no real harm in setting it tho.

PR #3494, opened 2023-02-16
Map interaction entity

apple502j:


mappings/net/minecraft/entity/decoration/InteractionEntity.mapping
```diff
@@ -0,0 +1,29 @@
+CLASS net/minecraft/class_8150 net/minecraft/entity/decoration/InteractionEntity
+	FIELD field_42625 WIDTH Lnet/minecraft/class_2940;
+	FIELD field_42626 HEIGHT Lnet/minecraft/class_2940;
+	FIELD field_42627 RESPONSE Lnet/minecraft/class_2940;
+	FIELD field_42628 WIDTH_KEY Ljava/lang/String;
+	FIELD field_42629 HEIGHT_KEY Ljava/lang/String;
+	FIELD field_42630 ATTACK_KEY Ljava/lang/String;
+	FIELD field_42631 INTERACTION_KEY Ljava/lang/String;
+	FIELD field_42632 RESPONSE_KEY Ljava/lang/String;
```
Shnupbups, line 9: I've gotten in the habit of suffixing with `NBT_KEY` rather than just `KEY`, makes it a bit clearer IMO. Okay if you don't want to, but that's how I've been naming these recently.

Shnupbups (approved):

mappings/net/minecraft/entity/decoration/InteractionEntity.mapping
apple502j: Let's make an issue then.

PR #3520, opened 2023-02-28
Improve ambiguous brain tasks

### LINKED ISSUE #3519: The names of some brain tasks are ambiguous and easily misunderstood
The names of the `FollowMobTask` and `FollowMobWithIntervalTask` brain tasks are easily misunderstood to mean that their function is to walk towards other mobs and follow them around. Actually looking into their code however reveals they handle the look direction of the entity, causing it to look at other nearby entities, which explains why basically every mob with a brain has them. I suggest renaming these to something like `LookAtMobTask` or adding a javadoc to explain what they actually do.

`WalkTask` also seems like a strange naming choice, as it handles the panicking behaviour when passive mobs are attacked.

Benonardo:
Fixes #3519 with three class renames


apple502j (approved):
If we want this to be in 1.19.4 we need to merge ASAP. LGTM otherwise

Shnupbups (approved):

modmuss50 (approved):

PR #3553, opened 2023-04-27
Add mappings for font and light handling

bluebear94:
I was initially going to map just the new font stuff, but I got sucked into mapping some of the new light handling code.


apple502j (approved):
Will review later when I diff it

PR #3566, opened 2023-05-05
More buildscript updates

modmuss50:
Replace stitch with loom and mapping-io

There is more to do after this, but I want to take it one step at a time to make it easier to figure out what I break 😆 .


modmuss50:
This seems to be causing mapping poet to fail 🤔 

```
Exception in thread "main" java.lang.RuntimeException: Could not find parent class: net/minecraft/class_6740 for net/minecraft/class_6740$class_6741
	at net.fabricmc.mappingpoet.Main.writeClass(Main.java:247)
	at net.fabricmc.mappingpoet.Main.lambda$generate$0(Main.java:95)
	at net.fabricmc.mappingpoet.Main.lambda$forEachClass$5(Main.java:171)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at net.fabricmc.mappingpoet.Main.forEachClass(Main.java:171)
	at net.fabricmc.mappingpoet.Main.generate(Main.java:95)
```

liach (approved):

liach:
Quick question, does filament have any license?

gradle.properties
```diff
@@ -18,5 +16,6 @@ mappingpoet_version=0.3.2
 
 # Build logic
 tiny_mappings_parser_version=0.3.0+build.17
```
liach: Can be removed as well

modmuss50: Done 👍

liach:
Can you do us a favor and add `mavenLocal()` to filament's repositories as well? That way I just need to update mappingpoet version in gradle.properties only in order to test +local builds.

modmuss50:
> Can you do us a favor and add `mavenLocal()` to filament's repositories as well? That way I just need to update mappingpoet version in gradle.properties only in order to test +local builds.

Done 👍

modmuss50:
Ill try and get this tested and merged quite soon. I had hoped we would have a few more weeks of snapshots where breaking stuff wasnt so bad. 👍

modmuss50:
Tested with Fabric API, everything seems to work OK. Please let me know ASAP if you have any issues after I merge this 👍

PR #3575, opened 2023-05-11
Fix class name possibly being null during RemapUnpickDefinitionsTask

thecatcore:
Mappings tree doesn't seems to be constructed the same way in mappings-io and tiny-parser during mappings read.
When an entry is the same for all namespaces in tinyv2, it is written only one time.
Then when it is read by tiny-parser, all namespace names of the entry are filled with the same value.
When it is read by mappings-io, only the first namespace of the mappings has a value and the list of names for other namespaces is empty for the entry.

### DIFF
--- a/filament/src/main/java/net/fabricmc/filament/task/RemapUnpickDefinitionsTask.java
+++ b/filament/src/main/java/net/fabricmc/filament/task/RemapUnpickDefinitionsTask.java
@@ -105,10 +105,8 @@ public abstract class RemapUnpickDefinitionsTask extends DefaultTask {
 					String fromClass = classDef.getName(fromM);
 					String toClass = classDef.getName(toM);
 
-					if (fromClass == null || fromClass.isEmpty()) {
-						fromClass = toClass;
-					} else if (toClass == null || toClass.isEmpty()) {
-						toClass = fromClass;
+					if (fromClass == null || toClass == null) {
+						contiune;
 					}
 
 					classMappings.put(fromClass, toClass);
--- a/filament/src/main/java/net/fabricmc/filament/task/RemapUnpickDefinitionsTask.java
+++ b/filament/src/main/java/net/fabricmc/filament/task/RemapUnpickDefinitionsTask.java
@@ -102,18 +102,27 @@ public abstract class RemapUnpickDefinitionsTask extends DefaultTask {
 				final int toM = mappingTree.getNamespaceId(getParameters().getTargetNamespace().get());
 
 				for (MappingTree.ClassMapping classDef : mappingTree.getClasses()) {
-					classMappings.put(classDef.getName(fromM), classDef.getName(toM));
+					String fromClass = classDef.getName(fromM);
+					String toClass = classDef.getName(toM);
+
+					if (fromClass == null || fromClass.isEmpty()) {
+						fromClass = toClass;
+					} else if (toClass == null || toClass.isEmpty()) {
+						toClass = fromClass;
+					}
+
+					classMappings.put(fromClass, toClass);
 
 					for (MappingTree.MethodMapping methodDef : classDef.getMethods()) {
 						methodMappings.put(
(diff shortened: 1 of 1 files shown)


modmuss50 (commented):
Have you got a way to reproduce your issue, this doesnt feel correct. If the to or from class doesnt exist, surely there is nothing to remap? Setting them to match is pointless.

thecatcore:
Can reproduce with: https://github.com/Legacy-Fabric/yarn/tree/error-reproduction
the entry with the null key has the value "net/minecraft/class_0_1348$class_0_1350".
net/minecraft/class_0_1348$class_0_1350 doesn't have a mapping file.

NameProposal generated those mappings which are RemapUnpickDefinitionsTask input:
```
c	net/minecraft/class_0_1348$class_0_1350	
	f	Lnet/minecraft/class_0_1348$class_0_1350;	field_0_5659	FAILED
	f	Lnet/minecraft/class_0_1348$class_0_1350;	field_0_5658	CANCELLED
	f	Lnet/minecraft/class_0_1348$class_0_1350;	field_0_5655	DONE_TESTING_SERVER
	f	Lnet/minecraft/class_0_1348$class_0_1350;	field_0_5654	TESTING_SERVER
	f	Lnet/minecraft/class_0_1348$class_0_1350;	field_0_5657	CANCELLING
	f	Lnet/minecraft/class_0_1348$class_0_1350;	field_0_5656	FINISHED
	f	Lnet/minecraft/class_0_1348$class_0_1350;	field_0_5651	UNINITALIZED
	f	Lnet/minecraft/class_0_1348$class_0_1350;	field_0_5653	CONNECTING_TO_SERVER
	f	Lnet/minecraft/class_0_1348$class_0_1350;	field_0_5652	STARTING
```
The NPE is for the FieldKey "STARTING" because the owner class passed to the ctr is null instead of net/minecraft/class_0_1348$class_0_1350.

thecatcore:
This issue never occured before the switch from tiny-mappings-parser to mappings-io in the task in filament.
filament 0.5.0 doesn't have this issue.

filament/src/main/java/net/fabricmc/filament/task/RemapUnpickDefinitionsTask.java
```diff
@@ -102,18 +102,27 @@ public void execute() {
...
 				for (MappingTree.ClassMapping classDef : mappingTree.getClasses()) {
-					classMappings.put(classDef.getName(fromM), classDef.getName(toM));
+					String fromClass = classDef.getName(fromM);
+					String toClass = classDef.getName(toM);
+
+					if (fromClass == null || fromClass.isEmpty()) {
+						fromClass = toClass;
+					} else if (toClass == null || toClass.isEmpty()) {
+						toClass = fromClass;
+					}
```
modmuss50: Would the following work?
```suggestion
					if (fromClass == null || toClass == null) {
						contiune;
					}
```

thecatcore: Doesn't the remap need the field remap tho ? The class and all its elements would be skipped in this case?

thecatcore: In the case of those NPE, the from namespace is named and the to namespace is intermediary.

thecatcore:
I've used the debugger to see the comportment of filament 0.5.0 in this case.
The mappings are still added with the intermediary name as the owner in this case.

thecatcore:
Note: This issue is the only thing that prevent LegacyFabric Yarn from updating to filament 0.6.

thecatcore:
Any chance this get merged? I don’t want to maintain a fork of filament just for this fix.

modmuss50:
Im still not convined this is the correct fix, I think my suggestion is the solution here?

thecatcore:
Alright, I'm going to apply it then, I'll let you know if it fixes it or not.

thecatcore:
bump after applying recommended changes

PR #3579, opened 2023-05-16
1.20-pre2 stuff

apple502j:


mappings/net/minecraft/recipe/Ingredient.mapping
```diff
@@ -12,6 +12,8 @@ CLASS net/minecraft/class_1856 net/minecraft/recipe/Ingredient
 	METHOD method_35226 empty ()Lnet/minecraft/class_1856;
 	METHOD method_47374 (Lnet/minecraft/class_1856$class_1859;)Ljava/util/stream/Stream;
 		ARG 0 entry
+	METHOD method_52177 fromJsonAirAllowed (Lcom/google/gson/JsonElement;)Lnet/minecraft/class_1856;
```
ErrorCraft: I think treating it as an overload for `fromJson` would make more sense here? So `fromJson(JsonElement)` and `fromJson(JsonElement, boolean)`.

mappings/net/minecraft/screen/AnvilScreenHandler.mapping
```diff
@@ -20,4 +20,7 @@ CLASS net/minecraft/class_1706 net/minecraft/screen/AnvilScreenHandler
 		ARG 0 stack
 	METHOD method_48351 (Lnet/minecraft/class_1799;)Z
 		ARG 0 stack
+	METHOD method_52176 satanize (Ljava/lang/String;)Ljava/lang/String;
```
ErrorCraft: Small typo, should be `sanitize` (unless you want it to summon demons)

ErrorCraft (changes requested):

mappings/net/minecraft/recipe/Ingredient.mapping
modmuss50: Yes, I agree 👍

apple502j: I don't like things with the same signature but different behavior from old one being named the same. Previously this disallowed air, and now this allows air. By ensuring that compile fails, they can examine their code to see which one is better.

(also @modmuss50, I believe the ingredient API port used this method, but you should've used the boolean one. See, this is the kind of thing I am trying to prevent here.)

mappings/net/minecraft/screen/AnvilScreenHandler.mapping
apple502j:
```suggestion
	METHOD method_52176 sanitize (Ljava/lang/String;)Ljava/lang/String;
```

mappings/net/minecraft/recipe/Ingredient.mapping
Shnupbups: While I see your reasoning apple, I do feel it should be an overload. Having a bad name for the sake of temporary backwards-compat (or rather backwards-incompat?) is something I'd rather avoid.

apple502j:
```suggestion
	METHOD method_52177 fromJson (Lcom/google/gson/JsonElement;)Lnet/minecraft/class_1856;
```

Shnupbups (approved):

PR #3586, opened 2023-05-24
Map class_8012

apple502j:
This is one of the three mystery constant classes.

I diffed the pre4-pre5 and noticed that a new field was added here. I checked the usage of the newly added `field_44941` in the code diff, and it was used in `RealmsInviteScreen` as a gray color.

I then converted each field to hexadecimal unsigned int - and tada! It was all something that were meaningful as colors; white, black, and red.


modmuss50 (approved):

Shnupbups (approved):

YanisBft:
I'm a bit late, but shouldn't it be in the client package? If it's used in screens

apple502j:
@YanisBft Client package is reserved for actually client-only class; this is not.

PR #3598, opened 2023-06-22
Rename `unlockCraftedRecipe` to `onRecipeCrafted` in `PlayerEntity`

Jerozgen:
This method doesn't unlock any recipes, it triggers the `recipe_crafted` advancement criteria in `ServerPlayerEntity`

```java
@Override
public void unlockCraftedRecipe(Recipe<?> recipe, List<ItemStack> ingredients) {
	Criteria.RECIPE_CRAFTED.trigger(this, recipe.getId(), ingredients);
}
```

### DIFF
--- a/mappings/net/minecraft/entity/player/PlayerEntity.mapping
+++ b/mappings/net/minecraft/entity/player/PlayerEntity.mapping
@@ -122,7 +122,7 @@ CLASS net/minecraft/class_1657 net/minecraft/entity/player/PlayerEntity
 		ARG 1 lastDeathPos
 	METHOD method_43122 getLastDeathPos ()Ljava/util/Optional;
 	METHOD method_45015 shouldCloseHandledScreenOnRespawn ()Z
-	METHOD method_51283 unlockCraftedRecipe (Lnet/minecraft/class_1860;Ljava/util/List;)V
+	METHOD method_51283 onRecipeCrafted (Lnet/minecraft/class_1860;Ljava/util/List;)V
 		ARG 1 recipe
 		ARG 2 ingredients
 	METHOD method_7254 unlockRecipes (Ljava/util/Collection;)I


apple502j:
@Jerozgen Hmm, isn't recipe unlocking triggered by an advancement criteria?

Jerozgen:
Advancements can unlock a recipe using any criterion. Basically, they use `inventory_changed` and `recipe_unlocked` criteria to do that. 
The `recipe_crafted` criterion is used in these non-recipe vanilla advancements:
- **Careful Restoration**: Make a Decorated Pot out of 4 Pottery Sherds
- **Crafting a New Look**: Craft a trimmed armor at a Smithing Table
- **Smithing with Style**: Apply these smithing templates at least once: Spire, Snout, Rib, Ward, Silence, Vex, Tide, Wayfinder

apple502j:
@Jerozgen Hmm,I thought vanilla recipe unlocking was just a bunch of advancements, you sure `RECIPE_CRAFTED` is not the criteria used in unlocking right?

That said, it might be better to rename this anyway.

Jerozgen:
Recipe unlocking can be done by these actions:
- crafting
- using a knowledge book
- getting from an advancement reward
- using the `/recipe give` command

If a recipe book doesn't have an unlocked recipe yet, the `recipe_unlocked` criterion is triggered. `recipe_crafted` is used only in non-recipe advancements right now and can be triggered more than once by the crafting action only.

> That said, it might be better to rename this anyway.

Yeah, do it if needed. The suggested name is based on the `onBlockCollision`, `onStatusEffectApplied` etc. methods which trigger criteria

Shnupbups (approved):

Shnupbups:
This is indeed a yarnbug, the RECIPE_CRAFTED trigger is the new one from 1.20 that triggers when a recipe is crafted, and isn't used for unlocking recipes whatsoever.

Issue #3196, opened 2022-05-29
Enforcing sidedness for client package


apple502j:
Currently, yarn packages most of the classes available only in client jar under `net.minecraft.client`, and most of the common classes outside that. With Loom enforcing sidedness more strictly, and the addition of `yarnCommon` in enigma, this should be enforced more strictly ("most of" to "all of") here as well; however, doing that will require us to fix existing ones beforehand. Here is a list of classes which do not follow the sidedness rule.

Classes that are widely used (from GitHub code search using full class name in Java; 50+ files classified as impactful, 20+ as semi-impactful, others as not impactful; note that multiple files in one mod are counted separately, so number of affected mods might be smaller) and impactful-to-rename are marked as such.

This also includes a suggested new package.

# Common classes in client package
- [ ] `/color/world/FoliageColors` (Impactful) -> `world/biome`?
- [ ] `/color/world/GrassColors` (Impactful) -> `world/biome`?
- [ ] `/font/TextVisitFactory` (Semi-impactful) -> `text`
- [ ] `/gl/ShaderParseException` (Not impactful) -> `resource`
- [x] `/gui/WorldGenerationProgressTracker` (Semi-impactful) -> `server/world`
- [ ] `/item/BundleTooltipData` (Semi-impactful) -> `item/tooltip` (new package)
- [ ] `/item/TooltipContext` (VERY impactful) -> `item/tooltip`
- [ ] `/item/TooltipData` (Semi-impactful) -> `item/tooltip`
- [x] `/option/ChatVisibility` (Semi-impactful) -> `network` or `network/message`
- [ ] `/particle/ParticleGroup` (Not impactful) -> `particle`
- [ ] `/render/BlockBreakingInfo` (Semi-impactful) -> `entity/player`?
- [ ] `/render/CameraSubmersionType` (Impactful) -> `block` or `block/enums`
- [x] `/render/debug/NameGenerator` (Not impactful) -> `util`
- [ ] `/render/entity/feature/SkinOverlayOwner` (Semi-impactful) -> `entity`
- [ ] `/render/entity/PlayerModelPart` (Impactful) -> `entity/player`
- [x] `/resource/language/TextReorderingProcessor` (Not impactful) -> `text` (done as part of #3329)
- [ ] `/sound/MusicType` (Impactful) -> `sound`
- [x] `/util/CharPredicate` (Not impactful) -> `util/function`
[...]

PR #3601, opened 2023-07-02
Enforcing sidedness for client package

natanjunges:
Implements changes proposed in #3196, starting by the least impactful ones.

### DIFF
--- a/filament/src/test/resources/projects/sharedData/yarn-mappings-v2.tiny
+++ b/filament/src/test/resources/projects/sharedData/yarn-mappings-v2.tiny
@@ -4421,7 +4421,7 @@ c	net/minecraft/class_3748	net/minecraft/block/JigsawBlock
 		p	0		info1
 c	net/minecraft/class_2417	net/minecraft/data/server/StoryTabAdvancementGenerator
 	m	(Ljava/util/function/Consumer;)V	method_10347	accept
-c	net/minecraft/class_5582	net/minecraft/client/world/ClientEntityManager
+c	net/minecraft/class_5582	net/minecraft/world/entity/ClientEntityManager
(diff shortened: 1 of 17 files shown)


filament/src/test/resources/projects/sharedData/yarn-mappings-v2.tiny
Juuxel, line 1: This data doesn't need to be adjusted since it's an older copy of Yarn purely for testing that Filament's tasks work properly.

natanjunges: Ah, OK. Do I need to revert those changes?

modmuss50: > Ah, OK. Do I need to revert those changes?

Yes please 👍

modmuss50:
This looks like a good start likely should go into the next badly breaking Minecraft version.

Once this is ready we should also look into adding some tools to validate these packages. I can possibly look into moving yarn to use a split jar instead of a merged jar, should make this easier.

filament/src/test/resources/projects/sharedData/yarn-mappings-v2.tiny
natanjunges: Done.

natanjunges:
Where should I move the following classes to:
- `net.minecraft.client.option.ChatVisibility` should go under `net.minecraft.network` or `net.minecraft.network.message`?
- `net.minecraft.client.render.BlockBreakingInfo` should really go under `net.minecraft.entity.player`, or is there any other place where it fits better?
- `net.minecraft.client.util.NetworkUtils` should go under `net.minecraft.util`or `net.minecraft.network`?

PR #3608, opened 2023-07-20
Rename registerPressurePlate to registerWeightedPressurePlate 

Benonardo:
This method in `BlockStateModelGenerator` uses the `POWERED` property, which is not present in `PressurePlateBlock`, only `WeightedPressurePlateBlock`.
Normal pressure plates are instead handled by `BlockTexturePool#pressurePlate`.
A case of confusion caused by the previous name can be seen [here](https://discord.com/channels/507304429255393322/507982478276034570/1131634487512543382).

### DIFF
--- a/mappings/net/minecraft/data/client/BlockStateModelGenerator.mapping
+++ b/mappings/net/minecraft/data/client/BlockStateModelGenerator.mapping
@@ -281,8 +281,8 @@ CLASS net/minecraft/class_4910 net/minecraft/data/client/BlockStateModelGenerato
 	METHOD method_25664 registerTallSeagrass ()V
 	METHOD method_25665 registerOrientableTrapdoor (Lnet/minecraft/class_2248;)V
 		ARG 1 trapdoorBlock
-	METHOD method_25666 registerPressurePlate (Lnet/minecraft/class_2248;Lnet/minecraft/class_2248;)V
-		ARG 1 pressurePlate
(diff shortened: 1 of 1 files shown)


mappings/net/minecraft/data/client/BlockStateModelGenerator.mapping
```diff
@@ -300,8 +300,8 @@ CLASS net/minecraft/class_4910 net/minecraft/data/client/BlockStateModelGenerato
 		COMMENT Used for a block that shares a block model with another block, for example waxed copper or infested stone bricks.
 		ARG 1 modelSource
 		ARG 2 child
-	METHOD method_25673 createPressurePlateBlockState (Lnet/minecraft/class_2248;Lnet/minecraft/class_2960;Lnet/minecraft/class_2960;)Lnet/minecraft/class_4917;
-		ARG 0 pressurePlateBlock
+	METHOD method_25673 createWeightedPressurePlateBlockState (Lnet/minecraft/class_2248;Lnet/minecraft/class_2960;Lnet/minecraft/class_2960;)Lnet/minecraft/class_4917;
+		ARG 0 weightedPressurePlateBlock
```
Juuxel:
```suggestion
		ARG 0 weightedPressurePlate
```
The other ones only seem to use either `block` or a more specific name.

Juuxel (approved):

PR #3624, opened 2023-08-09
Network related notes and changes

liach:
play ping/pong -> common ping/pong
brand custom payload is both s2c and c2s
ConnectionIntent is only used by c2s handshake
Notes for ticking handlers and packet-based state transitions


mappings/net/minecraft/server/network/ServerPlayNetworkHandler.mapping
```diff
@@ -187,6 +187,8 @@ CLASS net/minecraft/class_3244 net/minecraft/server/network/ServerPlayNetworkHan
 	METHOD method_46366 (Lnet/minecraft/class_7822;Ljava/util/concurrent/Executor;)Ljava/util/concurrent/CompletableFuture;
 		ARG 2 executor
 	METHOD method_48107 getConnectionAddress ()Ljava/net/SocketAddress;
+	METHOD method_52414 reconfigurate ()V
```
apple502j:
```suggestion
	METHOD method_52414 reconfigure ()V
```

mappings/net/minecraft/server/network/ServerConfigurationNetworkHandler.mapping
```diff
@@ -13,6 +13,6 @@ CLASS net/minecraft/class_8610 net/minecraft/server/network/ServerConfigurationN
 	METHOD method_52407 (Lnet/minecraft/server/MinecraftServer$class_7460;)V
 		ARG 1 properties
 	METHOD method_52409 sendConfigurations ()V
-	METHOD method_52410 queueJoinWorldTask ()V
+	METHOD method_52410 endConfiguration ()V
```
modmuss50, line 16: I dont think this is correct, this method is only called from the unconfig debug command, and is not used during normal login/play. `cancelConfiguration` may be slightly better?

liach: This doesn't immediately end configuration indeed, but it prevents further tasks effectively. And its body is the same as the end of the sendConfiguration method.

modmuss50: Its not wrong, and it is better than queueJoinWorldTask. Unless you have another idea im happy to go with this.

liach: Guess will go with this and see mojang changes in snapshot

apple502j (approved):

modmuss50 (approved):

PR #3640, opened 2023-08-24
Rename an argument of Merchant#sendOffers

xanderstuff:
Basically go from:
`void sendOffers(PlayerEntity player, Text test, int levelProgress)`
to:
`void sendOffers(PlayerEntity player, Text text, int levelProgress)`
(test -> text. I know, it's a small change lol)


xanderstuff:
Also this is my first mappings PR, I hope this is the correct way to make changes

xanderstuff:
decided that "displayName" is more descriptive than just "text", since in the only places it's used, both `VillagerEntity` and `WanderingTraderEntity`, it gets `Entity#getDisplayName` passed to it. My guess is this is the text that goes at the top of the GUI trading screen

YanisBft:
Would be better to have consistency with `SimpleNamedScreenHandlerFactory#name` imo.

YanisBft (approved):

Shnupbups (approved):

PR #3644, opened 2023-08-27
Create client.session package

apple502j:
Client packages are mostly well-organized, except for one thing: there is no package dedicated to authlib interactions. This was fine a few years ago, when the only class that would interact with authlib was Session. Since then, however, the role of authlib in client code has expanded. This has forced us to package them in the (already-big) util package, or in the case of abuse reports, make a package directly under client package, for one relatively small feature.

I propose a repackaging, to establish "session" package and to move all authlib-interacting code there. This should have little impact outside of mods that disable reports or telemetry; however, since there are tons of mods touching this very specific stuff, it is nevertheless impactful.


Juuxel (dismissed):

liach:
Looks great! Two comments:
1. Should the telemetry go under session package? Many of the events have little to do with sessions, so I think it can still directly go under client.
2. The log package in report can probably just be merged into the report package.

apple502j:
- Telemetry is pretty much a feature used to talk to Mojang servers, so this is probably the best place.
- Think the log package can stay as well.

liach:
> Telemetry is pretty much a feature used to talk to Mojang servers

It's like "block and entities are part of the world". The telemetry package is already huge, almost larger than sessions because much of telemetry does not immediately talk to Mojang server, just like most code of blocks/entities doesn't immediately interact with rca region files.

In addition, how much of report depends on sessions? It might merit its own package as well. Reports and telemetries are users of session, and you don't group your users into your subpackage (like how report.log is used by report, not using report)

liach (approved):
This package nesting isn't too deep, so it's ok I think

YanisBft (approved):

PR #3655, opened 2023-09-05
1.20.2-pre1 stuff

apple502j:
The record shows that Mojang calls the settings "information", like, sure...

### DIFF
--- a/mappings/net/minecraft/network/packet/c2s/common/ClientInformation.mapping
+++ b/mappings/net/minecraft/network/packet/c2s/common/ClientInformation.mapping
@@ -1,7 +1,0 @@
-CLASS net/minecraft/class_8791 net/minecraft/network/packet/c2s/common/ClientInformation
-	FIELD field_46155 MAX_LANGUAGE_CODE_LENGTH I
-	METHOD <init> (Lnet/minecraft/class_2540;)V
-		ARG 1 buf
-	METHOD method_53821 createDefault ()Lnet/minecraft/class_8791;
-	METHOD method_53822 write (Lnet/minecraft/class_2540;)V
-		ARG 1 buf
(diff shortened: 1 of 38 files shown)


mappings/net/minecraft/network/packet/c2s/common/ClientInformation.mapping
haykam821, line 1: This record contains the following information:

- Language code (e.g. `en_us`)
- View distance
- Chat visibility
- Whether the client has chat colors enabled
- Skin customization
- Main hand
- Whether text filtering is enabled
- Whether the player should be hidden from the public player list

Since all of these fields are configurable within the client's options, rather than information about the client itself, I think a name such as `SyncedClientOptions` would suit this class better.

haykam821, line 1: Could this record's components be renamed to better match the existing names? In particular, the `ServerPlayerEntity#setClientInformation` method should have matching names:

```java
this.language = information.language();
this.viewDistance = information.viewDistance();
this.clientChatVisibility = information.chatVisibility();
this.clientChatColorsEnabled = information.chatColors();
this.filterText = information.textFilteringEnabled();
this.allowServerListing = information.allowsListing();
this.getDataTracker().set(PLAYER_MODEL_PARTS, (byte) information.modelCustomisation());
this.getDataTracker().set(MAIN_ARM, (byte) information.mainHand().getId());
```

apple502j: The "client" prefixes will be dropped, otherwise sounds goof

apple502j (commented):
~~TradeOffers mapping needs to be rechecked~~ Seems okay.

Shnupbups (approved):

PR #3664, opened 2023-09-13
Map changes to block pointers in 1.20.2 pre-release 4

haykam821:


mappings/net/minecraft/util/math/BlockPointer.mapping
```diff
@@ -1,9 +1,10 @@
...
-	METHOD comp_1969 getBlockState ()Lnet/minecraft/class_2680;
-	METHOD comp_1970 getBlockEntity ()Lnet/minecraft/class_2601;
+	FIELD comp_1967 world Lnet/minecraft/class_3218;
+	FIELD comp_1968 pos Lnet/minecraft/class_2338;
+	FIELD comp_1969 state Lnet/minecraft/class_2680;
+	FIELD comp_1970 blockEntity Lnet/minecraft/class_2601;
+	METHOD comp_1967 world ()Lnet/minecraft/class_3218;
+	METHOD comp_1968 pos ()Lnet/minecraft/class_2338;
+	METHOD comp_1969 state ()Lnet/minecraft/class_2680;
+	METHOD comp_1970 blockEntity ()Lnet/minecraft/class_2601;
```
liach, line 9: Should this simply be `entity`, now that `getBlockState` becomes `state`?

apple502j: IMO, BlockEntity should never be referred to as an entity; a block state is a type of state, a block entity is not an entity.

mappings/net/minecraft/util/math/BlockPointer.mapping
```diff
@@ -1,9 +1,10 @@
...
-	FIELD comp_1969 getBlockState Lnet/minecraft/class_2680;
-	FIELD comp_1970 getBlockEntity Lnet/minecraft/class_2601;
-	METHOD comp_1967 getWorld ()Lnet/minecraft/class_3218;
-	METHOD comp_1968 getPos ()Lnet/minecraft/class_2338;
-	METHOD comp_1969 getBlockState ()Lnet/minecraft/class_2680;
-	METHOD comp_1970 getBlockEntity ()Lnet/minecraft/class_2601;
+	FIELD comp_1967 world Lnet/minecraft/class_3218;
+	FIELD comp_1968 pos Lnet/minecraft/class_2338;
+	FIELD comp_1969 state Lnet/minecraft/class_2680;
+	FIELD comp_1970 blockEntity Lnet/minecraft/class_2601;
```
Shnupbups, line 5: These don't need to be mapped, do they? The names are exactly the same for these three with name proposal (obviously we want to keep the change from `level` to `world` though)

PR #3676, opened 2023-09-19
Some AbstractFurnaceScreenHandler constants.

gudenau:
I left `field_30744` and `field_30745` alone because they are the same constants, but they appear to be something like `PLAYER_MAIN_INVENTORY_END`.


mappings/net/minecraft/screen/AbstractFurnaceScreenHandler.mapping
```diff
@@ -2,6 +2,13 @@ CLASS net/minecraft/class_1720 net/minecraft/screen/AbstractFurnaceScreenHandler
 	FIELD field_17286 propertyDelegate Lnet/minecraft/class_3913;
 	FIELD field_17494 recipeType Lnet/minecraft/class_3956;
 	FIELD field_25762 category Lnet/minecraft/class_5421;
+	FIELD field_30738 BURN_TIME_PROPERTY_INDEX I
+	FIELD field_30739 FUEL_TIME_PROPERTY_INDEX I
+	FIELD field_30740 COOK_TIME_PROPERTY_INDEX I
+	FIELD field_30741 COOK_TIME_TOTAL_PROPERTY_INDEX I
+	FIELD field_30742 PROPERTY_COUNT I
+	FIELD field_30743 INVENTORY_SIZE I
+	FIELD field_30746 PLAYER_INVENTORY_END I
```
modmuss50, line 11: Some of these are likely to be the slot id, im not sure these names are correct. We generally dont map stuff like this unless we are 100% sure about it, in this case I dont think you can be sure.

PR #3698, opened 2023-10-14
23w41a stuff

apple502j:
See each commit for explanation. And yes, I did pick my own color name.

### DIFF
--- a/mappings/net/minecraft/entity/vehicle/VehicleEntity.mapping
+++ b/mappings/net/minecraft/entity/vehicle/VehicleEntity.mapping
@@ -14,6 +14,6 @@ CLASS net/minecraft/class_8836 net/minecraft/entity/vehicle/VehicleEntity
 		ARG 1 damageWobbleTicks
 	METHOD method_54300 setDamageWobbleSide (I)V
 		ARG 1 damageWobbleSide
-	METHOD method_7516 killVehicle (Lnet/minecraft/class_1282;)V
+	METHOD method_7516 killAndDropSelf (Lnet/minecraft/class_1282;)V
 		ARG 1 source
 	METHOD method_7557 asItem ()Lnet/minecraft/class_1792;
--- a/mappings/net/minecraft/entity/vehicle/VehicleEntity.mapping
+++ b/mappings/net/minecraft/entity/vehicle/VehicleEntity.mapping
@@ -2,7 +2,7 @@ CLASS net/minecraft/class_8836 net/minecraft/entity/vehicle/VehicleEntity
 	FIELD field_46655 DAMAGE_WOBBLE_TICKS Lnet/minecraft/class_2940;
 	FIELD field_46656 DAMAGE_WOBBLE_SIDE Lnet/minecraft/class_2940;
 	FIELD field_46657 DAMAGE_WOBBLE_STRENGTH Lnet/minecraft/class_2940;
-	METHOD method_54293 shouldAlwaysDropItem ()Z
+	METHOD method_54293 shouldAlwaysDestroy ()Z
 	METHOD method_54294 getDamageWobbleStrength ()F
 	METHOD method_54295 getDamageWobbleTicks ()I
 	METHOD method_54296 getDamageWobbleSide ()I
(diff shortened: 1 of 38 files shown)


mappings/net/minecraft/entity/vehicle/VehicleEntity.mapping
```diff
@@ -2,7 +2,7 @@ CLASS net/minecraft/class_8836 net/minecraft/entity/vehicle/VehicleEntity
 	FIELD field_46655 DAMAGE_WOBBLE_TICKS Lnet/minecraft/class_2940;
 	FIELD field_46656 DAMAGE_WOBBLE_SIDE Lnet/minecraft/class_2940;
 	FIELD field_46657 DAMAGE_WOBBLE_STRENGTH Lnet/minecraft/class_2940;
-	METHOD method_54293 shouldAlwaysDropItem ()Z
+	METHOD method_54293 shouldAlwaysDestroy ()Z
```
Shnupbups: I disagree with this name change, though the existing name maybe isn't exactly right either?

When this returns `true` and the vehicle is damaged, if the player who hit it is in creative, the vehicle is killed (`Entity#kill`) and drops an item. (the same behavior as if a vehicle was broken normally, outside of creative mode)
When this returns `false` and the vehicle is damaged, if the player who hit it is in creative, the vehicle is discarded (`Entity#discard`) without an item drop.

This value is only ever queried when the vehicle is damaged by a player in creative mode.
This value only affects whether to kill and drop item (true), or to discard (false).
No matter the value, the entity is destroyed.

mappings/net/minecraft/entity/vehicle/VehicleEntity.mapping
```diff
@@ -14,6 +14,6 @@ CLASS net/minecraft/class_8836 net/minecraft/entity/vehicle/VehicleEntity
 		ARG 1 damageWobbleTicks
 	METHOD method_54300 setDamageWobbleSide (I)V
 		ARG 1 damageWobbleSide
-	METHOD method_7516 dropItems (Lnet/minecraft/class_1282;)V
-		ARG 1 damageSource
+	METHOD method_7516 destroy (Lnet/minecraft/class_1282;)V
+		ARG 1 source
```
Shnupbups, line 18: See above, perhaps a better name for this would be `killAndDropAsItem`

Shnupbups (changes requested):

mappings/net/minecraft/entity/vehicle/VehicleEntity.mapping
apple502j: There isn't really a good way to describe the behavior, without handling the TNT minecart case. This method in particular is only overridden by TNT carts - which, you know, explode.

apple502j:
```suggestion
	METHOD method_54293 shouldAlwaysKill ()Z
```

mappings/net/minecraft/entity/vehicle/VehicleEntity.mapping
```diff
@@ -14,6 +14,6 @@ CLASS net/minecraft/class_8836 net/minecraft/entity/vehicle/VehicleEntity
 		ARG 1 damageWobbleTicks
 	METHOD method_54300 setDamageWobbleSide (I)V
 		ARG 1 damageWobbleSide
-	METHOD method_7516 dropItems (Lnet/minecraft/class_1282;)V
-		ARG 1 damageSource
+	METHOD method_7516 destroy (Lnet/minecraft/class_1282;)V
```
apple502j:
```suggestion
	METHOD method_7516 killVehicle (Lnet/minecraft/class_1282;)V
```

mappings/net/minecraft/entity/vehicle/VehicleEntity.mapping
```diff
@@ -14,6 +14,6 @@ CLASS net/minecraft/class_8836 net/minecraft/entity/vehicle/VehicleEntity
 		ARG 1 damageWobbleTicks
 	METHOD method_54300 setDamageWobbleSide (I)V
 		ARG 1 damageWobbleSide
-	METHOD method_7516 dropItems (Lnet/minecraft/class_1282;)V
-		ARG 1 damageSource
+	METHOD method_7516 killVehicle (Lnet/minecraft/class_1282;)V
```
Shnupbups:
```suggestion
	METHOD method_7516 killAndDropSelf (Lnet/minecraft/class_1282;)V
```

apple502j:
```suggestion
	METHOD method_7516 killAndDropSelf (Lnet/minecraft/class_1282;)V
```

Shnupbups (approved):

PR #3712, opened 2023-10-31
23w43b NetherWartBlock

geralddhill:
I was making an similar to a Nether Wart and assumed that field_31199 in the NetherWartBlock class was just MAX_AGE similar the the CropBlock class, so I fixed that one thing.


YanisBft (approved):
Seems right

PR #3733, opened 2023-11-20
Rename grass to short grass

YanisBft:
And update enigma

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,7 +3,7 @@ org.gradle.jvmargs=-Xmx1G
 org.gradle.parallel=true
 org.gradle.configuration-cache=true
 
-enigma_version=2.3.2
+enigma_version=2.3.3
 unpick_version=2.3.0
 cfr_version=0.2.1
 name_proposal_version=0.1.5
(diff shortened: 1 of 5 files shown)


Benonardo (approved):

ErrorCraft:
`Blocks.GRASS` should also be renamed to `SHORT_GRASS`.

haykam821:
Is `Blocks.SHORT_GRASS` not already the name? The name should come from the string used in the method call to register the block.

YanisBft:
Yeah double-checked, the name is auto-generated

ErrorCraft:
Ah, alright. Wouldn't the item names be auto-generated too then, or does it not generate `SHORT_GRASS` because it doesn't have a string in the method it calls where it can take the name from?

YanisBft:
> Ah, alright. Wouldn't the item names be auto-generated too then, or does it not generate `SHORT_GRASS` because it doesn't have a string in the method it calls where it can take the name from?

Exactly, names of block items are not generated because there's no string in the register method

haykam821:
The item fields are initialized like `Items.register(SHORT_GRASS)`, so there isn't a string to derive a name from. However, FabricMC/name-proposal#9 expands the field naming system to support additional name providers, and naming fields transitively is a possibility.

Shnupbups (approved):

Shnupbups:
Requires developer review for changing gradle properties

modmuss50 (approved):

PR #3761, opened 2024-01-17
Add basic class mappings for new packets added in 24w04a

wgaylord:
These classes were discovered based on the 24w03a snapshot article and the names of enums in the source code.

### DIFF
--- a/mappings/net/minecraft/network/packet/c2s/common/CookieReqestResponseS2CPacket.mapping
+++ b/mappings/net/minecraft/network/packet/c2s/common/CookieReqestResponseS2CPacket.mapping
@@ -1,0 +1,1 @@
+CLASS net/minecraft/class_9091 net/minecraft/network/packet/c2s/common/CookieReqestResponseS2CPacket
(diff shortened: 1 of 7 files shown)


mappings/net/minecraft/network/packet/c2s/common/CookieReqestResponseS2CPacket.mapping
```diff
@@ -0,0 +1 @@
+CLASS net/minecraft/class_9091 net/minecraft/network/packet/c2s/common/CookieReqestResponseS2CPacket
```
Shnupbups: Typo in the name, be sure to fix both the file name and the content itself

wgaylord: Should be fixed now.

Shnupbups:
Could you please map their `CODEC` fields and `write` methods?

wgaylord:
Mapped the CODEC field and write methods on all 4 packets also mapped the static final field on that represents the max length of a cookie.

Also moved the transfer packet into the correct package.

mappings/net/minecraft/network/packet/c2s/common/CookieRequestResponseS2CPacket.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS net/minecraft/class_9091 net/minecraft/network/packet/c2s/common/CookieRequestResponseS2CPacket
```
Shnupbups: Name should reflect it being C2S

wgaylord: Not sure how I ended up doing that, since it renamed and put it in the right package at the same time, either way. This is now fixed,

Shnupbups (approved):

PR #3784, opened 2024-02-07
A little more 24w06 stuff

Dev0Louis:


mappings/net/minecraft/class_9226.mapping
```diff
@@ -0,0 +1,6 @@
+CLASS net/minecraft/class_9226
+	FIELD field_49035 PACKET_CODEC Lnet/minecraft/class_9139;
+	FIELD field_49036 DEFAULT_NAMESPACE Ljava/lang/String;
+	METHOD method_56931 isVanilla ()Z
+	METHOD method_56932 (Ljava/lang/String;)Lnet/minecraft/class_9226;
+		ARG 0 id
```
wgaylord: I was calling this path based on how Identifier calls it path instead of id.  Thoughts?

wgaylord: Also I ended  up calling this class DatapackIdentifier, since its the same as what a normal Identifier has but also contains a version string.

Dev0Louis: > I was calling this path based on how Identifier calls it path instead of id. Thoughts?

don't have anything against that.


> Also I ended up calling this class DatapackIdentifier, since its the same as what a normal Identifier has but also contains a version string.

Good idea, do you want to create your own pr or should I include it in my own?

Dev0Louis: Did end up calling it `VersionedIdentifier` and made the changes you suggested.
I also mapped one more method as #createVanilla as it creates the class with the vanilla namespace (minecraft)

apple502j (approved):

PR #3811, opened 2024-03-06
Add formatMappings task

modmuss50:
Any changes to the mappings will fail a PR build.

### DIFF
--- a/filament/src/main/java/net/fabricmc/filament/task/mappingio/FormatMappingsTask.java
+++ b/filament/src/main/java/net/fabricmc/filament/task/mappingio/FormatMappingsTask.java
@@ -31,12 +31,11 @@ public abstract class FormatMappingsTask extends FilamentTask {
 	void run() throws IOException {
 		Path path = getInput().get().getAsFile().toPath();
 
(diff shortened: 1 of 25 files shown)


filament/src/main/java/net/fabricmc/filament/task/mappingio/FormatMappingsTask.java
```diff
@@ -0,0 +1,42 @@
...
+
+	public FormatMappingsTask() {
+		getOutput().set(getInput());
+	}
+
+	@TaskAction
+	void run() throws IOException {
+		Path path = getInput().get().getAsFile().toPath();
+
+		try (MappingWriter writer = MappingWriter.create(path, MappingFormat.ENIGMA_DIR)) {
```
NebelNidas: The try-with-resources is no longer needed since MIO 0.5, as the writer now automatically closes itself on `visitEnd()`: https://github.com/FabricMC/mapping-io/pull/42

modmuss50: Ah nice, thanks.

NebelNidas (approved):

PR #3816, opened 2024-03-12
Map two methods in `FireworkRocketEntity`

ErrorCraft:
These were added in 24w09a.


ErrorCraft:
me when you freshly create a branch from 24w10a and it doesn't want to be up-to-date

Shnupbups (approved):

PR #3822, opened 2024-03-22
Javadoc fixes for net.minecraft.nbt.NbtIo

Octol1ttle:
This PR fixes ~~two issues~~ one issue I found when reading the Javadocs in NbtIo:

~~1) `compress` and `decompress` had their `@return` comments wrapped with `{}`~~
image

2) The wrong method name was used when referring to `writeCompound(NbtCompound, DataOutput)`
image


apple502j (changes requested):
Wrapped return value is intended

apple502j (approved):

Juuxel (approved):

PR #3837, opened 2024-04-06
[1.20.1] Add mappings for AbstractNbtList.add/remove

NerjalNosk:
Fixes NBT lists `add` and `remove` methods getting mixed up with their synthetic obfuscation.

Ensures Gradle to pick JVM 17 when the local default is set to another version (prevents Enigma from running properly)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -268,9 +268,6 @@ combineUnpickDefinitions {
 java {
 	sourceCompatibility = JavaVersion.VERSION_17
 	targetCompatibility = JavaVersion.VERSION_17
-	toolchain {
-		languageVersion.set(JavaLanguageVersion.of(17))
-	}
 }
 
 tasks.withType(JavaCompile).configureEach {
--- a/build.gradle
+++ b/build.gradle
@@ -268,6 +268,9 @@ combineUnpickDefinitions {
 java {
 	sourceCompatibility = JavaVersion.VERSION_17
 	targetCompatibility = JavaVersion.VERSION_17
+	toolchain {
+		languageVersion.set(JavaLanguageVersion.of(17))
+	}
 }
 
 tasks.withType(JavaCompile).configureEach {
--- a/mappings/net/minecraft/nbt/AbstractNbtList.mapping
+++ b/mappings/net/minecraft/nbt/AbstractNbtList.mapping
@@ -1,6 +1,6 @@
 CLASS net/minecraft/class_2483 net/minecraft/nbt/AbstractNbtList
 	COMMENT Represents an abstraction of a mutable NBT list which holds elements of the same type.
-	METHOD add (ILjava/lang/Object;)V
+	METHOD add add (ILjava/lang/Object;)V
 		ARG 1 index
 		ARG 2 value
 	METHOD method_10533 addElement (ILnet/minecraft/class_2520;)Z
@@ -21,7 +21,7 @@ CLASS net/minecraft/class_2483 net/minecraft/nbt/AbstractNbtList
 		COMMENT Gets the {@linkplain NbtElement#getType type} of element that this list holds.
 		COMMENT
 		COMMENT @return the type of element that this list holds
-	METHOD remove (I)Ljava/lang/Object;
+	METHOD remove remove (I)Ljava/lang/Object;
 		ARG 1 index
 	METHOD set (ILjava/lang/Object;)Ljava/lang/Object;
 		ARG 1 index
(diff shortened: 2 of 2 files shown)


build.gradle
```diff
@@ -268,6 +268,9 @@ combineUnpickDefinitions {
 java {
 	sourceCompatibility = JavaVersion.VERSION_17
 	targetCompatibility = JavaVersion.VERSION_17
+	toolchain {
+		languageVersion.set(JavaLanguageVersion.of(17))
+	}
```
modmuss50: No, this is a sperare issue and likely not the solution, Enimga should run just fine on J17+

modmuss50 (changes requested):
Hi, the mapping change looks reasonable. I would sort the toolchain stuff out as a diffrent thing.

I see this is for 1.20.1 which is fine, just want to double check that these have been named in the main branch?

build.gradle
NerjalNosk: Well that is weird, then, as I have OpenJDK21 as default, but somehow it breaks with the following message:

```
> Could not open proj generic class cache for build file '/****/yarn/build.gradle' (/home/nerjal/.gradle/caches/8.1.1/scripts/89g4wxh1rs3okpbpslwfmo3gm).
   > BUG! exception in phase 'semantic analysis' in source unit '_BuildScript_' Unsupported class file major version 65
```

NerjalNosk: Although turns out that change does not fix anything anyway, as it would appear I had forgotten I had set the gradle java home in my global `gradle.properties` as a fix attempt, but did not revert it. The build script change does not fix anything.

mappings/net/minecraft/nbt/AbstractNbtList.mapping
```diff
@@ -1,6 +1,6 @@
 CLASS net/minecraft/class_2483 net/minecraft/nbt/AbstractNbtList
 	COMMENT Represents an abstraction of a mutable NBT list which holds elements of the same type.
-	METHOD add (ILjava/lang/Object;)V
+	METHOD add add (ILjava/lang/Object;)V
```
modmuss50: Just realised that this mapping makes no sense, and will do nothing. You are trying to map `add` to `add` this will do nothing? This class likely extends `List`?

modmuss50 (changes requested):

mappings/net/minecraft/nbt/AbstractNbtList.mapping
NerjalNosk: I must admit that I do not understand how this actually occurs, I only made the change using Enigma in both NbtByteArray and NbtLongArray, and it would turn out this is the change that resulted from it

Shnupbups:
We usually don't add mappings to older versions of the game...

Shnupbups:
More importantly, I don't think it's necessary (or desirable) to map these synthetic bridge methods.
I believe in practice, in an actual dev environment, the proper names should already show up just fine.
The only place where the intermediaries are shown is in Enigma, in places where the methods are called. This is a bug with Enigma, but we probably shouldn't band-aid that bug with unnecessary mappings.
And actually, from my testing, after making the requested change to apply the mappings the intermediaries (`method_10531 add` instead of `add add`) it doesn't even actually band-aid the Enigma issue anyway, instead doing nothing...

NerjalNosk:
> More importantly, I don't think it's necessary (or desirable) to map these synthetic bridge methods.

Well as meaningless as it may seem, I do need them to be mapped, in order to extend the class, all so to provide a mirror instance of another, albeit with some form of "listeners", in order to make each changes be reflected on another object within the mod I am currently working on.

> And actually, from my testing, after making the requested change to apply the mappings the intermediaries (`method_10531 add` instead of `add add`) it doesn't even actually band-aid the Enigma issue anyway, instead doing nothing...

Honestly I do not know enough of dealing with mappings to know really what should or shouldn't be done, I am only trying to do what feels the most logical.

Also, additional point, the original mapping was only `add`, which Enigma first changed to `add add` upon trying to fix the issue on the NbtByteArray and NbtLongArray's ends (simply renaming the `method_10531` to `add`, and respectively for `method_10536` to `remove`). I only then manually made the other change as modmuss's comment on that matter made me think it might not be the right thing to do.

zifiri99 (approved):

zifiri99 (approved):

NerjalNosk (commented):
lowkey wondering what's the hold-up here

modmuss50:
> lowkey wondering what's the hold-up here

Sorry, this needs manually testing IMO to make sure its the correct change. As its for an older minecraft version its lower on the priority list, not helped by the fact I have been busy with 1.20.5 stuff and my laptop broke. Sorry again, ill try and get to this soon.

liach:
I think the existence of these mappings are solely for naming the parameters; these synthetic bridge methods should have their names auto-populated by our tools (enigma, stitch/loom) already. I recall enigma has some problem with populating names while stitch/loom works totally fine.

PR #3871, opened 2024-05-21
Change misleading method name `getNextMapId` in `ServerWorld` and `IdCountsState`

lovexyn0827:
Although named as `getNextMapId`, these methods also increase the id count of maps, which is very misleading.

Following is the implementation of `IdCountsState.getNextMapId()` in the Minecraft 1.20.6:

````
    public MapIdComponent increaseAndGetMapId() {
        int i = this.idCounts.getInt((Object)"map") + 1;
        this.idCounts.put((Object)"map", i);
        this.markDirty();
        return new MapIdComponent(i);
    }
````


modmuss50:
Please make sure to get this changed in the default branch first, it can then be back ported as needed. Thanks 👍

PR #3875, opened 2024-05-23
Add checkMappings task

modmuss50:
Checks for the issues fixed in https://github.com/FabricMC/yarn/pull/3873


build.gradle
```diff
@@ -150,7 +151,7 @@ tasks.register('yarnCommon', EnigmaTask) {
 	mappings = mappingsDir
 }
 
-tasks.register('checkMappings') {
+def checkPackages = tasks.register('checkPackages') {
```
liach, line 158: Do we need to declare a variable here? I thought we can always just use the task name to refer to the task, or `task.getByName('checkPackages')` in the worst case

filament/src/main/java/net/fabricmc/filament/task/mappingio/CheckMappingsTask.java
```diff
@@ -0,0 +1,64 @@
...
+
+		for (MappingTree.ClassMapping classMapping : tree.getClasses()) {
+			method: for (MappingTree.MethodMapping method : classMapping.getMethods()) {
+				Set<String> names = new HashSet<>();
+
+				for (int id = minNamespaceId; id < maxNamespaceId; id++) {
+					String name = method.getName(id);
+
+					if (name == null) {
+						break method;
```
liach: I think we should break the `id` loop or continue the method loop instead.

Shnupbups:
Would be nice if this detected duplicate class and field mappings in addition to methods, especially given the recent bug causing renamed classes to not have their old .mapping files deleted.

filament/src/main/java/net/fabricmc/filament/task/mappingio/CheckMappingsTask.java
NebelNidas: Fixed with my refactor 👍

NebelNidas (approved):

Shnupbups (approved):

Shnupbups:
Needs developer review (i.e. modmuss or Player) for merge

Player3324 (approved):

Shnupbups:
Requires #3948 before checks will pass

filament/src/main/java/net/fabricmc/filament/task/mappingio/CheckMappingsTask.java
```diff
@@ -0,0 +1,137 @@
...
+
+import org.gradle.api.file.DirectoryProperty;
+import org.gradle.api.tasks.InputDirectory;
+import org.gradle.api.tasks.TaskAction;
+import org.jetbrains.annotations.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import net.fabricmc.filament.task.base.FilamentTask;
+import net.fabricmc.mappingio.MappingReader;
```
liach, line 19: Checkstyle is failing on these out-of-order imports.

Shnupbups: Fixed

Shnupbups:
Needs dev review again...

PR #3887, opened 2024-05-30
Small component mappings

CallMeEchoCodes:
These were super easy. Just getting the small stuff out of the way so you smart people can figure out the hard stuff :)

Very new to mapping. Sorry if I did anything wrong.


apple502j:
I believe these are all already mapped.

PR #3916, opened 2024-07-31
Name methods and locals in PacketCodecs.BYTE_ARRAY

ZWORX52:
Ensures consistency with the other members of PacketCodecs and, in the case of naming the functions, is actually out of a concern with code correctness (encode and decode need to be overridden).


apple502j:
Methods that override another methods should not be renamed.

ZWORX52:
Sorry, I must have misworded it. The methods that are being overridden are called encode() and decode() and are overridden properly elsewhere in PacketCodecs, but in this instance, the code tries to override methods named method_xxxxx instead of encode() or decode(). This is the code error that I'm trying to fix.
image

Here's what another implementation of PacketCodec looks like: 
image
The other PacketCodecs all have encode() and decode() named, so I thought I'd name this one too.

Dev0Louis:
This is due to a fault in the Decompiler as far as I can see. As it should use the boxed type. 
I think that this will just need to be fixed by vineflower.

PR #3937, opened 2024-08-16
Map DatapackStructureProvider

MattiDragon:
A bit of a weird class. Would definetly like for someone to actually check if the names make sense


apple502j (commented):
P is capital

apple502j (approved):

PR #3941, opened 2024-08-17
Continue mapping 24w33a (redstone, entity rendering)

bluebear94:


mappings/net/minecraft/class_9929.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS net/minecraft/class_9929
```
apple502j, line 1: uh, please don't make mappings like this...

apple502j (approved):

PR #3956, opened 2024-08-22
Standardise WrapperLookup params and fields to "registries"

Juuxel:
- Some old more descriptive names like `nonReloadables`, `first`/`second`, `fullRegistry` and `base`/`patches` were retained.
- I also added this class as well as a few similar cases to CONVENTIONS.md.
  - Another option is a general rule for these "providers" (and `MatrixStack` is sort of already covered under the collection rules). It would just lead to refactors like `registry, blockRegistry -> blocks`, which are out of scope here. 😅

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -115,6 +115,16 @@ front of the coordinate (`velocityX`, not `xVelocity`).
 
 Name screen coordinates `x` and `y`, rather than `left` and `top`.
 
+### Special classes
+
+Unless a more descriptive name is better, these Minecraft classes should get specific names as fields and parameters:
+
+| Class                                                  | Name              |
+| ------------------------------------------------------ | ----------------- |
+| `net.minecraft.client.render.VertexConsumerProvider`   | `vertexConsumers` |
+| `net.minecraft.client.util.math.MatrixStack`           | `matrices`        |
+| `net.minecraft.registry.RegistryWrapper.WrapperLookup` | `registries`      |
+
 ## Javadocs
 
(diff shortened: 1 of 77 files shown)


NebelNidas (dismissed):
Overall a welcome change! However, searching for `ARG \d registryLookup` reveals further untouched entries. Have you left them as-is intentionally or have you simply missed them? Long-term we should let Name Proposal handle this via default parameter names attached to the corresponding class mappings.

Juuxel:
I used MIO to look at all `WrapperLookup` fields and all methods with a `WrapperLookup` parameter and map all of the disallowed names (`lookup`, `registryLookup`, `registriesLookup` and `wrapperLookup`) to `registries`.

I excluded lambdas/synthetics to avoid name clashes with the enclosing context just in case - that's where some of them come from. I'll fix this.

~~(There's also one instance of a synthetic method's `WrapperLookup` param being called `nbt`?)~~ That was just a bug in my code.

Juuxel:
The PR now has all existing WrapperLookup params and new params for all missing non-synthetic cases. I excluded the synthetic ones since it'd also include mapping invisible capture params for lambdas.

apple502j:
Note: I think for block/item/etc registry it makes sense to use `Registry` suffix. There are too many instances of lists/maps/sets of blocks or items that differentiating between them by name appears justified.

I am neutral on `registryLookup` vs `registries`. Does need a standard, though.

Juuxel:
Yeah, I agree on registries benefitting from the suffix.

I used `registries` as the name of the wrapper lookups as they are basically dictionaries of registry key -> view of registry. (And it was one of the original names used in Yarn for this type, along with the `*lookup` ones).

modmuss50 (approved):

Shnupbups (approved):

modmuss50 (approved):

PR #3957, opened 2024-08-22
24w34a fixes and minor renames

Juuxel:
See the individual commits for details.


apple502j:
re: slot amount - I think `amount` is better in that context; perhaps the standard could be that, use `count` for the `ItemStack#getCount`, and `amount` for the delta. Compare "The count of items to take" and "The amount of items to take". (While using "amount" for a countable noun is proscribed by some, the "recommended" alternative of "number" is much worse for mapping purposes.)

Juuxel:
Makes sense, all the ones I renamed are deltas. Should I change it back?

Juuxel:
I reverted the stack amount changes. I don't think there's anything else that's particularly controversial. (Note: the item stack method renames are to match the pre-component method names and the conventions for serialisation)

Shnupbups (approved):

PR #3960, opened 2024-08-23
Map ByteArray codecs in PacketCodecs

### LINKED ISSUE #3917: method_59799 and method_59800 in PacketCodecs.BYTE_ARRAY don't implement PacketCodec.encode/decode()
Screenshot of error in IntelliJ IDEA
image
Can be fixed by renaming them to `decode` and `encode` respectively.
Fixed by #3916.

Octol1ttle:
Fixes #3917 

This issue was very weird to me. Other codecs had `encode` and `decode` as their method names, but were marked obfuscated and there were no mappings for these methods except for the PacketCodec returned by the `byteArray` method. I'm not sure why `BYTE_ARRAY` field and `byteArray` method require manual mapping.

Also I mapped a constant that defines the maximum size a collection can be initialized with when reading collections. I think it needs to be unpicked to work properly, but I don't know how.

### DIFF
--- a/mappings/net/minecraft/network/codec/PacketCodecs.mapping
+++ b/mappings/net/minecraft/network/codec/PacketCodecs.mapping
@@ -77,6 +77,7 @@ CLASS net/minecraft/class_9135 net/minecraft/network/codec/PacketCodecs
 		COMMENT
 		COMMENT @see net.minecraft.network.PacketByteBuf#readByteArray()
 		COMMENT @see net.minecraft.network.PacketByteBuf#writeByteArray(byte[])
+	FIELD field_49674 MAX_INITIAL_COLLECTION_SIZE I
 	FIELD field_49675 INTEGER Lnet/minecraft/class_9139;
 		COMMENT A codec for an integer value.
 		COMMENT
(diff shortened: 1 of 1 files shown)


modmuss50:
> I think it needs to be unpicked to work properly, but I don't know how.

Im not sure if unpick can unpick only within a certain class, as we wouldnt want to unpick this for all `Math.min` usages across all classes.

modmuss50 (approved):

apple502j:
I feel like this has to be a bug in the toolchain, not a mapping issue.

Juuxel:
Yeah, this feels like an issue with specialised methods, which has happened before (see also #3916 which makes largely the same changes and wasn't merged).

Octol1ttle:
I didn't notice that PR, sorry

modmuss50:
Is it broken in dev? Does this work around it? If so maybe it is fine to merge idk.

Octol1ttle:
See #3917 for how this issue manifests during development

modmuss50:
This fix / workaround seems fine then. Unless you want to go digging for the root cause of the issue I dont have a problem with this for now?

Octol1ttle:
I'm not familiar enough with the toolchain to fix this myself. I submitted this PR because I saw that there were already mappings for `encode` and `decode` in the `byteArray` method. If this way works for you, you should be able to re-open and merge

PR #3992, opened 2024-09-20
Revise some Entity Brain Tasks names to better reflect their function

falseresync:
Brain tasks related to Walk and Look targets have inconsistent names that sometimes poorly reflect their function as well. For example, `LookAroundTask` and `RandomLookAroundTask` are named as if they do the same thing, but one does it randomly. But actually `LookAroundTask` reads from `LOOK_TARGET` memory, while `RandomLookAroundTask` writes to it. 

Therefore I suggest to unify task names under the following principle:
- Tasks that update Look target should start with `Look*` or `LookAt*`
- Tasks that update Walk target should start with `Go` or `GoTo*`
- Tasks that update both Look and Walk targets should start with `WalkTowards*`
- The current `LookAroundTask` that does neither of these should be renamed to `UpdateLookControlTask`

And the tasks that were previously named `GoTo*`, etc but don't update Look or Walk targets should be named in a different way, that better reflects their function too.

Additionally:
- `TemptationCooldownTask` isn't used just for temptation, but to tick the `RandomLookAroundTask` cooldown as well, so I suggest to rename it to `TickCooldownTask` (and to rename it's fields and methods accordingly)
- `LookTargetUtil` doesn't relate to just Look targets, but to several kinds of targets and has location utils, so I suggest to rename it to just `TargetUtil`


falseresync:
Note to self: need to compare to other Look* Tasks to see what they do and see if these new names fit.

Another variant:
- `LookAroundTask` becomes `UpdateLookControlTask`
- `RandomLookAroundTask` becomes `LookAroundTask`

Shnupbups (dismissed):

apple502j (approved):

Shnupbups (approved):

PR #4002, opened 2024-09-28
fix: ChunkGenerator#getDebugHudText -> ChunkGenerator#appendDebugHudText

### LINKED ISSUE #3996: The ChunkGenerator#getDebugHudText method mutates a list rather than returning text
The `ChunkGenerator#getDebugHudText` method acts like the `Item#appendTooltip` method but uses the wrong verb.

Octol1ttle:
Fixes #3996

The List<String> parameter may be worth renaming to `output` or just `list` maybe? I'm not sure so I'll leave it as-is


haykam821:
How is the list displayed within the overlay? If each `Text` represents a separate line, perhaps `lines` would work as a parameter name.

Octol1ttle:
Each ~~`Text`~~ `String` is a separate line in the debug HUD, yes. I think we can leave that name alone though, it's not a big deal

Shnupbups (approved):

PR #4012, opened 2024-10-06
24w40a recipe and client

### LINKED ISSUE #4011: `RealmsLoadingWidget` name is misleading
The `RealmsLoadingWidget` class is used for realms *and* for the statistics screen.

apple502j:
Resolves #4011

Also fixes some more yarnbugs, including the only remaining common class in client pkg, and `drawItemInSlot` (to `drawStackOverlay`)


Shnupbups:
What's the logic behind `ModelTransformationMode` being in the entity package? It's about item models, if anything it should be in the item package, or maybe just throw it in util or something.

Shnupbups (approved):

PR #4036, opened 2024-10-27
Renamed `Direction.idHorizontal` and related methods

Estecka:
The field `Direction.idHorizontal` doesn't make sense to interpret as an id, and it certainly doesn't correspond the id of any other direction. This field's value is very frequently multiplied by 90 to obtain an angle in degree, so it makes more sense to to see it as the number of horizontal quarter turns required to face this direction. 

 `asRotation()` does precisely that, so I renammed it to `getPositiveHorizontalDegrees()`. It does almost the same job as the existing `getHorizontalDegrees(Direction)`, except that the later throws when fed a vertical direction, and returns angles wrapped between -90° and 180°. The "positive" counterpart returns angles wrapped between 0° and 270°.

I renamed `getHorizontalDegrees` to `getHorizontalDegreesOrThrow` to further emphasize the difference in behaviour.


Octol1ttle:
Since `asRotation` got renamed, maybe `fromRotation` should also be renamed?

Estecka:
Good spot, I hadn't noticed these.
Added `fromRotation`->`fromHorizontalDegrees` and `fromHorizontal`->`fromHorizontalQuarterTurns`

mappings/net/minecraft/util/math/Direction.mapping
```diff
@@ -51,7 +51,7 @@ CLASS net/minecraft/class_2350 net/minecraft/util/math/Direction
 		ARG 1 y
 		ARG 2 z
 	METHOD method_10148 getOffsetX ()I
-	METHOD method_10150 fromRotation (D)Lnet/minecraft/class_2350;
+	METHOD method_10150 fromHorizontalDegrees (D)Lnet/minecraft/class_2350;
 		ARG 0 rotation
```
Octol1ttle: rename the argument as well?

Estecka: Done. I believe `angle` would suffice for this one in particular.

Octol1ttle (approved):

apple502j (dismissed):

modmuss50 (approved):

PR #4042, opened 2024-10-31
Follow convention regarding "last[A-Z]" instead of "prev"

Octol1ttle:


mappings/net/minecraft/block/vault/VaultClientData.mapping
```diff
@@ -1,7 +1,7 @@
 CLASS net/minecraft/class_9202 net/minecraft/block/vault/VaultClientData
 	FIELD field_48878 DISPLAY_ROTATION_SPEED F
 	FIELD field_48879 displayRotation F
-	FIELD field_48880 prevDisplayRotation F
+	FIELD field_48880 lastDisplayRotation F
 	METHOD method_56761 getDisplayRotation ()F
 	METHOD method_56762 getPreviousDisplayRotation ()F
```
NebelNidas:
```suggestion
	METHOD method_56762 getLastDisplayRotation ()F
```

NebelNidas: Unless you only want to replace `prev`, not `previous`

Shnupbups (approved):

PR #4052, opened 2024-11-17
Add 'Client' suffix to some 'World' methods to indicate that they only work on the client

Octol1ttle:
- **change: add 'Client' suffix to some 'World' methods to indicate that they only work on the client**
Self-explanatory.
- **change(World#playSound): explain the behavior on client vs. server in more detail**
Explains that, on the client, the sound will play if it's the same as the client's player (`source == this.client.player`). On the server, every player *except the source* will get a sound packet.
- **Add Javadocs to World#add(Important)ParticlesClient**
The Javadocs explain what an "important" particle is, what these methods do on the server (nothing), and reference the `ServerWorld` method used for spawning particles in case the developer is looking for it.

### DIFF
--- a/mappings/net/minecraft/world/World.mapping
+++ b/mappings/net/minecraft/world/World.mapping
@@ -45,6 +45,10 @@ CLASS net/minecraft/class_1937 net/minecraft/world/World
 		ARG 2 old
 		ARG 3 updated
 	METHOD method_17452 addImportantParticleClient (Lnet/minecraft/class_2394;ZDDDDDD)V
+		COMMENT Adds an important particle to the client's world renderer. Important particles get added even if the Particles graphics option is set to Minimal
+		COMMENT Does nothing on the server.
+		COMMENT
+		COMMENT @see ServerWorld#spawnParticles
 		ARG 1 parameters
 		ARG 2 force
 		ARG 3 x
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/world/World.mapping
```diff
@@ -144,8 +144,8 @@ CLASS net/minecraft/class_1937 net/minecraft/world/World
 	METHOD method_43127 getSpawnAngle ()F
 	METHOD method_43128 playSound (Lnet/minecraft/class_1657;DDDLnet/minecraft/class_3414;Lnet/minecraft/class_3419;FF)V
 		ARG 1 source
-			COMMENT the player that caused the sound and therefore plays the sound directly in
-			COMMENT the client, or {@code null}
+			COMMENT the player that caused this sound, or {@code null}. On the client, the sound will only if if the source is the same as the client's player.
```
Syst3ms:
```suggestion
			COMMENT the player that caused this sound, or {@code null}. On the client, the sound will only play if the source is the same as the client's player.
```
repeated across the other instances

Syst3ms (changes requested):

Syst3ms (approved):

Shnupbups (approved):

apple502j:
Uhhh, this shouldn't be merged blindly as Mojang did have some rewrites here, but guess I can fix that...

PR #4065, opened 2024-12-03
Fix/Refactor `VillagerEntity.canBreed` and `wantsToStartBreeding`

Estecka:
The method `VillagerEntity.canBreed` returns true if the villager is **unable** to breed.
The method `VillagerEntity.wantsToStartBreeding` may sometimes return false even if the villager is willing to breed. It only returns true if the villager has *twice over* the amount of food required to start breeding.

In practice, these methods are only ever used to check whether two villagers should share food with each other. They're never used to trigger or prevent the actual breeding task. (The method used for that purpose is `isReadyToBreed`.)

I propose:
`can(not)Breed` -> `needsFoodForBreeding`
`wantsToStartBreeding` -> `canShareFoodForBreeding`

### DIFF
--- a/mappings/net/minecraft/entity/passive/VillagerEntity.mapping
+++ b/mappings/net/minecraft/entity/passive/VillagerEntity.mapping
@@ -77,7 +77,7 @@ CLASS net/minecraft/class_1646 net/minecraft/entity/passive/VillagerEntity
 		ARG 2 villager
 	METHOD method_20696 decayGossip ()V
 	METHOD method_20697 eatForBreeding ()V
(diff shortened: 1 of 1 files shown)


Estecka:
Just noticed that a `lacksFood` also exists so I renamed it to `canEatFood` to avoid confusion.

`needsFoodForBreeding` and `canShareFoodForBreeding` check the villager's inventory, whereas `canEatFood` checks the villager's "hunger bar", and is used to consume food from the inventory.

Shnupbups (approved):

PR #4077, opened 2025-01-08
25w02a changes

MattiDragon:
Draft as I still have more to map


mappings/net/minecraft/block/SegmentedFlowerBlock.mapping
```diff
@@ -0,0 +1,6 @@
+CLASS net/minecraft/class_10594 net/minecraft/block/SegmentedFlowerBlock
```
Shnupbups, line 1: Should be named `FlowerbedBlock` to match its model's name and previous yarn name I think

Shnupbups (approved):

PR #4088, opened 2025-01-13
Dep updates

modmuss50:


filament/src/main/java/net/fabricmc/filament/nameproposal/MappingNameCompleter.java
```diff
@@ -136,9 +137,6 @@ private static MemoryMappingTree readMappings(Path path) throws IOException {
 	private static void inheritMappedNamesOfEnclosingClasses(MemoryMappingTree tree) {
 		int namedIdx = tree.getNamespaceId("named");
 
-		// The tree does not have an index by intermediary names by default
-		tree.setIndexByDstNames(true);
```
NebelNidas: What is the reasoning behind deleting these lines?

modmuss50:
Going to merge this, tested in a fabric api dev env.

PR #4094, opened 2025-01-21
Rename tickDelta to tickProgress

Octol1ttle:
This PR replaces all mentions of 'tickDelta' with 'tickProgress'. For an explanation of 'tickDelta', see https://github.com/FabricMC/fabric-docs/issues/173

I hesitated to make this pull request for a while. I think the name 'tickProgress' is not as good as it can be. Parchment uses 'partialTick', but some Fabricord members argued that it is too vague and it should be named 'fractionalTick' instead
I went with 'tickProgress' because it lines up with its main chunk of usages - MathHelper#lerp (`MathHelper.lerp(tickProgress, prevPos, pos)` sounds more obvious than `MathHelper.lerp(partialTick, prevPos, pos)`)
If you have any suggestions for a better name, please don't be afraid to leave a comment

### DIFF
--- a/mappings/com/mojang/blaze3d/systems/RenderSystem.mapping
+++ b/mappings/com/mojang/blaze3d/systems/RenderSystem.mapping
@@ -170,7 +170,7 @@ CLASS com/mojang/blaze3d/systems/RenderSystem
 		ARG 3 alpha
 	METHOD setShaderGameTime (JF)V
 		ARG 0 time
-		ARG 2 tickDelta
+		ARG 2 tickProgress
 	METHOD setShaderTexture (II)V
 		ARG 0 texture
 		ARG 1 glId
--- a/mappings/net/minecraft/block/ChestBlock.mapping
+++ b/mappings/net/minecraft/block/ChestBlock.mapping
@@ -43,4 +43,4 @@ CLASS net/minecraft/class_2281 net/minecraft/block/ChestBlock
 		ARG 0 state
 	CLASS 3
 		METHOD method_23900 (Lnet/minecraft/class_2595;Lnet/minecraft/class_2595;F)F
-			ARG 2 tickDelta
+			ARG 2 tickProgress
--- a/mappings/net/minecraft/block/entity/ChestLidAnimator.mapping
+++ b/mappings/net/minecraft/block/entity/ChestLidAnimator.mapping
@@ -5,6 +5,6 @@ CLASS net/minecraft/class_5560 net/minecraft/block/entity/ChestLidAnimator
 	FIELD field_27214 lastProgress F
 	METHOD method_31672 step ()V
 	METHOD method_31673 getProgress (F)F
-		ARG 1 delta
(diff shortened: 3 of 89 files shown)


IMB11 (approved):

Earthcomputer (approved):
🚀

NebelNidas:
What about the remaining usages?
- https://github.com/Octol1ttle/yarn/blob/02cf0a78eec8bdcc966c67baa42f6c3aa93d94cd/mappings/net/minecraft/client/render/Camera.mapping#L16
- https://github.com/Octol1ttle/yarn/blob/02cf0a78eec8bdcc966c67baa42f6c3aa93d94cd/mappings/net/minecraft/client/render/Camera.mapping#L56
- https://github.com/Octol1ttle/yarn/blob/02cf0a78eec8bdcc966c67baa42f6c3aa93d94cd/mappings/net/minecraft/client/render/RenderTickCounter.mapping#L5
- https://github.com/Octol1ttle/yarn/blob/02cf0a78eec8bdcc966c67baa42f6c3aa93d94cd/mappings/net/minecraft/entity/passive/CamelBrain.mapping#L28
- https://github.com/Octol1ttle/yarn/blob/02cf0a78eec8bdcc966c67baa42f6c3aa93d94cd/mappings/net/minecraft/entity/passive/CamelEntity.mapping#L24
- https://github.com/Octol1ttle/yarn/blob/02cf0a78eec8bdcc966c67baa42f6c3aa93d94cd/mappings/net/minecraft/entity/vehicle/ExperimentalMinecartController.mapping#L9

Not sure if there are more, just did a quick keyword search

Octol1ttle:
Good catch, my replace-all command was case-sensitive 😅

NebelNidas:
I'd also check whether parameters have been shortened from `tickDelta` to `delta`, and if there are any derivative names that need changing too (perhaps https://github.com/Octol1ttle/yarn/blob/02cf0a78eec8bdcc966c67baa42f6c3aa93d94cd/mappings/net/minecraft/entity/passive/CamelBrain.mapping#L30? Would have to look at the code)

Octol1ttle:
> I'd also check whether parameters have been shortened from `tickDelta` to `delta`, and if there are any derivative names that need changing too (perhaps https://github.com/Octol1ttle/yarn/blob/02cf0a78eec8bdcc966c67baa42f6c3aa93d94cd/mappings/net/minecraft/entity/passive/CamelBrain.mapping#L30? Would have to look at the code)

I've checked what I could and replaced it. Note that in `Screen`s and associated classes the `delta` is *actually* `deltaTicks` - the time difference between two frames, in ticks

I recommend that the maintainers review this PR by checking each commit separately to not get lost too much

Shnupbups (approved):
I think this is as good a time as any to make this change - gonna wait for another approval before merge though

modmuss50 (approved):

apple502j (approved):

PR #4116, opened 2025-02-13
25w07a changes

MattiDragon:
Draft as I'm still adding more


MattiDragon:
Nothing exhaustive, but what I had time for. Should help a bit with fapi porting even if the names aren't perfect

Shnupbups (approved):

PR #4164, opened 2025-04-03
Add names for GameInstance related stuff in 25w14craftmine

zly2006:


modmuss50 (approved):

modmuss50:
Im happy with this for now, its a good name. If this change does happen in a regular snapshot I expect we will debate the name some more.

PR #4189, opened 2025-05-04
Fix outdated unpick mappings

Earthcomputer:
There were some outdated unpick mappings. I'm not sure where the `ScreenHandlerSlotUpdateS2CPacket` constants went, I couldn't find them.


Earthcomputer:
Ah looks like some of these were changed while I was working on it

apple502j (approved):

PR #4209, opened 2025-06-12
Update HoneyBlock.mapping

Calboot:
These two methods have appeared since 1.21.2's snapshots, causing [MC-278572](https://bugs.mojang.com/browse/MC/issues/MC-278572)


Calboot:
these two methods have appeared since 1.21.2's snapshots, causing [MC-278572](https://bugs.mojang.com/browse/MC/issues/MC-278572)

PR #4212, opened 2025-06-21
fix: Item.inventoryTick is only called on the server side

Hunam6:


haykam821:
The `Item#inventoryTick` method only being called on the server is indeed the case since its second parameter is `ServerWorld` as of Minecraft snapshot 25w06a. Since this method is called by the `ItemStack#inventoryTick` method, which also handles the bobbing animation on the client, I wonder if the two methods should have different names.

haykam821 (dismissed):

modmuss50 (dismissed):

PR #4216, opened 2025-07-01
Fixed order of argument names of `TestContext.assertEquals`

Papierkorb2292:
This pull request swaps the order of the `expected` and the `value` argument of `TestContext.assertEqual`, such that `expected` is the first argument. 

This order is correct, since the error message translation (`"Expected %s to be %s, was %s"`) uses the first argument (now `expected`) in place of the second `%s` and the second argument (now `value`) in place of the third `%s`.


modmuss50 (approved):
Thanks

PR #4226, opened 2025-07-29
25w31a entity render command queue

mschae23:
I'm not familiar with Minecraft's rendering, so please do check if everything here is correct :)

It's in a new package because of all the classes that will actually render the commands at the end, which are not mapped in this PR.

I've also thought about giving `class_11659` a different name, since the command queue is technically internals (like `EntityRendering`, is that too generic?), but this is probably fine for now.

### DIFF
--- a/mappings/net/minecraft/class_11690.mapping
+++ b/mappings/net/minecraft/class_11690.mapping
@@ -1,6 +1,6 @@
 CLASS net/minecraft/class_11690
 	METHOD method_73016 vertex (Lorg/joml/Matrix4f;Lnet/minecraft/class_4588;IFFFFF)V
-		ARG 0 trasformationMatrix
+		ARG 0 matrix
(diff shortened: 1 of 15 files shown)


mappings/net/minecraft/class_11690.mapping
```diff
@@ -0,0 +1,10 @@
+CLASS net/minecraft/class_11690
+	METHOD method_73016 vertex (Lorg/joml/Matrix4f;Lnet/minecraft/class_4588;IFFFFF)V
+		ARG 0 trasformationMatrix
```
Shnupbups: typo

mschae23: I'll rename it to just `matrix`, “transformation” is actually redundant, I think

modmuss50 (approved):

PR #4228, opened 2025-07-29
25w31a render command queue renderer

mschae23:
Depends on #4226.


Shnupbups:
What about these renderers are command related? These just seem to be entity feature renderers, no?

mschae23:
> What about these renderers are command related? These just seem to be entity feature renderers, no?

Maybe there's a better term than “command” to prevent confusion, but basically, entities no longer render these things immediately, but add them to a queue now so they can all be rendered at once later. Maybe something related to batching could be better?

I don't know how this code was mapped before this snapshot, it might be worth it preserving some names for consistency (like “feature”?), although the modder-facing names were in a different PR that's already merged... but since Fabric API hasn't released yet, changing names shouldn't be a problem yet at this point.

mschae23:
I looked at it again. Entity feature renderers still exist; entity render commands are used in the *implementation* of entity feature renderers (as well as normal entity renderers).

The end goal is probably to be able to have all the render commands for the next frame ready while the current one is still being rendered.

modmuss50 (approved):

PR #4234, opened 2025-08-05
25w31a more rendering things

mschae23:
These are already a few days old (was blocked by the other PR), but I figured it's worth submitting them even though there's probably a snapshot releasing today.


modmuss50:
Awesome, yes im trying to get everything wrapped up before the next snapshot. Thanks :)

modmuss50 (approved):

PR #4239, opened 2025-08-08
Port to unpick v3

Earthcomputer:
See [the unpick wiki](https://github.com/FabricMC/unpick/wiki/Unpick-Format-V3) for a specification on the unpick v3 format.


filament/src/main/java/net/fabricmc/filament/task/RemapUnpickDefinitionsTask.java
```diff
@@ -92,58 +126,189 @@ public void execute() {
...
+														.add(mappedClassName.replace('/', '.'));
+									}
+
+									@Override
+									public FieldVisitor visitField(int access, String name, String descriptor, String signature, Object value) {
+										String mappedName = name;
+										MappingTree.FieldMapping fieldMapping = mappingTree.getField(className, name, descriptor, classpathM);
+
+										if (fieldMapping != null) {
+											mappedName = fieldMapping.getName(fromM);
```
modmuss50: Dont be afraid to create new classes/methods, this is indented 11 times! Class/method names also help explain what the class is does. This class visitor should become its own class imo.

filament/src/main/java/net/fabricmc/filament/task/RemapUnpickDefinitionsTask.java
```diff
@@ -92,58 +126,189 @@ public void execute() {
...
+									}
+								}, ClassReader.SKIP_CODE);
+							}, executor));
+						}
+					}
+				}
+
+				for (CompletableFuture<Void> future : futures) {
+					future.join();
+				}
```
modmuss50: Is this needed? I think closing the exeuctor does this already.

filament/src/main/java/net/fabricmc/filament/task/UnpickJarTask.java
```diff
@@ -56,25 +78,102 @@ public interface UnpickParameters extends WorkParameters {
...
+
+						if (entry.data != null) {
+							outputZip.write(entry.data);
+						}
+
+						outputZip.closeEntry();
+					}
+				}
+			} catch (IOException e) {
+				throw new RuntimeException(e);
```
modmuss50: Nit: UncheckedIoException

filament/src/main/java/net/fabricmc/filament/task/UnpickJarTask.java
```diff
@@ -56,25 +78,102 @@ public interface UnpickParameters extends WorkParameters {
...
+					try {
+						classpathZip.close();
+					} catch (IOException e) {
+						// ignore
+					}
+				}
+			}
+		}
+
+		private static byte[] readAllBytes(InputStream in) throws IOException {
```
modmuss50: Is this needed, id prefer to use `InputStream.readAllBytes()`

filament/src/main/java/net/fabricmc/filament/task/UnpickJarTask.java
```diff
@@ -56,25 +78,102 @@ public interface UnpickParameters extends WorkParameters {
...
-				Main.main(args.toArray(String[]::new));
+				FileUtil.deleteIfExists(outputFile);
 			} catch (IOException e) {
 				throw new RuntimeException(e);
 			}
+
+			List<ZipFile> classpathZips = new ArrayList<>();
+
+			try (
+					ZipFile inputZip = new ZipFile(getParameters().getInput().get().getAsFile());
```
modmuss50, line 90: This is fine for yarn, but in loom we tend to always use NIO zip fs via tiny remappers `FileSystemReference.openJar` (loom has loads of helpers around this) instead of the "legacy" ZipFile apis. Dont rewrite this now, but keep in mind for the future.

It may make more sense to try and share some of this code with loom, such as `ZipFsClasspath` and `AsyncZipProcessor`. It will likely need some loom changes to make it a better library.

filament/src/main/java/net/fabricmc/filament/task/RemapUnpickDefinitionsTask.java
```diff
@@ -92,58 +126,189 @@ public void execute() {
...
+
+						@Override
+						protected List<String> getClassesInPackage(String pkg) {
+							return jarIndex.classesInPackages.getOrDefault(pkg, List.of());
+						}
+
+						@Override
+						protected String getFieldDesc(String className, String fieldName) {
+							final String fieldKey = className + "." + fieldName;
+							String fieldDesc = jarIndex.fieldDescs.get(fieldKey);
```
modmuss50, line 178: In loom I used TR for this, may that be easier? I'm slightly worried about how much code is duplicated (in a slightly different way) between here and loom. Long term the less code we have to maintain the better, but then I dont want to go too extreme with it.

https://github.com/FabricMC/fabric-loom/blob/573bc119ecf434666e0fbbfc1b2bbe9730905d29/src/main/java/net/fabricmc/loom/task/service/UnpickRemapperService.java#L137-L147

filament/src/main/java/net/fabricmc/filament/task/RemapUnpickDefinitionsTask.java
Earthcomputer: Ah, I was under the impression that it could cancel them but you're right

filament/src/main/java/net/fabricmc/filament/task/RemapUnpickDefinitionsTask.java
Earthcomputer: Does a `TrClass` also include classes that are not in the mappings, even JDK classes?

modmuss50:
The unpick json should be updated to the new V2 spec

https://github.com/FabricMC/yarn/blob/25w32a/unpick-definitions/unpick.json such as: https://github.com/FabricMC/fabric-loom/blob/573bc119ecf434666e0fbbfc1b2bbe9730905d29/src/test/groovy/net/fabricmc/loom/test/unit/providers/mappings/unpick/UnpickMetadataTest.groovy#L64-L66

Let me know if you need help with that. We also need to test that this works end-to-end with loom before merging.

filament/src/main/java/net/fabricmc/filament/task/UnpickJarTask.java
Earthcomputer: Honestly forgot that existed lol

filament/src/main/java/net/fabricmc/filament/task/RemapUnpickDefinitionsTask.java
Earthcomputer: Not immediately sure how to reuse that. I'll leave it for now as this code will already have to change quite a bit when it becomes an Enigma plugin.

TR might be a bit easier, I didn't think of it, although I'm already having to scan the jar for the packages. Looking at it, it doesn't seem to include JDK classes, which I do via reflection here, it's required because Yarn has some unpick definitions referencing `java.util.Spliterator`.

modmuss50 (approved):
Tested, seems to work great. Lets get this merged if there are any issues we can fix them, the whole reason for waiting for snapshots before doing this.

unpick-definitions/unpick.json
```diff
@@ -1,5 +1,4 @@
 {
-  "version": 1,
-  "unpickGroup": "net.fabricmc.unpick",
-  "unpickVersion": "${version}"
-}
\ No newline at end of file
+  "version": 2,
+  "namespace": "named"
```
modmuss50: Just noticed this should have the constants maven def defined here

Earthcomputer: I removed it for now as we don't need it currently and it wasn't published at least by the `publishToMavenLocal` task. Does not having it cause issues?

modmuss50: Ive added it back, it seems to work just fine. Just seen its tottally empty, lets merge this with it, and I can clean it up in a follow up PR to fully remove it (e.g stop building+publishing it)

PR #4265, opened 2025-08-27
[25w35a] Mapping remaining `/server/dedicated/management`

rotgruengelb:
not sure i always understood the exact implementation here, but I did my best to find fitting mappings.
feel free to edit


rotgruengelb:
not exactly happy with some of the names, but cant thing of anything better

mschae23 (approved):

PR #4281, opened 2025-09-04
25w36b `MinecraftClient#is?Pressed` mappings

0x000006:


0x000006:
That wasn't intentional!

PR #4291, opened 2025-09-06
Add field name providers to name proposal

haykam821:
This pull request adds field name providers, which are a declarative way to modify how field names are proposed. Field name providers can be applied in sequence to apply a specific name to fields matching certain conditions or apply a string argument-based name like the existing behavior.

The same mappings are built by this pull request and the previous revision.


modmuss50 (approved):
This feels a little overengineered/enterprisy at least for what NP currently does. Im happy to merge this, as we can always improve it as we go on.

PR #4293, opened 2025-09-06
Unpick: refactor and add new GLFW/InputUtil entries

0x000006:


unpick-definitions/glfw/key.unpick
```diff
@@ -0,0 +1,136 @@
+unpick v3
+
+group int key
```
Earthcomputer, line 3: While you're at it you could maybe use the constants in `InputUtil` where possible

0x000006: Acknowledged.

0x000006:
Reason: mapping `InputUtil` unmapped fields

apple502j (approved):

PR #4306, opened 2025-09-17
Don't apply the unpick plugin outside the yarn task

Earthcomputer:
Applying the unpick plugin includes starting the unpick directory watcher thread. This means the thread now won't be started outside the yarn task. The yarn task forks the JVM, so the daemon thread should stop once the task ends.


filament/src/main/java/net/fabricmc/filament/unpick/enigma/UnpickEnigmaPlugin.java
```diff
@@ -47,6 +47,14 @@ public class UnpickEnigmaPlugin implements EnigmaPlugin {
 
 	@Override
 	public void init(EnigmaPluginContext ctx) {
+		String unpickPath = System.getProperty("unpick.directory");
+
+		if (unpickPath == null) {
+			return;
+		}
+
+		unpickDir = Path.of(System.getProperty("unpick.directory"));
```
Juuxel:
```suggestion
		unpickDir = Path.of(unpickPath);
```

Player3324 (approved):

PR #4307, opened 2025-09-17
Back to 1GB memory

modmuss50:
This reverts the memory required back to the previous 1GB


Player3324:
Someone needs to run a gradle clean build with visualvm on the daemon open to verify, i can do that later maybe

Player3324 (approved):
the leak seems to be fixed

PR #4319, opened 2025-09-22
Rename entityRenderStates to worldRenderState in GameRenderer and WorldRenderer

DennisOchulor:
Also renames WorldRenderState#entityStates to entityRenderStates for consistency.


mappings/net/minecraft/client/render/state/WorldRenderState.mapping
haykam821, line 1: Perhaps the 'state' or even 'render state' suffix is unnecessary since the class itself is a render state, but the improved consistency is good for now.

haykam821 (approved):

PR #4329, opened 2025-10-10
Add gizmo mappings

bluebear94:


mappings/net/minecraft/client/render/DrawStyle.mapping
```diff
@@ -0,0 +1,15 @@
+CLASS net/minecraft/class_12179 net/minecraft/client/render/DrawStyle
+	FIELD field_63655 DEFAULT_STROKE_WIDTH F
+	METHOD method_75534 hasFill ()Z
+	METHOD method_75535 stroked (I)Lnet/minecraft/class_12179;
```
Juuxel, line 4: Or `withStroke` etc? (same pattern as in the `Style` type for texts)

bluebear94: This is a static factory method, so I think `stroked` is more appropriate here.

Juuxel: Ah, I thought it was like the `Style` methods. (Another option is `ofStroke` like JDK factory methods, but the `-ed` names are fine for statics)

apple502j (approved):

PR #4333, opened 2025-10-11
Automatically map parameter names in the default constructor in record classes

O7410:


filament/src/main/java/net/fabricmc/filament/nameproposal/MappingNameCompleter.java
```diff
@@ -98,6 +103,43 @@ public static void completeNames(Path intermediaryJar, Path inputYarnMappings, P
...
+
+			yarn.visitMethod("<init>", initDesc);
+			MappingTree.MethodMapping methodMapping = classMapping.getMethod("<init>", initDesc, yarnIntermediaryNs);
+			int lvIndex = 1;
+
+			for (RecordComponentNode recordComponentNode : recordComponents) {
+				int currentLvIndex = lvIndex;
+				lvIndex += Type.getType(recordComponentNode.descriptor).getSize();
+				String name = recordNames.get(recordComponentNode.name);
+				if (name == null) continue;
```
Earthcomputer: Braces for the if statement, not sure why this isn't in checkstyle

Juuxel: The checkstyle lets these single line ifs skip the braces:
https://github.com/FabricMC/yarn/blob/49e01fb86265baa8417307b4c7789e11af7b25dc/filament/checkstyle.xml#L82-L86

Earthcomputer: Hmm, I don't think it should

Earthcomputer (approved):

Juuxel (approved):

modmuss50 (approved):
Awesome, looks great.

filament/src/main/java/net/fabricmc/filament/nameproposal/MappingNameCompleter.java
modmuss50: Yeah, we might want to change this. I'm not a big fan of it.

PR #4336, opened 2025-10-13
Map more record constructors

O7410:
Previously, it would just check the recordNames.
Now:
- if the obfuscated name of the record component doesn't start with "comp_", use it
- if there is a mapped method or field with the same intermediary name, use it

the recordNames maps methods and fields right before this code, so it'll find names from recordNames too
only if those 2 checks fail, then it doesn't map the parameter

also in this PR, it no longer maps methods or classes to themselves, it now leaves them unmapped, with \t\n


modmuss50:
Can you update the PR description to explain how you have got it to map more constructors? To whats different between the old and new code?

O7410:
oh sorry i'll update it tomorrow

O7410:
updated, sorry for the wait

modmuss50 (approved):

PR #4342, opened 2025-10-14
25w42a stuff

apple502j:

PR #4344, opened 2025-10-15
Map bed rules

haykam821:


mappings/net/minecraft/class_12193.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS net/minecraft/class_12193
```
Shnupbups, line 1: Please name this class

haykam821: I do not like this class very much.

apple502j: If you don't map I'll just merge this into #4342

Shnupbups: Should this not just be `net/minecraft/world/attribute/AttributeTypes`?

apple502j (approved):

PR #4346, opened 2025-10-15
Map more record constructors (fixed)

O7410:
maps classes to themselves if they don't have a name, always maps \<init\> to \<init\>


modmuss50 (approved):
This is fine for now, as I have said on discord we should fix the issues requiring stuff to be named to its self, but thats likely a bit of a pain to do.

PR #4351, opened 2025-10-19
Annotations editor

Earthcomputer:
I know this is a lot... I would recommend trying it out to get a feel for the features that exist before reviewing.

To open the annotations editor, right click on any declaration in Enigma and press "Edit Annotations", or use the ctrl+T shortcut.

The GUI where you type in a new annotation or edit an existing one (called `SingleAnnotationEditor` in the code) could still be improved usability-wise, but it's usable enough for now.


modmuss50 (dismissed):
Happy to merge this whenever, we might want to move this code in the future but at least here its easily changed.

modmuss50 (approved):

modmuss50 (approved):

PR #4352, opened 2025-10-20
Contract annotations for net.minecraft.util.math classes

skycatminepokie:
These classes trip me up sometimes, with immutability and mutability mixed in with each other. Hopefully, adding contract annotations will help.

I avoided annotating anything that returns a primitive type - that seems unnecessary and would add bloat. I also didn't annotate static methods, as they can usually be assumed to be pure.

Depends on #4351 , and has more classes to do, so I'm leaving this as a draft.


annotations.json
```diff
@@ -577,6 +577,168 @@
...
+      "methods": {
+        "<init>(Lorg/joml/Vector3fc;)V": {
+          "add": [
+            {
+              "desc": "Lorg/jetbrains/annotations/Contract;",
+              "values": {
+                "value": {
+                  "type": "string",
+                  "value": "!null->new"
+                },
```
Earthcomputer: This is unnecessary, it's the constructor, of course it returns something new.

In general, I would focus on annotations that will prevent bugs, rather than annotating every contract that is technically true, that just creates noise.

Also, the contract values should have spaces around the `->` arrows, and after the `,` commas, to make them easier to read.

Earthcomputer (changes requested):

annotations.json
skycatminepokie: That's my mistake, I know I don't want to be annotating constructors. I also thought the contract didn't support spaces, but I was wrong there. Will make changes in next push.

skycatminepokie:
OK, that's a lot better. I went a little crazy at first xd

cputnam-a11y:
~~The git history seems broken~~ see below apparently I did a stupid

skycatminepokie:
Where? I wouldn't be surprised though, I will fix it when your PR is merged. I checked out the latest yarn, then merged your branch on top, then made my changes. This is really my first experience with working with multiple collaborative branches - it's always been just me or a random one-off change :P

Earthcomputer:
The git history looks fine to me, I think he missed that this PR has a dependency. It will need rebasing once my PR is merged.

skycatminepokie:
OK just so I don't do a silly, I rebase onto 25w44a, then force-push, right? If I update first, I don't have to force-push, but that adds an empty commit and that seems unhelpful.

skycatminepokie:
OK, I kinda forgot about this. Can someone update the base? I'll fix conflicts and then this should be ready.

mappings/net/minecraft/util/math/AffineTransformation.mapping
```diff
@@ -5,7 +5,7 @@ CLASS net/minecraft/class_4590 net/minecraft/util/math/AffineTransformation
 	COMMENT {@linkplain #translation translation} 3-vector. It is also known as "TRSR"
 	COMMENT transformation, meaning "translation rotation scale rotation".
 	COMMENT
-	COMMENT <p>This class is immutable; its matrix is lazily decomposed upon demand.
+	COMMENT <p>This class is immutable, but not thread-safe; its matrix is lazily decomposed upon demand.
```
Earthcomputer, line 8: How can it be immutable but not thread safe?

annotations.json
```diff
@@ -1,8 +1,132 @@
...
+    "net/minecraft/util/math/BlockBox": {
+      "methods": {
+        "encompass(Lnet/minecraft/util/math/BlockBox;)Lnet/minecraft/util/math/BlockBox;": {
+          "add": [
+            {
+              "desc": "Lorg/jetbrains/annotations/Contract;",
+              "values": {
+                "value": {
+                  "type": "string",
+                  "value": "!null -> this"
```
Earthcomputer:
```suggestion
                  "value": "_ -> this"
```
You should only use `!null` if null is a valid input according to nullability annotations (params are not null by default in Minecraft). You've done this a lot

Earthcomputer (changes requested):

annotations.json
skycatminepokie: Fixed on next push

mappings/net/minecraft/util/math/AffineTransformation.mapping
skycatminepokie: > its matrix is lazily decomposed upon demand

This laziness is what makes it not thread-safe. It's immutable by outside users, but modifies its own private fields when computing.

Example race condition:
1. Create on thread 1, pass to thread 2
2. Call `getTranslation` on thread 1, follow through the `init` call until `initialized = true`
3. On thread 2, call `getTranslation`, move into the `init` call
4. Attempt to read `this.initialized` - it's currently being written to by thread 1, which leads to undefined behavior

I'm not 100% positive on whether that interaction is atomic or not, if I'm being honest. There are other writes, but since you're writing the same value, I don't *think* that is a problem (other than a bit of wasted computation).

Code:
```java
private void init() {
	if (!this.initialized) {
		float f = 1.0F / this.matrix.m33();
		Triple<Quaternionf, Vector3f, Quaternionf> triple = MatrixUtil.svdDecompose(new Matrix3f(this.matrix).scale(f));
		this.translation = this.matrix.getTranslation(new Vector3f()).mul(f);
		this.leftRotation = new Quaternionf(triple.getLeft());
		this.scale = new Vector3f(triple.getMiddle());
		this.rightRotation = new Quaternionf(triple.getRight());
		this.initialized = true;
	}
}
public Vector3fc getTranslation() {
	this.init();
	return this.translation;
}
```

Earthcomputer: I see

Earthcomputer (dismissed):
The git history is still kinda weird but it will fix itself when getting squashed on merge anyway

Earthcomputer (approved):

PR #4359, opened 2025-10-25
Publish unpick definitions alone as a separate artifact

jpenilla:
This is useful for external consumers that do not need or want the actual mappings, which take up the vast majority of the jar publication size

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -242,28 +242,12 @@ tasks.register('tinyJar', Jar) {
 	}
 }
 
-tasks.register('compressTiny', FileInputOutput) {
+tasks.register('compressTiny', net.fabricmc.filament.task.GzipTask) {
 	dependsOn tinyJar, mergeTiny
 	group = buildMappingGroup
 
 	input = mergeTiny.output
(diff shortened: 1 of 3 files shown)


modmuss50:
I'm happy with the idea, I just wonder if a jar would be better than a gzip file? A jar makes it trivial to put on the classpath of an application if needed and likely a little bit easier to handle (as basically everything on maven is a jar/zip).

jpenilla:
Swapped to a jar using the same structure as the merged jar

modmuss50:
Thanks. One last question: Would having this as a -extras.jar that also contains the annotation data be useful, the other option would be to just add a -annotations jar in a later PR.

Im unsure idk if you or anyone else has any thoughts?

jpenilla:
Sounds reasonable to combine them as they are both small relative to the main mappings files. Won't be able to apply that change myself until tomorrow.

Earthcomputer:
It's probably useful to include them both, unpick is applied after annotations, and the output of the annotations transforming can affect the output of unpick.

modmuss50 (approved):

PR #4360, opened 2025-10-25
Unpick: add new GL definitions and nullable_pointer

0x000006:


0x000006:
Thanks for updating the base, I forgot about this PR.

PR #4363, opened 2025-10-26
Update to unpick v4

Earthcomputer:
Updated to unpick V4, which adds support for `target_annotation`.

I have changed `set_block_state_flags.unpick` to use this feature, kind of as a demonstration, there are other places it could be used as well, which I'll leave to others to map. I had to leave in a few of the existing targets where Mojang forgot to annotate.


Earthcomputer:
When this gets merged, the new unpick file in #4360 will need updating to unpick v4, whichever way round they're merged.

Earthcomputer:
I've made the `checkUnpickDefinitions` task now ensure that all files are v4, so that we don't forget to update any files

modmuss50:
Note: This cannot be merged until we have a stable Loom 1.12 (hopefully in a few days)

0x000006:
> When this gets merged, the new unpick file in #4360 will need updating to unpick v4, whichever way round they're merged.

I would prefer this PR being merged first so I can rebase mine on this afterwards.

unpick-definitions/set_block_state_flags.unpick
Lulu13022002, line 1: Seems they also forgot `WorldRenderer#updateBlock` flags param if you want to add it here.

Earthcomputer: I edited your comment, please don't mention mojmap names in the yarn repository :)

I only kept our definitions that are still needed, didn't search for where new ones are needed. I'll add it now

Earthcomputer: I ended up getting sidetracked and not adding it, can always be added by yarn mappers (of which I am not one ;) ) later

modmuss50 (dismissed):

unpick-definitions/set_block_state_flags.unpick
Lulu13022002: That's fair ^^, maybe the enigma plugin should be updated too, to copy directly the target reference for annotations?

PR #4367, opened 2025-10-29
25w44a GameRules

ekulxam:
Adds mappings for GameRules. Since this is my first time contributing to yarn, I hope this pull request is satisfactory.

I'm somewhat unsure about these names, especially all the `updateValue` methods in GameRules, `copyValuesIf` in ServerGameRules, and `toShortString` in Identifier. I'd consider naming ServerGameRules to GameRuleContainer, but I'm not sure which is better. Additionally, I'm not sure whether putValue or setValue is better for some methods, so there is a mix of both throughout the classes.

All gamerule-related classes have been moved under `net/minecraft/world/rule`.


offbeat-stuff:
The value field in Gamerule should be called defaultValue
From what I have read in code minecraft keeps track of actual values in a seperate reference map

ekulxam:
True, I overlooked the final modifier on the field. Thank you!

modmuss50 (approved):

PR #4401, opened 2026-02-06
Fix RegionFile constructor argument order

0x000006:


haykam821 (approved):
This parameter swap is very sneaky:

```java
public RegionFile(StorageKey storageKey, Path directory, Path path, boolean dsync) throws IOException {
	this(storageKey, directory, path, ChunkCompressionFormat.getCurrentFormat(), dsync);
}

public RegionFile(StorageKey storageKey, Path path, Path directory, ChunkCompressionFormat compressionFormat, boolean dsync) throws IOException {
	// ...
}
```

PR #4407, opened 2026-05-20
Resolve `Pool` name collision

bluebear94:
## Executive summary

Resolves duplicate classes named `Pool` introduced by 721bbfa75fbb991c4533bcf3e7f3acf2593cac19.

* `net.minecraft.client.util.Pool` has been renamed to `ObjectPool` to remove the name collision with `net.minecraft.util.collection.Pool`.
* `net.minecraft.util.collection.Pool` has been renamed to `WeightedPool` (the official name is likely `WeightedList` as mentioned in error messages, but there is an existing class with this as its Yarn name).
* `WeightedList` has been renamed to `ShufflingList` (mentioned in its `toString` implementation).
* Several memory-related client util classes have been moved to their own package.
  * ~~`BufferAllocator` is moved to `com.mojang.blaze3d.buffers` instead. There is precedent for obfuscated classes under `com.mojang.blaze3d`: namely `ProjectionType` and `VertexSorter`.~~

### DIFF
--- a/mappings/com/mojang/blaze3d/buffers/BufferAllocator.mapping
+++ b/mappings/com/mojang/blaze3d/buffers/BufferAllocator.mapping
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_9799 net/minecraft/client/util/BufferAllocator
+CLASS net/minecraft/class_9799 com/mojang/blaze3d/buffers/BufferAllocator
 	FIELD field_52078 LOGGER Lorg/slf4j/Logger;
 	FIELD field_52079 ALLOCATOR Lorg/lwjgl/system/MemoryUtil$MemoryAllocator;
 	FIELD field_52080 MIN_GROWTH I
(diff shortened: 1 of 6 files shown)


mappings/com/mojang/blaze3d/buffers/BufferAllocator.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/class_9799 net/minecraft/client/util/BufferAllocator
+CLASS net/minecraft/class_9799 com/mojang/blaze3d/buffers/BufferAllocator
```
mschae23: Can anyone who happens to know how the blaze3d distinction is made tell if this is correct?

mschae23 (approved):
Seems good to avoid the name conflict. I don't know if the package move is correct (as mentioned in my comment), but it doesn't sound like too big of an issue even if it's wrong.

mappings/com/mojang/blaze3d/buffers/BufferAllocator.mapping
modmuss50: I believe this is wrong as all blaze3d classes are already unobfuscated, and thus dont get remapped. Any obfuscated class such as this should stay in net/minecraft.

bluebear94: I’ve moved that class back to its old place.

modmuss50 (approved):

